PR #38, opened 2025-07-07
Add basic Github pipelines

gabotechs:
The pipelines will run:
- cargo clippy
- cargo fmt -- --check
- cargo test

on every:
- PR push
- main merge

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -773,9 +773,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.27"
+version = "1.2.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d487aa071b5f64da6f19a3e848e3578944b726ee5a4854b82172f02aa876bfdc"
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Update and install APT packages
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo apt-get update
+        xargs -a .github/actions/setup/apt-packages.txt sudo apt-get install -y
+
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        toolchain: stable
```
fmonjalet, line 40: Does this always take the latest stable, or will this read what the project requests?

gabotechs: This will use the latest stable version yes. The project requests a minimum version of 1.82, but I don't think it's enforcing any specific version newer than that

gabotechs: Do you think it might be better to lock to a specific version?

PR #41, opened 2025-07-07
Support for substrait plans

LiaCastaneda:
This PR adds support to input substrait plans, the substrait plan `select_one` corresponding to `select 1 as test_col` translates to the distributed plan:

```
[ output_partitions: 1]MaxRowsExec[max_rows=8192]
[ output_partitions: 1]  CoalesceBatchesExec: target_batch_size=8192
[ output_partitions: 1]    ProjectionExec: expr=[1 as test_col]
[ output_partitions: 1]      DataSourceExec: partitions=1, partition_sizes=[1]
```

Even if the distributed plan is built, most substrait queries will fail when the plan is assigned to the workers, because distributed-datafusion doesn’t support serializing DataSourceExec physical nodes.  (see [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/84ed49c17174afeb1039e6143766d9a8d47f66d9/src/codec.rs#L147)).

For a standard sql string query that executes successfully, the distributed plan for the same query would be:

```
[ output_partitions: 1]MaxRowsExec[max_rows=8192]
[ output_partitions: 1]  CoalesceBatchesExec: target_batch_size=8192
[ output_partitions: 1]    ProjectionExec: expr=[1 as test_col]
[ output_partitions: 1]      PlaceholderRowExec
```

I had trouble finding a Substrait plan in which triggred a physical plan with  `PlaceholderRowExec` instead of `DataSourceExec`. Because of the way the consumer [handles](https://github.com/apache/datafusion/blob/12c40ca9499d23a3459c5f9f7712e9f62b3443a1/datafusion/substrait/src/logical_plan/consumer/rel/read_rel.rs#L116) a virtual table, it’s quite difficult to force the physical plan to include PlaceholderRowExec rather than DataSourceExec. I think we can account for this issue in a separate PR, for now I'm not sure if it should be a fix in Datafusion or in datafusion-distributed.

### DIFF
--- a/src/query_planner.rs
+++ b/src/query_planner.rs
@@ -1,11 +1,12 @@
 use std::sync::Arc;
 
 use anyhow::anyhow;
-use arrow::datatypes::SchemaRef;
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use datafusion::{
     logical_expr::LogicalPlan, physical_plan::ExecutionPlan, prelude::SessionContext,
 };
 use datafusion_substrait::logical_plan::consumer::from_substrait_plan;
+use datafusion_substrait::substrait::proto::Plan;
 
 use crate::{
     explain::{is_explain_query, DistributedExplainExec},
@@ -131,10 +132,7 @@ impl QueryPlanner {
     }
 
     /// Prepare a distributed query (Substrait entry point)
-    pub async fn prepare_substrait_query(
-        &self,
-        substrait_plan: datafusion_substrait::substrait::proto::Plan,
-    ) -> Result<QueryPlan> {
+    pub async fn prepare_substrait_query(&self, substrait_plan: Plan) -> Result<QueryPlan> {
         let base_result = self
             .prepare_substrait_query_base(substrait_plan, "SUBSTRAIT")
             .await?;
@@ -143,7 +141,7 @@ impl QueryPlanner {
 
     pub async fn prepare_substrait_query_base(
         &self,
-        substrait_plan: datafusion_substrait::substrait::proto::Plan,
+        substrait_plan: Plan,
         query_type: &str,
     ) -> Result<QueryPlanBase> {
         debug!(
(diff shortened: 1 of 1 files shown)


src/flight.rs
```diff
@@ -134,6 +134,12 @@ pub trait FlightSqlHandler: Send + Sync {
         request: Request<FlightDescriptor>,
     ) -> Result<Response<FlightInfo>, Status>;
 
+    async fn get_flight_info_substrait(
```
NGA-TRAN: Should we also name this `get_flight_info_substrait_plan` got it to match with the name in `FlightSqlService`?

src/proxy_service.rs
```diff
@@ -74,7 +76,7 @@ impl FlightSqlHandler for DfRayProxyHandler {
 
     async fn do_get_statement(
         &self,
-        ticket: arrow_flight::sql::TicketStatementQuery,
+        ticket: TicketStatementQuery,
```
NGA-TRAN, line 79: 👍

src/query_planner.rs
```diff
@@ -65,6 +66,37 @@ impl QueryPlanner {
...
+        .await?;
+
+        Ok(QueryPlan {
+            query_id: base_result.query_id,
+            worker_addresses: final_workers,
+            final_stage_id,
+            schema,
+            explain_data: None,
+        })
+    }
```
NGA-TRAN, line 99: Nice refactor

src/query_planner.rs
```diff
@@ -257,6 +309,41 @@ mod tests {
...
+                "Projection: Int64(1) AS test_col\n  Values: (Int64(0))"
+            );
+            // physical plan of select 1 on empty releation is ProjectionExec
+            assert_eq!(query_plan_base.physical_plan.name(), "ProjectionExec");
+        } else {
+            // If worker discovery fails, we expect a specific error
+            let error_msg = format!("{:?}", result.unwrap_err());
+            assert!(error_msg.contains("worker") || error_msg.contains("address"));
+        }
+    }
```
NGA-TRAN, line 342: Nice test

PR #43, opened 2025-07-11
remove old naming scheme and use DD where appropriate

robtandy:
Removing any of the old naming schemes.

`TODO` Update the readme as well, we should write it by hand, so it can be saved for a subsequent PR when we convert this to have an entry via a lib as well as an arrow flight sql proxy

Edit: Now that we no longer have Ray, we don't have collisions with the concept of Worker, so DDProcessorService has been renamed DDWorkerService.

### DIFF
--- a/src/worker_service.rs
+++ b/src/worker_service.rs
@@ -1,0 +1,696 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


k8s/cluster.yaml
```diff
@@ -52,16 +52,14 @@ spec:
...
-              value: trace
-            - name: DFRAY_WORKER_DEPLOYMENT
-              value: deployment-dfray-worker
-            - name: DFRAY_WORKER_DEPLOYMENT_PORT
+            - name: dd_WORKER_DEPLOYMENT
+              value: deployment-dd-worker
+            - name: dd_WORKER_DEPLOYMENT_PORT
               value: "20201"
-            - name: DFRAY_WORKER_DEPLOYMENT_NAMESPACE
+            - name: dd_WORKER_DEPLOYMENT_NAMESPACE
```
NGA-TRAN: So you mean we should keep `dd` lowercase to stay consistent?

robtandy: Oh, let me fix that.

PR #46, opened 2025-07-15
Robtandy/extension point

robtandy:
This PR attempts to make a clean integration point for this project and customizations.

Such customizations might be:
- adding a new file format
- adding any new PhysicalPlan operator

Such customizations might require the ability to transmit the new physical plan over the wire, so we need to accept a Codec as well for this.

Now `DDProxyService` and `DDWorkerService` accept a `Customizer` trait with three methods to implement to handle this.

### DIFF
--- a/src/customizer.rs
+++ b/src/customizer.rs
@@ -4,10 +4,10 @@ use datafusion_proto::physical_plan::PhysicalExtensionCodec;
 #[async_trait::async_trait]
 pub trait Customizer: PhysicalExtensionCodec + Send + Sync {
     /// Customize the context before planning a query.
-    /// This may include registering new file formats or introducing additional 
-    /// `PhysicalPlan` operators. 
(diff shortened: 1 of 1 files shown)


src/customizer.rs
```diff
@@ -0,0 +1,8 @@
+use datafusion::prelude::SessionContext;
+use datafusion_proto::physical_plan::PhysicalExtensionCodec;
+
+#[async_trait::async_trait]
+pub trait Customizer: PhysicalExtensionCodec + Send + Sync {
+    /// Customize the context before planning a a query.
```
NGA-TRAN:
```suggestion
    /// Customize the context before planning a query.
    /// This may include registering new file formats or introducing additional 
    /// `PhysicalPlan` operators. 
    ///
    /// To support serialization of customized plans for distributed execution, 
    /// a `Codec` may also be required.
```

PR #48, opened 2025-07-18
Refactor QueryPlanner

LiaCastaneda:
Refactors `QueryPlanner` to avoid distributing tasks in `prepare`. All the logic for resolving and sending the plan to workers is now handled within the `distribute_plan` function. This decouples the logic and makes it easier to test.

Tested the refactor by launching the TPCH cluster and doing `explain`, `select` & `describe` statements

### DIFF
--- a/src/query_planner.rs
+++ b/src/query_planner.rs
@@ -119,7 +119,13 @@ impl QueryPlanner {
     /// distributed plan, and distributed stages, but it does not yet contain
     /// worker addresses or tasks, as they are filled in later by `distribute_plan()`.
     pub async fn prepare_substrait(&self, substrait_plan: Plan) -> Result<QueryPlan> {
-        let ctx = get_ctx().map_err(|e| anyhow!("Could not create context: {e}"))?;
+        let mut ctx = get_ctx().map_err(|e| anyhow!("Could not create context: {e}"))?;
+        if let Some(customizer) = &self.customizer {
+            customizer
+                .customize(&mut ctx)
+                .await
+                .map_err(|e| anyhow!("Customization failed: {e:#?}"))?;
+        }
 
         let logical_plan = from_substrait_plan(&ctx.state(), &substrait_plan).await?;
 
@@ -257,7 +263,7 @@ impl QueryPlanner {
         })
     }
 
-    /// Performs worker discovery, and distributes the query plan to workers.
+    /// Performs worker discovery, and distributes the query plan to workers,
     /// also sets the final worker addresses and distributed tasks in the query plan.
     pub async fn distribute_plan(&self, initial_plan: &mut QueryPlan) -> Result<()> {
         // Perform worker discovery
@@ -267,7 +273,7 @@ impl QueryPlanner {
         // into chunks of partitions (partition_groups)
         let (final_workers, tasks) = distribute_stages(
             &initial_plan.query_id,
-            initial_plan.distributed_stages.clone(),
+            &initial_plan.distributed_stages,
             worker_addrs,
             self.codec.as_ref(),
         )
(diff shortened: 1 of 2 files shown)


src/query_planner.rs
```diff
@@ -176,56 +209,62 @@ impl QueryPlanner {
...
+    /// also sets the final worker addresses and distributed tasks in the query plan.
+    pub async fn distribute_plan(&self, initial_plan: &mut QueryPlan) -> Result<()> {
+        // Perform worker discovery
+        let worker_addrs = get_worker_addresses()?;
+
+        // Distribute the stages to workers, further dividing them up
+        // into chunks of partitions (partition_groups)
+        let (final_workers, tasks) = distribute_stages(
+            &initial_plan.query_id,
+            initial_plan.distributed_stages.clone(),
```
LiaCastaneda: I couldn't find a way to not clone this :(

src/query_planner.rs
```diff
@@ -84,8 +86,10 @@ impl QueryPlanner {
 
     /// Common planning steps shared by both query and its EXPLAIN
     ///
-    /// Prepare a query by parsing the SQL, planning it, and distributing the
-    /// physical plan into stages that can be executed by workers.
+    /// Prepare a query by parsing the SQL, performing logical planning,
+    /// and then physical planning to create a distributed plan.
+    /// Builds an initial `QueryPlan` with the logical and physical plans,
+    /// but without worker addresses and tasks, which will be set later in distribute_plan().
```
NGA-TRAN:
```suggestion
    /// Prepare a query by parsing the SQL, performing logical planning,
    /// and then physical planning to create a distributed plan.
    /// Builds an initial `QueryPlan` with the logical plan, physical plan, distributed plan and distributed stages
    /// but without worker addresses and tasks, which will be set later in distribute_plan().
```

src/query_planner.rs
```diff
@@ -151,7 +170,20 @@ impl QueryPlanner {
...
+        // build the initial plan, without the worker addresses and DDTasks
+        return self
+            .build_initial_plan(
+                distributed_plan,
+                distributed_stages,
+                ctx,
+                logical_plan,
+                physical_plan,
+            )
+            .await;
```
NGA-TRAN, line 203: Can you also add an comment explaining what the function `prepare_local` does?

src/query_planner.rs
NGA-TRAN: I think if you modify the `distribute_stages` function making it receive `stages` as a reference, you won't need to clone it here. From skimming that function, it seems not require `stages` as value but I may miss something.

src/query_planner.rs
LiaCastaneda: I will modify the description as it is not very clear

src/query_planner.rs
```diff
@@ -105,6 +109,15 @@ impl QueryPlanner {
...
+    ///
+    /// 1. Convert the incoming Substrait plan into a `LogicalPlan` with DataFusion’s
+    ///    default Substrait consumer.
+    /// 2. Derive the corresponding physical plan and distributed variant.
+    ///
+    /// The resulting `QueryPlan` contains the logical plan, physical plan,
+    /// distributed plan, and distributed stages, but it does not yet contain
+    /// worker addresses or tasks, as they are filled in later by `distribute_plan()`.
     pub async fn prepare_substrait(&self, substrait_plan: Plan) -> Result<QueryPlan> {
         let ctx = get_ctx().map_err(|e| anyhow!("Could not create context: {e}"))?;
```
robtandy, line 129: Can you also add the line to customize the ctx here? I forgot it when I added that capability?

src/query_planner.rs
robtandy: Agree.  distribute_stages can probably take a `&[DDStage]`

LiaCastaneda: thanks!

PR #49, opened 2025-07-18
Add serialization and deserialization tests for all tpc-ch queries

NGA-TRAN:
## Changes in this PR:

1. Use recent commit in DF to get the fix for https://github.com/apache/datafusion/issues/16665
2. Add serialization and deserialization tests on physical plans because they are so important in our distributed system

**Note:** I added small tpc-h parquet files (20 rows each file) to ensure we simulate exact schema. The data will be good for testing correctness that I plan to do next, too


tests/round_trip_physical_plan.rs
```diff
@@ -0,0 +1,142 @@
...
+        .split(';')
+        .map(|s| s.trim())
+        .filter(|s| !s.is_empty())
+        .map(|s| s.to_string())
+        .collect())
+}
+
+/// Test that we are able to build the physical plan and its version after serialization and deserialization
+#[tokio::test]
+async fn test_serialize_deserialize_tpch_queries() -> Result<()> {
```
NGA-TRAN, line 96: This test will fail without recent commit from DF

PR #50, opened 2025-07-21
Add planning tests

LiaCastaneda:
Adds some tests for distributed planning.


src/query_planner.rs
```diff
@@ -286,3 +287,127 @@ impl QueryPlanner {
...
+        )]));
+        assert_eq!(qp.distributed_plan.schema(), expected_schema);
+
+        // Check the distributed physical plan.
+        let distributed_plan_str =
+            format!("{}", displayable(qp.distributed_plan.as_ref()).indent(true));
+        assert_snapshot!(distributed_plan_str, @r"
+        DDStageExec[0] (output_partitioning=UnknownPartitioning(1))
+          ProjectionExec: expr=[1 as test_col]
+            DataSourceExec: partitions=1, partition_sizes=[1]
```
LiaCastaneda, line 322: I noticed that even though we set `datafusion.execution.target_partitions = 3` in `make_state(),` the plan still shows partitions = 1 in all cases. If this changes, we could force 1 partition explicitly for tests in the future.

NGA-TRAN: Maybe because there are not enough data for the plan to make make 3 partitions? This is actually anther good case for us to test. I think we can either create a mock test and data to enforce the number of partitions or use larger data set.  I think the mock test is usually the choice to avoid adding more data

NGA-TRAN: Also, I think we will start having a lot of things to do soon If you cannot work on this test next, it is good to create a ticket to remind us to work on it

Cargo.toml
```diff
@@ -55,6 +55,8 @@ kube = { version = "1.1", features = ["derive", "runtime"] }
 log = "0.4"
 rand = "0.8"
 uuid = { version = "1.6", features = ["v4"] }
+serde_json = "1.0"
+insta = { version = "1.43.1"}
```
NGA-TRAN, line 59: ❤️

src/query_planner.rs
LiaCastaneda: I was planning on looking into integration tests with mocking after merging the planning tests, unless there was something more urgent to do. I'll create a ticket to track it still

PR #52, opened 2025-07-22
Fix compile and clippy warnings

NGA-TRAN:
To avoid `compile` and `clippy` warnings, I need to changes these;

1. Remove unnecessary reference `&` from `stages`
2. Add `#[cfg(test)]` to import test-only stuffs
3. Add `#[allow(dead_code)]` for `customizer`
4. Make `Status`  stack `Box<Status.` heap.
5. Add ` #[allow(clippy::result_large_err)]` where we do not want t(or cannot) o make `Status` `Box<Status>`

### DIFF
--- a/src/planning.rs
+++ b/src/planning.rs
@@ -460,7 +460,7 @@ pub async fn distribute_stages(
 
         // all stages to workers
         let (task_datas, final_addrs) =
-            assign_to_workers(query_id, &stages, workers.values().collect(), codec)?;
+            assign_to_workers(query_id, stages, workers.values().collect(), codec)?;
 
         // we retry this a few times to ensure that the workers are ready
         // and can accept the stages
--- a/src/proxy_service.rs
+++ b/src/proxy_service.rs
@@ -79,21 +79,25 @@ impl DDProxyHandler {
         }
     }
 
-    pub fn create_flight_info_response(&self, query_plan: QueryPlan) -> Result<FlightInfo, Status> {
+    pub fn create_flight_info_response(
+        &self,
+        query_plan: QueryPlan,
+    ) -> Result<FlightInfo, Box<Status>> {
         let mut flight_info = FlightInfo::new()
             .try_with_schema(&query_plan.schema)
-            .map_err(|e| Status::internal(format!("Could not create flight info {e:?}")))?;
+            .map_err(|e| {
+                Box::new(Status::internal(format!(
+                    "Could not create flight info {e:?}"
+                )))
+            })?;
 
         let ticket_data = TicketStatementData {
             query_id: query_plan.query_id,
             stage_id: query_plan.final_stage_id,
             stage_addrs: Some(query_plan.worker_addresses.into()),
-            schema: Some(
-                query_plan
-                    .schema
-                    .try_into()
-                    .map_err(|e| Status::internal(format!("Could not convert schema {e:?}")))?,
-            ),
+            schema: Some(query_plan.schema.try_into().map_err(|e| {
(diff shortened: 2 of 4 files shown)


src/planning.rs
```diff
@@ -460,7 +460,7 @@ pub async fn distribute_stages(
 
         // all stages to workers
         let (task_datas, final_addrs) =
-            assign_to_workers(query_id, &stages, workers.values().collect(), codec)?;
+            assign_to_workers(query_id, stages, workers.values().collect(), codec)?;
```
NGA-TRAN, line 463: @LiaCastaneda made this `&[DDStage]` in this https://github.com/datafusion-contrib/datafusion-distributed/pull/48/files and would not need another `&` herer

src/proxy_service.rs
```diff
@@ -79,21 +79,25 @@ impl DDProxyHandler {
         }
     }
 
-    pub fn create_flight_info_response(&self, query_plan: QueryPlan) -> Result<FlightInfo, Status> {
+    pub fn create_flight_info_response(
+        &self,
+        query_plan: QueryPlan,
+    ) -> Result<FlightInfo, Box<Status>> {
```
NGA-TRAN, line 85: `Cargo clippy` complains this status is over 176 bytes and we want to put this in heap instead

src/query_planner.rs
```diff
@@ -288,10 +287,12 @@ impl QueryPlanner {
     }
 }
 
-pub mod tests {
+#[cfg(test)]
```
NGA-TRAN, line 290: Add this to avoid compile warnings that we import wrong places

src/worker_service.rs
```diff
@@ -99,6 +99,7 @@ struct DDWorkerHandler {
     done: Arc<Mutex<bool>>,
 
     /// Optional customizer for our context and proto serde
+    #[allow(dead_code)]
```
NGA-TRAN, line 102: I know we use `customizer` else where but get complain here because it is not used. Put the deadcode here to avoid it. 
Also, I wonder if the unit or integration test will help avoid this warning  @LiaCastaneda

src/proxy_service.rs
```diff
@@ -357,6 +362,7 @@ impl DDProxyService {
             handler: self.handler.clone(),
         };
 
+        #[allow(clippy::result_large_err)]
```
NGA-TRAN, line 365: Since this function is passed to external FlightServiceServer::with_interceptor that use Status, we cannot change it to Box. So add clippy here to avoid complains

src/worker_service.rs
```diff
@@ -502,6 +502,7 @@ impl DDWorkerHandler {
 
         Ok(Box::pin(out_stream))
     }
+    #[allow(clippy::result_large_err)]
```
NGA-TRAN, line 505: Ditto. This function implements trait FlightHandler that return Status

src/worker_service.rs
LiaCastaneda: maybe I missed something, but do we need to keep the customizer field in `DDWorkerHandler`? I don't see its being used in `DDWorkerHandler`, except for creating the `PhysicalExtensionCodec` , so in theory any customization will be automatically applied/used when we call `try_decode` on `DDCodec` 🤔

NGA-TRAN: It is used in execute_plan_and_build_stream https://github.com/datafusion-contrib/datafusion-distributed/blob/07e26de9067f17e9acb72d11b74563148e86a04f/src/proxy_service.rs#L130.

I think we need this to talk with different datasources

PR #54, opened 2025-07-24
Build end-to-end test infrastructure and add some tests

NGA-TRAN:
# This PR:

- Enables us to create 3 different types of test clusters:
      1. Two-worker cluster that understands small tpc-h files  stored in ./tpch/data/
      2. Two-worker cluster  with mock data of 2 tables (and we add as many as we want)
      3. Custom number of workers and custom data files
- Adds some tests
      1. Tests for cluster setups
      2. ~~Tests for all tpc-h queries that run on small data set (20 rows each file)~~
      3. Basic tests on mock data

### Best way to review this PR
1. Start with `cluster_setup.rs` that is a util to create cluster for us. Main jobs of this file:
      - Help create 3 different choices of clusters for different testing purposes in one line
      - Help execute a given query in one line
      - Tests in this files testing cluster setup and operation
2. ~~Then `tpch_small.rs` that tests all 22 tpc-h queries on a small set of data~~
3. Then `mock_data.rs` that creates 2 tables and add some data
4. Then `basic_Tests.rs` that shows some example how to test the mock data

# Next step
todo: Will link a ticket for useful tests to add shortly

### DIFF
--- a/tests/common/cluster_setup.rs
+++ b/tests/common/cluster_setup.rs
@@ -41,6 +41,11 @@ const STARTUP_TIMEOUT: Duration = Duration::from_secs(30);
 const CHANNEL_TIMEOUT: Duration = Duration::from_secs(60);
 const QUERY_TIMEOUT: Duration = Duration::from_secs(120);
 const CONNECTION_CLEANUP_TIMEOUT: Duration = Duration::from_millis(100);
+const PROCESS_CLEANUP_WAIT: Duration = Duration::from_secs(1);
+const WORKER_STARTUP_WAIT: Duration = Duration::from_secs(5);
+const READY_POLL_INTERVAL: Duration = Duration::from_millis(500);
+const WORKER_DISCOVERY_WAIT: Duration = Duration::from_secs(3);
+const SHUTDOWN_WAIT: Duration = Duration::from_secs(2);
 
 /// Supported table formats for the cluster
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -382,7 +387,7 @@ impl TestCluster {
         }
 
         // Give processes time to die
-        thread::sleep(Duration::from_secs(1));
+        thread::sleep(PROCESS_CLEANUP_WAIT);
         Ok(())
     }
 
@@ -441,7 +446,7 @@ impl TestCluster {
 
         // Give workers time to start and be ready for connections
         println!("  Waiting for workers to be ready...");
-        thread::sleep(Duration::from_secs(5));
+        thread::sleep(WORKER_STARTUP_WAIT);
 
         // Start proxy
         let proxy_port = self.config.proxy_port();
@@ -473,7 +478,7 @@ impl TestCluster {
 
         let proxy = cmd.spawn()?;
         self.proxy_process = Some(proxy);
-        self.is_running.store(true, Ordering::Relaxed);
+        self.is_running.store(true, Ordering::SeqCst);
 
         Ok(())
     }
@@ -500,14 +505,14 @@ impl TestCluster {
                 );
             }
 
-            thread::sleep(Duration::from_millis(500));
+            thread::sleep(READY_POLL_INTERVAL);
         }
 
         Err("Cluster failed to become ready within timeout".into())
     }
 
     async fn is_cluster_ready(&self) -> bool {
-        if !self.is_running.load(Ordering::Relaxed) {
+        if !self.is_running.load(Ordering::SeqCst) {
             return false;
         }
 
@@ -570,7 +575,7 @@ impl TestCluster {
 
         // Give additional time for worker discovery to complete
         println!("  Allowing time for worker discovery...");
-        thread::sleep(Duration::from_secs(3));
+        thread::sleep(WORKER_DISCOVERY_WAIT);
 
         Ok(())
     }
@@ -583,7 +588,7 @@ impl TestCluster {
         &self,
         sql: &str,
     ) -> Result<Vec<RecordBatch>, Box<dyn std::error::Error>> {
-        if !self.is_running.load(Ordering::Relaxed) {
+        if !self.is_running.load(Ordering::SeqCst) {
             return Err("Cluster is not running".into());
         }
 
@@ -600,7 +605,7 @@ impl TestCluster {
         &self,
         sql: &str,
     ) -> Result<Vec<RecordBatch>, Box<dyn std::error::Error>> {
-        if !self.is_running.load(Ordering::Relaxed) {
+        if !self.is_running.load(Ordering::SeqCst) {
             return Err("Cluster is not running".into());
         }
 
@@ -701,13 +706,13 @@ impl TestCluster {
     /// Check if the cluster is running
     #[allow(dead_code)]
     pub fn is_running(&self) -> bool {
-        self.is_running.load(Ordering::Relaxed)
+        self.is_running.load(Ordering::SeqCst)
     }
 }
 
 impl Drop for TestCluster {
     fn drop(&mut self) {
-        if self.is_running.load(Ordering::Relaxed) {
+        if self.is_running.load(Ordering::SeqCst) {
             println!("🧹 Shutting down integration test cluster...");
 
             // Kill proxy
@@ -722,7 +727,7 @@ impl Drop for TestCluster {
                 let _ = worker.wait();
             }
 
-            self.is_running.store(false, Ordering::Relaxed);
+            self.is_running.store(false, Ordering::SeqCst);
             println!("✅ Cluster shutdown complete");
         }
     }
@@ -792,7 +797,7 @@ mod tests {
         drop(cluster);
 
         // Give a moment for processes to die
-        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+        tokio::time::sleep(SHUTDOWN_WAIT).await;
 
         // Verify the port is now available (should fail to connect)
         use std::net::TcpStream;


tests/basic_tests.rs
```diff
@@ -0,0 +1,228 @@
...
+    assert_snapshot!(result, @r"
+    +-------------+---------------+----------+-----------+
+    | customer_id | name          | city     | country   |
+    +-------------+---------------+----------+-----------+
+    | 1           | Alice Johnson | New York | USA       |
+    | 2           | Bob Smith     | London   | UK        |
+    | 3           | Carol Davis   | Paris    | France    |
+    | 4           | David Wilson  | Tokyo    | Japan     |
+    | 5           | Eve Brown     | Sydney   | Australia |
+    +-------------+---------------+----------+-----------+
```
NGA-TRAN: @LiaCastaneda : Can you review the results of these tests? I will double check before the merge but another set of eyes will help.

Also with this Mock and Custom cluster setup, feel free to modify or add new data for you new tests

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+//! async fn test_simple_query() {
+//!     let cluster = TestCluster::start().await.unwrap();
+//!     
+//!     // Execute a simple query
+//!     let batches = cluster.execute_sql("SELECT 1 as test_column").await.unwrap();
+//!     // your assertions here
+//!     
+//!     // Cluster automatically shuts down when dropped
+//! }
+//! ```
```
NGA-TRAN: This file not only creates test clusters but also tests our cluster creation and operation 🙂

tests/tpch_small.rs
```diff
@@ -0,0 +1,278 @@
...
+    +--------------+--------------+---------+----------------+----------------+---------------+-----------+--------------+----------+-------------+
+    ");
+}
+
+#[tokio::test]
+async fn test_tpch_q02() {
+    let result = execute_tpch_query(2).await;
+    assert_snapshot!(result, @r"
+    ++
+    ++
```
NGA-TRAN: In a future PR, I will try to generate small dataset that includes data for these queries for us to get some results back

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+
+/// Configuration for the integration test cluster
+#[derive(Debug, Clone)]
+pub struct ClusterConfig {
+    /// Number of worker nodes
+    pub num_workers: usize,
+    /// Base port for services (proxy will use this, workers will use base_port + 1, base_port + 2, etc.)
+    pub base_port: u16,
+    /// Tables to register in the cluster (name -> (format, path))
+    pub tables: HashMap<String, (String, String)>,
```
jayshrivastava: nit: ambiguous tuple. maybe we should have a format enum. also, in theory, you could add `(table1, parquet, dir1), (table2, csv, dir1)`. Not sure if that's okay or not.

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+    proxy_process: Option<Child>,
+    worker_processes: Vec<Child>,
+    is_running: Arc<AtomicBool>,
+    session_context: SessionContext,
+}
+
+impl TestCluster {
+    /// Start a cluster with default configuration
+    #[allow(dead_code)]
+    pub async fn start() -> Result<Self, Box<dyn std::error::Error>> {
```
jayshrivastava: Should we use DDError in result.rs?

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+                    if let Ok(pid) = line.trim().parse::<u32>() {
+                        let _ = Command::new("kill").args(["-9", &pid.to_string()]).output();
+                        println!("  🔥 Killed process {} on port {}", pid, port);
+                    }
+                }
+            }
+        }
+
+        // Give processes time to die
+        thread::sleep(Duration::from_secs(1));
```
jayshrivastava: A sub issue to improve this to poll the workers until they die would be good 👍🏽

jayshrivastava: Or, it can be done in this PR similar to `is_cluster_ready`

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+            .enumerate()
+            .map(|(i, &port)| format!("worker{}/127.0.0.1:{}", i + 1, port))
+            .collect::<Vec<_>>()
+            .join(",");
+
+        let mut cmd = Command::new(&binary_path);
+        cmd.args(["--mode", "proxy", "--port", &proxy_port.to_string()])
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .env("RUST_LOG", "info,distributed_datafusion=debug")
```
jayshrivastava: This environment vars are a bit confusing. How do they get set?

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+                let _ = proxy.wait();
+            }
+
+            // Kill workers
+            for mut worker in self.worker_processes.drain(..) {
+                let _ = worker.kill();
+                let _ = worker.wait();
+            }
+
+            self.is_running.store(false, Ordering::Relaxed);
```
jayshrivastava: nit: might as well use SeqCst?

tests/basic_tests.rs
```diff
@@ -0,0 +1,228 @@
...
+mod mock_data;
+
+use cluster_setup::{ClusterConfig, TestCluster};
+use mock_data::generate_mock_data;
+
+/// Ensure mock data is generated only once across all tests
+static MOCK_DATA_INIT: Once = Once::new();
+
+/// Global cluster instance shared across basic tests
+static BASIC_CLUSTER: OnceCell<TestCluster> = OnceCell::const_new();
```
fmonjalet: Since these are sub processes, how is this cluster cleaned up in case of panic or harder (sigsegv, sigabrt) fault?

tests/basic_tests.rs
```diff
@@ -0,0 +1,228 @@
...
+    BASIC_CLUSTER
+        .get_or_init(|| async {
+            println!("🚀 Initializing shared basic test cluster...");
+
+            // Ensure mock data exists
+            ensure_mock_data();
+
+            // Configure cluster with mock data tables
+            let config = ClusterConfig::new()
+                .with_base_port(33000) // Use different port range from TPC-H tests
```
fmonjalet: Ideally in the future we have in-process workers and no network to avoid this kind of hassle (using [in-memory channels](https://docs.rs/tokio/latest/tokio/io/struct.DuplexStream.html))

tests/basic_tests.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    println!("✅ Query completed successfully");
+
+    pretty_format_batches(&batches).unwrap().to_string()
+}
+
+#[tokio::test]
+async fn test_select_customers() {
+    let result = execute_basic_query("SELECT * FROM customers ORDER BY customer_id").await;
+    assert_snapshot!(result, @r"
```
fmonjalet: This is a really nice and simple API to write tests!

minor (since it's easy to address at any point in the future): Looking at it, I am even wondering if we should not adhere to sqllogic test format instead, so that we can reuse existing test beds.

tests/cluster_setup.rs
LiaCastaneda: its used [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/07e26de9067f17e9acb72d11b74563148e86a04f/src/lib.rs#L64C5-L64C15) upon [startup](https://github.com/datafusion-contrib/datafusion-distributed/blob/07e26de9067f17e9acb72d11b74563148e86a04f/src/bin/distributed-datafusion.rs#L30)

tests/cluster_setup.rs
```diff
@@ -0,0 +1,655 @@
...
+        self.base_port
+    }
+
+    fn worker_ports(&self) -> Vec<u16> {
+        (1..=self.num_workers)
+            .map(|i| self.base_port + i as u16)
+            .collect()
+    }
+
+    fn tables_env_string(&self) -> String {
```
LiaCastaneda: what about renaming this one to `tables_from_env` as its more intuitive?

tests/basic_tests.rs
LiaCastaneda: i just checked and they are all marked as ok/updated

tests/tpch_small.rs
```diff
@@ -0,0 +1,278 @@
...
+    );
+
+    // Get the shared cluster instance (no mutex - concurrent access allowed!)
+    let cluster = get_cluster().await;
+
+    let sql = read_query(query_num);
+    let batches = cluster
+        .execute_sql(&sql)
+        .await
+        .expect("Failed to execute TPC-H query");
```
LiaCastaneda: I'm not sure I understand how we communicate to the worker cluster during execution, `execute_sql` doesn't use the flight endpoints I think, is that expected from these tests?

tests/cluster_setup.rs
LiaCastaneda: but i agree we should document it somewhere

tests/tpch_small.rs
NGA-TRAN: Good catch, Lia. I need to change `execute_sql` to make it work as expected

src/planning.rs
```diff
@@ -208,6 +212,31 @@ pub async fn add_tables_from_env(state: &mut SessionState) -> Result<()> {
...
+        let ctx = SessionContext::new_with_state(state.clone());
+        ctx.sql(view_sql.trim()).await?;
+
+        // Update the state with the new view
+        *state = ctx.state();
+    }
+
+    Ok(())
+}
+
```
NGA-TRAN: These changes are necessary to support view registration in the same way we currently register tables. They'll remain in place until we establish a proper catalog implementation.

tests/common/mod.rs
```diff
@@ -1,1222 +1,3 @@
-use std::fs;
```
NGA-TRAN: The red lines here are moved to `tpch_validation_helpers.rs` to keep this file short and only for importing modules

tests/tpch_validation_helpers.rs
```diff
@@ -0,0 +1,1222 @@
+use std::fs;
```
NGA-TRAN: The content of this file is  from `mod.rs`. No need to review it. Also, since now we have a way to run queries without going thru Python, soon I will remove this file after I simplify the tpch validation tests

tests/cluster_setup.rs
NGA-TRAN: Done

tests/cluster_setup.rs
NGA-TRAN: I tried but the cluster setup has different errors

tests/cluster_setup.rs
NGA-TRAN: Done

tests/tpch_small.rs
NGA-TRAN: Fixed

tests/cluster_setup.rs
NGA-TRAN: So many needed sleeps so I agree we should have one sub issue to make them better

tests/cluster_setup.rs
NGA-TRAN: I agree we should add `RUST_LOG` into the document. We need a PR to write and update our document

tests/cluster_setup.rs
NGA-TRAN: Done

tests/basic_tests.rs
```diff
@@ -0,0 +1,275 @@
...
+        | Alice Johnson | 3           |
+        | Bob Smith     | 2           |
+        | Carol Davis   | 2           |
+        | David Wilson  | 2           |
+        | Eve Brown     | 1           |
+        +---------------+-------------+
+        ");
+    }
+
+    // TODO: Investigate why only at most 4 or 5 queries can be executed.
```
NGA-TRAN: I will create a ticket for this

tests/common/mock_data.rs
```diff
@@ -0,0 +1,129 @@
...
+    let customers_data = vec![
+        "customer_id,name,city,country",
+        "1,Alice Johnson,New York,USA",
+        "2,Bob Smith,London,UK",
+        "3,Carol Davis,Paris,France",
+        "4,David Wilson,Tokyo,Japan",
+        "5,Eve Brown,Sydney,Australia",
+    ];
+
+    fs::write(path, customers_data.join("\n"))?;
```
fmonjalet: nit: do we need to write this to the FS vs registering this as a `MemTable` directly?

tests/common/test_utils.rs
```diff
@@ -0,0 +1,85 @@
...
+/// use crate::common::test_utils::allocate_port_range;
+///
+/// let base_port = allocate_port_range();
+/// println!("Proxy will use port: {}", base_port);
+/// println!("Workers will use ports: {}-{}", base_port + 1, base_port + 5);
+/// ```
+pub fn allocate_port_range() -> u16 {
+    let mut current_port = CURRENT_MAX_PORT.lock().unwrap();
+    let base_port = *current_port;
+    *current_port += MAX_CLUSTER_SIZE;
```
fmonjalet: This is fine as long as we don't create too many clusters and tests. Do we have tests that run concurrently and compete for ports?

NGA-TRAN: There is a big range of available ports for us to use and the goal is all test in the same file/module should use only one cluster so the range won't grow too much but I agree this can be an issue when we add a lot more tests.

I will open ticket for us to improve this. In general, the test infrastructure for a distributed system is way more complicated than a single-node and will need a lot of work before we feel comfortable

tests/basic_tests.rs
NGA-TRAN: This is a very good point. Let me dig in a bit more how we handle this

tests/basic_tests.rs
NGA-TRAN: Do you mean we use sqllogictest instead of of this insta assert_snapshot? I like  sqllogictest and I think it is a good idea

fmonjalet: Yes exactly this. They have their own quirks so we'll also need the type of test you just wrote in some cases, but there is a huge number of existing tests that we could easily port, which is useful.

Issue #23, opened 2025-06-24
Centralize Configuration Using Global Config with LazyLock or OnceLock


NGA-TRAN:
Currently, configuration settings are scattered across the codebase and surfaced via:

- Direct use of command-line arguments
- Environment variables
- Occasional hardcoded constants (e.g., timeouts, thresholds)

This makes it harder to discover, document, and reason about config behavior across the system.

**Proposed Improvement:**

- Introduce a central Config struct that captures all runtime settings
- Store it in a global LazyLock or OnceLock to allow safe and efficient access across modules
- Populate Config at startup using layered config sources (env, CLI, defaults, etc.)
- Replace ad hoc lookups and hardcoded values with references to Config

This will simplify configuration management, increase code clarity, and make it easier to document and extend in the future.

PR #57, opened 2025-07-27
Refactor partition groups and add tests

jayshrivastava:
This change adds a new abstraction called a `Grouper` in `distribution_strategy.rs`. This is responsible for grouping partitions into partition groups. The new `distribution_strategy` module can be used to store other splitters and, in the future, distribution strategies like round robin etc.

A "partition group" is now a range [starting_partition, ending_partition) instead of a Vec, so we improve space efficiency. In the proto layer, the `DDTask` proto still uses a Vec, so we simply expand the range into a Vec when creating them in `planning.rs::assign_to_workers`.

This also fixes a bug in the isolator where it may not have reported the right number of partitions.

Testing
- unit tests in isolator.rs
- unit tests for `build_replacement` in `planning.rs`
- unit tests for the grouper in `distribution_strategy.rs`

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use arrow::datatypes::Schema;
+use arrow::datatypes::{Field, Schema};
 use datafusion::{
     common::{internal_datafusion_err, internal_err, Result},
     execution::FunctionRegistry,
@@ -243,10 +243,10 @@ mod test {
         isolator::PartitionIsolatorExec, max_rows::MaxRowsExec, stage_reader::DDStageReaderExec,
     };
 
-    fn create_test_schema() -> Arc<arrow::datatypes::Schema> {
-        Arc::new(arrow::datatypes::Schema::new(vec![
-            arrow::datatypes::Field::new("a", DataType::Int32, false),
-            arrow::datatypes::Field::new("b", DataType::Int32, false),
+    fn create_test_schema() -> Arc<Schema> {
+        Arc::new(Schema::new(vec![
+            Field::new("a", DataType::Int32, false),
+            Field::new("b", DataType::Int32, false),
         ]))
     }
 
--- a/src/distribution_strategy.rs
+++ b/src/distribution_strategy.rs
@@ -1,0 +1,129 @@
+// A Grouper is used to group partitions.
+pub trait Grouper {
+    // group groups the number of partitions into a vec of groups.
+    fn group(&self, num_partitions: usize) -> Vec<PartitionGroup>;
+}
+
+// PartitionGroup is a struct that represents a range of partitions from [start, end). This is
+// more space efficient than a vector of u64s.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct PartitionGroup {
+    start: usize,
+    end: usize,
+}
+
+impl PartitionGroup {
+    // new creates a new PartitionGroup containing partitions in the range [start..end).
+    pub fn new(start: usize, end: usize) -> Self {
+        Self { start, end }
+    }
+
+    // start is the first in the range
+    pub fn start(&self) -> usize {
+        self.start
+    }
+
+    // end is the exclusive end partition in the range
+    pub fn end(&self) -> usize {
+        self.end
+    }
+}
+
+// PartitionGrouper groups a number partitions together depending on a partition_group_size.
+// Ex. 10 partitions with a group size of 3 will yield groups [(0..3), (3..6), (6..9), (9)].
+// - A partition_group_size of 0 will panic
+// - Grouping 0 partitions will result an empty vec
+// - It's possible for bad groupings to exist. Ex. if the group size is 99 and there are 100
+//   partitions, then you will get unbalanced partitions [(0..99), (99)]
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct PartitionGrouper {
+    partition_group_size: usize,
+}
+
+impl PartitionGrouper {
+    pub fn new(partition_group_size: usize) -> Self {
+        assert!(
+            partition_group_size > 0,
+            "partition groups cannot be size 0"
+        );
+        PartitionGrouper {
+            partition_group_size,
+        }
+    }
+}
+
+impl Grouper for PartitionGrouper {
+    // group implements the Grouper trait
+    fn group(&self, num_partitions: usize) -> Vec<PartitionGroup> {
+        (0..num_partitions)
+            .step_by(self.partition_group_size)
+            .map(|start| {
+                let end = std::cmp::min(start + self.partition_group_size, num_partitions);
+                PartitionGroup { start, end }
+            })
+            .collect()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_partition_grouper_basic() {
+        let grouper = PartitionGrouper::new(4);
+        let groups = grouper.group(10);
+
+        let expected = vec![
+            PartitionGroup { start: 0, end: 4 },
+            PartitionGroup { start: 4, end: 8 },
+            PartitionGroup { start: 8, end: 10 },
+        ];
+
+        assert_eq!(groups, expected);
+    }
+
+    #[test]
+    fn test_partition_grouper_uneven() {
+        let grouper = PartitionGrouper::new(2);
+        let groups = grouper.group(5);
+
+        let expected = vec![
+            PartitionGroup { start: 0, end: 2 },
+            PartitionGroup { start: 2, end: 4 },
+            PartitionGroup { start: 4, end: 5 },
+        ];
+
+        assert_eq!(groups, expected);
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_invalid_group_size() {
+        PartitionGrouper::new(0);
+    }
+
+    #[test]
+    fn test_num_partitions_smaller_than_group_size() {
+        let g = PartitionGrouper::new(2);
+        let groups = g.group(1);
+        let expected = vec![PartitionGroup { start: 0, end: 1 }];
+        assert_eq!(groups, expected);
+    }
+
+    #[test]
+    fn test_num_partitions_equal_to_group_size() {
+        let g = PartitionGrouper::new(2);
+        let groups = g.group(2);
+        let expected = vec![PartitionGroup { start: 0, end: 2 }];
+        assert_eq!(groups, expected);
+    }
+
+    #[test]
+    fn test_zero_partitions_to_group() {
+        let g = PartitionGrouper::new(2);
+        let groups = g.group(0);
+        let expected = vec![];
+        assert_eq!(groups, expected);
+    }
+}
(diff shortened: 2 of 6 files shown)


src/planning.rs
```diff
@@ -585,22 +596,23 @@ fn assign_to_workers(
...
                 }
             }
 
+            // TODO: It would be more efficient to avoid materializing the partition group below.
             let task_data = DDTask {
                 query_id: query_id.to_string(),
                 stage_id: stage.stage_id,
                 plan_bytes,
-                partition_group: partition_group.to_vec(),
+                partition_group: (partition_group.start() as u64..partition_group.end() as u64).collect(),
```
jayshrivastava: Please let me know if I can change the `DDTask` proto to use a range as well.

src/planning.rs
```diff
@@ -669,38 +681,31 @@ fn get_stage_addrs_from_tasks(target_stage_ids: &[u64], stages: &[DDTask]) -> Re
...
-            .collect(),
-        None => vec![(0..partition_count).collect()],
-    };
+    let partition_groups = splitter.split(partition_count);
 
     if isolate && partition_groups.len() > 1 {
         let new_child = Arc::new(PartitionIsolatorExec::new(
             child.clone(),
-            partitions_per_worker.unwrap(), // we know it is a Some, here.
+            partition_count,
```
jayshrivastava: I think this was a bug before. The isolator seems to want the total number of partitions here.

src/codec.rs
```diff
@@ -305,7 +305,7 @@ mod test {
         let schema = create_test_schema();
         let part = Partitioning::UnknownPartitioning(2);
         let reader_exec = Arc::new(DDStageReaderExec::try_new(part, schema, 1).unwrap());
-        let exec = Arc::new(PartitionIsolatorExec::new(reader_exec, 4));
+        let exec = Arc::new(PartitionIsolatorExec::new(reader_exec, 2));
```
NGA-TRAN: We probably don’t need to change the value from 4 to 2 in this test, since having just 2 input partitions is valid. Given our setup—where each worker can handle up to 4 partitions—those 2 inputs will be sent to and processed by a single worker.

On a related note, this raises the idea of adding tests that vary the number of input and output partitions. It might be worth establishing a best practice of opening a ticket whenever we identify follow-up work. Let me know if you'd like me to create one for this, or if you'd prefer to take the lead.

src/isolator.rs
```diff
@@ -73,9 +73,9 @@ impl ExecutionPlan for PartitionIsolatorExec {
     }
 
     fn with_new_children(
-        self: std::sync::Arc<Self>,
-        children: Vec<std::sync::Arc<dyn ExecutionPlan>>,
-    ) -> Result<std::sync::Arc<dyn ExecutionPlan>> {
+        self: Arc<Self>,
+        children: Vec<Arc<dyn ExecutionPlan>>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
```
NGA-TRAN, line 84: ❤️ Thanks for simplifying the code

src/distribution_strategy.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.start
+    }
+
+    // end is the exclusive end partition in the range
+    pub fn end(&self) -> usize {
+        self.end
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
```
NGA-TRAN, line 38: Can you also add comment explaining the purpose of this struct? It is clear from this PR what it is for but, in the future, if readers only come to this struct, they won't have context and the comment will help

src/distribution_strategy.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn test_partition_grouper_uneven() {
+        let grouper = PartitionGrouper::new(2);
+        let groups = grouper.split(5);
+
+        let expected = vec![
+            PartitionGroup { start: 0, end: 2 },
+            PartitionGroup { start: 2, end: 4 },
+            PartitionGroup { start: 4, end: 5 },
+        ];
+
```
NGA-TRAN, line 96: Can you add these tests:
1. `PartitionGrouper::new(2)` and `grouper.split(1);`. Idea: number of input partitions < number of partitions a work is set up to handle
2. `PartitionGrouper::new(2)` and `grouper.split(2);`. Idea: they are the same
3. `PartitionGrouper::new(2)` and `grouper.split(0);`. Idea: no input partitions or empty partition. Check the code to verify if empty partition is accepted. IIRC, this is a valid case and we need to see how we handle this in distributed plan

src/planning.rs
```diff
@@ -240,10 +241,10 @@ pub async fn physical_planning(
 }
 
 /// Returns distributed plan and execution stages for both query execution and EXPLAIN display
-pub async fn distributed_physical_planning(
+pub async fn distributed_physical_planning<S: Grouper + Copy>(
     physical_plan: Arc<dyn ExecutionPlan>,
+    splitter: S,
```
NGA-TRAN: I suggest to make the name more explicit. Maybe `partition_splitter`? or even more meaningful. Can you also add comment to explain that this splitter is for?

src/query_planner.rs
```diff
@@ -148,18 +149,18 @@ impl QueryPlanner {
 
         // construct the distributed physical plan and stages
         let (distributed_plan, distributed_stages) =
-            distributed_physical_planning(physical_plan.clone(), 8192, Some(2)).await?;
+            distributed_physical_planning(physical_plan.clone(), PartitionGrouper::new(2), 8192)
```
NGA-TRAN, line 152: **FYI:** We have this ticket to remind us to work on global config https://github.com/datafusion-contrib/datafusion-distributed/issues/23. These number 2 and 8192 should be a part of the config work which can be either in the same ticket or a separate and new ticket. If you think these should be in need a separate ticket, please feel free to create one or let me know, I am happy to do it top remind us to do so

src/isolator.rs
```diff
@@ -143,3 +143,73 @@ impl ExecutionPlan for PartitionIsolatorExec {
...
+        }
+
+        let task_context = ctx.task_ctx();
+
+        // Success case: execute valid partition
+        let result = isolator.execute(0, task_context.clone());
+        assert!(result.is_ok());
+
+        // Error case: try to execute partition beyond partition_count
+        let result = isolator.execute(4, task_context.clone());
```
NGA-TRAN, line 207: 👍

src/isolator.rs
```diff
@@ -143,3 +143,73 @@ impl ExecutionPlan for PartitionIsolatorExec {
...
+        let empty_ctx = SessionContext::new();
+        let empty_task_context = empty_ctx.task_ctx();
+
+        let result = isolator.execute(0, empty_task_context);
+        assert!(result.is_err());
+        if let Err(e) = result {
+            assert!(e
+                .to_string()
+                .contains("PartitionGroup not set in session config"));
+        }
```
NGA-TRAN, line 269: Excellent test. Can you also add `let result = isolator.execute(1, empty_task_context);` right here?

src/isolator.rs
```diff
@@ -143,3 +143,73 @@ impl ExecutionPlan for PartitionIsolatorExec {
...
+    }
+}
```
NGA-TRAN, line 271: Can you add a few more tests with different value of `partition_count` and `partition_group`. Make sure enough combination with different values of them. You have tested their count `equal`. Adding `less` and `greater` will help capture bugs. Think about all kinds of edge cases

src/planning.rs
```diff
@@ -309,7 +310,7 @@ pub async fn distributed_physical_planning(
             // left side of the join and is not suitable to be executed in a
             // partitioned manner.
             let mut replacement = plan.clone();
-            let partition_count = plan.output_partitioning().partition_count() as u64;
+            let partition_count = plan.output_partitioning().partition_count();
```
NGA-TRAN, line 313:
```suggestion
            let partition_group_size = 1; // NestedLoopJoinExec must be on a stage by itself
            let partition_count = plan.output_partitioning().partition_count();
```

src/planning.rs
```diff
@@ -318,15 +319,15 @@ pub async fn distributed_physical_planning(
                 batch_size,
             )) as Arc<dyn ExecutionPlan>;
 
-            partition_groups = vec![(0..partition_count).collect()];
+            partition_groups = PartitionGrouper::new(1).split(partition_count);
```
NGA-TRAN:
```suggestion
            partition_groups = PartitionGrouper::new(partition_group_size).split(partition_count);
```

src/planning.rs
```diff
@@ -318,15 +319,15 @@ pub async fn distributed_physical_planning(
...
             full_partitions = true;
             Ok(Transformed::yes(replacement))
         } else {
             trace!("not special case partition_groups: {:?}", partition_groups);
 
-            let partition_count = plan.output_partitioning().partition_count() as u64;
+            let partition_count = plan.output_partitioning().partition_count();
             // set this back to default
-            partition_groups = vec![(0..partition_count).collect()];
+            partition_groups = PartitionGrouper::new(1).split(partition_count);
```
NGA-TRAN: Ditto: assign a value as above next to the comment to say why it is 1

src/planning.rs
```diff
@@ -585,22 +586,24 @@ fn assign_to_workers(
...
                         .entry(stage.stage_id)
                         .or_default()
-                        .entry(*part)
+                        .entry(part as u64)
                         .or_default()
                         .push(host.clone());
                 }
             }
 
+            // TODO: It would be more efficient to avoid materializing the partition group below.
```
NGA-TRAN: 👍

src/planning.rs
```diff
@@ -669,39 +672,29 @@ fn get_stage_addrs_from_tasks(target_stage_ids: &[u64], stages: &[DDTask]) -> Re
 }
 
 #[allow(clippy::type_complexity)]
-fn build_replacement(
+fn build_replacement<S: Grouper>(
     plan: Arc<dyn ExecutionPlan>,
-    partitions_per_worker: Option<usize>,
+    splitter: S,
```
NGA-TRAN: Same as above. Make the name more explicit

src/planning.rs
```diff
@@ -713,3 +706,75 @@ fn build_replacement(
...
+        let repartition = coalesce.children()[0];
+        let isolator = repartition.children()[0]
+            .as_any()
+            .downcast_ref::<PartitionIsolatorExec>()
+            .unwrap();
+        assert_eq!(
+            isolator.partition_count,
+            plan.properties().partitioning.partition_count()
+        )
+    }
```
NGA-TRAN, line 779: I am very interested in a lot more tests here.  Either you can add int his PR or in a different/follow-on PR. Here is some ideas:

1. Add great combination of inputs to the function `build_replacement` so we know they work well. Try edge cases
2. Make some mock data test if needed so instead of 1000 max row count or 8192 batch size, make them a lot smaller to control number of partitions out we want to test

src/query_planner.rs
jayshrivastava: Created a new sub-issue on #23

PR #69, opened 2025-08-01
Adds error serialization-deserialization

gabotechs:
Allows propagating `DataFusionError`s over the wire so that upstream stages can read the original error without loosing information.

This is done by providing some mirroring protobuf messages for all the current variants of `DataFusionError`s. As these errors might also hold references to other errors from other libraries, also those other error message get a mirroring protobuf message.

This PR ships a bunch of boilerplate code, but the actual important parts and the only ones that are exposed to the rest of the project are just two functions:

```rust
/// Encodes a [DataFusionError] into a [tonic::Status] error. The produced error is suitable
/// to be sent over the wire and decoded by the receiving end, recovering the original
/// [DataFusionError] across a network boundary with [tonic_status_to_datafusion_error].
pub fn datafusion_error_to_tonic_status(err: &DataFusionError) -> tonic::Status;

/// Decodes a [DataFusionError] from a [tonic::Status] error. If the provided [tonic::Status]
/// error was produced with [datafusion_error_to_tonic_status], this function will be able to
/// recover it even across a network boundary.
///
/// The provided [tonic::Status] error might also be something else, like an actual network
/// failure. This function returns `None` for those cases.
pub fn tonic_status_to_datafusion_error(status: &tonic::Status) -> Option<DataFusionError>;
```

Everything else is hidden in the `src/errors/` modules

---

The PR ships both unit tests for all the error variations and an integration tests that checks that errors are properly propagated through the networks and can be recovered on the receiving end

### DIFF
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -11,6 +11,9 @@ mod parquet_error;
 mod parser_error;
 mod schema_error;
 
+/// Encodes a [DataFusionError] into a [tonic::Status] error. The produced error is suitable
+/// to be sent over the wire and decoded by the receiving end, recovering the original
+/// [DataFusionError] across a network boundary with [tonic_status_to_datafusion_error].
 pub fn datafusion_error_to_tonic_status(err: &DataFusionError) -> tonic::Status {
     let err = DataFusionErrorProto::from_datafusion_error(err);
     let err = err.encode_to_vec();
@@ -18,6 +21,12 @@ pub fn datafusion_error_to_tonic_status(err: &DataFusionError) -> tonic::Status 
     status
 }
 
+/// Decodes a [DataFusionError] from a [tonic::Status] error. If the provided [tonic::Status]
+/// error was produced with [datafusion_error_to_tonic_status], this function will be able to
+/// recover it even across a network boundary.
+///
+/// The provided [tonic::Status] error might also be something else, like an actual network
+/// failure. This function returns `None` for those cases.
 pub fn tonic_status_to_datafusion_error(status: &tonic::Status) -> Option<DataFusionError> {
     if status.code() != tonic::Code::Internal {
         return None;


src/errors/arrow_error.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ArrowErrorProto {
+    #[prost(string, optional, tag = "1")]
+    pub ctx: Option<String>,
+    #[prost(
+        oneof = "ArrowErrorInnerProto",
+        tags = "2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19"
+    )]
+    pub inner: Option<ArrowErrorInnerProto>,
+}
```
NGA-TRAN, line 13: Question: If a new error number 20 is added in the future, can we simply add 20 here without worrying about backward compatibility?
Also this means whenever we upgrade datafusion and there are new arrow (datafusion, parquet, ...) errors, we have to add them here (or in the right error file) right?

src/errors/arrow_error.rs
```diff
@@ -0,0 +1,256 @@
...
+            let proto_no_ctx =
+                ArrowErrorProto::decode(proto_no_ctx.encode_to_vec().as_ref()).unwrap();
+            let (recovered_error_no_ctx, recovered_ctx_no_ctx) = proto_no_ctx.to_arrow_error();
+
+            assert_eq!(
+                original_error.to_string(),
+                recovered_error_no_ctx.to_string()
+            );
+            assert_eq!(recovered_ctx_no_ctx, None);
+        }
```
NGA-TRAN, line 244: Like these tests. Simple and show how to use them in our future code

src/errors/datafusion_error.rs
```diff
@@ -0,0 +1,404 @@
...
+    Context(DataFusionContextErrorProto),
+    #[prost(string, tag = "16")]
+    Substrait(String),
+    #[prost(message, boxed, tag = "17")]
+    Diagnostic(Box<DataFusionErrorProto>),
+    #[prost(message, tag = "18")]
+    Collection(DataFusionCollectionErrorProto),
+    #[prost(message, boxed, tag = "19")]
+    Shared(Box<DataFusionErrorProto>),
+}
```
NGA-TRAN, line 61: It is very nice that you make and put all Arrow, Parquet, Schema, Context, SQL parser,... error protos in here so we only worry about serialize and deserialize one type.

src/flight_service/do_get.rs
```diff
@@ -140,22 +141,24 @@ impl ArrowFlightEndpoint {
...
         let flight_data_stream = FlightDataEncoderBuilder::new()
             .with_schema(stream_partitioner.schema())
-            .build(stream.map_err(|err| FlightError::ExternalError(Box::new(err))));
-
-        Ok(Response::new(Box::pin(flight_data_stream.map_err(|err| {
-            Status::internal(format!("Error during flight stream: {err}"))
-        }))))
+            .build(stream.map_err(|err| {
+                FlightError::Tonic(Box::new(datafusion_error_to_tonic_status(&err)))
+            }));
```
NGA-TRAN, line 154: Very nit and easy to serialize error

src/plan/arrow_flight_read.rs
```diff
@@ -189,14 +190,18 @@ impl ExecutionPlan for ArrowFlightReadExec {
...
                 .into_inner()
                 .map_err(|err| FlightError::Tonic(Box::new(err)));
 
             Ok(FlightRecordBatchStream::new_from_flight_data(stream)
-                // TODO: propagate the error from the service to here, probably serializing it
-                //  somehow.
-                .map_err(|err| DataFusionError::External(Box::new(err))))
+                .map_err(|err| match err {
+                    FlightError::Tonic(status) => tonic_status_to_datafusion_error(&status)
+                        .unwrap_or_else(|| DataFusionError::External(Box::new(status))),
```
NGA-TRAN, line 202: And also very nice and easy to get error back

tests/error_propagation.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let stream = execute_stream(plan, ctx.task_ctx())?;
+
+        let Err(err) = stream.try_collect::<Vec<_>>().await else {
+            panic!("Should have failed")
+        };
+        assert_eq!(
+            DataFusionError::Execution("something failed".to_string()).to_string(),
+            err.to_string()
+        );
```
NGA-TRAN, line 66: I am amazed reading this test. We do not have to convert anything here. You already deserialize and covert the error in `arrow_flight_read.rs`

src/errors/arrow_error.rs
gabotechs: That's right. The only thing we cannot do is to tweak the order of the messages, or deleting numbers, but adding new numbers (and therefore error variants), is backwards compatible.

If there's a DataFusion upgrade and a new error variant is added, we'll need to come here and add it ourselves, otherwise the Rust compiler will tell us there are some error variants not handled.

src/errors/datafusion_error.rs
gabotechs: Yeah, it was quite a bit of work, but it's something that's probably going to need close 0 maintenance, and it's going to help us a lot in knowing what goes wrong.

I expect initially to see more errors than actual queries running.

tests/error_propagation.rs
gabotechs: Yeah, I expect these integration tests to be key not only for ensuring our changes remain healthy, but also for development and debugging.

If they feel like magic it means that we are on the right track.

src/errors/arrow_error.rs
NGA-TRAN: It will be perfect if we get compile error

PR #70, opened 2025-08-01
delegation: add gc task to clean up abandoned contexts

jayshrivastava:
This change addresses a TODO in the `StageDelegation` struct. Previously, it was possible for `StageContext`s to leak if a context was produced and not consumed (or if a waiter created a map entry and timed out without any corresponding producer call). This change introduces a GC task to clear contexts from the map if they have been there for longer than a certain TTL. The TTL proposed is 2 * `wait_timeout`.

When the `StageDelegation` is dropped, the GC task is terminated.

Testing
- Adds two unit tests (one for `add_delegate_info` and one for `wait_for_delegate_info`) which test that abandoned map entries are garbage collected appropriately.

### DIFF
--- a/src/stage_delegation/delegation.rs
+++ b/src/stage_delegation/delegation.rs
@@ -2,8 +2,13 @@ use super::StageContext;
 use dashmap::{DashMap, Entry};
 use datafusion::common::{exec_datafusion_err, exec_err};
 use datafusion::error::DataFusionError;
-use std::time::Duration;
+use std::ops::Add;
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tokio::sync::oneshot;
+use tokio::sync::Notify;
+use tokio::time;
+use tokio::time::Instant;
 
 /// In each stage of the distributed plan, there will be N workers. All these workers
 /// need to coordinate to pull data from the next stage, which will contain M workers.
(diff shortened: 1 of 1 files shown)


src/stage_delegation/delegation.rs
```diff
@@ -30,20 +35,70 @@ use tokio::sync::oneshot;
...
+/// It's possible for [StageContext] to "get lost" if `add_delegate_info` is called without
+/// a corresponding call to `wait_for_delegate_info` or vice versa. In this case, a task will
+/// reap any contexts that live for longer than the `gc_ttl`.
 pub struct StageDelegation {
-    stage_targets: DashMap<(String, usize), Oneof>,
+    stage_targets: Arc<DashMap<(String, usize), Value>>,
     wait_timeout: Duration,
+
+    /// notify is used to shut down the garbage collection task when the StageDelegation is dropped.
+    notify: Arc<Notify>,
```
gabotechs, line 47: There should be no need for a `Notify` here, there's already tooling available in the tokio and DataFusion ecosystem for cancelling tasks on Drop, for example you could simply do

```rust
pub struct StageDelegation {
    stage_targets: Arc<DashMap<(String, usize), Value>>,
    wait_timeout: Duration,
    /// notify is used to shut down the garbage collection task when the StageDelegation is dropped.
    _task: SpawnedTask<()>,
}

impl Default for StageDelegation {
    fn default() -> Self {
        let stage_targets = Arc::new(DashMap::default());
        Self {
            stage_targets: stage_targets.clone(),
            wait_timeout: Duration::from_secs(5),
            _task: SpawnedTask::spawn(run_gc(
                stage_targets.clone(),
                Duration::from_secs(30), /* gc period */
            )),
        }
    }
}

// run_gc will continuously clear expired entries from the map, checking every `period`. The
// function terminates if `shutdown` is signalled.
async fn run_gc(stage_targets: Arc<DashMap<(String, usize), Value>>, period: Duration) {
    loop {
        tokio::time::sleep(period).await;
        stage_targets.retain(|_key, value| value.expiry.gt(&Instant::now()));
    }
}
```

And it should work the same

src/stage_delegation/delegation.rs
```diff
@@ -30,20 +35,70 @@ use tokio::sync::oneshot;
...
+            stage_targets.clone(),
+            notify.clone(),
+            Duration::from_secs(30), /* gc period */
+        ));
+
+        result
+    }
+}
+
+const GC_PERIOD_SECONDS: usize = 30;
```
gabotechs, line 72: Looks like this `const` is unused?

src/stage_delegation/delegation.rs
```diff
@@ -30,20 +35,70 @@ use tokio::sync::oneshot;
...
+    loop {
+        tokio::select! {
+            _ = shutdown.notified() => {
+                break;
+            }
+            _ = tokio::time::sleep(period) => {
+               // Performance: This iterator is sharded, so it won't lock the whole map.
+                stage_targets.retain(|_key, value| {
+                  value.expiry.gt(&Instant::now())
+                  });
```
gabotechs, line 90: `.retain()` iterates the whole map, so it do will end up locking the whole map unfortunately. 

I think there should be ways of just garbage collecting abandoned tasks without iterating the full map, maybe spawning a task with a big `tokio::sleep()` at the beginning for each for each added entry that gets either cancelled due to a drop (the task was not abandoned) or completely executed included the tokio::sleep (the task was abandoned)

PR #75, opened 2025-08-05
Small cleanup after #74

gabotechs:
Some small cleanup after https://github.com/datafusion-contrib/datafusion-distributed/pull/74:
- removes some unused imports
- removes an empty resut.rs file
- uncomments the tests in favor of ignoring them
- moves tcph/mod.rs -> to common/tpch.rs following the Rust book example for integration tests

### DIFF
--- a/src/stage/proto.rs
+++ b/src/stage/proto.rs
@@ -99,9 +99,7 @@ pub fn stage_from_proto(
 }
 
 // add tests for round trip to and from a proto message for ExecutionStage
-/* TODO: broken for now
 #[cfg(test)]
-
 mod tests {
     use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


src/common/mod.rs
gabotechs, line 1: This file seems to be empty, I just deleted it

tests/common/plan.rs
gabotechs, line 1: As instead of commenting the tests I have `#[ignore]`-ed them, this file was still needed, otherwise a compilation error will happen.

tests/common/tpch.rs
gabotechs, line 1: Following the Rust book recommendation https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html, I moved all the common modules that are meant to be used by the integration tests to `common/`

tests/custom_extension_codec.rs
```diff
@@ -1,6 +1,6 @@
 #[allow(dead_code)]
 mod common;
-/*
+
```
gabotechs, line 3: Instead of having big chunks of commented code, a more Rusty way of handling this is to just #[ignore] the tests

PR #76, opened 2025-08-05
Fix serialization error

LiaCastaneda:


src/plan/codec.rs
```diff
@@ -138,3 +146,98 @@ pub struct ArrowFlightReadExecProto {
...
+    fn schema_i32(name: &str) -> Arc<Schema> {
+        Arc::new(Schema::new(vec![Field::new(name, DataType::Int32, false)]))
+    }
+
+    fn repr(plan: &Arc<dyn ExecutionPlan>) -> String {
+        displayable(plan.as_ref()).indent(true).to_string()
+    }
+
+    #[test]
+    fn distributed_codec_roundtrips() -> datafusion::common::Result<()> {
```
gabotechs: Usually in Rust, is more comment to create one test per test case rather than having several test cases inside the same test.

Unfortunately, unlike Go that allows nesting tests, if a single assertion fails in one test case, no further test cases will be executed, giving only partial information about what went wrong.

I recommend to split each test case into its own Rust test so that if something fails we have granular information about which specific cases are broken

LiaCastaneda: done!

PR #77, opened 2025-08-05
Fix ArrowFlightReadExec result streaming

gabotechs:
Wires up the FligthServiceClient with the `ArrowFlightReadExec` node so that actual gRPC flight calls are issued to the `ArrowFlightEndpoint`.

It uses the gRPC generated `FlightServiceClient` instead of the `FlightClient` in order to used the `BoxCloneSyncChannel` created by the `ChannelManager`

### DIFF
--- a/src/plan/codec.rs
+++ b/src/plan/codec.rs
@@ -84,22 +84,30 @@ impl PhysicalExtensionCodec for DistributedCodec {
         buf: &mut Vec<u8>,
     ) -> datafusion::common::Result<()> {
         if let Some(node) = node.as_any().downcast_ref::<ArrowFlightReadExec>() {
-            ArrowFlightReadExecProto {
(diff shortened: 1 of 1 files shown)


src/plan/arrow_flight_read.rs
```diff
@@ -125,8 +127,6 @@ impl ExecutionPlan for ArrowFlightReadExec {
         let schema = child_stage.plan.schema();
 
         let stream = async move {
-            // concurrenly build streams for each stage
-            // TODO: tokio spawn instead here?
```
gabotechs, line 129: I think this is fine. As this is pure IO work, I think it's fair to not spawn it and just join it in a single thread

robtandy: That was my thought too, but I left the comment so that we'd get an additional perspective.  Ty 👍

PR #80, opened 2025-08-06
add doc comment for execution stage struct

robtandy:
Re-add back doc comment for execution stage struct.

### DIFF
--- a/src/stage/stage.rs
+++ b/src/stage/stage.rs
@@ -25,7 +25,7 @@ use crate::ChannelManager;
 /// When an [`ExecutionStage`] is execute()'d if will execute its plan and return a stream
 /// of record batches.
 ///
-/// If the stage has input stages, then it those input stages will be executed on remote resources
+/// If the stage has input stages, then those input stages will be executed on remote resources
 /// and will be provided the remainder of the stage tree.
 ///
 /// For example if our stage tree looks like this:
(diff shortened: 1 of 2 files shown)


src/stage/stage.rs
```diff
@@ -15,10 +15,64 @@ use crate::task::ExecutionTask;
...
+/// stream of record batches.
 ///
-/// see https://howqueryengineswork.com/13-distributed-query.html
+/// An ExecutionTask is a finer grained unit of work compared to an ExecutionStage.
+/// One ExecutionStage will create one or more ExecutionTasks
 ///
+/// When an [`ExecutionStage`] is execute()'d if will execute its plan and return a stream
+/// of record batches.
+///
+/// If the stage has input stages, then it those input stages will be executed on remote resources
```
NGA-TRAN:
```suggestion
/// If the stage has input stages, those input stages will be executed on remote resources
```

src/stage/stage.rs
```diff
@@ -15,10 +15,64 @@ use crate::task::ExecutionTask;
...
+/// and will be provided the remainder of the stage tree.
+///
+/// For example if our stage tree looks like this:
```
NGA-TRAN, line 31: The sentence does not sound finish

src/stage/stage.rs
```diff
@@ -15,10 +15,64 @@ use crate::task::ExecutionTask;
...
+///             ┌──────┴────────┐
+///        ┌────┴────┐     ┌────┴────┐
+///        │ Stage 4 │     │ Stage 5 │
+///        └─────────┘     └─────────┘                    
+///
+/// ```
+///
+/// The receiving ArrowFlightEndpoint will then execute Stage 2 and will repeat this process.
+///
+/// When Stage 4 4 is executed, it has no input tasks, so it is assumed that the plan included in that
```
NGA-TRAN:
```suggestion
/// When Stage 4 is executed, it has no input tasks, so it is assumed that the plan included in that
```

PR #81, opened 2025-08-07
Support user provided codecs

gabotechs:
This PR ships the ability to provide user-defined codecs for users, for that two main pieces of work are shipped:
- Fix the propagation of the user provided codecs. This codec is now used and combined with the `DistributedCodec` in both the head stage of the plan that gets executed locally and any other further stages that get executed remotely in an ArrowFlightEndpoint
- Proposes a new API for letting users provide their own codecs

About the second point:

Previously, the user defined codec was provided as an optional input to the `DistributedPhysicalOptimizerRule`, which ergonomically made a lot of sense. However, the `DistributedPhysicalOptimizerRule` has no way of propagating the user defined codec down to remote workers.

In practice, users with custom codecs need to inject it twice:
1) one for the head of the plan that gets executed locally (currently addressed by `DistributedPhysicalOptimizerRule` that holds a reference to the codec and injects it into the `ExecutionStage`)
2) another one while instantiating the `ArrowFlightEndpoint` that runs in a different machine (currently unaddressed)

The proposal in this PR is to find an API that is as consistent as possible for 1) and 2) so that providing a custom codec is a consistent and ergonomic experience no matter if it's for the head stage running locally, or for a worker running an `ArrowFlightEndpoint`.

For that, just some plain functions are shipped:

#### providing a codec for an `ArrowFlightEndpoint` that will run remotely
```rust
#[derive(Clone)]
struct CustomSessionBuilder;
impl SessionBuilder for CustomSessionBuilder {
     fn on_new_session(&self, builder: SessionStateBuilder) -> SessionStateBuilder {
          with_user_codec(builder, MyCustomCodec) // <- this is how you'd inject a custom codec for a remote worker
     }
}
```

#### providing a codec for the head of the plan that runs locally
```rust
let builder = SessionStateBuilder::new().with_default_features();
let builder = with_user_codec(builder, MyCustomCodec); // <- this is how you'd inject a custom codec locally
let state = builder.build()
```

The challenge while keeping the old API (`DistributedPhysicalOptimizerRule::with_codec()`), is that it would lead to two different ways necessary for the user to learn to provide their codecs, and makes the `ExecutionStage.codec` field just a transportation field that adds no semantic meaning to the structure, and is just there to carry the codec around.


tests/custom_extension_codec.rs
```diff
@@ -43,22 +45,14 @@ mod tests {
         #[derive(Clone)]
```
gabotechs, line 45: This test cannot yet be un-ignored, as the expected results are failing, but the custom codec propagation works.

tests/error_propagation.rs
```diff
@@ -27,26 +29,17 @@ mod tests {
     use std::sync::Arc;
 
     #[tokio::test]
-    #[ignore]
```
gabotechs, line 30: This test now works! 🚀

src/stage/proto.rs
```diff
@@ -35,54 +35,42 @@ pub struct ExecutionStageProto {
...
-            tasks: stage.tasks.clone(),
-        })
-    }
-}
-
-impl TryFrom<ExecutionStage> for ExecutionStageProto {
-    type Error = DataFusionError;
+pub fn proto_from_stage(
+    stage: &ExecutionStage,
+    codec: &dyn PhysicalExtensionCodec,
```
gabotechs, line 40: As now a codec is needed here, we cannot just implement this in terms of a simple `TryFrom` implementation

src/stage/stage.rs
```diff
@@ -32,8 +31,6 @@ pub struct ExecutionStage {
     /// Our tasks which tell us how finely grained to execute the partitions in
     /// the plan
     pub tasks: Vec<ExecutionTask>,
-    /// An optional codec to assist in serializing and deserializing this stage
-    pub codec: Option<Arc<dyn PhysicalExtensionCodec>>,
```
gabotechs, line 36: Before, this fields ends up being `Some()` in the head stage, but `None` in any other. As it's always `None` in the `ArrowFlightEndpoint`, there's no way we can use a custom user provided codec there.

src/user_provided_codec.rs
```diff
@@ -0,0 +1,126 @@
...
+impl UserCodecTransport for SessionStateBuilder {
+    fn set(&mut self, codec: impl PhysicalExtensionCodec + 'static) {
+        self.config().get_or_insert_default().set(codec);
+    }
+
+    fn get(&self) -> Option<Arc<dyn PhysicalExtensionCodec>> {
+        // Nobody will never want to retriever a user codec from a SessionStateBuilder
+        None
+    }
+}
```
NGA-TRAN, line 126: The content of this file is awesome. New UserCodecTransport trait for setting and getting codec for existing DF struct. And comments showing how to use them

tests/custom_extension_codec.rs
```diff
@@ -43,22 +45,14 @@ mod tests {
...
-        let (ctx, _guard) =
+        let (mut ctx, _guard) =
             start_localhost_context([50050, 50051, 50052], CustomSessionBuilder).await;
-
-        let codec: Arc<dyn PhysicalExtensionCodec> = Arc::new(Int64ListExecCodec);
-        ctx.state_ref()
-            .write()
-            .config_mut()
-            .set_extension(Arc::new(codec));
+        add_user_codec(&mut ctx, Int64ListExecCodec);
```
NGA-TRAN, line 55: Nice and simple

src/plan/arrow_flight_read.rs
```diff
@@ -170,12 +173,14 @@ impl ExecutionPlan for ArrowFlightReadExec {
...
-                "ArrowFlightReadExec: failed to convert stage to proto: {}",
-                e
-            )
+        let mut combined_codec = ComposedPhysicalExtensionCodec::default();
+        combined_codec.push(DistributedCodec {});
+        if let Some(ref user_codec) = get_user_codec(context.session_config()) {
+            combined_codec.push_arc(Arc::clone(user_codec));
+        }
+
+        let child_stage_proto = proto_from_stage(child_stage, &combined_codec).map_err(|e| {
```
NGA-TRAN, line 182: Intuively this makes sense but I think I can only fully get the role of `combined_codec` when I start using it

PR #83, opened 2025-08-08
Add test comparing distributed + single node execution on TPCH data

jayshrivastava:
* add TPCH queries from DataFusion repo (see tests/fixtures/tpch/queries)

* add lazy TPCH parquet data generator using github.com/clflushopt/tpchgen-rs

* write consistency test that runs queries twice (distributed and single-node) and asserts match


tests/non_distributed_consistency_test.rs
```diff
@@ -0,0 +1,135 @@
...
+    use futures::TryStreamExt;
+    use std::error::Error;
+
+    // test_non_distributed_consistency runs each TPC-H query twice - once in a distributed manner
+    // and once in a non-distributed manner. For each query, it asserts that the results are identical.
+    #[tokio::test]
+    async fn test_non_distributed_consistency() -> Result<(), Box<dyn Error>> {
+        tpch::generate_tpch_data().await;
+
+        for i in 1..=tpch::NUM_QUERIES {
```
gabotechs: It would be nice if we could create 22 different tests for the 22 queries rather than for looping through them here. This is desirable because of several reasons:
- Rust's test runner will stop testing as soon as 1 query fails, so we'll not able to get a full report of what went wrong, only a partial result with the first failure
- We can hookup a debugger for the query we care about, and just debug that query in isolation without running the previous first
- All 22 queries can potentially be run in parallel, which ideally should be managed by the test runner
- Allows #[ignore]-ing specific tests that we know are failing (like 15 in this case)

<details><summary>Here's a snippet that hopefully you can just copy-paste</summary>

```rust
#[tokio::test]
async fn test_tpch_1() -> Result<(), Box<dyn Error>> {
    test_tpch_query(1).await
}

#[tokio::test]
async fn test_tpch_2() -> Result<(), Box<dyn Error>> {
    test_tpch_query(2).await
}

#[tokio::test]
async fn test_tpch_3() -> Result<(), Box<dyn Error>> {
    test_tpch_query(3).await
}

#[tokio::test]
async fn test_tpch_4() -> Result<(), Box<dyn Error>> {
    test_tpch_query(4).await
}

#[tokio::test]
async fn test_tpch_5() -> Result<(), Box<dyn Error>> {
    test_tpch_query(5).await
}

#[tokio::test]
async fn test_tpch_6() -> Result<(), Box<dyn Error>> {
    test_tpch_query(6).await
}

#[tokio::test]
async fn test_tpch_7() -> Result<(), Box<dyn Error>> {
    test_tpch_query(7).await
}

#[tokio::test]
async fn test_tpch_8() -> Result<(), Box<dyn Error>> {
    test_tpch_query(8).await
}

#[tokio::test]
async fn test_tpch_9() -> Result<(), Box<dyn Error>> {
    test_tpch_query(9).await
}

#[tokio::test]
async fn test_tpch_10() -> Result<(), Box<dyn Error>> {
    test_tpch_query(10).await
}

#[tokio::test]
async fn test_tpch_11() -> Result<(), Box<dyn Error>> {
    test_tpch_query(11).await
}

#[tokio::test]
async fn test_tpch_12() -> Result<(), Box<dyn Error>> {
    test_tpch_query(12).await
}

#[tokio::test]
async fn test_tpch_13() -> Result<(), Box<dyn Error>> {
    test_tpch_query(13).await
}

#[tokio::test]
async fn test_tpch_14() -> Result<(), Box<dyn Error>> {
    test_tpch_query(14).await
}

#[tokio::test]
async fn test_tpch_15() -> Result<(), Box<dyn Error>> {
    test_tpch_query(15).await
}

#[tokio::test]
async fn test_tpch_16() -> Result<(), Box<dyn Error>> {
    test_tpch_query(16).await
}

#[tokio::test]
async fn test_tpch_17() -> Result<(), Box<dyn Error>> {
    test_tpch_query(17).await
}

#[tokio::test]
async fn test_tpch_18() -> Result<(), Box<dyn Error>> {
    test_tpch_query(18).await
}

#[tokio::test]
async fn test_tpch_19() -> Result<(), Box<dyn Error>> {
    test_tpch_query(19).await
}

#[tokio::test]
async fn test_tpch_20() -> Result<(), Box<dyn Error>> {
    test_tpch_query(20).await
}

#[tokio::test]
async fn test_tpch_21() -> Result<(), Box<dyn Error>> {
    test_tpch_query(21).await
}

#[tokio::test]
async fn test_tpch_22() -> Result<(), Box<dyn Error>> {
    test_tpch_query(22).await
}

```
</details>

tests/non_distributed_consistency_test.rs
```diff
@@ -0,0 +1,135 @@
...
+
+            let batches2 = execute_stream(physical2.clone(), ctx2.task_ctx())?
+                .try_collect::<Vec<_>>()
+                .await?;
+
+            // Compare results
+            assert_eq!(
+                batches1.len(),
+                batches2.len(),
+                "Query {} returned different number of batches: {} vs {}",
```
gabotechs: I think it's fine if the are a different number of batches in distributed mode vs normal mode. What we ultimately care is that the results are the same. For example, it should be fine if normal mode returns 2 batches with 5 rows each and the distributed mode returns 5 batches with 2 rows each, as long as the data is the same.

DataFusion has this `pretty_format_batches` that will build a string out of the results that is human inspectable, maybe comparing those instead?

tests/common/tpch.rs
```diff
@@ -5,15 +5,36 @@ use datafusion::{
...
     let schema = Arc::new(get_tpch_table_schema(name));
     Arc::new(MemTable::try_new(schema, vec![]).unwrap())
 }
 
 pub fn tpch_query(num: u8) -> String {
-    // read the query from the test/tpch/queries/ directory and return it
-    let query_path = format!("testing/tpch/queries/q{}.sql", num);
-    std::fs::read_to_string(query_path)
+    // read the query from the queries directory in the fixtures dir and return it
+    let query_path = format!("{}/q{}.sql", QUERIES_DIR, num);
```
gabotechs: One way to make this agnostic from the relative working directory is to use `env!("CARGO_MANIFEST_PATH")`. The Rust compiler will put there the absolute path of the current project, so the actual working directory from which the tests are run does not matter

tests/common/tpch.rs
```diff
@@ -113,3 +134,61 @@ pub fn get_tpch_table_schema(table: &str) -> Schema {
...
     }
 }
+
+// generate_table creates a parquet file in the DATA_DIR directory from an arrow RecordBatch row
+// source.
+fn generate_table<A>(mut data_source: A, table_name: &str) -> Result<(), Box<dyn std::error::Error>>
+where
+    A: Iterator<Item = RecordBatch>,
+{
+    let output_path = format!("{}/{}.parquet", DATA_DIR, table_name);
```
gabotechs: If you prefer, you also can do this:
```rust
    let output_path = format!("{DATA_DIR}/{table_name}.parquet");
```

Cargo.toml
```diff
@@ -25,3 +25,7 @@ object_store = "0.12.3"
 
 [dev-dependencies]
 insta = { version = "1.43.1", features = ["filters"] }
+tpchgen = { git = "https://github.com/clflushopt/tpchgen-rs" }
+tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs" }
```
gabotechs: 🥳 awesome!

Now that we are using a git dependency, one way of making this more resilient to fix a specific commit hash, (last one is `c8d823432528eed4f70fca5a1296a66c68a389a8`). That way there's no surprises if a breaking change happens

tests/non_distributed_consistency_test.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    #[tokio::test]
+    async fn test_tpch_14() -> Result<(), Box<dyn Error>> {
+        test_tpch_query(14).await
+    }
+
+    #[tokio::test]
+    #[ignore]
+    // TODO: Support query 15?
+    // Skip because it contains DDL statements not supported in single SQL execution
```
NGA-TRAN, line 85: It is fine to skip this for now.
I am just curious how Vanilla DF test this case.

jayshrivastava: I filed https://github.com/datafusion-contrib/datafusion-distributed/issues/87 to track this. I will look into how to support views

PR #89, opened 2025-08-08
Execution working on all 22 TPCH queries

robtandy:
## Edit:
- Moved validation out of benchmarks and into @jayshrivastava 's tpch validation tests
- Incorporated improvements suggested in comments
- Ignored tpch query 22, while support for `NestedLoopJoinExec` is added.

This PR fixes an execution logical bug and also adds a `--validate` flag to the benchmarks to confirm that we calculate the correct result vs the single node case.

To run the TPCH benchmark in a distributed fashion and validate it against single node execution, follow the readme in `benchmarks`.

Note that the particular approach to distributed execution that this library takes requires all joins to be partition joins.   We do not support `CollectLeft` in particular.    So, the following modifications to the context before planning are required for correct operation
```rust
config
       .options_mut()
       .optimizer
       .hash_join_single_partition_threshold = 0;
config
       .options_mut()
       .optimizer
       .hash_join_single_partition_threshold_rows = 0;

config.options_mut().optimizer.prefer_hash_join = true;

```
At the moment this is set in the benchmark crate, but we really need to make this easy for the user and not allow them to mess up these values.  I'm not sure how to do this at the moment.  I think we can refactor subsequent to this PR.

Regarding adding support for other Hash join modes, I think we can do that, and then use the benchmark to compare and evaluate options for execution speed.

### DIFF
--- a/src/errors/mod.rs
+++ b/src/errors/mod.rs
@@ -17,7 +17,7 @@ mod schema_error;
 pub fn datafusion_error_to_tonic_status(err: &DataFusionError) -> tonic::Status {
     let err = DataFusionErrorProto::from_datafusion_error(err);
     let err = err.encode_to_vec();
-    
+
     tonic::Status::with_details(tonic::Code::Internal, "DataFusionError", err.into())
 }
 
--- a/src/physical_optimizer.rs
+++ b/src/physical_optimizer.rs
@@ -12,9 +12,7 @@ use datafusion::{
     config::ConfigOptions,
     error::Result,
     physical_optimizer::PhysicalOptimizerRule,
-    physical_plan::{
-        repartition::RepartitionExec, ExecutionPlan, ExecutionPlanProperties,
-    },
+    physical_plan::{repartition::RepartitionExec, ExecutionPlan, ExecutionPlanProperties},
 };
 use uuid::Uuid;
 
--- a/src/stage/execution_stage.rs
+++ b/src/stage/execution_stage.rs
@@ -1,0 +1,293 @@
+use std::sync::Arc;
+
+use datafusion::common::internal_err;
+use datafusion::error::{DataFusionError, Result};
+use datafusion::execution::TaskContext;
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::prelude::SessionContext;
+
+use itertools::Itertools;
+use rand::Rng;
+use url::Url;
+use uuid::Uuid;
+
+use crate::task::ExecutionTask;
+use crate::ChannelManager;
+
+/// A unit of isolation for a portion of a physical execution plan
+/// that can be executed independently and across a network boundary.  
+/// It implements [`ExecutionPlan`] and can be executed to produce a
+/// stream of record batches.
+///
+/// An ExecutionTask is a finer grained unit of work compared to an ExecutionStage.
+/// One ExecutionStage will create one or more ExecutionTasks
+///
+/// When an [`ExecutionStage`] is execute()'d if will execute its plan and return a stream
+/// of record batches.
+///
+/// If the stage has input stages, then it those input stages will be executed on remote resources
+/// and will be provided the remainder of the stage tree.
+///
+/// For example if our stage tree looks like this:
+///
+/// ```text
+///                       ┌─────────┐
+///                       │ stage 1 │
+///                       └───┬─────┘
+///                           │
+///                    ┌──────┴────────┐
+///               ┌────┴────┐     ┌────┴────┐
+///               │ stage 2 │     │ stage 3 │
+///               └────┬────┘     └─────────┘
+///                    │
+///             ┌──────┴────────┐
+///        ┌────┴────┐     ┌────┴────┐
+///        │ stage 4 │     │ Stage 5 │
+///        └─────────┘     └─────────┘                    
+///
+/// ```
+///  
+/// Then executing Stage 1 will run its plan locally.  Stage 1 has two inputs, Stage 2 and Stage 3.  We
+/// know these will execute on remote resources.   As such the plan for Stage 1 must contain an
+/// [`ArrowFlightReadExec`] node that will read the results of Stage 2 and Stage 3 and coalese the
+/// results.
+///
+/// When Stage 1's [`ArrowFlightReadExec`] node is executed, it makes an ArrowFlightRequest to the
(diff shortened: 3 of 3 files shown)


benchmarks/src/tpch/run.rs
```diff
@@ -100,10 +100,12 @@ pub struct RunOpt {
...
     sorted: bool,
 
-    /// Mark the first column of each table as sorted in ascending order.
-    /// The tables should have been created with the `--sort` option for this to have any effect.
+    /// The maximum number of partitions per task.
     #[structopt(long = "ppt")]
     partitions_per_task: Option<usize>,
+
+    #[structopt(long = "validate")]
+    validate: bool,
```
gabotechs: What do you think about moving forward with @jayshrivastava's changes in https://github.com/datafusion-contrib/datafusion-distributed/pull/83 for validating TPCH correctness instead of this? it might be slightly better to ensure validation there because:
- It would be nice to touch this code as little as possible, as this is pretty much vendored code from upstream DataFusion, and if we decide to move this project there or upstream DataFusion decides to make their benchmarks crate public, it would be difficult to port because of conflicts
- We want to ensure TPCH correctness in the CI, so it might be more suitable to do it as a mandatory test suite using Cargo test tools rather than an optional step during the benchmarks

src/flight_service/do_get.rs
```diff
@@ -36,59 +48,35 @@ impl ArrowFlightEndpoint {
...
+        state.config_mut().set_extension(Arc::new(partition_group));
+
+        let inner_plan = stage.plan.clone();
+
+        /*println!(
+            "{} Task {:?} executing partition {}",
+            stage.name(),
+            task.partition_group,
+            partition
+        );*/
```
gabotechs: remove this maybe?

src/flight_service/do_get.rs
```diff
@@ -104,4 +92,68 @@ impl ArrowFlightEndpoint {
...
+        &self,
+        doget: DoGet,
+    ) -> Result<(SessionState, Arc<ExecutionStage>), Status> {
+        let key = doget
+            .stage_key
+            .ok_or(Status::invalid_argument("DoGet is missing the stage key"))?;
+        let once_stage = self.stages.entry(key).or_default();
+
+        let (state, stage) = once_stage
+            .get_or_try_init(|| async {
```
gabotechs, line 100: This will lock the `once_stage` RefMut across the initialization, locking a shard in the `self.stages` dashmap across an asynchronous gap, which might be too much locking.

Fortunately, it's very easy to prevent this:
- we can make the `OnceCell` a shared reference:
```rust
    pub(super) stages: DashMap<StageKey, Arc<OnceCell<(SessionState, Arc<ExecutionStage>)>>>,
```
- and then immediately drop the reference to the dashmap entry
```rust
        let once_stage = {
            let entry = self.stages.entry(key).or_default();
            Arc::clone(&entry)
            // <- dashmap RefMut get's dropped, releasing the lock for the current shard
        };
```

src/flight_service/stream_partitioner_registry.rs
gabotechs, line 1: This file contained some tests that tested the behavior of sharing a single execution node across multiple callers with a dashmap. It's a shame to delete them, I would have expected them to still be valid here.

If you see no path forward in keeping those tests it's fine, we can build new ones eventually.

src/plan/arrow_flight_read.rs
```diff
@@ -183,26 +184,45 @@ impl ExecutionPlan for ArrowFlightReadExec {
...
+        let query_id = stage.query_id.to_string();
+
         let stream = async move {
-            let futs = child_stage_tasks.iter().enumerate().map(|(i, task)| async {
-                let url = task.url()?.ok_or(internal_datafusion_err!(
-                    "ArrowFlightReadExec: task is unassigned, cannot proceed"
-                ))?;
-                stream_from_stage_task(ticket.clone(), &url, schema.clone(), &channel_manager).await
+            let futs = child_stage_tasks.iter().enumerate().map(|(i, task)| {
+                let i_capture = i;
```
gabotechs: 🤔 `i_capture`? It should not be necessary to capture any variables that implement the `Copy` trait right?

tests/distributed_aggregation.rs
```diff
@@ -31,22 +30,24 @@ mod tests {
...
+                    RepartitionExec: partitioning=Hash([RainToday@0], 3), input_partitions=3
+                      RepartitionExec: partitioning=RoundRobinBatch(3), input_partitions=1
                         AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))]
                           DataSourceExec: file_groups={1 group: [[/testdata/weather.parquet]]}, projection=[RainToday], file_type=parquet
         ",
-        );
+        );*/
 
-        assert_snapshot!(physical_distributed_str,
+        /*assert_snapshot!(physical_distributed_str,
```
gabotechs: does this need to be commented? I think it should be fine to leave this uncommented, it should work.

FYI, this is using https://github.com/mitsuhiko/insta, which means that if the test fails, you can just do:

```
cargo insta review
```

And you will be prompted to accept the changes

You can install `cargo insta` with:

```
curl -LsSf https://insta.rs/install.sh | sh
```

benchmarks/src/tpch/run.rs
robtandy: Yep i thought the same thing and i moved it out of the benchmarks and aligned with @jayshrivastava 's PR

src/flight_service/do_get.rs
robtandy: ✅

src/plan/arrow_flight_read.rs
robtandy: Good eye and yes, not needed!

tests/distributed_aggregation.rs
robtandy: wow, insta is awesome! 💯  Done.

src/flight_service/stream_partitioner_registry.rs
robtandy: I've added it back but i'm not sure of its function or necessity, as its not referenced anywhere.

src/flight_service/do_get.rs
robtandy: A good improvement. added.

PR #91, opened 2025-08-12
Add delta report for benchmarks

gabotechs:
Adds a delta report that compares the last run benchmarks with the current ones, something like this:

```
==== Comparison with the previous benchmark from 2025-08-12 14:38:33 UTC ====
 Query 1: prev=  74 ms, new=  70 ms, diff=1.06 faster ✔
 Query 2: prev=  25 ms, new=  25 ms, diff=1.00 slower ✖
 Query 3: prev=  37 ms, new=  36 ms, diff=1.03 faster ✔
 Query 4: prev=  25 ms, new=  28 ms, diff=1.12 slower ✖
 Query 5: prev=  46 ms, new=  54 ms, diff=1.17 slower ✖
 Query 6: prev=  22 ms, new=  21 ms, diff=1.05 faster ✔
 Query 7: prev=  64 ms, new=  61 ms, diff=1.05 faster ✔
 Query 8: prev=  47 ms, new=  43 ms, diff=1.09 faster ✔
 Query 9: prev=  63 ms, new=  53 ms, diff=1.19 faster ✔
Query 10: prev=  66 ms, new=  53 ms, diff=1.25 faster ✅
Query 11: prev=  34 ms, new=  21 ms, diff=1.62 faster ✅
Query 12: prev=  58 ms, new=  41 ms, diff=1.41 faster ✅
Query 13: prev=  34 ms, new=  32 ms, diff=1.06 faster ✔
Query 14: prev=  35 ms, new=  30 ms, diff=1.17 faster ✔
Query 15: prev=  53 ms, new=  48 ms, diff=1.10 faster ✔
Query 16: prev=  23 ms, new=  19 ms, diff=1.21 faster ✅
Query 17: prev=  77 ms, new=  70 ms, diff=1.10 faster ✔
Query 18: prev= 103 ms, new=  92 ms, diff=1.12 faster ✔
Query 19: prev=  58 ms, new=  51 ms, diff=1.14 faster ✔
Query 20: prev=  52 ms, new=  41 ms, diff=1.27 faster ✅
Query 21: prev= 112 ms, new=  89 ms, diff=1.26 faster ✅
Query 22: prev=  22 ms, new=  17 ms, diff=1.29 faster ✅
```

The first time the benchmarks are run, no report will be printed to stdout, the second time, a report will be produced comparing it with the first run, the third time will be compared to the second one, etc...

This can be tested running this command twice:
```
cargo run -p datafusion-distributed-benchmarks --release -- tpch --path benchmarks/data/tpch_sf1 
```

### DIFF
--- a/benchmarks/src/tpch/run.rs
+++ b/benchmarks/src/tpch/run.rs
@@ -187,7 +187,7 @@ impl RunOpt {
                 }
             }
         }
-        benchmark_run.compare_with_previous(self.output_path.as_ref())?;
+        benchmark_run.maybe_compare_with_previous(self.output_path.as_ref())?;
         benchmark_run.maybe_write_json(self.output_path.as_ref())?;
         benchmark_run.maybe_print_failures();
         Ok(())
--- a/benchmarks/src/util/run.rs
+++ b/benchmarks/src/util/run.rs
@@ -209,7 +209,7 @@ impl BenchmarkRun {
         Ok(())
     }
 
-    pub fn compare_with_previous(&self, maybe_path: Option<impl AsRef<Path>>) -> Result<()> {
+    pub fn maybe_compare_with_previous(&self, maybe_path: Option<impl AsRef<Path>>) -> Result<()> {
         let Some(path) = maybe_path else {
             return Ok(());
         };


benchmarks/src/tpch/run.rs
```diff
@@ -139,14 +145,16 @@ impl SessionBuilder for RunOpt {
 }
 
 impl RunOpt {
-    pub async fn run(self) -> Result<()> {
-        let (ctx, _guard) = start_localhost_context([50051], self.clone()).await;
```
jayshrivastava: nit: we should put these in a config or `DEFAULT_HOST` const eventually

benchmarks/src/tpch/run.rs
```diff
@@ -164,6 +172,7 @@ impl RunOpt {
                 }
             }
         }
+        benchmark_run.compare_with_previous(self.output_path.as_ref())?;
```
jayshrivastava: nit: consider calling it `maybe_compare_with_previous`

benchmarks/Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde_json = "1.0.141"
 env_logger = "0.11.8"
 async-trait = "0.1.88"
 datafusion-proto = { version = "49.0.0", optional = true }
+chrono = "0.4.41"
```
jayshrivastava, line 19: Thoughts on using https://github.com/bheisler/criterion.rs?

benchmarks/src/tpch/run.rs
gabotechs: @robtandy will probably address this soon, the idea would be to pick some free ports rather than hardcoding them here

benchmarks/Cargo.toml
gabotechs: The strategy for now is to keep it as close to what upstream DataFusion does, so that eventually we can:
- request upstream to expose the benchmark crate so that we no longer need to maintain this code.
- introduced distributed-datafusion in apache/datafusion as a separate crate and use their existing benchmark crate.

For that reason, the less both approaches diverge, the easier it will be to port it at some point. 

That being said, it's starting to diverge already, as we have needs that are not satisfied upstream (or are satisfied in a different way)...

I'd defer this decision as much as possible, but if we find that we should remove all upstream copy-pasted code in favor of our own, this decision should probably be revisited.

jayshrivastava: Sounds good 👍🏽

benchmarks/src/tpch/run.rs
robtandy: Indeed, i addressed in #89

benchmarks/Cargo.toml
robtandy: Yes, agree on deferment for the reasons stated.

Issue #90, opened 2025-08-11
clean up orphaned ExecutionStage instances in the ArrowFlightEndpoint


jayshrivastava:
Right now ExecutionStages are held by stage_key which is `(query_id, stage_num, task_num)`

They can be discarded if:
- [x] We have executed all partitions in it and do not need it anymore (#108)
- [x] It gets orphaned because a node higher up in the plan died and never requested the data. Having a timeout clean this up is appropriate (#96)

Some prior art here: https://github.com/datafusion-contrib/datafusion-distributed/pull/70

PR #92, opened 2025-08-12
Create TTL map with time wheel architecture

jayshrivastava:
This change adds a DashMap-like struct which has a background tasks to clean
up entries that have outlived a configurable TTL.

This struct is simliar to https://github.com/moka-rs/moka, which also uses
time wheels. Having our own module avoids introducing a large dependency, which
keeps this project closer to vanilla datafusion.

This change is meant to be useful for https://github.com/datafusion-contrib/datafusion-distributed/pull/89, where it's
possible for `ExecutionStages` to be orphaned in `ArrowFlightEndpoint`. We need an async task to clean up old entries.

Informs: https://github.com/datafusion-contrib/datafusion-distributed/issues/90

### DIFF
--- a/src/common/ttl_map.rs
+++ b/src/common/ttl_map.rs
@@ -368,7 +368,6 @@ mod tests {
 
         let start_time = Instant::now();
         let task_count = 100_000;
-        let contention_factor = 10;
 
         // Spawn 10 tasks that repeatedly read the same keys
         let mut handles = Vec::new();
@@ -377,7 +376,7 @@ mod tests {
             let handle = tokio::spawn(async move {
                 // All tasks fight for the same keys - maximum contention
                 let start = Instant::now();
-                let _value = map.get_or_init(task_id % (task_count / contention_factor), || task_id * 1000).await;
+                let _value = map.get_or_init(rand::random(), || task_id * 1000).await;
                 start.elapsed().as_nanos()
             });
             handles.push(handle);


src/common/ttl_map.rs
```diff
@@ -0,0 +1,352 @@
...
+            tick: Duration::from_secs(3),
+            ttl: Duration::from_secs(60),
+        }
+    }
+}
+
+impl<K, V> TTLMap<K, V>
+where
+    K: Eq + Hash + Send + Sync + Clone + 'static,
+    V: Default + Clone + Send + Sync + 'static,
```
jayshrivastava, line 82: I can explain why V is clone below

src/common/ttl_map.rs
```diff
@@ -0,0 +1,352 @@
...
+        let mut new_entry = false;
+        let value = match self.data.entry(key.clone()) {
+            Entry::Vacant(entry) => {
+                let value = f();
+                entry.insert(value.clone());
+                new_entry = true;
+                value
+            }
+            Entry::Occupied(entry) => entry.get().clone(),
+        };
```
jayshrivastava, line 142: This is tricky.

When you hold an `Entry`, you are locking the map and the dashmap specifically says `**Locking behaviour:** May deadlock if called when holding any sort of reference into the map`

I believe holding the lock over any `.await` point may deadlock because another task may be scheduled and try to access the same entry (Tbh I think this may happen if the other task accesses a different entry on the same shard).

Alternatives considered:
- we return a reference to V. This is risky because the caller might hold it across an .await and deadlock
- we pass a closure which operates on V while we hold the lock. This is not ideal because it increases lock contention

Thus, I went with Clone. If the caller wants a shared reference, they can wrap their type in an Arc.

src/common/ttl_map.rs
```diff
@@ -0,0 +1,343 @@
+/*
+TTLMap is a DashMap that automatically removes entries after a specified time-to-live (TTL).
+
+How the Time Wheel Works
+
+Time Buckets:  [0] [1] [2] [3] [4] [5] [6] [7] ...
```
gabotechs, line 6: In Rust, module level docs are defined with `//!`. That way, documentation compilers and IDEs can interpret them as pieces of documentation and display them appropriately.

src/common/ttl_map.rs
```diff
@@ -0,0 +1,343 @@
...
+use dashmap::{DashMap, Entry};
+use datafusion::error::DataFusionError;
+use std::collections::HashSet;
+use std::hash::Hash;
+use std::sync::atomic::AtomicU64;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::Mutex;
+
+// TTLMap is a key-value store that automatically removes entries after a specified time-to-live.
```
gabotechs, line 38: This should probably be a doc comment with a triple `///`, otherwise IDEs will ignore it while hovering over the struct

src/common/ttl_map.rs
gabotechs: I think requiring `Clone` for the values is a fair ask, and it's completely fine for our use case 👍

src/common/ttl_map.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+            Entry::Occupied(entry) => entry.get().clone(),
+        };
+
+        // Insert the key into the previous bucket, meaning the key will be evicted
+        // when the wheel completes a full rotation.
+        if new_entry {
+            let time = self.time.load(std::sync::atomic::Ordering::SeqCst);
+            {
+                let mut buckets = self.buckets.lock().await;
```
gabotechs: 🤔 this looks like it's introducing a global lock shared by all the new requests. This means that all new requests will be competing for the same lock, and it's the main thing the `dashmap` tried to address in the first place.

I think we should be capable of finding a solution that does not imply a global lock. A hint I know is: spawning tokio tasks is almost free. What to do with that hint? no idea yet

jayshrivastava: Thanks for the review!

I've attempted an alternative implementation here, using your hint: https://github.com/datafusion-contrib/datafusion-distributed/pull/96/files#diff-835dd4f31132680b017b402166e24979a012bb5f1975c95c34de467706b964d2 
- Each shard in the time wheel has it's own task and `KeySet` and a clone of the `Arc<DashMap>`. Each task is responsible for managing some subset of keys in the map. This avoids the global mutex over all the KeySets
- Now, when we insert a key into the `TTLMap`, we send the key on a channel to the task which is responsible for its shard.
- There's one task responsible for issuing Clear commands (also over the channel) to the time wheel tasks

I pushed a benchmark commit to this branch (and same one to the other branch above) to see if I improved lock contention. Unfortunately, the results look similar regardless of what I do.

There's a lot of variance but generally I see 17531721 - 20000000 ops / sec.

This PR
```
Tasks: 100
Operations per task: 1000000
Total operations: 100000000
Total time: 5.70s
Throughput: 17531721 ops/sec
Average latency: 0.06 μs per operation
Entries remaining: 10
test common::ttl_map::tests::bench_lock_contention ... ok
```

Other PR with no mutex
```
=== TTLMap Lock Contention Benchmark ===
Tasks: 100
Operations per task: 1000000
Total operations: 100000000
Total time: 6.03s
Throughput: 16582541 ops/sec
Average latency: 0.06 μs per operation
Entries remaining: 10
test common::ttl_map::tests::bench_lock_contention ... ok
```

If I comment all locking and the GC task, the baseline performance is at most 20M ops / sec as well.
```
=== TTLMap Lock Contention Benchmark ===
Tasks: 100
Operations per task: 1000000
Total operations: 100000000
Total time: 4.93s
Throughput: 20291279 ops/sec
Average latency: 0.05 μs per operation
Entries remaining: 10
test common::ttl_map::tests::bench_lock_contention ... ok
```

I'm not sure how to proceed. Either the benchmark needs to be better or locking the time wheel is not the bottle neck. The latter is possible because:
- The critical sections under `self.buckets.lock().await` are very small (constant # of instructions)
- The dashmap mutex (which we hold while initializing entries) is probably a bigger source of contention because we hold an Entry's lock while initializing it. 
    - This is okay because the first request initializes the key, so the other ones have to wait anyways and can read the cached `Value` once its done.
    - This might be bad if the shards in the DashMap are large, introducing cross-key contention. Generally, this is an okay tradeoff and we defer this decision to the DashMap

Please let me know your thoughts!

gabotechs: It's going to be a bit hard to create an accurate benchmark for this, but I think it's worth getting that right in order to compare approaches. I think there are some improvement points to the benchmark in order to make it more realistic:
- The `worker_threads` in the test should be as high as your machine allows you, with just the current `worker_threads = 2`, the chances of just two physical threads competing concurrently for the same lock are slim.
- `key_count` should probably be much higher, as in reality almost all keys are going to be different, I think even making all the keys random could work
- `task_count` should be several orders of magnitude higher, something like (number of concurrent queries) * (number of stages per query) * (number of tasks per stage) / (number of available workers), in practice, this can easily go in the order of millions
- In practice, each task is handled in its own `do_get` call, and tonic will spawn a tokio task for it, so it should be more realistic that each task in the bench only performs 1 `get_or_init`
- I'd remove any source of noise like string allocations (`format!`), I'd try to just benchmark a pure `get_or_init()` access

gabotechs: here's a draft approach https://github.com/datafusion-contrib/datafusion-distributed/pull/97

The result of running that is:
```
=== TTLMap Lock Contention Benchmark ===
Tasks: 100000
Total time: 88.52ms
Average latency: 425 μs per operation
Entries remaining: 0
DashMap Lock contention time: 12ms
Mutex Lock contention time: 42503ms
```

PR #96, opened 2025-08-13
Create TTL map with time wheel architecture

jayshrivastava:
This change adds a DashMap-like struct which has a background tasks to clean
up entries that have outlived a configurable TTL.

This data structure is added so it can be used in the `ArrowFlightEndpoint`, where it's
possible for `ExecutionStages` to be orphaned due to errors. This change adds the ability to
clean up tasks async.

The implementation is simliar to https://github.com/moka-rs/moka, which also uses
time wheels. Benchmarks show that this implementation has less overhead per operation
than moka (see results https://github.com/datafusion-contrib/datafusion-distributed/pull/99).
```
=== TTLMap Moka Benchmark ===
Tasks: 100000
Total time: 162.53ms
Average latency: 45 μs per operation
Throughput: 615257.30 ops/sec

=== TTLMap Lock Contention Benchmark ===
Tasks: 100000
Total time: 137.07ms
Average latency: 985 ns per operation
Entries remaining: 0
DashMap Lock contention time: 21ms
Accounting time: 47ms
```

There's also an implementation in https://github.com/datafusion-contrib/datafusion-distributed/pull/92, which
has the worst performance by far.
```
Tasks: 100000
Total time: 105.65ms
Average latency: 20453 μs per operation
Entries remaining: 0
DashMap Lock contention time: 23ms
Mutex Lock contention time: 2045251ms
```

Informs: https://github.com/datafusion-contrib/datafusion-distributed/issues/90

### DIFF
--- a/src/common/ttl_map.rs
+++ b/src/common/ttl_map.rs
@@ -1,0 +1,479 @@
+//! TTLMap is a DashMap that automatically removes entries after a specified time-to-live (TTL).
+//!
+//! How the Time Wheel Works
+//!
+//! Time Buckets:  [0] [1] [2] [3] [4] [5] [6] [7] ...
+//! Current Time:           ^
+//!                         |
+//!                    time % buckets.len()
+//!
+//! When inserting key "A" at time=2:
+//! - Key "A" goes into bucket[(2-1) % 8] = bucket[1]
+//! - Key "A" will be expired when time advances to bucket[1] again
+//!
+//! Generally, keys in a bucket expire when the wheel makes a full rotation, making
+//! the total TTL equal to the tick duration * buckets.len().
+//!
+//! Usage
+//! ```rust,ignore
+//! use std::time::Duration;
+//! use datafusion_distributed::common::ttl_map::{TTLMapConfig, TTLMap};
+//! let config = TTLMapConfig { tick: Duration::from_secs(5), ttl: Duration::from_secs(60) };
+//! let ttl_map = TTLMap::try_new(config).unwrap();
+//! let value = ttl_map.get_or_init("key", || "value");
+//! ```
+//! TODO(#101): If an existing entry is accessed, reset its TTL timer.
+use dashmap::{DashMap, Entry};
+use datafusion::error::DataFusionError;
+use std::collections::HashSet;
+use std::hash::Hash;
+use std::mem;
+use std::sync::atomic::AtomicU64;
+#[cfg(test)]
+use std::sync::atomic::{AtomicUsize, Ordering::Relaxed};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+
+/// A bucket is a set of keys that are managed edited asynchronously.
+#[derive(Clone)]
+struct Bucket<K> {
+    /// tx is used to send a BucketOp to the task
+    tx: UnboundedSender<BucketOp<K>>,
+    /// This task is responsible for managing keys
+    _task: Arc<tokio::task::JoinHandle<()>>,
+}
+
+/// BucketOp is used to communicate with the task responsible for managing keys.
+enum BucketOp<K> {
+    Insert { key: K },
+    Clear,
+}
+
+impl<K> Bucket<K>
+where
+    K: Hash + Eq + Send + Sync + Clone + 'static,
+{
+    /// new creates a new Bucket
+    fn new<V>(data: Arc<DashMap<K, V>>) -> Self
+    where
+        V: Default + Clone + Send + Sync + 'static,
+    {
+        // TODO: To avoid unbounded growth, consider making this bounded. Alternatively, we can
+        // introduce a dynamic GC period to ensure that GC can keep up.
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+        Self {
+            tx,
+            _task: Arc::new(tokio::spawn(
+                async move { Bucket::task(rx, data.clone()).await },
+            )),
+        }
+    }
+
+    fn register_key(&self, key: K) {
+        // We can safely ignore the error here because the receiver is never closed.
+        // If the receiver is dropped, it means this struct is being dropped.
+        let _ = self.tx.send(BucketOp::Insert { key });
+    }
+
+    fn clear(&self) {
+        // We can safely ignore the error here because the receiver is never closed.
+        // If the receiver is dropped, it means this struct is being dropped.
+        let _ = self.tx.send(BucketOp::Clear);
+    }
+
+    /// task is responsible for managing a subset of keys in the map.
+    async fn task<V>(mut rx: UnboundedReceiver<BucketOp<K>>, data: Arc<DashMap<K, V>>)
+    where
+        V: Default + Clone + Send + Sync + 'static,
+    {
+        let mut shard = HashSet::new();
+        while let Some(op) = rx.recv().await {
+            match op {
+                BucketOp::Insert { key } => {
+                    shard.insert(key);
+                }
+                BucketOp::Clear => {
+                    let keys_to_delete = mem::replace(&mut shard, HashSet::new());
+                    for key in keys_to_delete {
+                        data.remove(&key);
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// TTLMap is a key-value store that automatically removes entries after a specified time-to-live.
+pub struct TTLMap<K, V> {
+    /// The buckets in the time wheel
+    buckets: Vec<Bucket<K>>,
+
+    /// gc_scheduler_task is responsible scheduling GC operations among the Buckets in the time wheel.
+    gc_scheduler_task: Option<Vec<tokio::task::JoinHandle<()>>>,
+
+    /// The actual key-value storage using DashMap for concurrent access
+    data: Arc<DashMap<K, V>>,
+
+    /// Atomic counter tracking the current time. Incremented by the background GC task every `tick` duration.
+    time: Arc<AtomicU64>,
+
+    config: TTLMapConfig,
+
+    #[cfg(test)]
+    metrics: TTLMapMetrics,
+}
+
+#[cfg(test)]
+#[derive(Default)]
+struct TTLMapMetrics {
+    dash_map_lock_contention_time: AtomicUsize,
+    ttl_accounting_time: AtomicUsize,
+}
+
+/// TTLMapConfig configures the TTLMap parameters such as the TTL and tick period.
+pub struct TTLMapConfig {
+    /// How often the map is checks for expired entries.
+    /// This must be less than `ttl`. It's recommended to set `ttl` to a multiple
+    /// of `tick`.
+    pub tick: Duration,
+    /// The time-to-live for entries
+    pub ttl: Duration,
+}
+
+impl Default for TTLMapConfig {
+    fn default() -> Self {
+        Self {
+            tick: Duration::from_secs(3),
+            ttl: Duration::from_secs(60),
+        }
+    }
+}
+
+impl TTLMapConfig {
+    fn is_valid(&self, tick: Duration, ttl: Duration) -> Result<(), DataFusionError> {
+        if tick > ttl && !tick.is_zero() {
+            return Err(DataFusionError::Configuration(
+                "`tick` must be nonzero and <= `ttl`".to_string(),
+            ));
+        }
+        Ok(())
+    }
+}
+
+impl<K, V> TTLMap<K, V>
+where
+    K: Eq + Hash + Send + Sync + Clone + 'static,
+    V: Default + Clone + Send + Sync + 'static,
+{
+    // try_new creates a new TTLMap.
+    pub fn try_new(config: TTLMapConfig) -> Result<Self, DataFusionError> {
+        config.is_valid(config.tick, config.ttl)?;
+        let mut map = Self::_new(config);
+        map._start_gc();
+        Ok(map)
+    }
+
+    fn _new(config: TTLMapConfig) -> Self {
+        let stage_targets = Arc::new(DashMap::new());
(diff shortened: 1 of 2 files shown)


src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+//
+// Usage
+// ```rust
+// let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_mins(5) };
+// let ttl_map = TTLMap::new(config).await.unwrap();
+// let value = ttl_map.get_or_init(key, || initial_value).await;
+// ```
+//
+// TODO: If an existing entry is accessed, we don't extend its TTL. It's unclear if this is
+// necessary for any use cases. This functionality could be added if needed.
```
gabotechs: I can imagine this being useful, but I think can be done in a follow up if we find it necessary.

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+// Generally, keys in a bucket expire when the wheel makes a full rotation, making
+// the total TTL equal to the tick duration * buckets.len().
+//
+// Usage
+// ```rust
+// let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_mins(5) };
+// let ttl_map = TTLMap::new(config).await.unwrap();
+// let value = ttl_map.get_or_init(key, || initial_value).await;
+// ```
+//
```
gabotechs: For the doc comment to be valid, all the lines need to start with `//!`. Just so that you can see the difference in doc generation:

### Just the first line

Image

### All lines

Image

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+
+/// A bucket is a set of keys that are managed edited asynchronously.
+#[derive(Clone)]
+struct Bucket<K> {
+    // tx is used to send a BucketOp to the task
+    tx: UnboundedSender<BucketOp<K>>,
+    // This task is responsible for managing keys
```
gabotechs: Same here, for these to be valid doc comments that a doc generation can interpret, they need to be triple `///`

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+    }
+
+    /// task is responsible for managing a subset of keys in the map.
+    async fn task<V>(mut rx: UnboundedReceiver<BucketOp<K>>, data: Arc<DashMap<K, V>>)
+    where
+        V: Default + Clone + Send + Sync + 'static,
+    {
+        let mut shard = HashSet::new();
+        loop {
+            tokio::select! {
```
gabotechs: As you are selecting from just 1 task, this is equivalent to just omit `tokio::select!` and just do a normal iteration over the channel:

```rust
        while let Some(op) = rx.recv().await {
             // do stuff
        }
```

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+                Some(op) = rx.recv() => {
+                    match op {
+                        BucketOp::Insert{key} => {
+                            shard.insert(key);
+                        }
+                        BucketOp::Clear => {
+                           // Remove expired keys from the map.
+                           // TODO: it may be worth exploring if we can group keys by shard and do a batched
+                           // remove.
```
gabotechs: This would be nice, but I don't see `dashmap` having any API for batch deleting a list of keys. I think it should be fine though.

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+// TTLMap is a key-value store that automatically removes entries after a specified time-to-live.
```
gabotechs: Same as above, this probably should be a doc comment starting with `///`

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+// let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_mins(5) };
+// let ttl_map = TTLMap::new(config).await.unwrap();
+// let value = ttl_map.get_or_init(key, || initial_value).await;
+// ```
+//
+// TODO: If an existing entry is accessed, we don't extend its TTL. It's unclear if this is
+// necessary for any use cases. This functionality could be added if needed.
+use dashmap::{DashMap, Entry};
+use datafusion::error::DataFusionError;
+use futures::SinkExt;
```
gabotechs: this `use futures::SinkExt;` import seems unused, maybe remove it?

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// How often the map is checks for expired entries.
+    /// This must be less than `ttl`. It's recommended to set `ttl` to a multiple
+    /// of `tick`.
+    tick: Duration,
+    // The time-to-live for entries
+    ttl: Duration,
+
+    /// Optional metrics.
+    dash_map_lock_contention_time: Option<AtomicUsize>,
+    ttl_accounting_time: Option<AtomicUsize>,
```
gabotechs: As these are not really config parameters, probably they should be somewhere else. One idea that comes to mind is to make the compiler completely trim the out if we are not testing, something like this:

```rust
#[cfg(test)]
#[derive(Default)]
struct TTLMapMetrics {
    dash_map_lock_contention_time: AtomicUsize,
    ttl_accounting_time: AtomicUsize,
}
```

```rust
pub struct TTLMap<K, V> {
    ...
    // conf
    config: TTLMapConfig,

    #[cfg(test)]
    metrics: TTLMapMetrics,
}
```

```rust
    async fn _new(config: TTLMapConfig) -> Self {
        ...
        Self {
            ...
            #[cfg(test)]
            metrics: Default::default(),
        }
    }
```

```rust
    pub async fn get_or_init<F>(&self, key: K, f: F) -> V
        ...
        #[cfg(test)]
        self.metrics
            .dash_map_lock_contention_time
            .fetch_add(start.elapsed().as_nanos() as usize, Relaxed);
        ...
        if new_entry {
            ...
            #[cfg(test)]
            self.metrics
                .ttl_accounting_time
                .fetch_add(start.elapsed().as_nanos() as usize, Relaxed);
        }
   }
```

src/common/ttl_map.rs
```diff
@@ -0,0 +1,500 @@
...
+        loop {
+            tokio::time::sleep(period).await;
+            Self::gc(time.clone(), buckets);
+        }
+    }
+
+    fn gc(time: Arc<AtomicU64>, buckets: &Vec<Bucket<K>>) {
+        let index = time.load(std::sync::atomic::Ordering::SeqCst) % buckets.len() as u64;
+        // TODO: error
+        buckets[index as usize].clear().unwrap();
```
gabotechs: Any chance to remove this `.unwrap()` in favor of something else? ideally we should not panic intentionally.

jayshrivastava: Done. I think we can ignore the error since the we never close the `rx`. If the `rx` is dropped, it means the `tx` is being dropped too.

src/common/ttl_map.rs
jayshrivastava: Nice! I was wondering how to do this. Appreciate the tip :)

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Removed comment

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: I filed https://github.com/datafusion-contrib/datafusion-distributed/issues/101. Can look into this in a follow up PR!

src/common/ttl_map.rs
```diff
@@ -0,0 +1,481 @@
...
+//! When inserting key "A" at time=2:
+//! - Key "A" goes into bucket[(2-1) % 8] = bucket[1]
+//! - Key "A" will be expired when time advances to bucket[1] again
+//!
+//! Generally, keys in a bucket expire when the wheel makes a full rotation, making
+//! the total TTL equal to the tick duration * buckets.len().
+//!
+//! Usage
+//!
+//! let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_mins(5) };
```
jayshrivastava: I removed the ``` here because it was causing compile errors in CI

gabotechs: In Rust, doctests are expected to be compilable and runnable, I can think of two more elegant ways exposing this doc comment:

1. marking the `common` module visible to the outside only if `cfg(doctest)` or something like that (I actually don't know how to do that), and make the code in the doc test actually compile:
```rust
//! ```rust
//! # use std::time::Duration;
//! # use datafusion_distributed::common::ttl_map::{TTLMapConfig, TTLMap};
//! # async fn main() {
//!
//! let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_secs(5 * 60) };
//! let ttl_map = TTLMap::new(config).await.unwrap();
//! let value = ttl_map.get_or_init("foo", || "bar").await;
//! # }
//! ```
```
2. explicitly tell the Rust compiler to ignore it:
```rust
//! ```ignore
//! let config = TTLMapConfig { tick: Duration::from_secs(30), ttl: Duration::from_mins(5) };
//! let ttl_map = TTLMap::new(config).await.unwrap();
//! let value = ttl_map.get_or_init(key, || initial_value).await;
//! ```
```

Probably the second solution is good enough, and it will still display nicely in doc generators

src/common/ttl_map.rs
```diff
@@ -0,0 +1,481 @@
...
+{
+    // new creates a new TTLMap.
+    pub async fn new(config: TTLMapConfig) -> Result<Self, DataFusionError> {
+        config.is_valid(config.tick, config.ttl)?;
+        let mut map = Self::_new(config).await;
+        map._start_gc();
+        Ok(map)
+    }
+
+    async fn _new(config: TTLMapConfig) -> Self {
```
gabotechs: We can make this function synchronous, there's nothing asynchronous in it, so it should be fine and more ergonomic to the outside:


```suggestion
    fn _new(config: TTLMapConfig) -> Self {
```

src/common/ttl_map.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl<K, V> TTLMap<K, V>
+where
+    K: Eq + Hash + Send + Sync + Clone + 'static,
+    V: Default + Clone + Send + Sync + 'static,
+{
+    // new creates a new TTLMap.
+    pub async fn new(config: TTLMapConfig) -> Result<Self, DataFusionError> {
+        config.is_valid(config.tick, config.ttl)?;
+        let mut map = Self::_new(config).await;
```
gabotechs: It should be fine to make this synchronous:

```suggestion
    pub fn new(config: TTLMapConfig) -> Result<Self, DataFusionError> {
        config.is_valid(config.tick, config.ttl)?;
        let mut map = Self::_new(config);
```

src/common/ttl_map.rs
```diff
@@ -0,0 +1,481 @@
...
+        let gc_task = tokio::spawn(async move {
+            Self::run_gc_loop(time, period, &buckets).await;
+        });
+
+        self.gc_scheduler_task = Some(vec![gc_task]);
+    }
+
+    /// get_or_default executes the provided closure with a reference to the map entry for the given key.
+    /// If the key does not exist, it inserts a new entry with the default value.
+    pub async fn get_or_init<F>(&self, key: K, f: F) -> V
```
gabotechs: Same here, as there's nothing asynchronous here, we can remove this `async`:

```suggestion
    pub fn get_or_init<F>(&self, key: K, f: F) -> V
```

src/common/ttl_map.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+}
+
+impl<K, V> TTLMap<K, V>
+where
+    K: Eq + Hash + Send + Sync + Clone + 'static,
+    V: Default + Clone + Send + Sync + 'static,
+{
+    // new creates a new TTLMap.
+    pub async fn new(config: TTLMapConfig) -> Result<Self, DataFusionError> {
```
gabotechs: Usually, `::new()` methods that return a `Result<Self>` instead of a plain `Self` are named `try_new()`. Just something people in the Rust community tend to do.

src/common/ttl_map.rs
jayshrivastava: I tried a few iterations of this, but couldn't quite get it working. I went with `ignore` instead.
```
#[cfg(doc)]
pub mod common;
#[cfg(not(doc))]
mod common;
```

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

src/common/ttl_map.rs
jayshrivastava: Done

PR #100, opened 2025-08-18
Introduce `ConfigExtensionExt`, allowing the propagation of arbitrary `ConfigExtension`s across network boundaries

gabotechs:
DataFusion allows users to provide their own `ConfigExtension`s that get carried by the `SessionConfig` through the execution of a query.

https://github.com/apache/datafusion/blob/b84ddfde0689e8ecb9b71521fd95c7245d68d4ca/datafusion/common/src/config.rs#L1174-L1223

Users can thread their own `ConfigExtension` in order to provide additional context to their own execution nodes or other custom implementations.

---

In Distributed DataFusion, sessions are not automatically propagated across network boundaries, and any `ConfigExtension` set in the head stage will be lost when doing a network hop.

This PR allows propagating user-provided `ConfigExtension`s in a distributed context, using GRPC metadata for propagating these extensions across network boundaries.

For example, if a user has this `ConfigExtension`:

```rust
    extensions_options! {
        pub struct CustomExtension {
            pub foo: String, default = "".to_string()
            pub bar: usize, default = 0
            pub baz: bool, default = false
        }
    }

    impl ConfigExtension for CustomExtension {
        const PREFIX: &'static str = "custom";
    }
```

It can be:

1) added to the session in the head stage:

```rust
        let custom_extension = CustomExtension {
            foo: "foo".to_string(),
            bar: 1,
            baz: true,
        };
-       state.config_mut().options().extensions.insert(custom_extension); // <- this is normal datafusion
+       state.add_distributed_option_extension(custom_extension)?;
```

2) recursively propagated across network calls in the Arrow Flight Endpoint

```rust
        #[async_trait]
        impl SessionBuilder for CustomSessionBuilder {
            async fn session_state(&self, mut state: SessionState) -> Result<SessionState, DataFusionError> {
+               state.propagate_distributed_option_extension::<CustomExtension>()?;
                Ok(state)
            }
        }
```

With just the two lines above, any user-provided `CustomExtension` will be able to cross network boundaries and be accessible anywhere in the plan, regardless of the physical node executing the query.

### DIFF
--- a/src/config_extension_ext.rs
+++ b/src/config_extension_ext.rs
@@ -278,9 +278,9 @@ mod tests {
         assert!(flight_metadata.is_some());
 
         let metadata = &flight_metadata.unwrap().0;
-        assert!(metadata.contains_key("x-datafusion-distributed-custom.foo"));
-        assert!(metadata.contains_key("x-datafusion-distributed-custom.bar"));
-        assert!(metadata.contains_key("x-datafusion-distributed-custom.baz"));
+        assert!(metadata.contains_key("x-datafusion-distributed-config-custom.foo"));
+        assert!(metadata.contains_key("x-datafusion-distributed-config-custom.bar"));
+        assert!(metadata.contains_key("x-datafusion-distributed-config-custom.baz"));
 
         let get = |key: &str| metadata.get(key).unwrap().to_str().unwrap();
         assert_eq!(get("x-datafusion-distributed-config-custom.foo"), "");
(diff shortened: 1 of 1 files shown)


src/config_extension_ext.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// impl SessionBuilder for MyCustomSessionBuilder {
+    ///     async fn session_state(&self, mut state: SessionState) -> Result<SessionState, DataFusionError> {
+    ///         // while providing this MyCustomSessionBuilder to an Arrow Flight endpoint, it will
+    ///         // know how to deserialize the CustomExtension from the gRPC metadata.
+    ///         state.propagate_distributed_option_extension::<CustomExtension>()?;
+    ///         Ok(state)
+    ///     }
+    /// }
+    /// ```
+    fn propagate_distributed_option_extension<T: ConfigExtension + Default>(
```
fmonjalet: nit:
```suggestion
    fn retrieve_distributed_option_extension<T: ConfigExtension + Default>(
```
Or something else that makes it clear that we read the extensions into the context, as opposed to injecting the options into the headers that will be propagated (which is what I understood when reading "propagate")

src/config_extension_ext.rs
```diff
@@ -0,0 +1,456 @@
...
+            DataFusionError::Internal(format!("Failed to add config extension: {err}"))
+        }
+        let mut meta = HeaderMap::new();
+
+        for entry in t.entries() {
+            if let Some(value) = entry.value {
+                meta.insert(
+                    HeaderName::from_str(&format!(
+                        "{}{}.{}",
+                        FLIGHT_METADATA_PREFIX,
```
fmonjalet: nit: you may want to add a `config-` suffix to the prefix here so that we can carry other information using the same `x-datafusion-distributed-` header prefix without conflict.

(making things up: we want to propagate the map of nodes where work has already been scheduled so that we make decisions based on this, we may want to reuse the `x-datafusion-distributed-` prefix but scope the information to `workers-`)

src/config_extension_ext.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+
+        let mut result = T::default();
+        let mut found_some = false;
+        for (k, v) in flight_metadata.0.iter() {
+            let key = k.as_str().trim_start_matches(FLIGHT_METADATA_PREFIX);
+            if key.starts_with(T::PREFIX) {
+                found_some = true;
+                result.set(
+                    &key.trim_start_matches(T::PREFIX).trim_start_matches("."),
```
fmonjalet: We never checked the key started with `T::PREFIX` + `.`, maybe we should check this before to avoid surprising behaviors

src/plan/arrow_flight_read.rs
```diff
@@ -195,6 +203,7 @@ impl ExecutionPlan for ArrowFlightReadExec {
                 let channel_manager_capture = channel_manager.clone();
                 let schema = schema.clone();
                 let query_id = query_id.clone();
+                let flight_metadata = flight_metadata.clone().unwrap_or_default().as_ref().clone();
```
fmonjalet: minor: why the double clone?

gabotechs: First for taking ownwership, of the Arc, second for cloning the struct inside the Arc. I can make it just 1 clone

PR #108, opened 2025-08-22
do_get: use TTL map to store task state

jayshrivastava:
### ttl_map: add remove() method and tests
The remove() method allows explicit deletion of entries from the TTLMap before they expire naturally.

### do_get: use TTL map to store task state

The do_get call now evicts task state from the map after N calls where N
is the number of partitions. This is an approximation because we don't
track that unique partition ids are used, so we might evict early in case
of retries.

The TTLMap will also GC orphaned task state after the configured TTL period.

### DIFF
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -14,7 +14,9 @@ use arrow_flight::Ticket;
 use datafusion::execution::SessionState;
 use futures::TryStreamExt;
 use prost::Message;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use tokio::sync::OnceCell;
 use tonic::metadata::MetadataMap;
 use tonic::{Request, Response, Status};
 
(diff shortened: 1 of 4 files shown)


src/flight_service/service.rs
```diff
@@ -35,18 +35,21 @@ pub struct ArrowFlightEndpoint {
     pub(super) channel_manager: Arc<ChannelManager>,
     pub(super) runtime: Arc<RuntimeEnv>,
     #[allow(clippy::type_complexity)]
-    pub(super) stages: DashMap<StageKey, Arc<OnceCell<(SessionState, Arc<ExecutionStage>)>>>,
+    pub(super) stages: TTLMap<StageKey, Arc<OnceCell<TaskData>>>,
```
jayshrivastava, line 34: Oncecell must be in an Arc so we can copy the oncecell after it has been initialized without resetting it

src/flight_service/do_get.rs
```diff
@@ -37,6 +39,19 @@ pub struct DoGet {
...
+/// TaskData stores state for a single task being executed by this Endpoint. It may be shared
+/// by concurrent requests for the same task which execute separate partitions.
+pub struct TaskData {
+    pub(super) state: SessionState,
+    pub(super) stage: Arc<ExecutionStage>,
+    /// Initialized to the total number of partitions in the task (not only tasks in the partition
+    /// group). This is decremented for each request to the endpoint for this task. Once this count
+    /// is zero, the task is likely complete. This count does not account for retried requests
+    /// for the same partition.
+    approx_partitions_remaining: Arc<AtomicUsize>,
```
gabotechs: 🤔 This says "approx", but we are relying on an exact number being here for triggering the key removal logic. Maybe just choosing a more descriptive name for what it is exactly? something like `partitions_not_yet_started`

src/flight_service/do_get.rs
```diff
@@ -134,10 +151,206 @@ impl ArrowFlightEndpoint {
...
+        if remaining_partitions <= 1 {
+            self.stages.remove(key.clone());
+        }
+
+        Ok(stage_data.clone())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
gabotechs, line 177: It looks like there's a couple of clippy rule violations in this test, for some reason pipelines are not catching them.

If you try running:

```
cargo clippy --tests --features integration
```

you'll see several errors

PR #109, opened 2025-08-25
Refactor arrow_flight_read.rs and friends

gabotechs:
In preparation of propagating execution metrics up in the stages stack, I ended up cleaning up some ArrowFightExec stuff.

The `arrow_flight_read.rs` file along with the `do_get.rs` will need to be extended with execution metrics propagation, so cleaning this up now will make a future PR smaller and more digestible.


src/plan/combined.rs
gabotechs, line 1: `futures::stream::select()` does exactly this, so I'm using that instead in this PR.

Issue #4, opened 2025-06-24
Build a Distributed-DataFusion-CLI


NGA-TRAN:
We currently use the script `./scripts/launch_python_arrowflightsql_client.sh` to start a `Python` client for issuing queries via `ArrowFlightSQL`.

To improve the developer experience, we’d like to build a dedicated Client CLI, similar in spirit to `datafusion-cli`. This would allow for easier experimentation with queries, viewing results, and running EXPLAIN commands—directly from the command line.

The goal is to create a more intuitive and interactive interface for users working with the distributed DataFusion setup.

PR #111, opened 2025-08-25
Add `localhost_run.rs` and `localhost_worker.rs` examples

gabotechs:
Adds a basic example for running distributed queries on top of a couple of parquet files, the same we use for tests.

### DIFF
--- a/examples/localhost.md
+++ b/examples/localhost.md
@@ -36,8 +36,8 @@ Now, DataFusion queries can be issued using these workers as part of the cluster
 cargo run --example localhost_run -- 'SELECT count(*), "MinTemp" FROM weather GROUP BY "MinTemp"' --cluster-ports 8080,8081
 ```
 
-The head stage will be executed locally in the same process as that `cargo run` command, but further stages will be
-delegated to the workers running on ports 8080 and 8081.
+The head stage (the one that outputs data to the user) will be executed locally in the same process as that `cargo run`
+command, but further stages will be delegated to the workers running on ports 8080 and 8081.
 
 Additionally, the `--explain` flag can be passed to render the distributed plan:
 
@@ -50,4 +50,40 @@ cargo run --example localhost_run -- 'SELECT count(*), "MinTemp" FROM weather GR
 Two tables are available in this example:
 
 - `flights_1m`: Flight data with 1m rows
+
+```
+FL_DATE [INT32]
+DEP_DELAY [INT32]
+ARR_DELAY [INT32]
+AIR_TIME [INT32]
+DISTANCE [INT32]
+DEP_TIME [FLOAT]
+ARR_TIME [FLOAT]
+```
+
 - `weather`: Small dataset of weather data
+
+```
+MinTemp [DOUBLE]
+MaxTemp [DOUBLE]
+Rainfall [DOUBLE]
+Evaporation [DOUBLE]
+Sunshine [BYTE_ARRAY]
+WindGustDir [BYTE_ARRAY]
+WindGustSpeed [BYTE_ARRAY]
+WindDir9am [BYTE_ARRAY]
+WindDir3pm [BYTE_ARRAY]
+WindSpeed9am [BYTE_ARRAY]
+WindSpeed3pm [INT64]
+Humidity9am [INT64]
+Humidity3pm [INT64]
+Pressure9am [DOUBLE]
+Pressure3pm [DOUBLE]
+Cloud9am [INT64]
+Cloud3pm [INT64]
+Temp9am [DOUBLE]
+Temp3pm [DOUBLE]
+RainToday [BYTE_ARRAY]
+RISK_MM [DOUBLE]
+RainTomorrow [BYTE_ARRAY]
+```


examples/localhost.md
```diff
@@ -0,0 +1,53 @@
...
+```shell
+cargo run --example localhost_run -- 'SELECT count(*), "MinTemp" FROM weather GROUP BY "MinTemp"' --cluster-ports 8080,8081 --explain
+```
+
+### Available tables
+
+Two tables are available in this example:
+
+- `flights_1m`: Flight data with 1m rows
+- `weather`: Small dataset of weather data
```
NGA-TRAN, line 64: Are the schema of these tables defined somewhere in the code and can we provide a link there? If not yet, can you have a brief description for table columns so user know to issue query?

A few tricks I did for myself to know them in case you do not want to document the schema but show user how to get them.

- I usually run `datafusion-cli` which is CLI for single node. In this doc, maybe you use `cargo run ...` instead?
- Inside the datafusion-cli, I create external table for the parquet

```SQL
CREATE EXTERNAL TABLE flight
STORED AS PARQUET
LOCATION 'path/to/datafusion-distributed/testdata/flights-1m.parquet';
```

- Then there are 2 ways to get the schema:
   1. Quick dirty trick but won't know column data type
        ``` SQL
        SELECT * FROM flight LIMIT 0;
        +---------+-----------+-----------+----------+----------+----------+----------+
         | FL_DATE | DEP_DELAY | ARR_DELAY | AIR_TIME | DISTANCE | DEP_TIME | ARR_TIME |
         +---------+-----------+-----------+----------+----------+----------+----------+
         +---------+-----------+-----------+----------+----------+----------+----------+
         0 row(s) fetched.
        ```
   3. List all columns and data type:
       ```SQL
       show columns from flight;
       +---------------+--------------+------------+-------------+-----------+-------------+
        | table_catalog | table_schema | table_name | column_name | data_type | is_nullable |
       +---------------+--------------+------------+-------------+-----------+-------------+
        | datafusion    | public       | flight     | FL_DATE     | Date32    | YES         |
        | datafusion    | public       | flight     | DEP_DELAY   | Int16     | YES         |
        | datafusion    | public       | flight     | ARR_DELAY   | Int16     | YES         |
        | datafusion    | public       | flight     | AIR_TIME    | Int16     | YES         |
        | datafusion    | public       | flight     | DISTANCE    | Int16     | YES         |
        | datafusion    | public       | flight     | DEP_TIME    | Float32   | YES         |
        | datafusion    | public       | flight     | ARR_TIME    | Float32   | YES         |
        +---------------+--------------+------------+-------------+-----------+-------------+
        7 row(s) fetched. 
        ```

examples/localhost.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+The head stage will be executed locally in the same process as that `cargo run` command, but further stages will be
+delegated to the workers running on ports 8080 and 8081.
+
+Additionally, the `--explain` flag can be passed to render the distributed plan:
+
+```shell
+cargo run --example localhost_run -- 'SELECT count(*), "MinTemp" FROM weather GROUP BY "MinTemp"' --cluster-ports 8080,8081 --explain
+```
```
NGA-TRAN, line 46: These commands are so cool. Do you think for near future work, we are ready to work on supporting `distributed-datafusion-cli` defined in https://github.com/datafusion-contrib/datafusion-distributed/issues/4?

Maybe we add a new folder `distributed-datafusion-cli` similar to `datafusion-cli` to support this?

examples/localhost_run.rs
```diff
@@ -0,0 +1,106 @@
...
+struct Args {
+    #[structopt()]
+    query: String,
+
+    // --cluster-ports 8080,8081,8082
+    #[structopt(long = "cluster-ports", use_delimiter = true)]
+    cluster_ports: Vec<u16>,
+
+    #[structopt(long)]
+    explain: bool,
```
NGA-TRAN, line 29: Nice

examples/localhost.md
```diff
@@ -0,0 +1,53 @@
...
+### Issuing a distributed SQL query
+
+Now, DataFusion queries can be issued using these workers as part of the cluster.
+
+```shell
+cargo run --example localhost_run -- 'SELECT count(*), "MinTemp" FROM weather GROUP BY "MinTemp"' --cluster-ports 8080,8081
+```
+
+The head stage will be executed locally in the same process as that `cargo run` command, but further stages will be
+delegated to the workers running on ports 8080 and 8081.
```
NGA-TRAN: Can you define `head stage`? This makes me ask myself which part of the plan runs in head stage

examples/localhost.md
```diff
@@ -0,0 +1,53 @@
+# Localhost workers example
```
NGA-TRAN, line 1: I have checked out your branch and run the below commands and get query result and explain back. Super cool!

examples/localhost.md
gabotechs: Makes sense! added them to the `localhost.md` file

examples/localhost.md
gabotechs: If you visualize a plan as a tree of stages, the "head" or "root" stage is the top-level one, the first one counting from top to bottom. Added a clarifying comment.

examples/localhost.md
gabotechs: 🤔 I'm not sure what value would `distributed-datafusion-cli` bring on top of the normal `datafusion-cli`. As this is just a library for distributing queries, the concept of CLI becomes less relevant in this context.

If people anyways want to use the CLI, hopefully we can just reuse the normal `datafusion-cli` rather than building our own thing.

NGA-TRAN: Reusing `datafusion-cli` is a good option as long as we provide a good way to have a default (e.g 3 workers) and easy-custom distributed settings

PR #114, opened 2025-09-02
Add README.md and LICENSE.txt

gabotechs:
Adds a README.md and a LICENSE file

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ capabilities with minimal changes.
 
 Before diving into the architecture, it's important to clarify some terms and what they mean:
 
-- `worker`: a physical machine listening to serialized plans over an Arrow Flight interface.
+- `worker`: a physical machine listening to serialized execution plans over an Arrow Flight interface.
 - `network boundary`: a node in the plan that streams data from a network interface rather than directly from its
   children. Implemented as an `ArrowFlightReadExec` physical DataFusion node.
 - `stage`: a portion of the plan separated by a network boundary from other parts of the plan. Implemented as any
@@ -37,8 +37,8 @@ Before diving into the architecture, it's important to clarify some terms and wh
 A distributed DataFusion query is executed in a very similar fashion as a normal DataFusion query with one key
 difference:
 
-The physical plan is divided into stages that can be executed on different workers and exchange data using Arrow
-Flight. All of this is done at the physical plan level, and is implemented as a `PhysicalOptimizerRule` that:
+The physical plan is divided into stages, each stage is assigned tasks that run in parallel in different workers. All of
+this is done at the physical plan level, and is implemented as a `PhysicalOptimizerRule` that:
 
 1. Inspects the non-distributed plan, placing network boundaries (`ArrowFlightReadExec` nodes) in the appropriate places
 2. Based on the placed network boundaries, divides the plan into stages and assigns tasks to them.
@@ -158,8 +158,8 @@ will issue 3 concurrent Arrow Flight requests to the appropriate physical nodes.
 This means that:
 
 1. The head stage is executed normally as if the query was not distributed.
-2. Upon calling `.execute()` on the `ArrowFlightReadExec`, instead of recursively calling `.execute()` on its children,
-   the child subplan will be serialized and sent over the wire to another node.
+2. Upon calling `.execute()` on `ArrowFlightReadExec`, instead of propagating the `.execute()` call on its child,
+   the subplan is serialized and sent over the wire to be executed on another worker.
 3. The next node, which is hosting an Arrow Flight Endpoint listening for gRPC requests over an HTTP server, will pick
    up the request containing the serialized chunk of the overall plan and execute it.
 4. This is repeated for each stage, and data will start flowing from bottom to top until it reaches the head stage.


README.md
```diff
@@ -0,0 +1,250 @@
...
+
+The plan is immediately executable, and the same process that planned the distributed query can start executing the head
+stage (stage 2). The `ArrowFlightReadExec` in that stage will know from which tasks to gather data from stage 1, and
+will issue 3 concurrent Arrow Flight requests to the appropriate physical nodes.
+
+This means that:
+
+1. The head stage is executed normally as if the query was not distributed.
+2. Upon calling `.execute()` on the `ArrowFlightReadExec`, instead of recursively calling `.execute()` on its children,
+   they will be serialized and sent over the wire to another node.
```
NGA-TRAN: I asked myself what the word `they` means here. The next sentence (point 3) lets me know it is a `plan`. Maybe explain `they` here will make it clearer? And it is a `sub-plan` that needs to be executed in stage1, right?

What if the plan has 3 stages: stage3 us the head stage that will get data from stage2 and stage2 gets data from stage1? The subplan stage3 sends to stage2 also include the sub-sub-plan that stage2 will send to stage1, right?

README.md
```diff
@@ -0,0 +1,250 @@
...
+
+- [localhost_worker.rs](examples/localhost_worker.rs): code that spawns an Arrow Flight Endpoint listening for physical
+  plans over the network.
+- [localhost_run.rs](examples/localhost_run.rs): code that distributes a query across the spawned Arrow Flight Endpoints
+  and executes it.
+
+The integration tests also provide an idea about how to use the library and what can be achieved with it:
+
+- [tpch_validation_test.rs](tests/tpch_validation_test.rs): executes all TPCH queries and performs assertions over the
+  distributed plans.
```
NGA-TRAN: Maybe add bit more saying we validate by comparing distributed results with single node vanilla ones; and also saying we use scale factor 0.001?

README.md
gabotechs: Yes, that's right. Here "they" refers to the immediate previous word "children". As you pointed out, I think this can be clearer, let me fix that

README.md
```diff
@@ -0,0 +1,260 @@
...
+  a familiar API for building applications.
+- Unopinionated about networking. This crate does not take any opinion about the networking stack, and users are
+  expected to leverage their own infrastructure for hosting DataFusion nodes.
+- No coordinator-worker architecture. To keep infrastructure simple, any node can act as a coordinator or a worker.
+
+## Architecture
+
+Before diving into the architecture, it's important to clarify some terms and what they mean:
+
+- `worker`: a physical machine listening to serialized plans over an Arrow Flight interface.
```
geoffreyclaude: logical or execution plans?

README.md
```diff
@@ -0,0 +1,260 @@
...
+  children. Implemented as an `ArrowFlightReadExec` physical DataFusion node.
+- `stage`: a portion of the plan separated by a network boundary from other parts of the plan. Implemented as any
+  other physical node in DataFusion.
+- `task`: a unit of work inside a stage that executes a subset of its partitions in a specific worker.
+- `subplan`: a slice of the overall plan
+
+A distributed DataFusion query is executed in a very similar fashion as a normal DataFusion query with one key
+difference:
+
+The physical plan is divided into stages that can be executed on different workers and exchange data using Arrow
```
geoffreyclaude: Not super clear: who exchanges data? The workers? The stages? The plan?

README.md
```diff
@@ -0,0 +1,260 @@
...
+
+The plan is immediately executable, and the same process that planned the distributed query can start executing the head
+stage (stage 2). The `ArrowFlightReadExec` in that stage will know from which tasks to gather data from stage 1, and
+will issue 3 concurrent Arrow Flight requests to the appropriate physical nodes.
+
+This means that:
+
+1. The head stage is executed normally as if the query was not distributed.
+2. Upon calling `.execute()` on the `ArrowFlightReadExec`, instead of recursively calling `.execute()` on its children,
+   the child subplan will be serialized and sent over the wire to another node.
```
geoffreyclaude: This seems like a convoluted way of saying that ` ArrowFlightReadExec` is just a standard input node, with the child plan as part of the request?

README.md
```diff
@@ -0,0 +1,260 @@
...
+- `task`: a unit of work inside a stage that executes a subset of its partitions in a specific worker.
+- `subplan`: a slice of the overall plan
+
+A distributed DataFusion query is executed in a very similar fashion as a normal DataFusion query with one key
+difference:
+
+The physical plan is divided into stages that can be executed on different workers and exchange data using Arrow
+Flight. All of this is done at the physical plan level, and is implemented as a `PhysicalOptimizerRule` that:
+
+1. Inspects the non-distributed plan, placing network boundaries (`ArrowFlightReadExec` nodes) in the appropriate places
```
geoffreyclaude, line 43: "in the appropriate places" is the core problem: might be worth it to expand a bit on this. Also, how is the number of partitions chosen now that it can't just be the number of CPU on the single node?

README.md
gabotechs: added

README.md
gabotechs: changed that phrase, it should be clearer now

README.md
gabotechs: Rephrased that to make it a bit clearer

README.md
gabotechs: I'd really like to give a more informative description about what's happening here, but I'm afraid this is in so early stages that whatever I put here might changed several times in the following days, so keeping it ambiguous for now might be the best card to play.

Issue #110, opened 2025-08-25
"partition not used yet" panics during tests


gabotechs:
Currently, if the integration tests are run:

```
cargo test --features integration --test distributed_aggregation
```

Even if the test succeed, several panics happen in the ArrowFlightEndpoint handler:

```
thread 'tests::distributed_aggregation' panicked at /Users/foo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/datafusion-physical-plan-49.0.0/src/repartition/mod.rs:694:22:
partition not used yet
```

PR #120, opened 2025-09-03
Fix panics in tests and un-ignore working tests

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/110

Previously, the `ArrowFlightReadExec` node could choose any partitioning scheme regardless of the child that was passed, but in reality it should be exposing the child partitioning scheme unmodified, as it's pretty just a network passthrough.

Also, `ArrowFLightReadExec` is not really prepared to have something other than a `RepartitionExec` as a child, so this PR also accounts for that in the tests while manually building a distributed plan.

Additionally, some tests are unignored, as they are working now.


src/test_utils/plan.rs
```diff
@@ -1,65 +0,0 @@
-use crate::{ArrowFlightReadExec, DistributedPhysicalOptimizerRule};
-use datafusion::common::plan_err;
-use datafusion::common::tree_node::{Transformed, TreeNode};
-use datafusion::error::DataFusionError;
-use datafusion::physical_expr::Partitioning;
-use datafusion::physical_plan::aggregates::{AggregateExec, AggregateMode};
-use datafusion::physical_plan::ExecutionPlan;
-use std::sync::Arc;
-
-pub fn distribute_aggregate(
```
gabotechs, line 10: They way function was distributing a plan for the tests is just wrong, so now the normal planner is used.

tests/distributed_aggregation.rs
```diff
@@ -48,21 +52,23 @@ mod tests {
...
+          │partitions [out:1  <-- in:3  ]   PartitionIsolatorExec [providing upto 1 partitions]
+          │partitions [out:3            ]     ArrowFlightReadExec: Stage 1  
           └──────────────────────────────────────────────────
-            ┌───── Stage 1   Task: partitions: 0,unassigned]
-            │partitions [out:1            ] DataSourceExec: file_groups={1 group: [[/testdata/weather.parquet]]}, projection=[RainToday], file_type=parquet
+            ┌───── Stage 1   Task: partitions: 0,unassigned],Task: partitions: 1,unassigned],Task: partitions: 2,unassigned]
+            │partitions [out:3  <-- in:1  ] RepartitionExec: partitioning=RoundRobinBatch(3), input_partitions=1
+            │partitions [out:1  <-- in:1  ]   PartitionIsolatorExec [providing upto 1 partitions]
+            │partitions [out:1  <-- in:1  ]     AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))]
+            │partitions [out:1            ]       DataSourceExec: file_groups={1 group: [[/testdata/weather.parquet]]}, projection=[RainToday], file_type=parquet
```
NGA-TRAN, line 71: I am glad we already have available tests to verify the fix

PR #122, opened 2025-09-04
Bigger TPCH tests

gabotechs:
Uses a bigger scale factor in the TPCH tests so that queries get more heavily distributed. 

As these tests are more expensive now, the run only with the `tpch` feature, and there's a dedicated workflow in GitHub actions specific for it.


src/errors/mod.rs
```diff
@@ -40,10 +40,7 @@ pub fn tonic_status_to_datafusion_error(status: &tonic::Status) -> Option<DataFu
     }
 
     match DataFusionErrorProto::decode(status.details()) {
-        Ok(err_proto) => {
-            dbg!(&err_proto);
-            Some(err_proto.to_datafusion_err())
-        }
+        Ok(err_proto) => Some(err_proto.to_datafusion_err()),
```
gabotechs, line 43: Not related to this PR, but saw some `dbg!` statement pending here.

tests/common.rs
gabotechs, line 1: I moved the contents of this file to `tpch_validation_test.rs`. As the `common` module is not hidden behind the "integration" flag, compilations can fail as the functions used by it are do hidden behind the "integration" flag.

As this is not really common code, and it's just used in `tpch_validation_test.rs`, I just moved it there.

tests/tpch_validation_test.rs
```diff
@@ -897,11 +1131,17 @@ mod tests {
...
+        ctx2.state_ref()
+            .write()
+            .config_mut()
+            .options_mut()
+            .execution
+            .target_partitions = 15;
 
         // Context 1: Non-distributed execution.
-        let config1 = SessionConfig::new().with_target_partitions(3);
+        let config1 = SessionConfig::new().with_target_partitions(15);
```
NGA-TRAN, line 1144: Are you suggesting keeping the number odd (15) to validate test behavior?

For unit tests like with_target_partitions, I believe we already have good coverage across different partition counts. For this result validation test, though, it makes sense to use a more structured setup—like setting the number of workers to a power of 2 (which you’ve set to 8 👍), and the number of partitions to a multiple of that. So I’d suggest using 16 partitions here.

tests/common.rs
jayshrivastava: Sounds good. Thanks!

src/test_utils/tpch.rs
```diff
@@ -158,33 +156,33 @@ where
...
+        ($generator:ident, $arrow:ident, $name:literal) => {
+            let data_dir = data_dir.join($name);
+            fs::create_dir_all(data_dir.clone()).expect("Failed to create data directory");
+            // create three partitions for the table
+            (1..=parts).for_each(|part| {
+                generate_table(
+                    // TODO: Consider adjusting the partitions and batch sizes.
+                    $arrow::new($generator::new(sf, part, parts)).with_batch_size(1000),
+                    &format!("{part}"),
+                    &data_dir,
```
jayshrivastava, line 173: 👍🏽

.github/workflows/ci.yml
```diff
@@ -33,6 +33,13 @@ jobs:
       - uses: ./.github/actions/setup
       - run: cargo test --features integration
 
+  tpch-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - run: cargo test --features integration --features tpch --test tpch_validation_test
```
jayshrivastava: mega nit: since `tpch` depends on `integration` in the Cargo.toml, I think you can remove `--features integration` here.

tests/tpch_validation_test.rs
gabotechs: The number of workers does not matter that much here, as all of them are localhost (this test currently uses 2).

The rationale for using 15, and odd number, is to stress out the case when the last task needs to deal with an odd number of partitions (1 in this case), something that was not happening before and we probably should test anyways.

.github/workflows/ci.yml
gabotechs: nice! I didn't knew you could do that

tests/tpch_validation_test.rs
NGA-TRAN: Actually because these are `validation` tests, it is good to keep it odd to capture edge bugs. For benchmarking and performance tests, we want the settings that makes more sense

PR #124, opened 2025-09-04
File name and folder restructure

gabotechs:
This PR provides a better organization for the project colocating into modules the different pieces based on the features they are providing.

The new folder structure, as it colocates things related to each other more efficiently, allows to tweak the privacy of some structs/methods so that they are not expose to outside unrelated modules, and therefore reducing the chances of future coupling.

All the code is mostly untouched, only a couple of code changes are shipped for making this possible:
- The `ExecutionTask` now has two forms:
  - `ExecutionTask`: struct with the `Url` already parsed and the partition groups already as `usize`s, so that we don't need to do juggling in the code for parsing Urls or casing u64s to usizes
  - `ExecutionTaskProto`: the protobuf representation, decoupled from the actual useful struct
- The test in the `do_get` method was using directly the proto representation of the `ExecutionStage` and the `ExecutionTask`, but they are no longer exposed to the outside, so instead the normal structs are used, which then are converted to protos with the `proto_to_stage` function. This actually results in a nice cleanup that saves some lines of code.

```
common/
    mod.rs
    composed_extension_codec.rs
    ttl_map.rs
errors/
    mod.rs
    arrow_error.rs
    datafusion_error.rs
    io_error.rs
    objectstore_error.rs
    parquet_error.rs
    parser_error.rs
    schema_error.rs
execution_plans/
    mod.rs
    arrow_flight_read.rs
    partition_isolator.rs
    stage.rs
flight_service/
    mod.rs
    do_get.rs
    service.rs
    session_builder.rs
protobuf/
    mod.rs
    distributed_codec.rs
    stage_proto.rs
    user_codec.rs
test_utils/
    mod.rs
    insta.rs
    localhost.rs
    mock_exec.rs
    parquet.rs
    tpch.rs
lib.rs
channel_resolver_ext.rs
config_extension_ext.rs
distributed_ext.rs
distributed_physical_optimizer_rule.rs
```


src/distributed_physical_optimizer_rule.rs
```diff
@@ -188,7 +187,7 @@ pub fn can_be_divided(plan: &Arc<dyn ExecutionPlan>) -> Result<bool> {
 #[cfg(test)]
 mod tests {
     use crate::assert_snapshot;
-    use crate::physical_optimizer::DistributedPhysicalOptimizerRule;
+    use crate::distributed_physical_optimizer_rule::DistributedPhysicalOptimizerRule;
```
NGA-TRAN, line 190: I like you rename it to `distributed_physical_optimizer_rule`. Make it clearer

src/distributed_physical_optimizer_rule.rs
```diff
@@ -55,7 +54,7 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
         _config: &ConfigOptions,
     ) -> Result<Arc<dyn ExecutionPlan>> {
         // We can only optimize plans that are not already distributed
-        if plan.as_any().is::<ExecutionStage>() {
+        if plan.as_any().is::<StageExec>() {
```
NGA-TRAN, line 57: `StageExec` makes it more like DF style 👍

src/flight_service/do_get.rs
```diff
@@ -289,38 +287,9 @@ mod tests {
...
-                            partition_method: Some(PartitionMethod::RoundRobin(partitions as u64)),
-                        }),
-                    })
-                )
-            ),
-        }
+    // Helper to create a mock physical plan
+    fn create_mock_physical_plan(partitions: usize) -> Arc<dyn ExecutionPlan> {
+        let node = Arc::new(EmptyExec::new(SchemaRef::new(Schema::empty())));
+        Arc::new(RepartitionExec::try_new(node, Partitioning::RoundRobinBatch(partitions)).unwrap())
```
NGA-TRAN, line 293: This isn’t directly related to this PR, but might be something we consider updating or adding in a follow-up.

This `RoundRobinBatch` got me thinking 🤔. On one hand, it's a valid scenario we want our tests to cover. On the other, round-robin repartitioning isn’t ideal from a performance standpoint. How about we introduce a variety of partitioning schemes in our mock tests? That way, we can surface more realistic and appropriate cases while still including edge scenarios like this one.

NGA-TRAN (approved):
Nice refactor and rename.

src/distributed_physical_optimizer_rule.rs
gabotechs: yeah, I've tried to rename files to be a bit more consistent with the content.

src/distributed_physical_optimizer_rule.rs
gabotechs: Yeah, it seems like an unwritten rule to suffix all `ExecutionPlan` implementations with `*Exec`

src/flight_service/do_get.rs
gabotechs: Yeah, we should do something with RoundRobin repartitions...

We probably should extend our parquet files test datasets so that we can cover more cases and more esoteric distribution patterns.

For example, it would be nice to have two test parquet files with related data that we can use to stress joins in weird ways.

gabotechs: Probably something to follow up soon

PR #129, opened 2025-09-08
Add support for in-memory TPCH tests

gabotechs:
The original DataFusion repo has the ability to run TPCH benchmarks by pre-loading the TPCH dataset in-memory. This is good because it removes disk IOs out of the picture during benchmarks.

Unfortunately, this project cannot serialize/deserialize in-memory nodes, as that would imply sending over the wire the full data loaded in memory.

This PR introduces a new node that overcomes this by storing the data in-memory in a global `CACHE` variable rather than in the node itself, that way, the node can be serialized and sent over the wire without baking all the in-memory data in. Upon receiving a request, the node will load the data from the global cache, which works no matter if the node was deserialized or was built natively.

An in-memory benchmark can be executed with the `-m` flag:

```shell
cargo run -p datafusion-distributed-benchmarks --release -- tpch -m  --path benchmarks/data/tpch_sf1
```

### DIFF
--- a/benchmarks/src/util/memory.rs
+++ b/benchmarks/src/util/memory.rs
@@ -24,6 +24,16 @@ static CACHE: LazyLock<DashMap<Key, Value>> = LazyLock::new(DashMap::default);
 
 /// Caches all the record batches in a global [CACHE] on the first run, and serves
 /// them from the cache in any subsequent run.
+/// The order of events looks like this:
+///   1. A first query is run.
+///   2. Data is not cached, so it is gathered from the underlying node.
+///   3. The cache is populated with the recently gathered data.
+///   4. A second query is run.
+///   5. The cache is hit, and data is returned from there.
+///
+/// The cache key includes the result of "explaining" the underlying node, so different
+/// nodes applying different filters under the same parquet files will be cached
+/// independently.
 #[derive(Debug, Clone)]
 pub struct InMemoryCacheExec {
     inner: Arc<dyn ExecutionPlan>,


benchmarks/src/util/memory.rs
```diff
@@ -1,21 +1,215 @@
...
+
+/// Caches all the record batches in a global [CACHE] on the first run, and serves
+/// them from the cache in any subsequent run.
+#[derive(Debug, Clone)]
+pub struct InMemoryCacheExec {
+    inner: Arc<dyn ExecutionPlan>,
+}
+
+extensions_options! {
+    /// Marker used by the [InMemoryCacheExec] that determines wether its fine
```
NGA-TRAN:
```suggestion
    /// Marker used by the [InMemoryCacheExec] that determines whether its fine
```

benchmarks/src/util/memory.rs
```diff
@@ -1,21 +1,215 @@
...
+        };
+        let Ok(_) = proto.encode(buf) else {
+            return plan_err!("no InMemoryDataSourceExecProto");
+        };
+
+        Ok(())
+    }
+}
+
+/// Wraps any plan without children with an [InMemoryCacheExec] node.
```
NGA-TRAN, line 195: Does it mean `InMemoryCacheExec` is only applied to plan without children and those plans should be datasource that reads data or empty one?

benchmarks/src/tpch/run.rs
```diff
@@ -191,8 +200,18 @@ impl RunOpt {
...
-        self.register_tables(&single_node_ctx).await?;
+        // Warmup the cache for the in-memory mode.
+        if self.mem_table {
+            // put the WarmingUpMarker in the context, otherwise, queries will fail as the
+            // InMemoryCacheExec node will think they should already be warmed up.
+            let ctx = ctx
+                .clone()
+                .with_distributed_option_extension(WarmingUpMarker::warming_up())?;
+            for query in sql.iter() {
+                self.execute_query(&ctx, query).await?;
```
NGA-TRAN, line 211: So the first query will wait for the cache to warm up before running? Will the query only warms up the tables it needs or all tables?  What will be cached? The full table or the part of the table the query needs? And the warm-up work will be the same every time we run it, right?

Caching and warming up are good. I ask questions to understand what exactly happen

benchmarks/src/util/memory.rs
gabotechs: More generically, it applies to any plan node without children (a leaf node). In practice, TPCH plans only has `DataSourceExec` leaf nodes, but theoretically this should work seamlessly for any other leaf node.

benchmarks/src/tpch/run.rs
gabotechs: The order of events looks like this:
1. A first query is run
2. Data is not cached, so it is gathered from the parquet files
3. The cache is populated with the recently gathered data
4. A second query is run
5. The cache is hit, and data is returned from there

The warmup covers step 1, 2 and 3. And what will get cached is just what the query would need to be fully executed. If the query is pruning some data and not returning everything, then that partial amount of data is what will get cached.

The cache key includes the serialized "explain" of the underlying node, so two nodes querying the same parquet files but with different filters will get cached independently.

gabotechs: I realized that I could have done a better job explaining all this, so just added some extended docs

Issue #123, opened 2025-09-04
Feature: distributed EXPLAIN ANALYZE


jayshrivastava:
# Problem

`EXPLAIN ANALYZE` in vanilla DF uses a visitor pattern on the physical plan. You can implement `metrics()` on `ExecutionPlan` nodes and the visitor collects them (see `DisplayableExecutionPlan::with_metrics`). The issue with distributed DF is that we copy the plan when sending it across the wire, so the coordinator sees no metrics once you cross network boundaries.

# Solution

## Execution Stage Metrics

A solution which integrates well with vanilla DF is to treat the physical plan as if it were one instance in memory even while crossing network boundaries. 

To do this, the `ArrowFlightEndpoint` can send a copy of the local plan's metrics (specifically, the inner plan of the `ExecutionStage`) when a request stream is completed (a request stream exists for each partition for each task). The `ArrowFlightReader` can "enrich" it's local `ExecutionStage` copy by merging these metrics in. We would likely save a plan per task and update the metrics for each plan copy as metrics arrive from the endpoint. Eventually, the coordinator's `ExecutionStage` copy gets the entire metrics set across all tasks and partitions.

This pattern should scale recursively since an `ExecutionStage` may contain more `ExecutionStage`s. We just need to treat `ExecutionStage` nodes in a special manner.

Caveats
- You cannot serialize metrics, so we would have to implement our own metrics proto
- When you serialize a plan, metrics do not get serialized. Instead of the `ArrowFlightEndpoint` sending a copy of its local plan with metrics, it would actually send metrics only
- The `ArrowFlightReader` can merge / store metrics for an `ExecutionStage` by task. Metrics in DF already store optional partition ids, so merging by task should preserve any per-partition information.
- You cannot retroactively add metrics to an `ExecutionPlan` node, so the `ArrowFlightReader` would have to transform the `ExecutionStage`'s inner plan to "wrap" nodes with metrics. 

## The Wire

On the endpoint side, we emit `FlightData`, which doesn’t have to be `RecordBatch`. When the batches are done, we can send a different message containing the mutated plan before closing it. There’s a few ways to do this such as implementing the `Stream` trait on a custom stream or I think there’s a `.chain()` method on streams to send some trailing data afterwards.

[...]

PR #141, opened 2025-09-17
metrics: add metrics module and protos

jayshrivastava:
This change adds a metrics module containing proto definitions to mirror datafusion metrics. These will be used for EXPLAIN ANALYZE. It also adds roundtrip serialization tests.

Informs #123.


src/lib.rs
```diff
@@ -8,6 +8,9 @@ mod distributed_physical_optimizer_rule;
 mod errors;
 mod execution_plans;
 mod flight_service;
+mod metrics;
+
+pub use metrics::proto::*;
```
gabotechs: For now, I think it's safest to just omit this line, as I don't think this is something we want to expose as part of the public API of the library.


```suggestion
```

You can remove this line for free right now, no need to have for internal usage

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
...
+pub struct ProtoLabel {
+    #[prost(string, tag = "1")]
+    pub name: String,
+    #[prost(string, tag = "2")]
+    pub value: String,
+}
+
+/// df_metrics_set_to_proto converts a [datafusion::physical_plan::metrics::MetricsSet] to a [ProtoMetricsSet].
+/// Custom metrics are filtered out, but any other errors are returned.
+/// TODO(#140): Support custom metrics.
```
gabotechs, line 139: Supporting custom metrics is probably going to be difficult, and will probably require people to hook in their own serde logic. +1 to just not do it right now.

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
+use chrono::DateTime;
+use datafusion::error::DataFusionError;
+use datafusion::physical_plan::metrics::{Metric, MetricValue, MetricsSet};
+use std::sync::Arc;
+
+/// A ProtoMetric is a protobuf mirror of [datafusion::physical_plan::metrics::Metric].
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ProtoMetric {
```
gabotechs: Nit: in other parts of the project we name the "proto" version of other structs by appending the `Proto` keyword as a suffix. How about also doing it here and naming things `MetricProto`, `MetricsSetProto`, etc...?

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
...
+            }
+            None => Ok(Arc::new(Metric::new_with_labels(
+                MetricValue::EndTimestamp(Timestamp::new()),
+                partition,
+                labels,
+            ))),
+        },
+        None => Err(DataFusionError::Internal(
+            "proto metric is missing the metric field".to_string(),
+        )),
```
gabotechs: Nit: A short version of this could be:

```suggestion
        None => internal_err!("proto metric is missing the metric field"),
```

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
...
+
+        struct TestCase {
+            name: &'static str,
+            metrics_set: Box<dyn Fn() -> MetricsSet>,
+        }
+
+        let test_cases = vec![
+            TestCase {
+                name: "empty",
+                metrics_set: Box::new(|| MetricsSet::new()),
```
gabotechs: 🤔 clippy should be failing here pointing you to do something like:

```suggestion
                metrics_set: Box::new(MetricsSet::new),
```

I wonder why pipelines are passing... maybe we are not correctly running clippy under tests?

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::physical_plan::metrics::{Metric, MetricValue};
+
+    #[test]
+    fn test_metric_roundtrip() {
```
gabotechs: Nice tests!

My only minor suggestion would be to create multiple actual tests (Rust style) rather than iterating over a list of test structs in the same test (Go style). Besides that really nice that we have this here!

src/metrics/proto.rs
```diff
@@ -0,0 +1,911 @@
...
+                    ),
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn test_custom_metrics_filtering() {
+        use datafusion::physical_plan::metrics::{Count, CustomMetricValue, MetricsSet};
+        use std::sync::Arc;
```
gabotechs, line 793: Nit: we can place the import statements along with the other import statements at the beginning of this `mod tests`

Cargo.toml
```diff
@@ -38,6 +38,7 @@ parquet = { version = "55.2.0", optional = true }
 arrow = { version = "55.2.0", optional = true }
 tokio-stream = { version = "0.1.17", optional = true }
 hyper-util = { version = "0.1.16", optional = true }
+chrono = "0.4.42"
```
gabotechs: You probably want to do the same thing we do for other test deps and add it as an optional dep here:

```toml
chrono = { version = "0.4.42", optional = true }
```

Add it as a dep-dependency like this:

```toml
chrono = "0.4.42"
```

And add it to the integration features:

```toml
[features]
integration = [
    "insta",
    "tpchgen",
    "tpchgen-arrow",
    "parquet",
    "arrow",
    "tokio-stream",
    "hyper-util",
    "chrono"
]

```

src/metrics/proto.rs
jayshrivastava: Oo nice catch

jayshrivastava: I updated the clippy command in CI to use `--all-targets` and `--features integration`

PR #142, opened 2025-09-17
fix bug in graphviz for determining output partitions

robtandy:
Prior to this PR, Hash joins with a Coalesce Partition node were rendering incorrectly
image

Now it renders correctly like:

image


gabotechs (approved):
🚀

PR #145, opened 2025-09-21
Distributed planning overhaul

gabotechs:
Still testing, don't look 🙈


src/execution_plans/arrow_flight_read.rs
gabotechs, line 1: Really annoying that GitHub doesn't realize that this is pretty much the same as the new `network_shuffle.rs` file... it's showing a huge diff instead

src/execution_plans/metrics.rs
```diff
@@ -390,6 +399,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore]
```
gabotechs, line 402: @jayshrivastava I might need some help here. As the planning nodes changed, all these tests started failing, as they no longer produce 5 nodes.

Ideally, we could come up with tests that do not need maintenance as planning changes, as I would expect our planning rules to greatly evolve with time.

src/execution_plans/partition_isolator.rs
```diff
@@ -1,45 +1,137 @@
...
     physical_plan::{
         DisplayAs, DisplayFormatType, EmptyRecordBatchStream, ExecutionPlan, Partitioning,
         PlanProperties,
     },
 };
+use std::{fmt::Formatter, sync::Arc};
 
-/// We will add this as an extension to the SessionConfig whenever we need
-/// to execute a plan that might include this node.
-pub struct PartitionGroup(pub Vec<usize>);
```
gabotechs, line 15: cc @robtandy 

As now all tasks in a stage handle the exact same number of partitions, there's a simpler way of identifying which partitions should be exposed by this `PartitionIsolatorExec`: just the task index within a stage is enough.

For example, If the input DataSource has 10 partitions, and we are on task index 0 out of a total amount of 5 tasks, we can deterministically say that this `PartitionIsolatorExec` is going to handle partitions [0,1]. If we are on task with index 3 of a total amount of 5 task, we know for sure that we are handling partitions [6,7].

This has two benefits:
- There are less things to thread, just propagating the task index is enough
- No assumptions are made regarding which partitions should be executed within a task. For example, for a parquet reader, it's fair to assume that each partition is always going to return the same set of data, but some other leaf nodes might not guarantee that (e.g. a leaf node that implements a work-stealing mpmc queue for distributing data gathering across partitions). Not forcing a specific partition indexes and letting each implementation decide accounts for this.

src/execution_plans/partition_isolator.rs
```diff
@@ -1,45 +1,137 @@
...
+/// ┌┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┐  ┌┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┐ ┌┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┴─┴┐     │
+/// │      DataSourceExec       │  │      DataSourceExec       │ │      DataSourceExec       │     │
+/// │         (task 1)          │  │         (task 2)          │ │         (task 3)          │     │
+/// └───────────────────────────┘  └───────────────────────────┘ └───────────────────────────┘     ■
+/// ```
+#[derive(Debug)]
+pub enum PartitionIsolatorExec {
+    Pending(PartitionIsolatorPendingExec),
+    Ready(PartitionIsolatorReadyExec),
+}
```
gabotechs, line 57: cc @robtandy 

By dividing the `PartitionIsolatorExec` into two states as the other nodes, we are capable of letting users just place a `PartitionIsolatorExec` wherever they want, and the distributed planner will then transition these to `Ready` when a stage is formed and the amount of tasks running in that stage is known.

This change was necessary in this PR for the integration tests: some integration tests are forming manually an execution plan rather than building it out of SQL, and in order for those tests to work, they need to be able to manually place `PartitionIsolatorExec` nodes in "pending" state in the same way they place the other distributed nodes also in "pending" state.

src/stage/display.rs
gabotechs, line 1: This file was unused, the graphviz visualization lives in `src/execution_plans/stage.rs`

src/distributed_physical_optimizer_rule.rs
```diff
@@ -1,48 +1,92 @@
...
-        self.partitions_per_task = Some(partitions_per_task);
+    /// Sets the amount of tasks employed in performing shuffles.
+    pub fn with_network_shuffle_tasks(mut self, tasks: usize) -> Self {
+        self.network_shuffle_tasks = Some(tasks);
+        self
+    }
+
+    /// Sets the amount of input tasks for every task coalescing operation.
+    pub fn with_network_coalesce_tasks(mut self, tasks: usize) -> Self {
+        self.network_coalesce_tasks = Some(tasks);
```
gabotechs, line 89: cc @robtandy 

Previously, we were splitting P partitions across N tasks. This means that each task handles an average of P/N partitions. The `partitions_per_task` parameter was used for telling the planner how many partitions should each task handle, and the number of tasks was inferred from it.

Now, it works pretty much the same as before, but with one subtle difference: before distributing, we multiply P by N so that when we fan out P * N partitions across N tasks, each task handles (P * N) / N = P partitions, making sure each task uses all the cores from the machine it runs on.

This means that the `partitions_per_task` is no longer relevant, as now `partitions_per_task` is always going to be P (~ the number of cores) of the machine running the task, and now the relevant parameters are these:

```rust
    /// Upon shuffling data, this defines how many tasks are employed into performing the shuffling.
    /// ```text
    ///  ( task 1 )  ( task 2 ) ( task 3 )
    ///      ▲           ▲          ▲
    ///      └────┬──────┴─────┬────┘
    ///       ( task 1 )  ( task 2 )       N tasks
    /// ```
    /// This parameter defines N
    network_shuffle_tasks: Option<usize>,
    /// Upon merging multiple tasks into one, this defines how many tasks are merged.
    /// ```text
    ///              ( task 1 )
    ///                  ▲
    ///      ┌───────────┴──────────┐
    ///  ( task 1 )  ( task 2 ) ( task 3 )  N tasks
    /// ```
    /// This parameter defines N
    network_coalesce_tasks: Option<usize>,
```

src/execution_plans/metrics.rs
jayshrivastava: No problem. I think we can dynamically check the size of the plan by doing a traversal. I can push some changes to the test in a bit

gabotechs: Cool! my plan was to tackle this in a follow up to not bloat this one more than what it already is, so anything you can provide is more than welcome

jayshrivastava: Oh feel free to `#[skip]` the test for now if that unblocks you faster

src/execution_plans/partition_isolator.rs
jayshrivastava: Very nice.

src/flight_service/do_get.rs
```diff
@@ -126,33 +87,52 @@ impl ArrowFlightEndpoint {
                 // Initialize partition count to the number of partitions in the stage
                 let total_partitions = stage.plan.properties().partitioning.partition_count();
                 Ok::<_, Status>(TaskData {
-                    session_state,
                     stage: Arc::new(stage),
                     num_partitions_remaining: Arc::new(AtomicUsize::new(total_partitions)),
```
jayshrivastava, line 91: I don't think this is correct anymore. We may need better test coverage here. We evict the stage when all partitions in the stage are read from. However, with `NetworkCoalesceExec`, not all partitions are read. We need to calculate how many partitions this task will execute here.
```
let stage_proto = doget.stage_proto.ok_or_else(missing("stage_proto"))?;
let stage = stage_from_proto(stage_proto, &session_state, &self.runtime, &codec)
    .map_err(|err| {
        Status::invalid_argument(format!("Cannot decode stage proto: {err}"))
    })?;

// Initialize partition count to the number of partitions in the stage
let total_partitions = stage.plan.properties().partitioning.partition_count();
```

jayshrivastava: `test_task_data_partition_counting` tests that we evict when all partitions are executed. We should add a case where we have a Network Coalesce scenario

gabotechs: > However, with NetworkCoalesceExec, not all partitions are read.

Unless there's a bug in the implementation, all partitions should be read. Take the first TPCH query's head as an example:

Screenshot 2025-09-22 at 8 25 21 PM

All partitions from all tasks are read exactly once, and this condition remains true independently of wether the reader node is a `NetworkShuffleExec` or a `NetworkCoalesceExec`

gabotechs: In fact `test_task_data_partition_counting` is independent from the reader node. If you look at the current test, no references to the `NetworkShuffleExec` or the previous `ArrowFlightReadExec` are made. I would say it should still be valid regardless of wether the reader is a `NetworkCoalesceExec` or `NetworkShuffleExec`.

src/execution_plans/network_coalesce.rs
```diff
@@ -0,0 +1,319 @@
...
+/// │  Arc<dyn ExecutionPlan>   │  │  Arc<dyn ExecutionPlan>   │ │  Arc<dyn ExecutionPlan>   │     │
+/// │         (task 1)          │  │         (task 2)          │ │         (task 3)          │     │
+/// └───────────────────────────┘  └───────────────────────────┘ └───────────────────────────┘     ■
+/// ```
+///
+/// The communication between two stages across a [NetworkCoalesceExec] has two implications:
+///
+/// - Stage N+1 must have exactly 1 task. The distributed planner ensures this is true.
+/// - The amount of partitions in the single task of Stage N+1 is equal to the sum of all
+///   partitions in all tasks in Stage N+1 (e.g. (1,2,3,4,5,6,7,8,9) = (1,2,3)+(4,5,6)+(7,8,9) )
```
NGA-TRAN, line 60: Very nice diagram and clear invariant description

tests/distributed_aggregation.rs
```diff
@@ -40,34 +40,28 @@ mod tests {
...
         └──────────────────────────────────────────────────
-          ┌───── Stage 2   Tasks: t0:[p0] t1:[p1] t2:[p2] 
-          │ RepartitionExec: partitioning=Hash([RainToday@0], 3), input_partitions=1
-          │   PartitionIsolatorExec Tasks: t0:[p0,__,__] t1:[__,p0,__] t2:[__,__,p0] 
-          │     ArrowFlightReadExec input_stage=1, input_partitions=3, input_tasks=3
+          ┌───── Stage 1   Tasks: t0:[p0,p1,p2] t1:[p3,p4,p5] 
+          │ RepartitionExec: partitioning=Hash([RainToday@0], 3), input_partitions=2
+          │   AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))]
+          │     PartitionIsolatorExec Tasks: t0:[p0,p1,__] t1:[__,__,p0] 
+          │       DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[RainToday], file_type=parquet
```
NGA-TRAN, line 63: Yay! No more unecessary roundrobin repartition

tests/tpch_validation_test.rs
```diff
@@ -11,30 +11,41 @@ mod tests {
...
+          │     AggregateExec: mode=FinalPartitioned, gby=[l_returnflag@0 as l_returnflag, l_linestatus@1 as l_linestatus], aggr=[sum(lineitem.l_quantity), sum(lineitem.l_extendedprice), sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount), sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount * Int64(1) + lineitem.l_tax), avg(lineitem.l_quantity), avg(lineitem.l_extendedprice), avg(lineitem.l_discount), count(Int64(1))]
+          │       CoalesceBatchesExec: target_batch_size=8192
+          │         NetworkShuffleExec read_from=Stage 1, output_partitions=6, n_tasks=4, input_tasks=3
           └──────────────────────────────────────────────────
+            ┌───── Stage 1   Tasks: t0:[p0,p1,p2,p3,p4,p5,p6,p7,p8,p9,p10,p11,p12,p13,p14,p15,p16,p17,p18,p19,p20,p21,p22,p23] t1:[p24,p25,p26,p27,p28,p29,p30,p31,p32,p33,p34,p35,p36,p37,p38,p39,p40,p41,p42,p43,p44,p45,p46,p47] t2:[p48,p49,p50,p51,p52,p53,p54,p55,p56,p57,p58,p59,p60,p61,p62,p63,p64,p65,p66,p67,p68,p69,p70,p71]
```
NGA-TRAN: I wonder if we need to name the partitions for each task different here? Can every task use the same p0 to p23 instead? They use the same algorithm. Their data will be different but must belong to the same hash/range

tests/tpch_validation_test.rs
```diff
@@ -11,30 +11,41 @@ mod tests {
...
-        │       AggregateExec: mode=FinalPartitioned, gby=[l_returnflag@0 as l_returnflag, l_linestatus@1 as l_linestatus], aggr=[sum(lineitem.l_quantity), sum(lineitem.l_extendedprice), sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount), sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount * Int64(1) + lineitem.l_tax), avg(lineitem.l_quantity), avg(lineitem.l_extendedprice), avg(lineitem.l_discount), count(Int64(1))]
-        │         CoalesceBatchesExec: target_batch_size=8192
-        │           ArrowFlightReadExec input_stage=1, input_partitions=15, input_tasks=8
+        │   NetworkCoalesceExec read_from=Stage 2, output_partitions=24, input_tasks=4
```
NGA-TRAN, line 31: Yay, we have a way to put partitions together without repartitioning

src/flight_service/do_get.rs
jayshrivastava: > In fact test_task_data_partition_counting is independent from the reader node. If you look at the current test, no references to the NetworkShuffleExec or the previous ArrowFlightReadExec are made. I would say it should still be valid regardless of wether the reader is a NetworkCoalesceExec or NetworkShuffleExec.

I agree the test should be independent. I was saying that the test asserts that the endpoint evicts a stage when all partitions in the stage are done. I suspect this needs to change slightly.

Please correct me if I'm wrong for this part: It looks like the `NetworkCoalesceExec::execute()` method would execute partitions 0-5 on stage 2 task 0 only. For a given partition, we look up the URL to send the request to (This is different than `NetworkShuffleExec` where we execute the partition on each worker). Stage 2 has 24 partitions. I think stage 2 task 0 needs to evict the stage after 6 partitions are run.

NGA-TRAN (approved):
I really like the way you define `NetworkCoalesceExec` and `NetworkShuffleExec`. The exact thing we need to shuffle and collect data efficiently 🎉 

Since your description makes sense and clearly solve the problems we found earlier,   I did not look into the details  but rely on @robtandy and @jayshrivastava to review them. I did look at the tests and had one suggestion to rename partitions of different tasks

src/distributed_physical_optimizer_rule.rs
```diff
@@ -73,115 +117,242 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+            // If this node is a DataSourceExec, we need to wrap it with PartitionIsolatorExec so
+            // that not all tasks have access to all partitions of the underlying DataSource.
+            if plan.as_any().is::<DataSourceExec>() {
+                let node = PartitionIsolatorExec::new_pending(plan);
+
+                return Ok(Transformed::yes(Arc::new(node)));
+            }
+
+            // If this is a hash RepartitionExec, introduce a shuffle.
+            if let Some(node) = plan.as_any().downcast_ref::<RepartitionExec>() {
```
jonathanc-n: Just a nit but I think its worth mentioning since we are currently using `Option`s to denote whether a variable exists. We can do:

```rust
if let (Some(repart), Some(tasks)) = (
    plan.as_any().downcast_ref::<RepartitionExec>(),
    self.network_shuffle_tasks.as_ref(),
) {
```

Keeps it more concise and easier to read. In The future we can possibly change options like these to use enums in the future; ex. for this case, `ShuffleDisabled` + `ShuffleEnabled(tasks)`

src/distributed_physical_optimizer_rule.rs
```diff
@@ -1,48 +1,92 @@
...
     error::Result,
     physical_optimizer::PhysicalOptimizerRule,
     physical_plan::{repartition::RepartitionExec, ExecutionPlan},
 };
+use std::error::Error;
+use std::fmt::{Display, Formatter};
+use std::sync::Arc;
 use uuid::Uuid;
 
+/// Physical optimizer rule that inspects the plan, places the appropriate network
```
jonathanc-n, line 23: Really like the new docs + diagrams, made it much easier to understand everything

src/flight_service/do_get.rs
gabotechs: > It looks like the NetworkCoalesceExec::execute() method would execute partitions 0-5 on stage 2 task 0 only

Not really, it will look for:
- partition 0-5 on stage 2 task 0
- partition 6-11 on stage 2 task 1
- partition 12-17 on stage 2 task 2
- partition 18-24 on stage 2 task 3

A very key piece of code for understanding how this happens is:

```rust
        let target_task = partition / partitions_per_task;
        let target_partition = partition % partitions_per_task;
```

That's what allows a single NetworkCoalesceExec to go to different tasks gathering all partitions from them

tests/tpch_validation_test.rs
gabotechs: Yeah, you are completely right there, I think we should change that

src/distributed_physical_optimizer_rule.rs
gabotechs: nice suggestion! I'll take it

src/distributed_physical_optimizer_rule.rs
gabotechs: I'm using https://monodraw.helftone.com/ for those. I'm finding it very useful.

tests/tpch_validation_test.rs
gabotechs: Done!

benchmarks/src/tpch/run.rs
```diff
@@ -102,8 +102,12 @@ pub struct RunOpt {
     sorted: bool,
 
     /// Number of partitions per task.
-    #[structopt(long = "ppt")]
-    partitions_per_task: Option<usize>,
+    #[structopt(long)]
+    shuffle_tasks: Option<usize>,
+
+    /// Number of partitions per task.
```
robtandy: we probably want to update the comments here.    

For the user, I think this is an improvement as `shuffle_tasks` and `coalesce_tasks` are easier to conceptualize.

src/distributed_physical_optimizer_rule.rs
```diff
@@ -73,115 +117,241 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+                let node = NetworkCoalesceExec::from_coalesce_partitions_exec(node, tasks)?;
+
+                let plan = plan.with_new_children(vec![Arc::new(node)])?;
+
+                return Ok(Transformed::yes(plan));
+            }
+
+            // The SortPreservingMergeExec node will try to coalesce all partitions into just 1.
+            // We need to account for it and help it by also coalescing all tasks into one, therefore
+            // a NetworkCoalesceExec is introduced.
```
robtandy, line 172: I like this because previously we were not able to divide the work for sorting among Tasks.   `datafusion-ray` had the functionality for this but it was buried in a conditional branch in planning, now it is easier to understand and reason about.

gabotechs:
> I'm not sure about each task having the same number of partitions
I think specifying the desired resources per task will aid in future scheduling of tasks which can consider resources for placement. This may be important in a heterogeneous environment which relies on resource aware task placement.

👍 this makes sense. With the current tools, I don't think there's something much better to do rather than assuming an homogeneous environment. This probably should be exposed to the user somehow, maybe in the `ChannerResolver` trait, so that the user can return not only the URL to the worker but further stats like CPU cores on that worker, memory, etc...

PR #146, opened 2025-09-22
update to DataFusion 50

adriangb:
DataFusion 50 has been released. I had to bump the MSRV and took the opportunity to upgrade the edition (I see no reason not to, unless I'm missing some internal reasoning).


src/protobuf/distributed_codec.rs
```diff
@@ -56,9 +56,12 @@ impl PhysicalExtensionCodec for DistributedCodec {
                     .map(|s| s.try_into())
                     .ok_or(proto_error("ArrowFlightReadExec is missing schema"))??;
 
+                // Create a default SessionContext for the protobuf parsing
+                // TODO: This loses the original function registry, but DataFusion 50.0.0 requires SessionContext
+                let session_ctx = SessionContext::new();
                 let partioning = parse_protobuf_partitioning(
                     partitioning.as_ref(),
-                    registry,
+                    &session_ctx,
```
adriangb: I'm not sure what DataFusion intends to do here, or what the consequences of this are. It needs some investigation.

src/test_utils/insta.rs
```diff
@@ -12,7 +12,9 @@ macro_rules! assert_snapshot {
 }
 
 pub fn settings() -> insta::Settings {
-    env::set_var("INSTA_WORKSPACE_ROOT", env!("CARGO_MANIFEST_DIR"));
+    unsafe {
```
adriangb: From 2024 edition

rust-toolchain.toml
```diff
@@ -1,3 +0,0 @@
-[toolchain]
-channel = "1.85.1"
-profile = "default"
```
adriangb, line 3: This locks in a specific version. The MSRV in `Cargo.toml` is a minimum version, much more flexible

Cargo.toml
```diff
@@ -29,16 +33,16 @@ dashmap = "6.1.0"
...
-tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "c8d823432528eed4f70fca5a1296a66c68a389a8", optional = true }
-parquet = { version = "55.2.0", optional = true }
-arrow = { version = "55.2.0", optional = true }
+tpchgen = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "482ee68", optional = true }
+tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "482ee68", optional = true }
+parquet = { version = "56.1.0", optional = true }
+arrow = { version = "56.1.0", optional = true }
 tokio-stream = { version = "0.1.17", optional = true }
 hyper-util = { version = "0.1.16", optional = true }
-chrono = { version = "0.4.42", optional = true }
```
adriangb: Same as https://github.com/datafusion-contrib/datafusion-distributed/pull/143, need to rebase

tests/tpch_validation_test.rs
```diff
@@ -88,7 +88,7 @@ mod tests {
...
+            │                       DataSourceExec: file_groups={6 groups: [[/testdata/tpch/data/partsupp/1.parquet:0..753508, /testdata/tpch/data/partsupp/10.parquet:0..748125, /testdata/tpch/data/partsupp/11.parquet:0..496934], [/testdata/tpch/data/partsupp/11.parquet:496934..744689, /testdata/tpch/data/partsupp/12.parquet:0..749068, /testdata/tpch/data/partsupp/13.parquet:0..750296, /testdata/tpch/data/partsupp/14.parquet:0..251448], [/testdata/tpch/data/partsupp/14.parquet:251448..748674, /testdata/tpch/data/partsupp/15.parquet:0..744298, /testdata/tpch/data/partsupp/16.parquet:0..748035, /testdata/tpch/data/partsupp/2.parquet:0..9008], [/testdata/tpch/data/partsupp/2.parquet:9008..753010, /testdata/tpch/data/partsupp/3.parquet:0..756405, /testdata/tpch/data/partsupp/4.parquet:0..498160], [/testdata/tpch/data/partsupp/4.parquet:498160..746724, /testdata/tpch/data/partsupp/5.parquet:0..746720, /testdata/tpch/data/partsupp/6.parquet:0..750184, /testdata/tpch/data/partsupp/7.parquet:0..253099], ...]}, projection=[ps_partkey, ps_suppkey, ps_supplycost], file_type=parquet, predicate=DynamicFilterPhysicalExpr [ true ] AND DynamicFilterPhysicalExpr [ true ]
```
adriangb: The diff in all of these is the extra `DynamicFilterPhysicalExpr [ true ]`

I'm guessing we're ending up with multiple because there are multiple optimizer passes being run? I'm not sure tbh.

They also obviously aren't working (will never be updated from `true`) but also shouldn't be much overhead (it's an extra literal `true` to evaluate in expression evaluation).

Changing this would require a lot of upstream work in DataFusion which we want to do anyway but I don't think it would be a good idea to block this PR for that.

src/errors/arrow_error.rs
```diff
@@ -129,6 +129,12 @@ impl ArrowErrorProto {
                 inner: Some(ArrowErrorInnerProto::RunEndIndexOverflowError(true)),
                 ctx: ctx.cloned(),
             },
+            ArrowError::OffsetOverflowError(offset) => ArrowErrorProto {
+                inner: Some(ArrowErrorInnerProto::ParseError(format!(
+                    "Offset overflow error: {offset}"
+                ))),
```
gabotechs: It looks like we are mapping here `ArrowError::OffsetOverflowError` to `ArrowErrorInnerProto::ParseError`, but that means we are losing the necessary information to convert back to `ArrowError::OffsetOverflowError` while deserializing.

As this is a new error variant in Arrow, we should add it to the proto version also, something like:
```rust
    #[prost(
        oneof = "ArrowErrorInnerProto",
    -   tags = "2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19"
    +   tags = "2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20"
    )]
...
pub enum ArrowErrorInnerProto {
    ...
    #[prost(uint64, tag = "20")]
    OffsetOverflowError(u64),
}
```

And map it here.

src/flight_service/do_get.rs
```diff
@@ -79,8 +82,8 @@ impl ArrowFlightEndpoint {
         let stage_data = once
             .get_or_try_init(|| async {
                 let stage_proto = doget.stage_proto.ok_or_else(missing("stage_proto"))?;
-                let stage = stage_from_proto(stage_proto, &session_state, &self.runtime, &codec)
-                    .map_err(|err| {
+                let stage =
+                    stage_from_proto(stage_proto, &ctx, &self.runtime, &codec).map_err(|err| {
```
gabotechs, line 98: We need the context here to contain all the UDFs, UDAFs and stuff necessary for properly decoding the stage node. Before, we were passing the `&session_state`, which is the one built by the user containing all their custom stuff, but now we are passing a recently built `SessionContext::new()` that knows nothing about the custom things provided by the user in `&session_state`.

This means that we should probably be passing:
```rust
let ctx = SessionContext::from(session_state);
```
instead

src/protobuf/distributed_codec.rs
```diff
@@ -62,7 +64,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
 
                 let partitioning = parse_protobuf_partitioning(
                     partitioning.as_ref(),
-                    registry,
+                    &ctx,
```
gabotechs, line 67: I think this is the problem https://github.com/apache/datafusion/pull/17601 is trying to solve upstream.

We cannot have access to a properly formed `SessionContext` here containing all the user-provided custom stuff because the `PhysicalExtensionCodec` trait signatures do not allow it.

For us here specifically, is not a big deal, as we probably don't need any user-provided custom stuff for just parsing partitioning, but probably worth a comment here somewhere stating the problem and that we are fine with just building a new default `SessionContext`.

tests/tpch_validation_test.rs
```diff
@@ -300,18 +300,18 @@ mod tests {
     async fn test_tpch_7() -> Result<(), Box<dyn Error>> {
         let plan = test_tpch_query(7).await?;
         assert_snapshot!(plan, @r"
-        ┌───── Stage 5   Tasks: t0:[p0] 
+        ┌───── Stage 9   Tasks: t0:[p0]
```
gabotechs, line 303: 🤔 interesting... this distributed plan seems to be changing significantly. 

It does not look wrong though, it seems like a HashJoinExec has now `Partitioned` mode instead of `CollectLeft`, and therefore we are capable of introducing a new shuffle step in the middle of the plan that we couldn't before.

jonathanc-n: Yes this is very nice, @adriangb any chance you know which pr brought this change?

adriangb: I'm not sure haha!

gabotechs:
🟢 green light to ship this upgrade! 

I think the only missing thing would be [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/146#discussion_r2378041186) and making the pipelines pass.

src/protobuf/distributed_codec.rs
adriangb: https://github.com/apache/datafusion/pull/17601 is merged now! I guess it will be included in DataFusion 50.1.0? We can update here then.

gabotechs:
> Great news! I did a rebase on this repo's main and addressed https://github.com/datafusion-contrib/datafusion-distributed/pull/146#discussion_r2378041186. Unfortunately there is now a failing test 😢

Yeap, I also so it failing when trying to point this repo to our https://github.com/DataDog/datafusion fork. cc @jayshrivastava it's one of the new tests for https://github.com/datafusion-contrib/datafusion-distributed/pull/160, it seems to not play well with DataFusion 50. 

It should be fine to leave it ignored in this PR I think.

jayshrivastava:
+1 please feel free to skip/ignore the failing unit test and file a corresponding ticket :) 

The issue is that this query does not get distributed anymore under this config
```
SELECT sum(balance) / 7.0 as avg_yearly
            FROM table2
            WHERE name LIKE 'customer%'
              AND balance < (
                SELECT 0.2 * avg(balance)
                FROM table2 t2_inner
                WHERE t2_inner.id = table2.id
              )
```
```
let config = SessionConfig::new().with_target_partitions(2);

    .with_default_features()
    .with_config(config)
    .with_distributed_channel_resolver(InMemoryChannelResolver::new())
    .with_physical_optimizer_rule(Arc::new(
        DistributedPhysicalOptimizerRule::default()
            .with_network_coalesce_tasks(2)
            .with_network_shuffle_tasks(2),
    ))
    .build();
```

adriangb:
skipped the test

adriangb:
I'm not sure what to do about the tpch tests. They pass locally, I tried regenerating the data

adriangb:
Thanks @gabotechs , we're 🟢 now!

gabotechs (approved):
💯 let's go

PR #149, opened 2025-09-22
refactor: Move `DistributedPlanError` to error folder

jonathanc-n:
For @gabotechs 's #145 branch

### DIFF
--- a/src/distributed_physical_optimizer_rule.rs
+++ b/src/distributed_physical_optimizer_rule.rs
@@ -1,5 +1,7 @@
 use super::{NetworkShuffleExec, PartitionIsolatorExec, StageExec};
-use crate::errors::distributed_plan_error::{get_distribute_plan_err, limit_tasks_err, DistributedPlanError};
+use crate::errors::distributed_plan_error::{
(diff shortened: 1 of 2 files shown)


jonathanc-n:
Does it seems like a valid change? Should I make another pr?

gabotechs:
Let me take a look at this one now

src/errors/mod.rs
```diff
@@ -8,6 +8,7 @@ use prost::Message;
 
 mod arrow_error;
 mod datafusion_error;
+pub mod distributed_plan_error;
```
gabotechs, line 11: Sorry if this looked confusing, the `errors` module here is not really a collection of errors used along the project (as it usually happens in Rust projects). These are meant to be just Protobuf variants for upstream's `DataFusionError` struct rather than actual errors that can be thrown, so that means probably `DistributedPlanError` should not belong here. 

That being said, we probably should move this `errors` folder inside of the `protobuf` now that we have one, that would scope things better and avoid confusion.

About whether having an `errors` module with an actual collection of errors (with just the `DistributedPlanError` for now), or just leaving the `DistributedPlanError` where it is, no strong opinion.

PR #150, opened 2025-09-22
execution_plans: add MetricsCollectingStream

jayshrivastava:
This change adds a new type called `MetricsCollectingStream`. It wraps a stream of `FlightData`
and collects any metrics that are passed in the `app_metadata`. This change also introduces
an `FlightAppMetadata` enum proto which can be used to define our app metadata protocol.

Informs #123

### DIFF
--- a/src/execution_plans/metrics_collecting_stream.rs
+++ b/src/execution_plans/metrics_collecting_stream.rs
@@ -1,0 +1,264 @@
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use crate::metrics::proto::MetricsSetProto;
+use crate::protobuf::StageKey;
+use crate::protobuf::{AppMetadata, FlightAppMetadata};
+use arrow_flight::{FlightData, error::FlightError};
+use dashmap::DashMap;
+use futures::stream::Stream;
+use pin_project::pin_project;
+use prost::Message;
+use std::sync::Arc;
+
+/// MetricsCollectingStream wraps a FlightData stream and extracts metrics from app_metadata
+/// while passing through all the other FlightData unchanged.
+#[pin_project]
+pub struct MetricsCollectingStream<S>
+where
+    S: Stream<Item = Result<FlightData, FlightError>> + Send,
+{
+    #[pin]
+    inner: S,
+    metrics_collection: Arc<DashMap<StageKey, Vec<MetricsSetProto>>>,
+}
+
+impl<S> MetricsCollectingStream<S>
+where
+    S: Stream<Item = Result<FlightData, FlightError>> + Send,
+{
+    #[allow(dead_code)]
+    pub fn new(
+        stream: S,
+        metrics_collection: Arc<DashMap<StageKey, Vec<MetricsSetProto>>>,
+    ) -> Self {
+        Self {
+            inner: stream,
+            metrics_collection,
+        }
+    }
+
+    fn extract_metrics_from_flight_data(
+        metrics_collection: Arc<DashMap<StageKey, Vec<MetricsSetProto>>>,
+        flight_data: &mut FlightData,
+    ) -> Result<(), FlightError> {
+        if flight_data.app_metadata.is_empty() {
+            return Ok(());
+        }
+
+        let metadata =
+            FlightAppMetadata::decode(flight_data.app_metadata.as_ref()).map_err(|e| {
+                FlightError::ProtocolError(format!("failed to decode app_metadata: {}", e))
+            })?;
+
+        let Some(content) = metadata.content else {
+            return Err(FlightError::ProtocolError(
+                "expected Some content in app_metadata".to_string(),
+            ));
+        };
+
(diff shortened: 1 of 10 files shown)


src/execution_plans/metrics_collecting_stream.rs
```diff
@@ -0,0 +1,311 @@
...
+                                    } else {
+                                        return Err(FlightError::ProtocolError("expected Some StageKey in MetricsCollectingStream, got None".to_string()));
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    flight_data.app_metadata.clear();
+                    Ok(())
+                }
```
NGA-TRAN: Nit: I wonder if you can make this fewer layers?

src/execution_plans/metrics_collecting_stream.rs
```diff
@@ -0,0 +1,311 @@
...
+                query_id: "test_query".to_string(),
+                stage_id: 1,
+                task_number: 1,
+            },
+            StageKey {
+                query_id: "test_query_2".to_string(),
+                stage_id: 2,
+                task_number: 2,
+            },
+        ];
```
NGA-TRAN, line 150: Can you describe the purpose of this test? Do you want to test collecting metrics for one query with 2 stages or you want to test collecting metrics for 2 queries each has one stage?

If you want to test both, you may want to split them in 2 different tests  and make it clear in the names of the tests or add comments to  explain so.  Also, do we collect metrics of different queries in one pass and need such tests?

Either way, you want to make the test valid by:
- Have the same `query_id` for the test on the same query. And have different `query_id` if they are for 2 different queries
- Use smaller number for child `stage_id`

src/execution_plans/metrics_collecting_stream.rs
```diff
@@ -0,0 +1,311 @@
...
+
+            // Verify the second metric value (Gauge)
+            if let Some(MetricValueProto::Gauge(gauge)) = &collected_metrics[0].metrics[1].metric {
+                assert_eq!(gauge.name, "test_gauge");
+                assert_eq!(gauge.value, 99);
+            } else {
+                panic!("expected Gauge metric");
+            }
+        }
+    }
```
NGA-TRAN, line 207: This test is a bit long. What do you think to make different functions for creating metrics, stages and result verification, then create a macro to use them. There are many examples for that in Vanilla DF tests and I think Gabriel has that in this repro, too. That way, you can reuse them and it is easier for us to review

NGA-TRAN: @gabotechs : I am a bit ambiguous here. It would be great, if you can hep point Jayant to some good examples how to do so

src/execution_plans/metrics_collecting_stream.rs
```diff
@@ -0,0 +1,311 @@
...
+use prost::Message;
+use std::sync::Arc;
+
+/// MetricsCollectingStream wraps a FlightData stream and extracts metrics from app_metadata
+/// while passing through all the other FlightData unchanged.
+pub struct MetricsCollectingStream<S>
+where
+    S: Stream<Item = Result<FlightData, FlightError>> + Send + Unpin,
+{
+    inner: S,
```
gabotechs, line 22: I wonder if this should instead be:

```rust
#[pin_project]
pub struct MetricsCollectingStream<S>
where
    S: Stream<Item = Result<FlightData, FlightError>> + Send + Unpin,
{
    #[pin]
    inner: S,
```

I don't know the details about when this is needed, but I usually see wrapping streams be implemented this way in Rust.

src/execution_plans/metrics_collecting_stream.rs
jayshrivastava: Fixed. Plus I made it much more concise.

jayshrivastava: Added a test utils for metrics creation as well :) Now, you can call `make_test_metrics_set_proto_from_seed` to make test data. Also all metrics protos implement `Eq` now so it's easy to assert metrics equality in tests

gabotechs (approved):
Awesome! not much to say, this looks just right

src/execution_plans/metrics_collecting_stream.rs
jayshrivastava: > Can you describe the purpose of this test?  

I added a comment. 

> Do you want to test collecting metrics for one query with 2 stages or you want to test collecting metrics for 2 queries each has one stage?

This metrics stream doesn't really know the semantics of how queries are executed.  The only "guarantee" it offers is that it collects metrics by StageKey but it doesn't know what those are. That's why I just used random StageKeys. However, I do agree that it's confusing, so I updated the test to use same `query_id` and `stage_id` with two different task ids. This should make more sense.

src/execution_plans/metrics_collecting_stream.rs
jayshrivastava: 👀 Will look into it.

PR #152, opened 2025-09-23
set msrv via Cargo.toml, use 2024 edition

adriangb:
The diff is mostly running `cargo fmt` on 2024 edition.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.86.0"
+channel = "1.85.1"
 profile = "default"


Cargo.toml
```diff
@@ -8,7 +8,8 @@ datafusion-proto = { version = "49.0.0" }
 [package]
 name = "datafusion-distributed"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
+rust-version = "1.86.0"
```
gabotechs: Our intention here is rather than enforcing ourselves a `rust-version` in Cargo.toml, let the `datafusion` crate enforce it instead. Automatically, because we are depending on `datafusion`, it's not necessary for us to apply further restrictions on the MSRV.

What we need to do though is to make sure that developers contributing to this project use exactly DataFusion's MSRV, therefore it's important to keep the `rust-toolchain.toml` file that was introduced in https://github.com/datafusion-contrib/datafusion-distributed/pull/72.

Any chance to keep the `rust-toolchain.toml` while changing it to the new `1.86.0` channel?

adriangb: Done!

src/protobuf/distributed_codec.rs
```diff
@@ -23,7 +23,7 @@ use std::sync::Arc;
 pub struct DistributedCodec;
 
 impl DistributedCodec {
-    pub fn new_combined_with_user(cfg: &SessionConfig) -> impl PhysicalExtensionCodec {
+    pub fn new_combined_with_user(cfg: &SessionConfig) -> impl PhysicalExtensionCodec + use<> {
```
gabotechs, line 26: 👀 `use<>`. Is this a thing in Rust? no idea what this means

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.85.1"
+channel = "1.86.0"
```
gabotechs: 🤔 Now that I think about it, probably it makes sense to leave this as 1.85.1 in this PR, as this is still running on DataFusion 49, and we need to make sure that both pipelines and developers adhere to support at least 1.85.1.

gabotechs (approved):
Nice! just some minor comments, but otherwise very grateful for this 😄

rust-toolchain.toml
adriangb: okay done

src/protobuf/distributed_codec.rs
adriangb: yeah, it means that the lifetime of the returned impl doesn't depend on the `&SessionConfig`

PR #156, opened 2025-09-24
fix: Move `error.rs` to protobuf

jonathanc-n:
cc @gabotechs


src/protobuf/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod distributed_codec;
+pub mod errors;
```
gabotechs: Note how in all the modules of the project, we try to be strict about what's visible to the outside by not making the inner modules public. It would be fine to be consistent here and:
- instead of blindly exposing the inner `errors` module:
```suggestion
mod errors;
```

- Just expose the few things that are strictly needed:
```rust
pub(crate) use errors::{
    datafusion_error_to_tonic_status, map_flight_to_datafusion_error,
    map_status_to_datafusion_error,
};
```

jonathanc-n: Yes this makes sense. I like the way this is done compared to Datafusion, lets you see everything that is public at a glance 😄

gabotechs (approved):
Nice! as soon as pipelines are green I'll merge this

PR #158, opened 2025-09-26
Address public api weak points

gabotechs:
Here are some improvement points I've discovered while playing with the public API:
- Users might want to provide more than 1 custom codec. Instead of forcing them to wrap their code with a combined codec, we can offer them the possibility to provide more than 1 natively, greatly improving UX.
- While providing user codecs, the current API does not allow to provide an `Arc<dyn PhysicalExtensionCodec>`, it only allows to provide non-dynamic codecs. This adds a new `with_distributed_user_codec_arc` and `set_distributed_user_codec_arc` variants for providing Arc-wrapped dynamic user codecs
- Expose some structs that were not public before: `DistributedTaskContext` and `NetworkCoalesceExec`. These structures where meant to be part of the public API, but they were not exposed.

### DIFF
--- a/src/common/composed_extension_codec.rs
+++ b/src/common/composed_extension_codec.rs
@@ -1,56 +1,96 @@
-use datafusion::common::not_impl_err;
+use datafusion::common::internal_datafusion_err;
 use datafusion::error::DataFusionError;
+use datafusion::error::Result;
 use datafusion::execution::FunctionRegistry;
 use datafusion::logical_expr::{AggregateUDF, ScalarUDF};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
+use prost::Message;
 use std::fmt::Debug;
 use std::sync::Arc;
+// Code taken from https://github.com/apache/datafusion/blob/10f41887fa40d7d425c19b07857f80115460a98e/datafusion/proto/src/physical_plan/mod.rs
+// TODO: It's not yet on DF 49, once upgrading to DF 50 we can remove this
 
-// Idea taken from
-// https://github.com/apache/datafusion/blob/0eebc0c7c0ffcd1514f5c6d0f8e2b6d0c69a07f5/datafusion-examples/examples/composed_extension_codec.rs#L236-L291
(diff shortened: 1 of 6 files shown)


src/common/composed_extension_codec.rs
```diff
@@ -1,56 +1,96 @@
...
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
+use prost::Message;
 use std::fmt::Debug;
 use std::sync::Arc;
+// Code taken from https://github.com/apache/datafusion/blob/10f41887fa40d7d425c19b07857f80115460a98e/datafusion/proto/src/physical_plan/mod.rs
+// TODO: It's not yet on DF 49, once upgrading to DF 50 we can remove this
 
-// Idea taken from
-// https://github.com/apache/datafusion/blob/0eebc0c7c0ffcd1514f5c6d0f8e2b6d0c69a07f5/datafusion-examples/examples/composed_extension_codec.rs#L236-L291
```
gabotechs, line 11: This code was previously taken from an example upstream. However, in version 50, there's a better version of this same code living in the datafusion-proto package, so I took it.

src/protobuf/distributed_codec.rs
```diff
@@ -24,12 +24,9 @@ pub struct DistributedCodec;
...
     pub fn new_combined_with_user(cfg: &SessionConfig) -> impl PhysicalExtensionCodec + use<> {
-        let mut combined_codec = ComposedPhysicalExtensionCodec::default();
-        combined_codec.push(DistributedCodec {});
-        if let Some(ref user_codec) = get_distributed_user_codec(cfg) {
-            combined_codec.push_arc(Arc::clone(user_codec));
-        }
-        combined_codec
+        let mut codecs: Vec<Arc<dyn PhysicalExtensionCodec>> = vec![Arc::new(DistributedCodec {})];
+        codecs.extend(get_distributed_user_codecs(cfg));
+        ComposedPhysicalExtensionCodec::new(codecs)
```
gabotechs, line 29: Now users can provide more than 1 codec

PR #159, opened 2025-09-26
Fix early drop stateful nodes

gabotechs:
Previously, once a stage was evicted, it was dropped immediately even if the `SendableRecordBatchStream` is still in progress. 

This is problematic because if a plan node inside the stage is holding some state (like an spawned task) necessary for the `SendableRecordBatchStream` to work, it might get dropped to soon, cancelling all tasks associated to that node too early.

This is something that does not happen in normal DataFusion, there, users can assume that references to plan nodes are going to live at least as long as the `SendableRecordBatchStream`.

### DIFF
--- a/src/common/callback_stream.rs
+++ b/src/common/callback_stream.rs
@@ -1,0 +1,84 @@
+use futures::Stream;
+use pin_project::{pin_project, pinned_drop};
+use std::fmt::Display;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// The reason why the stream ended:
+/// - [CallbackStreamEndReason::Finished] if it finished gracefully
+/// - [CallbackStreamEndReason::Aborted] if it was abandoned.
+#[derive(Debug)]
+pub enum CallbackStreamEndReason {
+    /// The stream finished gracefully.
+    Finished,
+    /// The stream was abandoned.
+    Aborted,
+}
+
+impl Display for CallbackStreamEndReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// Stream that executes a callback when it is fully consumed or gets cancelled.
+#[pin_project(PinnedDrop)]
+pub struct CallbackStream<S, F>
+where
+    S: Stream,
+    F: FnOnce(CallbackStreamEndReason),
+{
+    #[pin]
+    stream: S,
+    callback: Option<F>,
+}
+
+impl<S, F> Stream for CallbackStream<S, F>
+where
+    S: Stream,
+    F: FnOnce(CallbackStreamEndReason),
+{
+    type Item = S::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.project();
+
+        match this.stream.poll_next(cx) {
+            Poll::Ready(None) => {
+                // Stream is fully consumed, execute the callback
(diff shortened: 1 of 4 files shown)


gabotechs:
🤔 yeah, I think it should solve it. To make sure, you can port [tests/stateful_execution_plan.rs](https://github.com/datafusion-contrib/datafusion-distributed/pull/159/files#diff-0a8adaadd1cbb021ed413c146c91623b40f972f62127f75aae071ac02a9f9e61) to your PR and double check that it passes

src/common/callback_stream.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl Display for CallbackStreamEndReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// Stream that executes a callback when it is fully consumed or gets cancelled.
+#[pin_project(PinnedDrop)]
+pub struct CallbackStream<S, F>
```
NGA-TRAN, line 26: I read through the implementation and usage. I makes sense and a very nice fix.
My only question is about the name. Is it `CallBack` a common name for this kind of thing? The name does not spark immediate meaning of the work but if it is only me, it is fine to keep the name as-is

src/flight_service/do_get.rs
```diff
@@ -126,7 +128,17 @@ impl ArrowFlightEndpoint {
...
             .map_err(|err| Status::internal(format!("Error executing stage plan: {err:#?}")))?;
 
-        Ok(record_batch_stream_to_response(stream))
+        let schema = stream.schema();
+        let stream = with_callback(stream, move |_| {
+            // We need to hold a reference to the plan for at least as long as the stream is
+            // execution. Some plans might store state necessary for the stream to work, and
+            // dropping the plan early could drop this state too soon.
+            let _ = stage.plan;
+        });
```
NGA-TRAN, line 137: Very nice fix

tests/stateful_execution_plan.rs
```diff
@@ -0,0 +1,278 @@
...
+    use std::any::Any;
+    use std::fmt::Formatter;
+    use std::sync::{Arc, RwLock};
+    use std::time::Duration;
+    use tokio::sync::mpsc;
+    use tokio_stream::StreamExt;
+    use tokio_stream::wrappers::ReceiverStream;
+
+    #[tokio::test]
+    async fn stateful_execution_plan() -> Result<(), Box<dyn std::error::Error>> {
```
NGA-TRAN, line 50: It is hard for me to understand the problem of this test before the fix. Maybe adding comments in the plan where/why problem happens and why this fix solve it?

NGA-TRAN (approved):
The problem is well described and the fix is very nice. I just have a suggestion to describe where exactly the bug happens in your specific test

jayshrivastava:
@gabotechs Your test is passing on my PR here: https://github.com/datafusion-contrib/datafusion-distributed/pull/168. However, I think you should merge this fix. Even though my `TrailingFlightDataStream` fixes the issue, I will eventually make it so non `explain (analyze)` queries do not use it. In that case, the callback stream will be appropriate.

src/common/callback_stream.rs
gabotechs: Yeah, I'd say "callback" is pretty common for this in the software industry. A callback is a function whose execution gets deferred until a certain event happens. In this case, when the stream finishes.

tests/stateful_execution_plan.rs
gabotechs: Added a comment explaining what the test is about

PR #160, opened 2025-09-27
flight_service: emit metrics from ArrowFlightEndpoint

jayshrivastava:
This change updates the ArrowFlightEndpoint to collect metrics and emit them. When the last partition in a task is finished, the ArrowFlightEndpoint collects metrics and emits them via the TrailingFlightDataStream.

Previously, we would determine if a partition is finished when the request first hit the endpoint. Now, we do it on stream completition. This is crutial for metrics collection because we need to know that the stream is exhausted, meaning that there's no data flowing in the plan and metrics are not actively being updated.

Since the ArrowFlightEndpoint now emits metrics and NetworkBoundary plan nodes collect metrics, all coordinating StageExecs will now have the full collection of metrics for all tasks. This commit adds integration style tests that assert that the coordinator is recieving the full set of metrics.

Follow up work
- Only collect metrics if a configuration is set in the SessionContext, removing extra overhead
- Display metrics in the plan using EXPLAIN (ANALYZE) - consider using sqllogictest or similar to test the output


src/metrics/metrics_collecting_stream.rs
gabotechs, line 1: Some thought unrelated to this PR:

this `src/execution_plans` module is dedicated for DataFusion `ExecutionPlan` implementations. If a file inside here is not providing an `ExecutionPlan` implementation, it should probably not live here along with other `ExecutionPlan` implementations.

src/execution_plans/network_coalesce.rs
```diff
@@ -87,21 +90,25 @@ pub struct NetworkCoalesceReady {
...
-    /// metrics_collection is used to collect metrics from child tasks. It is empty when an
-    /// is instantiated (deserialized, created via [NetworkCoalesceExec::new_ready] etc...).
-    /// Metrics are populated in this map via [NetworkCoalesceExec::execute].
-    ///
-    /// An instance may receive metrics for 0 to N child tasks, where N is the number of tasks in
-    /// the stage it is reading from. This is because, by convention, the ArrowFlightEndpoint
-    /// sends metrics for a task to the last NetworkCoalesceExec to read from it, which may or may
-    /// not be this instance.
-    pub(crate) metrics_collection: Arc<DashMap<StageKey, Vec<MetricsSetProto>>>,
+    pub(crate) child_task_metrics: Arc<DashMap<StageKey, Vec<MetricsSetProto>>>,
```
gabotechs: Nit: It would be consistent with the naming in other parts of the project and call this `input_task_metrics`

src/flight_service/do_get.rs
```diff
@@ -126,15 +123,30 @@ impl ArrowFlightEndpoint {
...
     }
 }
 
 fn missing(field: &'static str) -> impl FnOnce() -> Status {
     move || Status::invalid_argument(format!("Missing field '{field}'"))
 }
 
-fn record_batch_stream_to_response(
+// Creates a tonic response from a stream of record batches. Handles
+// - RecordBatch to flight conversion partition tracking, stage eviction, and trailing metrics.
```
gabotechs: In Rust, documentation for functions is declared with triple `///`. A very immediate benefit is that IDEs are smart enough to display you the docs upon hovering usages.

src/flight_service/do_get.rs
```diff
@@ -126,15 +123,30 @@ impl ArrowFlightEndpoint {
...
+fn flight_stream_from_record_batch_stream(
+    stage_key: StageKey,
+    stage: Arc<StageExec>,
+    stage_data: TaskData,
```
gabotechs: It seems like we are unnecessarily passing `Arc<StageExec>` twice. `TaskData` already contains `Arc<StageExec>`

src/distributed_physical_optimizer_rule.rs
```diff
@@ -332,6 +321,20 @@ pub trait NetworkBoundary: ExecutionPlan {
...
+    /// collection is complete after the [NetworkBoundary] has been executed. It is undefined
+    /// what this returns during execution.
+    ///
+    /// An instance may receive metrics for 0 to N child tasks, where N is the number of tasks
+    /// in the stage it is reading from. This is because, by convention, the ArrowFlightEndpoint
+    /// sends metrics for a task to the last [NetworkBoundary] to read from it, which may or may
+    /// not be this instance.
+    fn metrics_collection(&self) -> Option<Arc<DashMap<StageKey, Vec<MetricsSetProto>>>> {
+        None
+    }
```
gabotechs: I think it might be better to not have this method here.

The only place where this is used is in:

https://github.com/datafusion-contrib/datafusion-distributed/blob/543a085489554446501b48c433cb724a1ded69c1/src/execution_plans/metrics.rs#L56-L72

And there, we are not making use of the `NetworkBoundary` trait, so this could perfectly just be normal methods in both `NetworkShuffleExec` and `NetworkCoalesceExec`.

It'd be nice to scope as much as possible the functionality of `NetworkBoundary`, leaving it exclusively dedicated to distributed planning, otherwise, unrelated features can creep in over time, bloating the trait and making the `SOLID` principles gods unhappy because we are not complying with the `I`

src/test_utils/plans.rs
gabotechs, line 1: I see that the functions in this files are not really test utils used in different places of the codebase, they are just used only in one place. 

I'd advocate towards:
- If a function is used only once, just place it where it's used
- If a function is used twice or more, promote it to a "utils"/"helpers"/"common" module for reuse

Otherwise, "utils"/"helpers"/"common" modules will grow unbounded with functions that are not really commonly used.

src/execution_plans/metrics.rs
```diff
@@ -29,35 +42,33 @@ pub struct TaskMetricsCollector {
 #[allow(dead_code)]
 pub struct MetricsCollectorResult {
```
gabotechs: I think this is no longer dead code right?

src/execution_plans/metrics.rs
```diff
@@ -270,219 +284,376 @@ impl ExecutionPlan for MetricsWrapperExec {
...
+        .task_metrics
+        .iter()
+        .map(|metrics| {
+            df_metrics_set_to_proto(metrics)
+                .map_err(|err| FlightError::ProtocolError(err.to_string()))
+        })
+        .collect::<Result<Vec<MetricsSetProto>, FlightError>>()?;
+    result
+        .child_task_metrics
+        .insert(stage_key.clone(), proto_task_metrics.clone());
```
gabotechs: These two clones seem unnecessary:

```suggestion
        .insert(stage_key, proto_task_metrics);
```

src/execution_plans/metrics.rs
```diff
@@ -270,219 +284,376 @@ impl ExecutionPlan for MetricsWrapperExec {
...
+    // Add the metrics for this task into the collection of task metrics.
+    // Skip any metrics that can't be converted to proto (unsupported types)
+    let proto_task_metrics = result
+        .task_metrics
+        .iter()
+        .map(|metrics| {
+            df_metrics_set_to_proto(metrics)
+                .map_err(|err| FlightError::ProtocolError(err.to_string()))
+        })
+        .collect::<Result<Vec<MetricsSetProto>, FlightError>>()?;
```
gabotechs: If you prefer, you can even let Rust infer the types:

```suggestion
        .collect::<Result<Vec<_>, _>>()?;
```

src/execution_plans/metrics.rs
```diff
@@ -270,219 +284,376 @@ impl ExecutionPlan for MetricsWrapperExec {
...
+    let (_, encoded_data) = data_gen
+        .encoded_batch(&empty_batch, &mut dictionary_tracker, &options)
+        .map_err(|e| {
+            FlightError::ProtocolError(format!("Failed to create empty batch FlightData: {e}"))
+        })?;
+    Ok(FlightData::from(encoded_data).with_app_metadata(buf))
+}
+
 #[cfg(test)]
 mod tests {
```
gabotechs: It looks like several structs in this file are unused (just used in tests):
- `TaskMetricsRewriter` is unused in production code, it's just used in the tests, so it probably should move there
- same for `MetricsWrapperExec`, it's unused in production code, so a better place for it could be under `mod tests {}`

gabotechs (commented):
Looking good! 🚀 this seems to go in the right direction.

src/execution_plans/metrics.rs
jayshrivastava: Done. Removed all unnecessary dead code markers.

src/execution_plans/metrics.rs
jayshrivastava: Done. Good catch :)

src/execution_plans/metrics.rs
jayshrivastava: Done

src/execution_plans/metrics.rs
jayshrivastava: The rewriter and wrapper will both be used in production soon when I implement the display portion of explain analyze, so I think it's okay to leave them.

src/distributed_physical_optimizer_rule.rs
jayshrivastava: Done.

I do think that network boundaries should be responsible for metrics collection, unfortunately, we don't gain the benefit of putting `metrics_collection()` in the trait here because we can't cast from `Arc<dyn ExecutionPlan>` to `NetworkBoundary` sadly. So, I've dropped the commit `distributed_physical_optimizer_rule: move metrics_collection to NetworkBoundary trait`

src/execution_plans/network_coalesce.rs
jayshrivastava: Updated in latest commit

src/metrics/metrics_collecting_stream.rs
jayshrivastava: This makes sense to me. I think it's okay to address in this PR. I pushed a commit 'refactor modules and files' to ensure that
- all metrics related code is in the `metrics` module
- only the `MetricsWrapperExec` lives in `execution_plans`

src/flight_service/do_get.rs
```diff
@@ -157,12 +177,109 @@ fn record_batch_stream_to_response(
...
+
+    Response::new(Box::pin(trailing_metrics_stream.map_err(|err| match err {
         FlightError::Tonic(status) => *status,
         _ => Status::internal(format!("Error during flight stream: {err}")),
     })))
 }
 
+// Collects metrics from the provided stage and encodes it into a stream of flight data using
+// the schema of the stage.
+pub fn collect_and_create_metrics_flight_data(
```
gabotechs: 👍 as this is only used here, it makes sense to have it here.

We can even make it private, as no one else is supposed to use it:

```suggestion
fn collect_and_create_metrics_flight_data(
```

src/flight_service/do_get.rs
```diff
@@ -157,12 +177,109 @@ fn record_batch_stream_to_response(
...
+        empty_flight_data_with_app_metadata(flight_app_metadata, stage.plan.schema())?;
+    Ok(Box::pin(stream::once(
+        async move { Ok(metrics_flight_data) },
+    )))
+}
+
+/// Creates a FlightData with the given app_metadata and empty RecordBatch using the provided schema.
+/// We don't use [arrow_flight::encode::FlightDataEncoder] (and by extension, the [arrow_flight::encode::FlightDataEncoderBuilder])
+/// since they skip messages with empty RecordBatch data.
+pub fn empty_flight_data_with_app_metadata(
```
gabotechs, line 263: Same here, only used once here, so it can be private:

```suggestion
fn empty_flight_data_with_app_metadata(
```

src/metrics/task_metrics_collector.rs
```diff
@@ -0,0 +1,338 @@
...
+            // Ensure each node has at least one metric which was collected.
+            for metrics_set in actual_metrics.iter() {
+                let metrics_set = metrics_set_proto_to_df(metrics_set).unwrap();
+                assert!(metrics_set.iter().count() > 0);
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_metrics_collection_e2e_1() {
```
gabotechs, line 300: Nice! any opinions on adding an integration test that does, for example, a TPCH query, and expects that all nodes have collected metrics?

It seems like this tests already provide good coverage, but giving ideas for if you want to take it one step further.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -0,0 +1,244 @@
...
+/// AggregateExec
+///  └── ProjectionExec
+///      └── NetworkShuffleExec
+///
+/// the task will be rewritten as
+///
+/// MetricsWrapperExec (wrapped: AggregateExec)
+///  └── MetricsWrapperExec (wrapped: ProjectionExec)
+///      └── NetworkShuffleExec
+/// (Note that the NetworkShuffleExec node is not wrapped)
```
gabotechs, line 27: I know this is not specific to this PR, but here's a question just for my understanding:

How is it that we cannot rely on the existing mechanism in `ExecutionPlans` for storing metrics?

https://github.com/apache/datafusion/blob/986cfcde0ee2bb2cd2e8cc684a497896b278a2d9/datafusion/physical-plan/src/execution_plan.rs#L455-L457

jayshrivastava: The reason we don't is because these metrics are populated during execution (cannot be written directly) and cannot be serialized. So, if a remote plan node runs, there's no way to have a copy of the same plan node return those metrics by calling `metrics()`. The workaround is to wrap the nodes.

It would be nice if we could use a more native method to inject metrics though if the `ExecutionPlan` interface were to allow it.

src/metrics/task_metrics_collector.rs
jayshrivastava: Yes. Ill add these when I can display the full explain output and assert the contents of the output. I'll probably need some way to obfuscate / hide the numbers since those are nondeterministic.

PR #162, opened 2025-09-29
Generalize functions for NetworkCoalesceExec creation

jonathanc-n:


src/execution_plans/network_coalesce.rs
```diff
@@ -101,21 +99,9 @@ pub struct NetworkCoalesceReady {
 }
 
 impl NetworkCoalesceExec {
-    pub fn from_coalesce_partitions_exec(
```
jonathanc-n, line 104: cc @gabotechs This look fine?

gabotechs: 😅 well, the idea here was to enforce at compile time that not any node could be passed, just nodes that merge all input partitions into 1 (`CoalescePartitionsExec` and `SortPreservingMergeExec`). 

I think the current API is a bit weird though, there are still ways for improving ergonomics, and probably this PR helps towards that goal, so let's ship it!

gabotechs (approved):

PR #164, opened 2025-09-29
Add partition coalescing at the head of the plan

gabotechs:
There's a couple of reasons why we want to manually enforce that the head node of the head stage of a distributed plan is a `CoalescePartitionsExec`:
- The `StageExec` node calls `try_assing_url`s lazily upon calling `.execute()`. This means that `.execute()` must only be called once, as we cannot afford to perform several random URL assignation while calling multiple partitions, as they will differ, producing an invalid plan
- Having a `CoalescePartitionsExec` on top of the plan allows us to place a new `NetworkCoalesceExec` on the top of the plan, executing the full plan in parallel.

### DIFF
--- a/src/execution_plans/stage.rs
+++ b/src/execution_plans/stage.rs
@@ -254,12 +254,22 @@ impl ExecutionPlan for StageExec {
         self.plan.properties()
     }
 
+    /// Executes a query in a distributed manner. This method will lazily perform URL assignation
+    /// to all the tasks, therefore, it must only be called once.
+    ///
+    /// [StageExec::execute] is only used for starting the distributed query in the same machine
+    /// that planned it, but it's not used for task execution in `ArrowFlightEndpoint`, there,
+    /// the inner `stage.plan` is executed directly.
     fn execute(
         &self,
         partition: usize,
         context: Arc<TaskContext>,
     ) -> Result<datafusion::execution::SendableRecordBatchStream> {
         if partition > 0 {
+            // The StageExec node calls try_assign_urls() lazily upon calling .execute(). This means
+            // that .execute() must only be called once, as we cannot afford to perform several
+            // random URL assignation while calling multiple partitions, as they will differ,
+            // producing an invalid plan
             return exec_err!(
                 "an executable StageExec must only have 1 partition, but it was called with partition index {partition}"
             );


src/execution_plans/stage.rs
```diff
@@ -170,15 +170,6 @@ impl StageExec {
         format!("Stage {:<3}{}", self.num, child_str)
     }
 
-    pub fn try_assign(self, channel_resolver: &impl ChannelResolver) -> Result<Self> {
```
gabotechs, line 173: Remove as it's unused

src/execution_plans/stage.rs
```diff
@@ -268,6 +259,12 @@ impl ExecutionPlan for StageExec {
         partition: usize,
         context: Arc<TaskContext>,
     ) -> Result<datafusion::execution::SendableRecordBatchStream> {
+        if partition > 0 {
+            return exec_err!(
+                "an executable StageExec must only have 1 partition, but it was called with partition index {partition}"
```
NGA-TRAN, line 274: Nice safety cache. We want to see this error/panic rather than running a wrong plan.

I wonder if it is useful to add comments somewhere, may be right above this `execute` function, describing something like you did:

> The StageExec node calls try_assing_urls lazily upon calling .execute(). This means that .execute() must only be called once, as we cannot afford to perform several random URL assignation while calling multiple partitions, as they will differ, producing an invalid plan

src/distributed_physical_optimizer_rule.rs
```diff
@@ -119,8 +120,14 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
     fn apply_network_boundaries(
         &self,
-        plan: Arc<dyn ExecutionPlan>,
+        mut plan: Arc<dyn ExecutionPlan>,
     ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+        if plan.output_partitioning().partition_count() > 1 {
+            // Coalescing partitions here will allow us to put a NetworkCoalesceExec on top
+            // of the plan, executing it in parallel.
+            plan = Arc::new(CoalescePartitionsExec::new(plan))
+        }
```
NGA-TRAN, line 129: 👍

tests/distributed_aggregation.rs
```diff
@@ -96,4 +96,60 @@ mod tests {
...
+              RepartitionExec: partitioning=Hash([RainToday@0], 3), input_partitions=3
+                AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))]
+                  DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[RainToday], file_type=parquet
+        ",
+        );
+
+        assert_snapshot!(physical_distributed_str,
+            @r"
+        ┌───── Stage 3   Tasks: t0:[p0] 
+        │ CoalescePartitionsExec
```
NGA-TRAN, line 135: 👍

src/distributed_physical_optimizer_rule.rs
```diff
@@ -501,15 +517,16 @@ mod tests {
...
-        │ CoalesceBatchesExec: target_batch_size=8192
-        │   HashJoinExec: mode=CollectLeft, join_type=Left, on=[(RainTomorrow@1, RainTomorrow@1)], projection=[MinTemp@0, MaxTemp@2]
-        │     CoalescePartitionsExec
-        │       NetworkCoalesceExec read_from=Stage 2, output_partitions=8, input_tasks=2
-        │     ProjectionExec: expr=[avg(weather.MaxTemp)@1 as MaxTemp, RainTomorrow@0 as RainTomorrow]
-        │       AggregateExec: mode=FinalPartitioned, gby=[RainTomorrow@0 as RainTomorrow], aggr=[avg(weather.MaxTemp)]
-        │         CoalesceBatchesExec: target_batch_size=8192
-        │           NetworkShuffleExec read_from=Stage 3, output_partitions=4, n_tasks=1, input_tasks=2
+        ┌───── Stage 4   Tasks: t0:[p0] 
+        │ CoalescePartitionsExec
```
NGA-TRAN, line 521: 👍

src/distributed_physical_optimizer_rule.rs
```diff
@@ -560,18 +577,22 @@ mod tests {
...
+          │   CoalesceBatchesExec: target_batch_size=8192
+          │     NetworkShuffleExec read_from=Stage 1, output_partitions=4, n_tasks=2, input_tasks=2
           └──────────────────────────────────────────────────
+            ┌───── Stage 1   Tasks: t0:[p0,p1,p2,p3,p4,p5,p6,p7] t1:[p0,p1,p2,p3,p4,p5,p6,p7] 
+            │ RepartitionExec: partitioning=Hash([RainToday@0, WindGustDir@1], 8), input_partitions=4
+            │   RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=2
+            │     AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday, WindGustDir@1 as WindGustDir], aggr=[]
+            │       PartitionIsolatorExec Tasks: t0:[p0,p1,__] t1:[__,__,p0] 
+            │         DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[RainToday, WindGustDir], file_type=parquet
+            └──────────────────────────────────────────────────
```
NGA-TRAN, line 595: This turns out to be a very good optimization for large data set that the final aggregation is done in many tasks/workers 👍  🎉 .

This isn’t directly related to the current PR, but worth noting as a future optimization:
If the input files are partitioned on the group-by columns `RainToday, WindGustDir` (i.e., the three files don’t overlap on those values), we can eliminate RoundRobin partitioning. This would allow stage 1 and stage 2 to merge into a single stage, executing both partial and final aggregation together. I’m currently exploring use cases for this and will share query plans soon to illustrate the idea more clearly.

src/execution_plans/stage.rs
jayshrivastava, line 274: nit: this is only the coordinator node right? it might be worth mentioning that here

src/distributed_physical_optimizer_rule.rs
```diff
@@ -268,6 +275,13 @@ impl DistributedPhysicalOptimizerRule {
             Ok(Transformed::new(node, true, TreeNodeRecursion::Jump))
         })?;
 
+        // The head stage is executable, and upon execution, it will lazily assign worker URLs to
+        // all tasks. This must only be done once, so the executable StageExec must only be called
+        // once on 1 partition.
+        if depth == 0 && distributed.data.output_partitioning().partition_count() > 1 {
+            distributed.data = Arc::new(CoalescePartitionsExec::new(distributed.data));
+        }
```
NGA-TRAN, line 283: Question: What did we do before this PR? I am asking because I do not see any removed or modified code in this PR

NGA-TRAN (approved):
This is actually a very nice fix and improvement. I just wonder what happened before this PR regarding the stage assigning task URLs.

src/execution_plans/stage.rs
gabotechs: added!

src/distributed_physical_optimizer_rule.rs
gabotechs: Previously we did pretty much nothing, and therefore, we were potentially executing multiple partitions with potentially conflicting URL assignations.

src/execution_plans/stage.rs
gabotechs: done!

gabotechs:
> It will be interesting to see how this evolves if we decide to make a "distributed endpoint" as @fmonjalet mentioned where we may stream different partitions to a distributed consumer / client like kafka etc.

👍 that makes sense. I'd say that's way too soon for the stage in which the project is in right now. Hopefully this feature can be driven by demand.

gabotechs:
> I'm curious if we can break the distributed optimizer rule into smaller rules (although having more than one rule makes it harder to migrate to distributed DF from vanilla DF)

I'd say right now is not too bad. The rule is pretty condensed and all the modifications provided go hand in hand with each other. 

As this grows, it's something we should definitely keep an eye on.

PR #165, opened 2025-09-29
fix: Incorrect weather parquet path in examples

zuston:
image


gabotechs (approved):
😄 Nice catch! thanks for this

PR #169, opened 2025-09-29
Js/sqllogictest

jayshrivastava:


robtandy (commented):
The CI looks like his passed, but clicking on the individual tests and inspecting output shows, 
```
...
❌ tests/sqllogictest/explain.slt: FAILED
...
```
which looked like it might have been missed.

jayshrivastava:
@robtandy Fixed! Good catch

gabotechs (commented):
It's nice to see more efforts towards testing 😄 , although I'm going to try to persuade you to not go down this path:

From [sqllogictest definition itself](https://sqlite.org/sqllogictest/doc/trunk/about.wiki):
> Sqllogictest is a program designed to verify that an SQL database engine computes correct results by comparing the results to identical queries from other SQL database engines.

Here are my reason to think sqllogictests are not suitable for testing this codebase:

- In this project, we care about bringing distributed capabilities to DataFusion, but it's not in our scope to test DataFusion correctness itself, that's done already upstream in their sqllogictest.

- This approach is not suitable for performing assertions over big pieces of data, where we want to assert results against normal DataFusion regardless of what they are (e.g. `assert_eq!(df_results, df_distributed_results)`).

- We lose `insta` as our snapshotting tool in favor of manually maintaining assertions over big portions of text. We can no longer do `cargo insta accept` for accepting changes in snapshots.

- Knowing that we still need our current snapshot testing, and we can't get rid of it, we would be maintaining two paths of testing with overlapping intentions, when we could just have 1.

- This approach requires adding additional custom scripts, pipelines, executable binaries and overall overhead to the project.

- It's more code to maintain. In terms of amount of code, this approach would pay off if we were to add a lot of sqllogictests, but I don't think we want or should do it, we should focus on testing things like: 
    - Are the metrics for each reconstructed across the network?
    - Are we able to propagate config properly across network boundaries?
    - Are we able to encode/decode custom user nodes?

In short:

I'd redirect the efforts towards reinforcing on the current integration testing tooling in order to make it even more ergonomic so that we can bring more tests if needed, but I'd advocated towards not using sqllogictests in this project.

jayshrivastava:
Ahh I didn't realize we had a solution for this. https://github.com/mitsuhiko/insta looks good.

Issue #177, opened 2025-10-06
Distributed plans do not support further optimization passes


gabotechs:
As mentioned in https://github.com/datafusion-contrib/datafusion-distributed/pull/163#discussion_r2400039303, this project leaves the distributed DataFusion plan in a state where it does not play well with other optimization rules in the DataFusion ecosystem.

The truth is that we are not playing by the DataFusion rules, and after distributed planning, we pretty much render the execution plan useless besides execution and display:
- The plan becomes just a tree of stages, making it impossible to perform further traversals to insides of a stage with just DataFusion tools.
- Our execution plans do not support to be called with new arbitrary children. The .with_new_children() call is either not supported or does not accept any arbitrary plan in our nodes.
- Our nodes need to be prepared to take any arbitrary node as a child. Some crates will wrap nodes in wrapper passthrough nodes, so downcasting children to specific types will fail.

This prevents the project from working well with other crates like https://github.com/datafusion-contrib/datafusion-tracing.

Ideally, the produced distributed plan should allow traversals and operations as any other non-distributed plan.

PR #178, opened 2025-10-06
Rework execution plan hierarchy for better interoperability

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/177.

This PR changes how we structure the distributed plans, without changing how we execute them or how we display them (mostly).

The idea is to bring us closer to a normal DataFusion plan, and:

- rather than nesting plans inside of `StageExec` plans:

```
┌─────────────────────────────┐
│          StageExec          │
│                             │
│┌───────────────────────────┐│
││                           ││
││      ProjectionExec       ││
││                           ││
│└───────────────────────────┘│
│┌───────────────────────────┐│
││                           ││
││   AggregateExec (final)   ││
││                           ││
│└───────────────────────────┘│
│┌───────────────────────────┐│
││                           ││
││    NetworkShuffleExec     ││
││                           ││
│└───────────────────────────┘│
└─────────────────────────────┘
               ▲               
               │               
               │               
               │               
┌─────────────────────────────┐
│          StageExec          │
│                             │
│┌───────────────────────────┐│
││                           ││
││      RepartitionExec      ││
││                           ││
│└───────────────────────────┘│
│┌───────────────────────────┐│
││                           ││
││  AggregateExec (partial)  ││
││                           ││
│└───────────────────────────┘│
│┌───────────────────────────┐│
││                           ││
││      DataSourceExec       ││
││                           ││
│└───────────────────────────┘│
└─────────────────────────────┘
```

To produce a flattened output as any other normal DataFusion plan

```
┌───────────────────────────┐
│                           │
│      DistributedExec      │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│      ProjectionExec       │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│   AggregateExec (final)   │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│    NetworkShuffleExec     │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│      RepartitionExec      │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│  AggregateExec (partial)  │
│                           │
└───────────────────────────┘
┌───────────────────────────┐
│                           │
│      DataSourceExec       │
│                           │
└───────────────────────────┘
```

This will play better with DataFusion tooling that tries to traverse the full plan, reassign children, repartition everything, etc...

In the way, it also results in less code and simpler

### DIFF
--- a/src/execution_plans/network_coalesce.rs
+++ b/src/execution_plans/network_coalesce.rs
@@ -127,6 +127,7 @@ impl NetworkBoundary for NetworkCoalesceExec {
             return plan_err!("can only return wrapped child if on Pending state");
         };
 
+        // As this node coalesces multiple tasks into 1, it must run in a stage with 1 task.
         if n_tasks > 1 {
             return Err(limit_tasks_err(1));
         }


src/stage.rs
```diff
@@ -367,103 +193,93 @@ const LDCORNER: &str = "└"; // Left bottom corner
...
-            let n_tasks = self.tasks.len();
-            let input_tasks = input_stage.tasks().len();
-            let partitions = plan.output_partitioning().partition_count();
-            let stage = ready.stage_num;
-            write!(
-                f,
-                " read_from=Stage {stage}, output_partitions={partitions}, n_tasks={n_tasks}, input_tasks={input_tasks}",
-            )?;
-        }
+pub fn display_plan_ascii(plan: &dyn ExecutionPlan) -> String {
```
jayshrivastava, line 196: Instead of using `display_plan_ascii`, can we make this more native? We can implement `Display` on `PartitionIsolator`, `NetworkCoalesceExec`, `NetworkShuffleExec`. Even though the display output may differ from what we have now, I think this plays well with DF principles. It will also work better with datafusion-tracing since it wraps every node. Using trait methods allows you to access `self` for each node without downcasting.

jayshrivastava: Generally, it would be nice if `displayable`, much like `execute()` worked out of the box without downcasting. I think truly "interoperating well" with datafusion actually means that we can use/implement these native methods. It gives me confidence that this would work with other people's DF extensions and custom nodes etc.

jayshrivastava: I think we can actually get everything in your PR working without downcasting outside of the optimizer step but I'm totally lost on metrics sadly. I think we will need to contribute upstream. 

The root issue that makes it hard is that I have no access to `self` nor the `TaskContext` while traversing a plan using the `TreeNodeRewriter`. We need a more native way to inject metrics IMO - in the `ExecutionPlan` trait.

gabotechs: > Instead of using display_plan_ascii, can we make this more native?

You could that now, and the output is nice also, but it does not contain ascii graphics. I kept this function for backwards compatibility with previous ways of visualizing plans.

gabotechs: > The root issue that makes it hard is that I have no access to self nor the TaskContext while traversing a plan using the TreeNodeRewriter. We need a more native way to inject metrics IMO - in the ExecutionPlan trait.

👍 that makes sense. Probably worth opening a discussion, maybe Ballista folks have something useful to say about that

src/distributed_physical_optimizer_rule.rs
```diff
@@ -333,6 +318,43 @@ pub trait NetworkBoundary: ExecutionPlan {
...
+pub trait NetworkBoundaryExt {
+    /// Downcasts self to a [NetworkBoundary] if possible.
+    fn as_network_boundary(&self) -> Option<&dyn NetworkBoundary>;
+    /// Returns whether self is a [NetworkBoundary] or not.
+    fn is_network_boundary(&self) -> bool {
+        self.as_network_boundary().is_some()
+    }
+}
+
+impl NetworkBoundaryExt for dyn ExecutionPlan {
```
jayshrivastava, line 331: I don't think this means we are able to support datafusion-tracing (even using this method). Any call to `as_network_boundary(&self)` on an `InstrumentationExec` which wraps a `Network*`, will not actually return a `NetworkBoundary`.

Even if you add this, we can never call it because we will never see a concrete `InstrumentationExec`. 
```rust
impl NetworkBoundaryExt for InstrumentationExec {
    fn as_network_boundary(&self) -> Option<&dyn NetworkBoundary> {
         self.inner.as_network_boundary()
    }
}
```
We will only see `dyn ExecutionPlan`, which will call the blanket implementation you wrote here. 

I also tried this
```
impl<T: ExecutionPlan + ?Sized> NetworkBoundaryExt for T {
    fn try_as_network_boundary(&self) -> Option<&dyn NetworkBoundary> {
        None
    }
}

impl NetworkBoundaryExt for InstrumentationExec {
    fn as_network_boundary(&self) -> Option<&dyn NetworkBoundary> {
         self.inner.as_network_boundary()
    }
}
```
but now the compiler throws an error because these are conflicting implementations (since `InstrumentationExec` implements `ExecutionPlan + ?Sized`.

This all boils down to specialization being unstable: https://github.com/rust-lang/rust/issues/31844. You can't have a "default blanket implementation" for a trait and override it in stable rust.

src/execution_plans/distributed.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn new(plan: Arc<dyn ExecutionPlan>) -> Self {
+        Self { plan }
+    }
+
+    fn prepare_plan(
+        &self,
+        urls: &[Url],
+        codec: &dyn PhysicalExtensionCodec,
+    ) -> datafusion::common::Result<Arc<dyn ExecutionPlan>> {
+        let prepared = Arc::clone(&self.plan).transform_up(|plan| {
```
jayshrivastava, line 38: Can this be done during planning? Totally fine to do as a follow up. Curious why we do it here.

src/distributed_physical_optimizer_rule.rs
```diff
@@ -333,6 +318,43 @@ pub trait NetworkBoundary: ExecutionPlan {
...
+        } else if let Some(node) = self.as_any().downcast_ref::<NetworkCoalesceExec>() {
+            Some(node)
+        } else {
+            None
+        }
+    }
+}
+
+/// Helper enum for storing either borrowed or owned trait object references
+enum Referenced<'a, T: ?Sized> {
```
jayshrivastava, line 344: This seems like it would be useful as a util. Seems a bit out of place here.

src/distributed_physical_optimizer_rule.rs
jayshrivastava: Ultimately, `datafusion-distributed` is at odds with `datafusion-tracing` because `as_any().downcast_ref` breaks under indirection caused by wrapper types like `InstrumentedExec`.

- are wrappers like in `datafusion-tracing` an anti pattern? 
- is downcasting after planning an anti-pattern?

I suspect that `InstrumentedExec` is actually the anti pattern.
- All other "wrappers" like `CooperativeExec` expose the children, so they would be (a) encountered during a plan traversal; and (b) downcastable. I asked claude and it seems that `InstrumentedExec` is the outlier - we could ask Geoffrey about why it's implemented this way.
- The old `StageExec` implementation was problematic for the same reason - we were obfuscating the parent-child relationships.
- When you want a transparent wrapper, why not just add it between two nodes and return an empty result in the `impl Display`?
- As you mentioned before, `as_any` is available in the `ExecutionPlan` trait.

src/stage.rs
jayshrivastava: Stepping back, I think it's impossible to actually render "natively". For any child node in vanilla DF, there's never more than one parent, I think. In our case, we could have `NetworkShuffleExecs` that read from the same task / stage, so we need to have our own renderer-visitor such as below to avoid rendering the same task twice. This seems right.

jayshrivastava: A more high level thought - maybe we should implement our own visitor which is capable of traversing tasks uniquely? Then we could display using that visitor. Not necessary in this PR but I'm just thinking about loud.

In the metrics use case, it would allow users to go find a particular node in a particular task and grab its metrics. Not sure how important that is - but that's certainly allowed in vanilla DF, so why not this project? 

(Also I would still like to be able to inject metrics retroactively if possible by adding `new_with_metrics`, but the `MetricsWrapperExec` can get the job done in the mean time)

src/distributed_physical_optimizer_rule.rs
jayshrivastava: If `InstrumentedExec` just delegated its `as_any` to the underlying type, we would avoid the issue. This would be a good question for DF maintainers...

It's only used here to avoid double wrapping. I wonder why they even have to worry about that.  It's supposed to be the last optimizer rule, so it shouldn't get run more than once...

https://github.com/datafusion-contrib/datafusion-tracing/blob/f0aee9ed2960fa101570ddc7ad11670c8ee64289/datafusion-tracing/src/instrument_rule.rs#L73

jayshrivastava (commented):
Sorry for the delay. I've been thinking about this change a lot. Overall, this change is good 👍🏽. I will give it a line-by-line review tmr morning. 

It would be valuable to ask Geoffrey and/or the DF maintainers about the downcasting thing which I've commented about below. I believe I'm pro-downcasting though, it seems like the DF way.

src/distributed_physical_optimizer_rule.rs
gabotechs: I imagine if `InstrumentedExec::as_any` returned the underlaying `ExecutionPlan`, downcasting would work, but upon reassigning a new version of the node in a `plan.transform_down()` operation the `InstrumentedExec` wrapper will be lost.

I'd say, that still sounds like a better option than completely obfuscating the plan like `InstrumentedExec` is currently doing.

gabotechs: I get the feeling we are not going to solve this problem in this PR though... we'll probably need some further work in other places.

src/execution_plans/distributed.rs
gabotechs: Here we do two things:
1. URL assignation
2. Stage plan encoding

we want to do 1) as late as possible, because otherwise, if we do it during planning, by the time we get to execute the plan, the assigned URLs might no longer be valid. By making the URL assignation happen as close to execution as possible, we reduce the risk of workers no longer being available. This is how it was done previously also, it's not really something new.

we want to do 2) right before execution because plans in the "encoded" state are not visible or inspectable or pretty much anything, and with a plan at hand, we might want to be able to display it or things like that.

src/distributed_physical_optimizer_rule.rs
gabotechs: 🤷‍♂️ it's only really used here. I'm trying to be faithful to the "not put in a `common` module things that are not common". 

Do you think this better belongs to the `common` module?

src/stage.rs
gabotechs: A more "native" render will look like this:
```
DistributedExec
  CoalescePartitionsExec
    [Stage 2] => NetworkCoalesceExec: output_partitions=8, input_tasks=2
      AggregateExec: mode=FinalPartitioned, gby=[RainToday@0 as RainToday, WindGustDir@1 as WindGustDir], aggr=[]
        CoalesceBatchesExec: target_batch_size=8192
          [Stage 1] => NetworkShuffleExec: output_partitions=4, input_tasks=2
            RepartitionExec: partitioning=Hash([RainToday@0, WindGustDir@1], 8), input_partitions=4
              RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=2
                AggregateExec: mode=Partial, gby=[RainToday@0 as RainToday, WindGustDir@1 as WindGustDir], aggr=[]
                  PartitionIsolatorExec: t0:[p0,p1,__] t1:[__,__,p0] 
                    DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[RainToday, WindGustDir], file_type=parquet

```

This is what you currently get if you do:
```rust
        let display = displayable(physical_plan.as_ref()).indent(true).to_string();
```

I'd say, that's how a native render should look like. All our ASCII graphics trickery is probably not suitable to how DF wants to render things by default, and for the same reason that we have a separate method for rendering Graphviz, we probably also should use a separate method for rendering ASCII

src/common/execution_plan_ops.rs
```diff
@@ -0,0 +1,13 @@
+use datafusion::common::plan_err;
+use datafusion::error::DataFusionError;
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::Arc;
+
+pub fn one_child(
```
jayshrivastava: nit: a doc comment would be nice
nit: would `src/common/execution_plan.rs` or `src/execution_plans/common.rs` make more sense?
nit: maybe `require_one_child` makes more sense

src/distributed_physical_optimizer_rule.rs
jayshrivastava: I think it's okay to put here. We can move things into common if we ever see repetition.

src/distributed_physical_optimizer_rule.rs
```diff
@@ -298,29 +287,25 @@ pub trait NetworkBoundary: ExecutionPlan {
     /// Returns the information necessary for building the next stage.
     /// - The head node of the stage.
     /// - the amount of tasks that stage will have.
-    fn to_stage_info(
-        &self,
-        n_tasks: usize,
-    ) -> Result<(Arc<dyn ExecutionPlan>, usize), DataFusionError>;
+    fn to_stage_info(&self, n_tasks: usize) -> Result<(Arc<dyn ExecutionPlan>, usize)>;
```
jayshrivastava, line 290: I'm a bit confused by this. The implementations return an error if n_tasks > 1 or they are not in the pending state. I'm not sure why. It would be nice if this was captured in the name and/or comment.

src/distributed_physical_optimizer_rule.rs
```diff
@@ -298,29 +287,25 @@ pub trait NetworkBoundary: ExecutionPlan {
...
     /// information to perform any internal transformations necessary for distributed execution.
-    fn to_distributed(
-        &self,
-        stage_num: usize,
-        stage_head: &Arc<dyn ExecutionPlan>,
-    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError>;
+    fn with_input_stage(&self, input_stage: Stage) -> Result<Arc<dyn ExecutionPlan>>;
+
+    /// Returns the assigned input [Stage], if any.
+    fn input_stage(&self) -> Option<&Stage>;
```
jayshrivastava, line 303: Not a blocker but this returns None IFF the boundary is pending, which only happens during planning. It would be nice to capture this semantic in the type system somehow. Just something to think about. I don't have any ideas at the moment.

src/distributed_physical_optimizer_rule.rs
```diff
@@ -422,23 +443,23 @@ mod tests {
             r#"SELECT count(*), "RainToday" FROM weather GROUP BY "RainToday" ORDER BY count(*)"#;
         let plan = sql_to_explain(query, 2).await.unwrap();
         assert_snapshot!(plan, @r"
-        ┌───── Stage 3   Tasks: t0:[p0] 
+        ┌───── DistributedExec ── Tasks: t0:[p0] 
         │ ProjectionExec: expr=[count(*)@0 as count(*), RainToday@1 as RainToday]
         │   SortPreservingMergeExec: [count(Int64(1))@2 ASC NULLS LAST]
-        │     NetworkCoalesceExec read_from=Stage 2, output_partitions=8, input_tasks=2
+        │     [Stage 2] => NetworkCoalesceExec: output_partitions=8, input_tasks=2
```
jayshrivastava, line 449: I really like this

src/common/execution_plan_ops.rs
gabotechs: `src/execution_plans/common.rs` sounds like a good place for helpers that are used in more than one plan, but that are scoped to `src/execution_plans`. 

Followed your suggestion with this, and with a couple other helpers that were also only used in `src/execution_plans`

src/distributed_physical_optimizer_rule.rs
gabotechs: Added a comment in `NetworkCoalesceExec`.

TL;DR: `NetworkCoalesceExec` coalesces N tasks into 1, therefore, it must run in exactly 1 task. That error is a controlled error that signals the planner to assign 1 task to it.

src/distributed_physical_optimizer_rule.rs
gabotechs: 🤔 yeah agree... I also have no ideas

PR #182, opened 2025-10-10
implement distributed EXPLAIN ANALYZE

jayshrivastava:
This change adds support for displaying a distributed EXPLAIN ANALYZE output with metrics. It updates the TPCH
validation tests to assert the EXPLAIN ANALYZE output for each query. I collected the single node
results [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/js/full-explain-analyze-rebased-comparison/tests/tpch_validation_test.rs) - using the `output_rows` metric values, we can cross check
that the distributed plan metrics are correct.

Implemenation notes:
- Adds `src/explain.rs` to stores the main entrypoint to rendering the output string
- We now re-write the whole plan before rendering using `rewrite_distributed_plan_with_metrics()` which
  is a new public API that takes an executed `DistributedExec` and wraps each node in the appropriate metrics
  - A user can call this method on an executed plan and traverse it to collect metrics from particular nodes
    (This may be hard though because all nodes are wrapped in MetricsWrapperExec...)
- Significantly refactors TaskMetricsRewriter -> StageMetricsWriter. See comment.

Informs: https://github.com/datafusion-contrib/datafusion-distributed/issues/123

Other follow up work:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/185
- https://github.com/datafusion-contrib/datafusion-distributed/issues/184
- https://github.com/datafusion-contrib/datafusion-distributed/issues/188
- https://github.com/datafusion-contrib/datafusion-distributed/issues/189
- https://github.com/datafusion-contrib/datafusion-distributed/issues/190


src/execution_plans/distributed.rs
```diff
@@ -23,11 +27,47 @@ use url::Url;
...
-        Self { plan }
+        Self {
+            plan,
+            prepared_plan: Arc::new(Mutex::new(None)),
+            display_ctx: None,
+        }
+    }
+
+    /// Returns a special stage key used to identify the root "stage" of the distributed plan.
+    /// TODO: reconcile this with display_plan_graphviz
```
jayshrivastava: Looking for thoughts here.

In display_plan_graphviz, we convert the `DistributedExec` to a stage for simplicity where that stage has a different stage id and query id. Maybe they should be on the same page, I'm not sure.

src/explain.rs
```diff
@@ -0,0 +1,40 @@
...
+/// explain_analyze renders an [ExecutionPlan] with metrics.
+pub fn explain_analyze(executed: Arc<dyn ExecutionPlan>) -> Result<String, DataFusionError> {
+    // Check if the plan is distributed by looking for a root DistributedExec.
+    match executed.as_any().downcast_ref::<DistributedExec>() {
+        None => Ok(DisplayableExecutionPlan::with_metrics(executed.as_ref())
+            .indent(true)
+            .to_string()),
+        Some(dist_exec) => {
+            // If the plan was distributed, collect metrics from the coordinating stage exec.
+            // TODO: Should we move this into the DistributedExec itself or a new ExplainAnalyzeExec?
```
jayshrivastava: Looking for thoughts here as well

tests/tpch_validation_test.rs
```diff
@@ -48,12 +48,74 @@ mod tests {
...
+            │     ProjectionExec: expr=[l_extendedprice@1 * (Some(1),20,0 - l_discount@2) as __common_expr_1, l_quantity@0 as l_quantity, l_extendedprice@1 as l_extendedprice, l_discount@2 as l_discount, l_tax@3 as l_tax, l_returnflag@4 as l_returnflag, l_linestatus@5 as l_linestatus], metrics=[output_rows=<metric>, elapsed_compute=<metric>]
+            │       CoalesceBatchesExec: target_batch_size=8192, metrics=[output_rows=<metric>, elapsed_compute=<metric>]
+            │         FilterExec: l_shipdate@6 <= 1998-09-02, projection=[l_quantity@0, l_extendedprice@1, l_discount@2, l_tax@3, l_returnflag@4, l_linestatus@5], metrics=[output_rows=<metric>, elapsed_compute=<metric>]
+            │           PartitionIsolatorExec: t0:[p0,p1,__,__,__,__] t1:[__,__,p0,p1,__,__] t2:[__,__,__,__,p0,p1] , metrics=[]
```
jayshrivastava, line 70: Metrics is empty, which is fine. One day, we can update the `PartitionIsolatorExec` to collect metrics.

Should we have this only render `PartitionIsolatorExec: t0:[p0,p1,__,__,__,__]` instead of all the tasks?

tests/tpch_validation_test.rs
```diff
@@ -48,12 +48,74 @@ mod tests {
...
+        │   [Stage 2] => NetworkCoalesceExec: output_partitions=24, input_tasks=4, metrics=[]
+        └──────────────────────────────────────────────────
+          ┌───── Stage 2 ── Task t0:[p0,p1,p2,p3,p4,p5] 
+          │ SortExec: expr=[l_returnflag@0 ASC NULLS LAST, l_linestatus@1 ASC NULLS LAST], preserve_partitioning=[true], metrics=[output_rows=<metric>, elapsed_compute=<metric>, spill_count=<metric>, spilled_bytes=<metric>, spilled_rows=<metric>, batches_split=<metric>]
+          │   ProjectionExec: expr=[l_returnflag@0 as l_returnflag, l_linestatus@1 as l_linestatus, sum(lineitem.l_quantity)@2 as sum_qty, sum(lineitem.l_extendedprice)@3 as sum_base_price, sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount)@4 as sum_disc_price, sum(lineitem.l_extendedprice * Int64(1) - lineitem.l_discount * Int64(1) + lineitem.l_tax)@5 as sum_charge, avg(lineitem.l_quantity)@6 as avg_qty, avg(lineitem.l_extendedprice)@7 as avg_price, avg(lineitem.l_discount)@8 as avg_disc, count(Int64(1))@9 as count_order], metrics=[output_rows=<metric>, elapsed_compute=<metric>]
```
robtandy: The `metrics=[output_rows=<metric>, elapsed_compute=<metric>]` should they have values there instead of `<metric>?`

jayshrivastava: I've hidden them in the insta snapshot config because certain ones change between runs (ex. ones that measure time). 

I could keep some, like output_rows. I think those stay the same.

src/execution_plans/distributed.rs
```diff
@@ -1,20 +1,24 @@
...
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
 use rand::Rng;
 use std::any::Any;
 use std::fmt::Formatter;
 use std::sync::Arc;
+use std::sync::Mutex;
 use url::Url;
 
-/// [ExecutionPlan] that executes the inner plan in distributed mode.
+/// [ExecutionPan] that executes the inner plan in distributed mode.
```
gabotechs: pan == 🍞 in spanish

src/execution_plans/distributed.rs
gabotechs: 🤔 we probably should not be converting it to a stage neither here or there, as it's not really a stage...

src/flight_service/do_get.rs
```diff
@@ -143,8 +143,9 @@ impl ArrowFlightEndpoint {
         let stream = map_last_stream(stream, move |last| {
             if num_partitions_remaining.fetch_sub(1, Ordering::SeqCst) == 1 {
                 task_data_entries.remove(key.clone());
+                return last.and_then(|el| collect_and_create_metrics_flight_data(key, plan, el));
             }
-            last.and_then(|el| collect_and_create_metrics_flight_data(key, plan, el))
+            last
```
gabotechs, line 148: 🤔 thinking about it, we can be in situations where this is never sent.

DataFusion, under certain situations, might decide to abandon some streams (early finishes do to `LIMIT X` being reached). In those scenarios, I don't think we are correctly decreasing by 1 the `num_partitions_remaining`, we just decrease if we successfully exhaust the full stream.

If this happens, `num_partitions_remaining` is never going to reach 0, and this code will never kick in.

Doesn't look like a problem with this specific PR though...

src/execution_plans/distributed.rs
jayshrivastava: pain without the i 🤕

src/execution_plans/distributed.rs
```diff
@@ -23,11 +26,37 @@ use url::Url;
...
     pub fn new(plan: Arc<dyn ExecutionPlan>) -> Self {
-        Self { plan }
+        Self {
+            plan,
+            prepared_plan: Arc::new(Mutex::new(None)),
+            display_ctx: None,
+        }
+    }
+
+    pub(crate) fn with_display_ctx(&self, display_ctx: DisplayCtx) -> Self {
```
jayshrivastava: Using this to smuggle the display ctx into `display_plan_ascii` without adding a parameter... Maybe it's more clean to add an argument to `display_plan_ascii`?

src/flight_service/do_get.rs
jayshrivastava: Yeah I think we need to address this. I wonder if the same problem occurs in datafusion-tracing, which collects metrics from the plan during `Drop`

tests/tpch_validation_test.rs
```diff
@@ -48,12 +48,35 @@ mod tests {
...
             └──────────────────────────────────────────────────
         ");
+
```
jayshrivastava, line 51: I re-ran all these tests in single node mode and the results are available here: https://github.com/datafusion-contrib/datafusion-distributed/blob/js/full-explain-analyze-rebased-comparison/tests/tpch_validation_test.rs

I looked over the results and it looks like `output_rows` values match between the distributed results here and the single node results (+ I asked claude to analyze everything), so it looks like the metrics should be correct 👍🏽

tests/tpch_validation_test.rs
jayshrivastava: Resolved. We now keep output_rows

src/flight_service/do_get.rs
jayshrivastava: I filed https://github.com/datafusion-contrib/datafusion-distributed/issues/187

jayshrivastava:
@gabotechs This is ready for another review :) 

Now, you can walk the plan programatically and collect metrics. Plus we don't display each task separately anymore.

I've left some comments and updated the PR desc to make it easier to review. Also happy to sync tmr on this too.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -188,48 +394,86 @@ mod tests {
...
+        // Assert that metrics from all tasks are present.
+        // actual_stage_node_metrics_set contains metrics for all task ex. [output_rows=1, elapsed_compute=1, output_rows=2, elapsed_compute=2...]
+        for (node_id, actual_stage_node_metrics_set) in actual_metrics.iter().enumerate() {
+            // actual_task_node_metrics_set contains metrics for one task ex. [output_rows=1, elapsed_compute=1]
+            for (task_id, actual_task_node_metrics_set) in actual_stage_node_metrics_set
+                .iter()
+                .chunks(num_metrics_per_task_per_node)
+                .into_iter()
+                .enumerate()
+            {
```
jayshrivastava, line 393: I will likely refactor some of this in the future. I think storing things in flat arrays and using `Arc<HashMap<StageKey, Vec<MetricsSetProto>>>` makes the code hard to reason about

src/execution_plans/distributed.rs
```diff
@@ -2,19 +2,22 @@ use crate::channel_resolver_ext::get_distributed_channel_resolver;
...
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
 use rand::Rng;
 use std::any::Any;
 use std::fmt::Formatter;
 use std::sync::Arc;
+use std::sync::Mutex;
 use url::Url;
 
-/// [ExecutionPlan] that executes the inner plan in distributed mode.
+/// [ExecutionPan] that executes the inner plan in distributed mode.
```
gabotechs:
```suggestion
/// [ExecutionPlan] that executes the inner plan in distributed mode.
```

src/execution_plans/distributed.rs
```diff
@@ -23,11 +26,36 @@ use url::Url;
 #[derive(Debug, Clone)]
 pub struct DistributedExec {
     pub plan: Arc<dyn ExecutionPlan>,
+    pub prepared_plan: Arc<Mutex<Option<Arc<dyn ExecutionPlan>>>>,
```
gabotechs: 🤔 thinking about this, if `DistributedExec` gets cloned, I don't think we want the two clones to share the same `prepared_plan`, as each clone could individually build it's own prepared plan. 

I see some DataFusion nodes do implement `Clone`:
https://github.com/apache/datafusion/blob/6368daf0b60611d5c6388b21def4d623b4c71fdc/datafusion/physical-plan/src/repartition/mod.rs#L476

While others don't:
https://github.com/apache/datafusion/blob/a68563b30ae03b074bdd687a181e42857e78f24f/datafusion/physical-plan/src/joins/cross_join.rs

I leave it to you. If you think cloning here could lead to unexpected situations, we can just remove the `Clone` derive. If you think it's fine, we can leave it.

src/execution_plans/distributed.rs
```diff
@@ -23,11 +26,36 @@ use url::Url;
...
+    pub(crate) fn with_display_ctx(&self, display_ctx: DisplayCtx) -> Self {
+        Self {
+            display_ctx: Some(display_ctx),
+            ..self.clone()
+        }
+    }
+
+    /// Returns the plan which is lazily prepared on execute() and actually gets executed.
+    /// It is updated on every call to execute(). Returns an error if .execute() has not been called.
+    pub(crate) fn pepared_plan(&self) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
```
gabotechs:
```suggestion
    pub(crate) fn prepared_plan(&self) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
```

src/execution_plans/distributed.rs
```diff
@@ -23,11 +26,36 @@ use url::Url;
...
+        self.prepared_plan
+            .lock()
+            .map_err(|e| DataFusionError::Internal(format!("Failed to lock prepared plan: {}", e)))?
```
gabotechs: There's a macro upstream for this:

```suggestion
            .map_err(|e| internal_datafusion_err!("Failed to lock prepared plan: {e}"))?
```

src/execution_plans/distributed.rs
```diff
@@ -23,11 +26,36 @@ use url::Url;
...
+            .clone()
+            .ok_or(DataFusionError::Internal(
+                "No prepared plan found. Was execute() called?".to_string(),
+            ))
```
gabotechs: Same here:
```suggestion
            .ok_or_else(|| {
                internal_datafusion_err!("No prepared plan found. Was execute() called?")
            })
```

I would try to use the `ok_or_else()` method generally rather than the `ok_or()`. Note that if we use `ok_or()` the error object and the String message get unnecessarily allocated if the call succeed, while using `ok_or_else()` the things just get allocated in case of failure.

src/execution_plans/distributed.rs
```diff
@@ -120,7 +150,14 @@ impl ExecutionPlan for DistributedExec {
         let channel_resolver = get_distributed_channel_resolver(context.session_config())?;
         let codec = DistributedCodec::new_combined_with_user(context.session_config());
 
-        let plan = self.prepare_plan(&channel_resolver.get_urls()?, &codec)?;
-        plan.execute(partition, context)
+        let prepared = self.prepare_plan(&channel_resolver.get_urls()?, &codec)?;
+        {
+            let mut guard = self.prepared_plan.lock().map_err(|e| {
+                DataFusionError::Internal(format!("Failed to lock prepared plan: {}", e))
```
gabotechs: Same as above:
```suggestion
                internal_datafusion_err!("Failed to lock prepared plan: {e}")
```

src/explain.rs
```diff
@@ -0,0 +1,20 @@
+use crate::display_plan_ascii;
+use crate::execution_plans::DistributedExec;
+use crate::metrics::rewrite_distributed_plan_with_metrics;
+use datafusion::error::DataFusionError;
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::display::DisplayableExecutionPlan;
+use std::sync::Arc;
+
+/// explain_analyze renders an [ExecutionPlan] with metrics.
+pub fn explain_analyze(executed: Arc<dyn ExecutionPlan>) -> Result<String, DataFusionError> {
```
gabotechs: I'd probably just put this in `stage.rs` along with the other visualizations. They probably should be somewhere else, maybe a new `display/` module or something like that, but for now I think it's preferable to put this there rather than having a new file just for 10 LOC

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,82 +1,235 @@
...
+) -> Result<Arc<dyn ExecutionPlan>> {
+    if !plan.as_any().is::<DistributedExec>() {
+        return Ok(plan);
+    }
+
+    // Metrics are available in the prepared plan of the DistributedExec.
+    let executed_plan = plan
+        .as_any()
+        .downcast_ref::<DistributedExec>()
+        .unwrap()
```
gabotechs: The docs say "does nothing if the root node is not DistributedExec", but I see here that it panics. Maybe you forgot to remove the `unwrap()` in favor of a graceful return?

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,82 +1,235 @@
...
+                None => {
+                    internal_err!("Expected input stage to be set in network boundary")
+                }
+            };
+        }
+        // Inject metrics into the DistributedExec itself.
+        if let Some(dist_exec) = plan.as_any().downcast_ref::<DistributedExec>() {
+            let dist_exec = dist_exec.with_display_ctx(DisplayCtx::new().with_metrics());
+            let plan_with_metrics =
+                rewrite_local_plan_with_metrics(dist_exec.plan.clone(), task_metrics.clone())?;
```
gabotechs: 🤔 But do we need to do this for the local plan also? I would expect the local plan to already have collected metrics successfully, as it happened locally.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,82 +1,235 @@
...
+                    Ok(Transformed::new(
+                        network_boundary.with_input_stage(Stage::new(
+                            stage.query_id,
+                            stage.num,
+                            plan_with_metrics,
+                            stage.tasks.len(),
+                        ))?,
+                        true,
+                        TreeNodeRecursion::Continue,
+                    ))
```
gabotechs: You can use the `Transformed::yes` shorthand for this:
```suggestion
                    Ok(Transformed::yes(network_boundary.with_input_stage(
                        Stage::new(
                            stage.query_id,
                            stage.num,
                            plan_with_metrics,
                            stage.tasks.len(),
                        ),
                    )?))
```

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,82 +1,235 @@
...
+            if idx >= metrics.len() {
+                return internal_err!("not enough metrics provided to rewrite plan");
+            }
+            let node_metrics = metrics[idx].clone();
+            idx += 1;
+            Ok(Transformed::new(
+                Arc::new(MetricsWrapperExec::new(node.clone(), node_metrics)),
+                true,
+                TreeNodeRecursion::Continue,
+            ))
```
gabotechs: Same here:

```suggestion
            Ok(Transformed::yes(Arc::new(MetricsWrapperExec::new(
                node.clone(),
                node_metrics,
            ))))
```

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,82 +1,235 @@
...
+/// TODO(#185): Add labels for each task
+pub struct StageMetricsRewriter<'a> {
+    /// A mapping from [StageKey] (which uniquely identifies a task)
+    /// to a list of [MetricsSetProto]. `metrics_collection[task][i]` is the metrics for
+    /// the i-th node in the task encountered via a pre-order traversal.
+    metrics_collection: Arc<HashMap<StageKey, Vec<MetricsSetProto>>>,
+    /// The stage containing the plan to be rewritten.
+    stage: &'a Stage,
+    /// The index of a node in the plan.
     idx: usize,
```
gabotechs: 🤔 isn't it a bit overkill using a `struct` for this? It looks like a plain function could be more than enough:

<details><summary>Shortened version with a plain function</summary>

```rust
pub fn stage_metrics_rewriter(
    stage: &Stage,
    metrics_collection: Arc<HashMap<StageKey, Vec<MetricsSetProto>>>,
    let mut node_idx = 0;

    let plan = stage.plan.decoded()?;

    plan.clone().transform_down(|plan| {
        // Stop at network boundaries.
        if plan.as_network_boundary().is_some() {
            return Ok(Transformed::new(plan, false, TreeNodeRecursion::Jump));
        }

        // Collect metrics for this node. It should contain metrics from each task.
        let mut stage_metrics = MetricsSetProto::new();

        for idx in 0..stage.tasks.len() {
            let stage_key = StageKey {
                query_id: Bytes::from(stage.query_id.as_bytes().to_vec()),
                stage_id: stage.num as u64,
                task_number: idx as u64,
            };
            match metrics_collection.get(&stage_key) {
                Some(task_metrics) => {
                    if node_idx >= task_metrics.len() {
                        return internal_err!(
                            "not enough metrics provided to rewrite task: {} metrics provided",
                            task_metrics.len()
                        );
                    }
                    let node_metrics = task_metrics[node_idx].clone();
                    for metric in node_metrics.metrics.iter() {
                        stage_metrics.push(metric.clone());
                    }
                }
                None => {
                    return internal_err!(
                        "not enough metrics provided to rewrite task: missing metrics for task {} in stage {}",
                        idx,
                        stage.num
                    );
                }
            }
        }

        node_idx += 1;

        let wrapped_plan_node: Arc<dyn ExecutionPlan> = Arc::new(MetricsWrapperExec::new(
            plan.clone(),
            metrics_set_proto_to_df(&stage_metrics)?,
        ));
        Ok(Transformed::yes(wrapped_plan_node))
    }).map(|v| v.data)
}

```

</details>

src/metrics/mod.rs
```diff
@@ -3,4 +3,7 @@ pub(crate) mod proto;
 mod task_metrics_collector;
 mod task_metrics_rewriter;
 pub(crate) use metrics_collecting_stream::MetricsCollectingStream;
-pub(crate) use task_metrics_collector::TaskMetricsCollector;
+pub(crate) use task_metrics_collector::{MetricsCollectorResult, TaskMetricsCollector};
+#[allow(unused_imports)]
+pub(crate) use task_metrics_rewriter::StageMetricsRewriter;
```
gabotechs: Why would this be exported? I don't think this is going to be used anywhere else right?

jayshrivastava: Updated. This was out of date.

src/metrics/task_metrics_rewriter.rs
jayshrivastava: Makes sense. Done!

src/metrics/task_metrics_rewriter.rs
jayshrivastava: Done

src/metrics/task_metrics_rewriter.rs
jayshrivastava: Done

src/metrics/task_metrics_rewriter.rs
jayshrivastava: Yes, we need to do this because we take the metrics from the `prepared_plan`, which is executed and has metrics, and put them in the `plan`, which does not have metrics. 

We want `plan` to have metrics because that's the real plan returned by `children()` that would be encountered during a traversal by a user. Also, the display code render the `plan` and not the `prepared_plan`, mainly because it has decoded stages it seems.

src/metrics/task_metrics_rewriter.rs
jayshrivastava: On the line above, we return if it's not `DistributedExec` so the unwrap is safe. However, the right way to write this in rust is 
```
let Some(distributed_exec) = plan.as_any().downcast_ref::<DistributedExec>() else {
      return Ok(plan);
};
```
I've updated this code to be the above

src/explain.rs
jayshrivastava: Done

src/execution_plans/distributed.rs
jayshrivastava: Done

src/execution_plans/distributed.rs
jayshrivastava: Done and good to know! Thanks.

src/execution_plans/distributed.rs
jayshrivastava: Done

src/execution_plans/distributed.rs
jayshrivastava: Done

src/execution_plans/distributed.rs
jayshrivastava: I think it makes sense to remove it. Done.

src/execution_plans/distributed.rs
jayshrivastava: Done! Sorry I lost this fix earlier.

src/execution_plans/distributed.rs
gabotechs: Yeah, I think it's a bit weird that `DisplayCtx` is a field in the `DistributedExec` struct, when it's irrelevant for execution purposes. 

If you ask me, I think threading a `bool` like this:

```rust
pub fn display_plan_ascii(plan: &dyn ExecutionPlan, show_metrics: bool) -> String;
```

Is a more than good enough solution for now, and if we find ourselves needing to pass around more fields in the future, we can bring back the extra struct as means of transportation for display configuration.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,115 +1,263 @@
...
+    let metrics_collection = Arc::new(input_task_metrics);
+
+    let transformed = plan.transform_down(|plan| {
+        // Transform all stages using NetworkShuffleExec and NetworkCoalesceExec as barriers.
+        if let Some(network_boundary) = plan.as_network_boundary() {
+            return match network_boundary.input_stage() {
+                Some(stage) => {
+                    // TODO: This transform is a bit inefficient because we traverse the plan nodes more than once.
+                    // For now, this is acceptable for clarity.
+                    let plan_with_metrics =
```
gabotechs: 🤔 more than once is exactly twice right? if that's the case, it does not sound too bad.

I'd try to either address this in this PR, or to remove the TODO comment replacing it with another that says that we are fine with this tradeoff.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -1,115 +1,263 @@
...
+/// Returns an error if the distributed plan was not executed.
+pub fn rewrite_distributed_plan_with_metrics(
+    plan: Arc<dyn ExecutionPlan>,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    let Some(distributed_exec) = plan.as_any().downcast_ref::<DistributedExec>() else {
+        return Ok(plan);
+    };
+
+    // Collect metrics from the DistributedExec.
+    // TODO: Should we move this into the DistributedExec itself or a new ExplainAnalyzeExec?
```
gabotechs: you mean move this all `rewrite_distributed_plan_with_metrics` there? if that's the case, I think it's fine as it is now.

I'll either address the TODO comment or remove it, but I would try to avoid leaving TODO comments that are not tracked by tickets that will get immediately solved after this PR.

gabotechs (approved):
💯 Leaving a +1 here, but let's make sure that all the remaining TODOs are either:
- Addressed in this PR.
- Tracked by GitHub issues that will get immediately worked on after this PR.
- Removed leaving a comment about why we are fine with a non-perfect state of things.

Issue #191, opened 2025-10-16
Information Schema Queries fail if distributed is enabled


marc-pydantic:
If an information schema query (e.g. `SHOW COLUMNS FROM records`) is ran with `datafusion-distributed` enabled, the resulting query fails due to no serialization support for `StreamingTableExec`, a similar problem of missing serialization like the one that was fixed in [#17200.](https://github.com/apache/datafusion/pull/17200).

Our initial attempt to solve that in a similar manner, i.e. by adding the missing serialization quickly turned out to be a rather large change. A more pragmatic approach was to detect if an information schema query was performed --- our first implementation does so checking if there is a table scan on a table named "information_schema" --- and simply disable distributed execution in this case.

This could potentially be lowered, i.e. made part of datafusion(-distributed), to not distribute these kinds of queries in the first place.

adriangb:
The point being that there's little point in distributing information schema queries.

The cases like `SHOW COLUMNS` are easy, they are specialized plans, can't be used in joins, etc.

On the other hand:

```sql
select * from information_schema.df_settings s1 JOIN information_schema.df_settings s2 using (name);
```

Does work.

Like Marc said, these plans generate `StreamingTableExec`s:

```sql
> explain analyze select * from information_schema.df_settings s1 JOIN information_schema.df_settings s2 using (name);
+-------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| plan_type         | plan                                                                                                                                                                                                                                                                                                                                                                     |
+-------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Plan with Metrics | CoalesceBatchesExec: target_batch_size=8192, metrics=[output_rows=109, elapsed_compute=19.333µs]                                                                                                                                                                                                                                                                         |
|                   |   HashJoinExec: mode=Partitioned, join_type=Inner, on=[(name@0, name@0)], projection=[name@0, value@1, description@2, value@4, description@5], metrics=[output_rows=109, elapsed_compute=1.778637ms, build_input_batches=12, build_input_rows=109, input_batches=12, input_rows=109, output_batches=12, build_mem_used=30205, build_time=861.457µs, join_time=185.588µs] |
[...]

PR #195, opened 2025-10-17
Add DistributedPlanError::NonDistributable rule and do not distribute SHOW COLUMNS

gabotechs:
https://github.com/datafusion-contrib/datafusion-distributed/issues/191.

There are certain nodes we cannot distribute, like the ones related to table introspection. This PR adds a new controlled error in the distributed planner that prompts it to not distribute the query at all in a controlled manner, and uses it for `StreamingTableExec`.

### DIFF
--- a/src/distributed_physical_optimizer_rule.rs
+++ b/src/distributed_physical_optimizer_rule.rs
@@ -10,6 +10,7 @@ use datafusion::physical_plan::ExecutionPlanProperties;
 use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
 use datafusion::physical_plan::joins::{HashJoinExec, PartitionMode};
 use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+use datafusion::physical_plan::streaming::StreamingTableExec;
 use datafusion::{
     common::tree_node::{Transformed, TreeNode},
     config::ConfigOptions,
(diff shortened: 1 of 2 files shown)


src/distributed_physical_optimizer_rule.rs
```diff
@@ -376,14 +395,17 @@ enum DistributedPlanError {
     /// Prompts the planner to limit the amount of tasks used in the stage that is currently
     /// being planned.
     LimitTasks(usize),
+    /// Signals the planner that this whole plan is non-distributable. This can happen if
+    /// certain nodes are present, like [StreamingTableExec], which are typically used in
+    /// queries that rather performing some execution, they perform some introspection.
+    NonDistributable,
```
jayshrivastava: we can put the plan name ex, `StreamingTableExec::name()` in here if it makes sense

src/distributed_physical_optimizer_rule.rs
```diff
@@ -259,6 +273,11 @@ impl DistributedPhysicalOptimizerRule {
                             }
                             dnode = Referenced::Arced(dnode.as_ref().with_input_task_count(*limit)?);
                         }
+                        Some(DistributedPlanError::NonDistributable) => {
+                            // This full plan is non-distributable, so abort any task and stage
+                            // assignation.
+                            return Err(non_distributable_err());
```
jayshrivastava: nit: return `e` to save an alloc?

jayshrivastava (approved):

src/distributed_physical_optimizer_rule.rs
jayshrivastava: also it may be useful to expose an option for users to specify nodes which they do not want to be distributed

src/distributed_physical_optimizer_rule.rs
```diff
@@ -630,6 +658,21 @@ mod tests {
...
+        let query = r#"SHOW COLUMNS from weather"#;
+        let plan = sql_to_explain(query, 2).await.unwrap();
+        assert_snapshot!(plan, @r"
+        CoalescePartitionsExec
+          NetworkCoalesceExec
+            ProjectionExec: expr=[table_catalog@0 as table_catalog, table_schema@1 as table_schema, table_name@2 as table_name, column_name@3 as column_name, data_type@5 as data_type, is_nullable@4 as is_nullable]
+              CoalesceBatchesExec: target_batch_size=8192
+                FilterExec: table_name@2 = weather
+                  RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=1
+                    StreamingTableExec: partition_sizes=1, projection=[table_catalog, table_schema, table_name, column_name, is_nullable, data_type]
```
NGA-TRAN: So a non-distributed plan can still include `NetworkCoalesceExec`? I had assumed it was exclusive to distributed plans. Was the intended fix to ensure that plans with `StreamingTableExec` remain non-distributed?

src/distributed_physical_optimizer_rule.rs
gabotechs: Yeah, users should probably be able to throw this errors themselves if they want to. Still have not figure out what's the right way of exposing that.

src/distributed_physical_optimizer_rule.rs
gabotechs: Oh wait, you are right, there's something wrong here. Let me fix it

gabotechs: nice catch! 👍

PR #196, opened 2025-10-20
Refactor distributed planner into its own folder

gabotechs:
Change factored out from https://github.com/datafusion-contrib/datafusion-distributed/pull/197.

As we find the need of adding more code to the distributed planner, it'd be nice to have a dedicated module for it. It's big enough and complex enough to organize it in a module rather than a single file.

### DIFF
--- a/src/distributed_planner/network_boundary.rs
+++ b/src/distributed_planner/network_boundary.rs
@@ -1,0 +1,86 @@
+use crate::{NetworkCoalesceExec, NetworkShuffleExec, Stage};
+use datafusion::common::plan_err;
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::Arc;
+
+/// Necessary information for building a [Stage] during distributed planning.
+///
+/// [NetworkBoundary]s return this piece of data so that the distributed planner know how to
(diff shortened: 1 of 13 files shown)


src/distributed_planner/network_boundary.rs
```diff
@@ -0,0 +1,86 @@
...
+pub trait NetworkBoundaryExt {
+    /// Downcasts self to a [NetworkBoundary] if possible.
+    fn as_network_boundary(&self) -> Option<&dyn NetworkBoundary>;
+    /// Returns whether self is a [NetworkBoundary] or not.
+    fn is_network_boundary(&self) -> bool {
+        self.as_network_boundary().is_some()
+    }
+}
+
+impl NetworkBoundaryExt for dyn ExecutionPlan {
```
jayshrivastava, line 76: Curious if this can be made generic

gabotechs: 🤔 what do you mean by generic here?

jayshrivastava: Ah right. Please ignore 🙇🏽

PR #197, opened 2025-10-20
Add dynamic task config based on DataFusion extension options

gabotechs:
Enhances the way we configure the `network_coalesce_tasks` and `network_shuffle_tasks` parameters. Rather than having these parameters as part of the `DistributedPhysicalOptimizerRule`, they are now it's own structure that implements `ConfigExtension`, the standard DataFusion way of passing configuration.

This allows us to ship two main enhancements:
- Now, both parameters can be set in SQL statements as any other parameters, for example:

```sql
SET distributed.network_coalesce_tasks = 2;
SET distributed.network_shuffle_tasks = 4;

SELECT ...
```

- Now, upon setting them programatically, a dynamic number that depends on the the underlaying plan can be specified:

```rust
            .with_default_features()
            .with_physical_optimizer_rule(Arc::new(DistributedPhysicalOptimizerRule))
            .with_distributed_network_coalesce_tasks(|plan: &Arc<dyn ExecutionPlan>| {
                if plan.as_any().is::<Something>() {
                    1
                } else {
                    4
                }
            })
            .build();

```

### DIFF
--- a/tests/udfs.rs
+++ b/tests/udfs.rs
@@ -10,14 +10,13 @@ mod tests {
     };
     use datafusion::physical_expr::expressions::lit;
     use datafusion::physical_expr::{Partitioning, ScalarFunctionExpr};
-    use datafusion::physical_optimizer::PhysicalOptimizerRule;
     use datafusion::physical_plan::empty::EmptyExec;
     use datafusion::physical_plan::repartition::RepartitionExec;
     use datafusion::physical_plan::{ExecutionPlan, execute_stream};
     use datafusion_distributed::test_utils::localhost::start_localhost_context;
     use datafusion_distributed::{
-        DistributedPhysicalOptimizerRule, DistributedSessionBuilderContext, assert_snapshot,
-        display_plan_ascii,
+        DistributedConfig, DistributedSessionBuilderContext, apply_network_boundaries,
+        assert_snapshot, display_plan_ascii, distribute_plan,
     };
     use futures::TryStreamExt;
     use std::any::Any;
@@ -59,10 +58,12 @@ mod tests {
 
         let node = wrap(wrap(Arc::new(EmptyExec::new(Arc::new(Schema::empty())))));
 
-        let physical_distributed = DistributedPhysicalOptimizerRule::default()
+        let cfg = DistributedConfig::default()
             .with_network_shuffle_tasks(2)
-            .with_network_coalesce_tasks(2)
-            .optimize(node, &Default::default())?;
+            .with_network_coalesce_tasks(2);
+        let node = apply_network_boundaries(node, &cfg)?;
+
+        let physical_distributed = distribute_plan(node)?;
 
         let physical_distributed_str = display_plan_ascii(physical_distributed.as_ref(), false);
 


src/distributed_planner/distributed_config.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub struct DistributedConfig {
+        /// Upon shuffling data, this defines how many tasks are employed into performing the shuffling.
+        /// ```text
+        ///  ( task 1 )  ( task 2 ) ( task 3 )
+        ///      ▲           ▲          ▲
+        ///      └────┬──────┴─────┬────┘
+        ///       ( task 1 )  ( task 2 )       N tasks
+        /// ```
+        /// This parameter defines N
+        pub(crate) network_shuffle_tasks: Option<PlanDependantUsize>, default = None
```
NGA-TRAN: I know this is a refactor but remind me:
-  Is N is the number of tasks of the input?  IIRC, isn't it the number of output tasks?
-  Where do we define the number of output tasks?

src/distributed_planner/distributed_config.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+extensions_options! {
+    pub struct DistributedConfig {
+        /// Upon shuffling data, this defines how many tasks are employed into performing the shuffling.
+        /// ```text
+        ///  ( task 1 )  ( task 2 ) ( task 3 )
+        ///      ▲           ▲          ▲
+        ///      └────┬──────┴─────┬────┘
+        ///       ( task 1 )  ( task 2 )       N tasks
```
NGA-TRAN, line 72: N = 2 in this example?

NGA-TRAN (approved):
Nice.
I only have minor questions to remind myself

src/distributed_planner/distributed_config.rs
gabotechs: yes, exactly

src/distributed_planner/distributed_config.rs
gabotechs: It's the number of input tasks. Whenever you define the tasks for a network boundary, you define how many input tasks it will have.

The number of tasks a network boundary and it's parent nodes will run on (the "output tasks", even if I don't think that's the correct term) is defined by the next upper network boundary, so, in short: you just define what's below the network boundary.

NGA-TRAN: Ah, I remember now. Thanks

src/distributed_planner/distributed_config.rs
```diff
@@ -0,0 +1,126 @@
+use datafusion::common::extensions_options;
+use datafusion::config::{ConfigExtension, ConfigField, Visit, default_config_transform};
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::prelude::SessionConfig;
+use std::fmt::{Debug, Display, Formatter};
+use std::sync::Arc;
+
+#[derive(Clone)]
+#[allow(clippy::type_complexity)]
+pub struct PlanDependantUsize(
```
jayshrivastava: nit: `PlanDependentUsize`

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -550,8 +563,13 @@ mod tests {
 
     #[tokio::test]
     async fn test_show_columns() {
-        let query = r#"SHOW COLUMNS from weather"#;
-        let plan = sql_to_explain(query, 2).await.unwrap();
+        let query = r#"
```
jayshrivastava, line 566: Very nice that we can do this in tests now

PR #198, opened 2025-10-21
Add arrow flight endpoint hooks

gabotechs:
Allow people to pass hooks that trigger on certain situations during the lifetime of an Arrow flight `do_get` query. This exposes a new API during `ArrowFlightEndpoint` construction that allows people to provide their own callbacks:

```rust
let mut endpoint = ArrowFlightEndpoint::try_new(DefaultSessionBuilder);
endpoint.on_plan(move |plan| {
    // log something based on the plan
    plan
});
```

I was tempted to introduce more hooks in this PR, but I think it'd be better to wait for use-cases rather than trying to guess them now.

The main use case for this right now is for wrapping nodes with https://github.com/datafusion-contrib/datafusion-tracing.

### DIFF
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -94,8 +94,10 @@ impl ArrowFlightEndpoint {
         let stage_data = once
             .get_or_try_init(|| async {
                 let proto_node = PhysicalPlanNode::try_decode(doget.plan_proto.as_ref())?;
-                let plan = proto_node.try_into_physical_plan(&ctx, &self.runtime, &codec)?;
-                let plan = (self.hooks.on_plan)(plan);
+                let mut plan = proto_node.try_into_physical_plan(&ctx, &self.runtime, &codec)?;
+                for hook in self.hooks.on_plan.iter() {
+                    plan = hook(plan)
+                }
 
                 // Initialize partition count to the number of partitions in the stage
                 let total_partitions = plan.properties().partitioning.partition_count();
@@ -232,7 +234,7 @@ mod tests {
         let plans_received = Arc::new(AtomicUsize::default());
         {
             let plans_received = Arc::clone(&plans_received);
-            endpoint.on_plan(move |plan| {
+            endpoint.add_on_plan_hook(move |plan| {
                 plans_received.fetch_add(1, Ordering::SeqCst);
                 plan
             });
--- a/src/flight_service/service.rs
+++ b/src/flight_service/service.rs
@@ -17,16 +17,10 @@ use tokio::sync::OnceCell;
 use tonic::{Request, Response, Status, Streaming};
 
 #[allow(clippy::type_complexity)]
+#[derive(Default)]
 pub(super) struct ArrowFlightEndpointHooks {
-    pub(super) on_plan: Arc<dyn Fn(Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan> + Sync + Send>,
-}
-
-impl Default for ArrowFlightEndpointHooks {
-    fn default() -> Self {
-        Self {
-            on_plan: Arc::new(|plan| plan),
-        }
-    }
+    pub(super) on_plan:
+        Vec<Arc<dyn Fn(Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan> + Sync + Send>>,
 }
 
 pub struct ArrowFlightEndpoint {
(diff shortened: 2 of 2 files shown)


src/flight_service/service.rs
```diff
@@ -10,15 +10,30 @@ use arrow_flight::{
...
+use datafusion::physical_plan::ExecutionPlan;
 use futures::stream::BoxStream;
 use std::sync::Arc;
 use tokio::sync::OnceCell;
 use tonic::{Request, Response, Status, Streaming};
 
+#[allow(clippy::type_complexity)]
+pub(super) struct ArrowFlightEndpointHooks {
+    pub(super) on_plan: Arc<dyn Fn(Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan> + Sync + Send>,
+}
```
gabotechs: IMO clippy here is a bit over dramatic. Not sure how this is too complex, but not line 59.

src/flight_service/do_get.rs
```diff
@@ -297,6 +308,8 @@ mod tests {
                 assert!(result.is_ok());
             }
         }
+        // As many plans as tasks should have been received.
+        assert_eq!(plans_received.load(Ordering::SeqCst), task_keys.len());
```
jayshrivastava, line 312: Not a blocker but an independent test where you swap out a plan node (ex. `DataSourceExec` with `EmptyExec`) would be nice

src/flight_service/service.rs
```diff
@@ -30,8 +39,21 @@ impl ArrowFlightEndpoint {
...
             session_builder: Arc::new(session_builder),
+            hooks: ArrowFlightEndpointHooks::default(),
         })
     }
+
+    /// Adds a callback for when an [ExecutionPlan] is received in the `do_get` call.
+    ///
+    /// The callback takes the plan and returns another plan that must be either the same,
+    /// or equivalent in terms of execution. Mutating the plan by adding nodes or removing them
+    /// will make the query blow up in unexpected ways.
```
jayshrivastava, line 50: You can add an assertion that the final schema matches within the `ArrowFlightEndpoint` so we fail early with "schema mismatch in hook" to make it clear that a hook caused this sort of error.

The harder part is asserting that the plan structure is the same, which is important for metrics. Traversing the plan to assert this would be expensive. Since we only support wrapping (because the plan structure cannot change), can we leverage `with_new_children` in the hook itself?

ex.
```rust
trait Hook {
    fn apply(Arc<dyn ExecutionPlan>) -> bool
    fn new(Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan>
}


// in the endpoint

fn apply_hook<H: Hook>(hook: H, plan: Arc<dyn ExecutionPlan>) {
    plan.transform_up(
        |node| {
            if hook.apply(node) {
                Transformed::yes(hook.new(node).with_new_children(node.children()))
            }
        } 
   )
}
```

This makes hooks a bit more expensive bc we necessarily traverse the whole plan, where previously we we were not. But for `datafusion-tracing`, we would have traversed the whole plan anyways.

This might be more effort than it's worth so I'll leave the decision up to you. If you do implement something like this, then I'm happy to take another look :)

src/flight_service/do_get.rs
gabotechs: 🤔 but that would fail pretty bad, part of the contract with this hook is that people should not do that. Or do you expect that to pass?

src/flight_service/service.rs
gabotechs: 🤔 I imagine in most cases there's just going to be no hook, I think it's worth not penalizing those cases.

The "schema mismatch" might be a good idea though

jayshrivastava: Sure that makes sense

src/flight_service/do_get.rs
jayshrivastava: I think swapping out a leaf node like `DataSourceExec` will work because it preserves the plan structure, as long as the schema matches.

gabotechs: But that's not something we are willing to support right? the contract is more something like: "don't touch anything that changes the plan", we just don't enforce it at runtime for performance reasons.

src/flight_service/service.rs
gabotechs: I think for now I'm going to keep it simple and not do anything. In case we find it necessary to add some checks we can refer back to this conversation.

PR #200, opened 2025-10-22
Fix user provided UDFs encoding

gabotechs:
When a UDF was inside a `Partitioning` field, we were failing to deserialize it as there we have a recently built `SessionContext` with no notion of UDFs.

This is a limitation of DF APIs that will be addressed in future releases, but for now, this fix should be enough.

### DIFF
--- a/src/protobuf/distributed_codec.rs
+++ b/src/protobuf/distributed_codec.rs
@@ -8,7 +8,7 @@ use datafusion::arrow::datatypes::Schema;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::common::internal_datafusion_err;
 use datafusion::error::DataFusionError;
-use datafusion::execution::FunctionRegistry;
+use datafusion::execution::{FunctionRegistry, SessionStateBuilder};
 use datafusion::physical_expr::EquivalenceProperties;
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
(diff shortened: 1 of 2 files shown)


tests/udfs.rs
gabotechs, line 1: This test replicates the issue and ensures it never happens again.

src/protobuf/distributed_codec.rs
```diff
@@ -54,7 +54,16 @@ impl PhysicalExtensionCodec for DistributedCodec {
         // TODO: The PhysicalExtensionCodec trait doesn't provide access to session state,
         // so we create a new SessionContext which loses any custom UDFs, UDAFs, and other
         // user configurations. This is a limitation of the current trait design.
-        let ctx = SessionContext::new();
+        let state = SessionStateBuilder::new()
+            .with_scalar_functions(
+                registry
+                    .udfs()
```
NGA-TRAN, line 60: Question: is UDF always registered with `with_scalar_functions`?

NGA-TRAN (approved):
LGTM, assuming the UDF is consistently registered using `with_scalar_functions`.

src/protobuf/distributed_codec.rs
gabotechs: I think there's other ways of registering them, but they are pretty much aliases to this one, all will result in the same: a `SessionContext` with some UDFs in it.

PR #202, opened 2025-10-23
metrics: make MetricsWrapperExec transparent

jayshrivastava:
Before this change, `name()` and `as_any()` on `MetricsWrapperExec` did not delegate calls to the wrapped node. This was a problem because a user could not traverse the plan and extract metrics from specific nodes.

After this change, those methods are delegated so a user can now extract metrics from specific nodes.


gabotechs (approved):
👍

PR #204, opened 2025-10-24
Return input task count in NetworkBoundary

gabotechs:
Will playing with the current API, I'm finding it useful to be able to traverse a distributed plan and see, for each `NetworkBoundary`, how many tasks it receives as an input.

The main use case for this right now is for letting people decide how many tasks will be spawned in a network boundary as a function as the input tasks of another network boundary, for example:

```
               ┌────────────────────┐             
               │NetworkCoalesceExec │             
               │  max(10, 5) tasks  │             
               └────────────────────┘             
                          │                       
           ┌──────────────┴────────────┐          
           ▼                           ▼          
┌────────────────────┐      ┌────────────────────┐
│ NetworkShuffleExec │      │NetworkCoalesceExec │
│      10 tasks      │      │      5 tasks       │
└────────────────────┘      └────────────────────┘
```

If I want to be able to do that in the upper network boundary, I need to be able to access how many input tasks the lower network boundaries are spawning

### DIFF
--- a/src/distributed_planner/network_boundary.rs
+++ b/src/distributed_planner/network_boundary.rs
@@ -35,6 +35,9 @@ pub trait NetworkBoundary: ExecutionPlan {
         input_tasks: usize,
     ) -> datafusion::common::Result<Arc<dyn NetworkBoundary>>;
 
+    /// Returns the input tasks assigned to this [NetworkBoundary].
+    fn input_task_count(&self) -> usize;
+
     /// Called when a [Stage] is correctly formed. The [NetworkBoundary] can use this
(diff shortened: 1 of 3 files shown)


NGA-TRAN (approved):
The concept looks good. The API is fine and we will need it for special use cases.

We have to be careful with the usage and may need to add document because it will be complicated if we want to shuffle to different number of tasks (see your `NetworkCoalesceExec` that needs `max(10, 5)` ). Usually, in optimal plan we do not do it except in some special cases. People use this have to know exactly what they are doing. 

I suggest we play with this in real examples and document it so it won't be abuse. I actually have a use case for this but it will take me a wile to show it

PR #205, opened 2025-10-26
remove schema adapter

adriangb:
This was added in #174 but I think a better fix is to just configure the IPC encoder to send dictionaries across the wire.

### DIFF
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -143,7 +143,7 @@ impl ArrowFlightEndpoint {
 
         let stream = FlightDataEncoderBuilder::new()
             .with_schema(schema)
-            // This tells the encoder to send dictioanries across the wire as-is.
+            // This tells the encoder to send dictionaries across the wire as-is.
             // The alternative (`DictionaryHandling::Hydrate`) would expand the dictionaries
             // into their value types, which can potentially blow up the size of the data transfer.
             // The main reason to use `DictionaryHandling::Hydrate` is for compatibility with clients


src/flight_service/do_get.rs
```diff
@@ -136,6 +136,7 @@ impl ArrowFlightEndpoint {
 
         let stream = FlightDataEncoderBuilder::new()
             .with_schema(stream.schema().clone())
+            .with_dictionary_handling(DictionaryHandling::Resend)
```
gabotechs, line 154: So, with this line, we are telling the flight data encoder to just send the dictionaries over the wire instead of hydrating them and re-encoding them. Is that right?

If you could include a very brief comment above this line that would be awesome 🙏

src/execution_plans/network_coalesce.rs
```diff
@@ -328,3 +327,37 @@ impl ExecutionPlan for NetworkCoalesceExec {
...
 }
+
+/// Garbage collects dictionary arrays in the given RecordBatch.
+/// 
+/// We apply this before sending RecordBatches over the network to avoid sending
+/// unused dictionary values.
+/// 
+/// Unused dictionary values can arise from operations such as filtering, where
+/// some dictionary keys may no longer be referenced in the filtered result.
+fn garbage_collect_dictionary_arrays(batch: RecordBatch) -> Result<RecordBatch, DataFusionError> {
```
gabotechs: 🤔 the comment says:

```rust
/// We apply this before sending RecordBatches over the network to avoid sending
/// unused dictionary values.
```

But `NetworkCoalesceExec` does not send RecordBatches over the network, it receives them.

Should this function be applied in `src/flight_service/do_get.rs` instead? that's the one that sends RecordBatches over the network.

src/flight_service/do_get.rs
NGA-TRAN: This is a great optimization. Thanks @adriangb

src/execution_plans/network_coalesce.rs
adriangb: Yep you're right. I moved it there.

adriangb: I also handled view arrays :)

Cargo.toml
```diff
@@ -15,6 +15,7 @@ chrono = { version = "0.4.42" }
 datafusion = { workspace = true }
 datafusion-proto = { workspace = true }
 arrow-flight = "56.1.0"
+arrow-select = "56.1.0"
```
adriangb, line 18: Note that this was already in the tree. I'm not sure why the dictionary gc stuff is in this crate, but it seems fine to add it for now, practically everyone is going to have it in their dep tree already.

gabotechs: completely reasonable IMO

src/flight_service/do_get.rs
```diff
@@ -134,8 +136,22 @@ impl ArrowFlightEndpoint {
...
 
+        let schema = stream.schema().clone();
+
+        // Apply garbage collection of dictionary and view arrays before sending over the network
+        let stream = stream.and_then(|rb| std::future::ready(garbage_collect_arrays(rb)));
+
         let stream = FlightDataEncoderBuilder::new()
-            .with_schema(stream.schema().clone())
+            .with_schema(schema)
+            // This tells the encoder to send dictioanries across the wire as-is.
```
gabotechs:
```suggestion
            // This tells the encoder to send dictionaries across the wire as-is.
```

PR #208, opened 2025-10-27
Adds test ensuring dictionary corruption does not occur anymore

marc-pydantic:
This test ensures #205 has the desired effect on distributed execution. I also snuck in a few docs and a single typo fix to make up for the ones I am adding :grimacing: .

It has a bit of half-life, as the fixes from https://github.com/apache/datafusion/pull/18303 will mask the issue again, but for now, it is better than nothing.

### DIFF
--- a/tests/distributed_aggregation.rs
+++ b/tests/distributed_aggregation.rs
@@ -157,6 +157,9 @@ mod tests {
     }
 
     /// Test that multiple first_value() aggregations work correctly in distributed queries.
+    // TODO: Once https://github.com/apache/datafusion/pull/18303 is merged, this test will lose
+    //       meaning, since the PR above will mask the underlying problem. Different queries or
+    //       a new approach must be used in this case.
     #[tokio::test]
     async fn test_multiple_first_value_aggregations() -> Result<(), Box<dyn Error>> {
         let (ctx, _guard) = start_localhost_context(3, DefaultSessionBuilder).await;


tests/distributed_aggregation.rs
```diff
@@ -149,4 +155,98 @@ mod tests {
 
         Ok(())
     }
+
+    /// Test that multiple first_value() aggregations work correctly in distributed queries.
+    // TODO: Once https://github.com/apache/datafusion/pull/18303 is merged, this test will lose
+    //       meaning, since the PR above will mask the underlying problem. Different queries or
+    //       a new approach must be used in this case.
```
adriangb, line 162: Should we just check for duplicate column names directly then? We could make a `MemoryExec` or something

marc-pydantic: I can look into that as a follow-up? I want to get the remaining issues with distributed activation first, this can be a side-quest.

adriangb: sure yes we'll see what the maintainers here think

tests/distributed_aggregation.rs
```diff
@@ -149,4 +155,98 @@ mod tests {
...
++----------+--------+-----+
+| 1        | trace1 | 100 |
+| 2        | trace2 | 200 |
+| 3        | trace3 | 300 |
+| 4        | trace4 | 400 |
++----------+--------+-----+";
+
+        // Print them out, the error message from `assert_eq` is otherwise hard to read.
+        println!("{}", expected_result);
+        println!("{}", actual_result);
```
jayshrivastava, line 244: nit: It would be preferable to only do this on failure. I think it's fine if we do 
```
if actual_result.to_string() != expected_result {
    println!("{}", expected_result);
    println!("{}", actual_result);
    panic!(...)
}
```

tests/distributed_aggregation.rs
jayshrivastava: I think this is fine. It's still a good test. If you could file and issue and put the number in the todo, that would help us track it :)

gabotechs (approved):
👍 makes sense. Thanks!

PR #209, opened 2025-10-30
Introduce with_distributed_execution 

gabotechs:
This PR refactors the public API people use for enriching a DataFusion `SessionState` with distributed capabilities.

Before, all the `with_distributed_*` and `set_distributed_*` methods were applicable to a lot of DataFusion structs:
- `SessionContext`
- `SessionConfig`
- `SessionState`
- `SessionStateBuilder`

But there's an ergonomic issue with that:

Users have no other option but to use `SessionStateBuilder` for adding configuring distributed capabilities, because that's the only possible place in DataFusion to add a `PhysicalOptimizerRule` implementation (`builder.with_physical_optimizer_rule()`).

This means that no matter how many ergonomic improvements we bring to other structs, people need to configure things through `SessionStateBuilder`.

This PR allows configuring distributed DataFusion with a new `with_distributed_execution()` that will automatically:
- set the `ChannelResolver` implementation
- inject a `DistributedConfig` struct in the `ConfigOptions`
- Add a `DistributedPhysicalOptimizerRule` as an optimization rule

As these are all things that users need to do proactively one way or another, we now just expose it in one method:

```rust
    .with_distributed_execution(CustomChannelResolver)
    .build();
```

---

This is a preparation PR for a bigger one: 
- https://github.com/datafusion-contrib/datafusion-distributed/pull/216

### DIFF
--- a/src/distributed_planner/distributed_physical_optimizer_rule.rs
+++ b/src/distributed_planner/distributed_physical_optimizer_rule.rs
@@ -66,6 +66,9 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
         let Some(cfg) = config.extensions.get::<DistributedConfig>() else {
             return Ok(plan);
         };
+        if cfg.network_coalesce_tasks.is_none() && cfg.network_shuffle_tasks.is_none() {
+            return Ok(plan);
+        }
         // We can only optimize plans that are not already distributed
         distribute_plan(apply_network_boundaries(plan, cfg)?)
     }
@@ -180,6 +183,9 @@ pub fn apply_network_boundaries(
 pub fn distribute_plan(
     plan: Arc<dyn ExecutionPlan>,
 ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    if plan.as_any().is::<DistributedExec>() {
+        return Ok(plan);
+    }
     let stage = match _distribute_plan_inner(Uuid::new_v4(), plan.clone(), &mut 1, 0, 1) {
         Ok(stage) => stage,
         Err(err) => {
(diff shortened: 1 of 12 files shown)


src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -66,6 +66,9 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
         let Some(cfg) = config.extensions.get::<DistributedConfig>() else {
             return Ok(plan);
         };
+        if cfg.network_coalesce_tasks.is_none() && cfg.network_shuffle_tasks.is_none() {
+            return Ok(plan);
+        }
```
gabotechs, line 71: As now `DistributedConfig` is always in the config, we cannot rely on it not being present as a signal that the query should not be distributed. Now we also need to check that it's not configured at all in order to skip the distribution.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -180,6 +183,9 @@ pub fn apply_network_boundaries(
 pub fn distribute_plan(
     plan: Arc<dyn ExecutionPlan>,
 ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    if plan.as_any().is::<DistributedExec>() {
+        return Ok(plan);
+    }
```
gabotechs, line 188: Not really related to this PR, but found this footgun while making the tests pass, so I thought that it might be fine to add it here, it would have helped me.

gabotechs:
@adriangb any feedback on this? for us (DataDog) this API change makes sense, but let me know what you guys think

PR #210, opened 2025-10-31
Struct is public only within crate.

### LINKED ISSUE #203: Make MetricsWrapperExec private

JSOD11:
Resolves #203.

The `MetricsWrapperExec` struct is [not exposed to users](https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/lib.rs). This notation makes it clear that the struct is public only within the crate.


gabotechs:
Thanks @JSOD11!

PR #211, opened 2025-10-31
Create stage key constructor

### LINKED ISSUE #186: Add helper to make stage keys
We reuse this code a lot.
```
StageKey {
    query_id: Bytes::from(input_stage.query_id.as_bytes().to_vec()),
    stage_id: input_stage.num as u64,
    task_number: target_task as u64,
}
```

adding `Stage::to_stage_key(task_number: usize)` would be a nice util

JSOD11:
Resolves #186 

Introduces a helper function for creating `StageKey` instances and replaces manual struct instantiations.

### DIFF
--- a/src/protobuf/distributed_codec.rs
+++ b/src/protobuf/distributed_codec.rs
@@ -239,7 +239,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
     }
 }
 
-/// A key that uniquely identifies a stage in a query
+/// A key that uniquely identifies a stage in a query.
 #[derive(Clone, Hash, Eq, PartialEq, ::prost::Message)]
 pub struct StageKey {
     /// Our query id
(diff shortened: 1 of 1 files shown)


src/protobuf/distributed_codec.rs
```diff
@@ -253,6 +253,16 @@ pub struct StageKey {
     pub task_number: u64,
 }
 
+impl StageKey {
+    pub fn new(query_id: Bytes, stage_id: u64, task_number: u64) -> StageKey {
```
jayshrivastava, line 258: I feel like the linter typically fails if pub fns do not have a comment 🤔. In this case, it's fine if there's no comment because it's so simple.

JSOD11: Updated 👍

PR #212, opened 2025-11-01
Update Readme with working git-lfs instructions

JSOD11:
Following the instructions in `README.md` didn't work for me, so I've added some new instructions. I also make a note that this step is required for tests to pass, since that wasn't clear to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,19 +250,15 @@ The integration tests also provide an idea about how to use the library and what
    ```
 
 2. **Install Git LFS and fetch test data:**
+
+This step downloads Parquet files used in some tests. It must be completed successfully for the testing suite to pass.
+
+Install Git LFS if you don't already have it: [Official docs](https://git-lfs.com/).
+
    ```bash
    git lfs install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -255,6 +255,15 @@ The integration tests also provide an idea about how to use the library and what
    git lfs checkout
    ```
 
+If the above does not work, try the following:
+   ```bash
+   brew install git-lfs
```
gabotechs: I don't think we can assume that people using this library are on MacOs, and these commands are irrelevant for Linux users.

What I'd do instead is to post a link that points to how to install `git-lfs`, ideally from an official source, and just let people follow that rather than putting the instruction for all the operating systems here.

gabotechs (changes requested):

README.md
JSOD11: Makes sense, updated. Thoughts?

gabotechs (approved):
👍 nice! in it goes

PR #216, opened 2025-11-04
Rework task assignation mechanism

gabotechs:
The previous task assignation mechanism was very rudimentary:

- take the `coalesce_tasks` and `shuffle_tasks` parameters as static numbers and use them indiscriminately for all stages.

This has a big problem: it's not possible to have a static number that makes sense for all the possible queries a system will see.

This PR proposes a new way of calculating tasks in the different stages of a query. For a plan divided into stages like this:
```
┌─────────────────┐         
│     Stage 3     │? tasks  
└────────▲────────┘         
         │                  
┌────────┴────────┐         
│     Stage 2     │? tasks  
└────────▲────────┘         
         │                  
┌────────┴────────┐         
│     Stage 1     │? tasks  
└─────────────────┘         
```

It will
1. Calculate the number of tasks for a bottom stage based on how much data the leaf node[s] (`DataSourceExec`s mostly) are expected to pull

```
┌─────────────────┐         
│     Stage 3     │? tasks  
└────────▲────────┘         
         │                  
┌────────┴────────┐         
│     Stage 2     │? tasks  
└────────▲────────┘         
         │                  
┌────────┴────────┐         
│     Stage 1     │3 tasks  
└─────────────────┘         
```

2. Based on the calculated tasks in the leaf stage (e.g. 3 tasks), calculate the amount of tasks in the next stage. This is done by multiplying the task count by a scale factor every time a node that increments or reduces the cardinality of the data appears, which is information present in the `fn cardinality_effect(&self) -> CardinalityEffect` method. For example, if "Stage 1" has a partial aggregation step, and the scale factor is 1.5, it will look like this:

```
┌─────────────────┐               
│     Stage 3     │? tasks        
└────────▲────────┘               
         │                        
┌────────┴────────┐               
│     Stage 2     │3/1.5 = 2 tasks
└────────▲────────┘               
         │                        
┌────────┴────────┐               
│     Stage 1     │3 tasks  (cardinality effect factor of 1.5)
└─────────────────┘               
```

3. This is repeated recursively until all tasks have been assigned to all stages, keeping into account the cardinality effect different nodes in subplans have. If there is no cardinality effect (e.g. `ProjectExec` nodes), then the task count is kept across stages:

```
┌─────────────────┐           
│     Stage 3     │2 tasks    
└────────▲────────┘           
         │                    
┌────────┴────────┐           
│     Stage 2     │2 tasks    
└────────▲────────┘           
         │                    
┌────────┴────────┐           
│     Stage 1     │3 tasks    
└─────────────────┘           
```

The calculation for the tasks needed for a leaf stage is decided by the user in the following trait:

```rust
/// Given a leaf node, provides an estimation about how many tasks should be used in the
/// stage containing it, and if the leaf node should be replaced by some other.
///
/// The distributed planner will try many [TaskEstimator]s in order until one provides an
/// estimation for a specific leaf node. Once that's done, upper stages will get its task
/// count calculated based on wether lower stages are reducing the cardinality of the data
/// or increasing it.
pub trait TaskEstimator {
    /// Function applied to leaf nodes that returns a [TaskEstimation] hinting how many
    /// tasks should be used in the [Stage] containing that leaf node, and if the leaf
    /// node itself should be modified somehow.
    fn estimate_tasks(
        &self,
        cfg: &ConfigOptions,
    ) -> Option<TaskEstimation>;
}
```

There is a sane default for `FileScanConfig` operations (reader parquet or csv files) that is configurable with a new `files_per_task` config parameter that will spawn the necessary amount of tasks so that no task takes more than `files_per_task` files, but users are anyways able to provide their own implementations overriding the default one with their own custom logic.

---

Performance impact of this PR

```
$ WORKERS=8 ./benchmarks/run.sh --threads 2 --files-per-task 2 --cardinality-task-sf 3.0 --path benchmarks/data/tpch_sf10

==== Comparison with the previous benchmark from 2025-11-13 13:51:19 UTC ====
os:        macos
arch:      aarch64
cpu cores: 16
threads:   2 -> 2
workers:   8 -> 8
 Query 1: prev=2886 ms, new=2829 ms, diff=1.02 faster ✔
 Query 2: prev= 394 ms, new= 400 ms, diff=1.02 slower ✖
 Query 3: prev=1220 ms, new=1231 ms, diff=1.01 slower ✖
 Query 4: prev=1137 ms, new= 803 ms, diff=1.42 faster ✅
 Query 5: prev=2010 ms, new=1826 ms, diff=1.10 faster ✔
 Query 6: prev=1186 ms, new= 722 ms, diff=1.64 faster ✅
 Query 7: prev=2622 ms, new=1836 ms, diff=1.43 faster ✅
 Query 8: prev=2854 ms, new=2225 ms, diff=1.28 faster ✅
 Query 9: prev=3634 ms, new=2895 ms, diff=1.26 faster ✅
Query 10: prev=1749 ms, new=1664 ms, diff=1.05 faster ✔
Query 11: prev= 598 ms, new= 397 ms, diff=1.51 faster ✅
Query 12: prev=1608 ms, new=1268 ms, diff=1.27 faster ✅
Query 13: prev=2346 ms, new=1368 ms, diff=1.71 faster ✅
Query 14: prev=1479 ms, new=1016 ms, diff=1.46 faster ✅
Query 15: prev=1599 ms, new=1202 ms, diff=1.33 faster ✅
Query 16: prev= 358 ms, new= 254 ms, diff=1.41 faster ✅
Query 17: prev=5101 ms, new=2936 ms, diff=1.74 faster ✅
Query 18: prev=4722 ms, new=3117 ms, diff=1.51 faster ✅
Query 19: prev=2886 ms, new=1757 ms, diff=1.64 faster ✅
Query 20: prev=2096 ms, new=1283 ms, diff=1.63 faster ✅
Query 21: prev=4907 ms, new=4244 ms, diff=1.16 faster ✔
Query 22: prev= 481 ms, new= 452 ms, diff=1.06 faster ✔
```

### DIFF
--- a/src/distributed_ext.rs
+++ b/src/distributed_ext.rs
@@ -7,7 +7,8 @@ use crate::protobuf::{set_distributed_user_codec, set_distributed_user_codec_arc
 use crate::{ChannelResolver, DistributedConfig, TaskEstimator};
 use datafusion::common::DataFusionError;
 use datafusion::config::ConfigExtension;
-use datafusion::execution::SessionStateBuilder;
+use datafusion::execution::{SessionState, SessionStateBuilder};
+use datafusion::prelude::{SessionConfig, SessionContext};
 use datafusion_proto::physical_plan::PhysicalExtensionCodec;
 use delegate::delegate;
 use http::HeaderMap;
@@ -359,53 +360,48 @@ pub trait DistributedExt: Sized {
     ) -> Result<(), DataFusionError>;
 }
 
-impl DistributedExt for SessionStateBuilder {
+impl DistributedExt for SessionConfig {
     fn set_distributed_option_extension<T: ConfigExtension + Default>(
         &mut self,
         t: T,
     ) -> Result<(), DataFusionError> {
-        set_distributed_option_extension(self.config().get_or_insert_default(), t)
+        set_distributed_option_extension(self, t)
     }
 
     fn set_distributed_option_extension_from_headers<T: ConfigExtension + Default>(
         &mut self,
         headers: &HeaderMap,
     ) -> Result<(), DataFusionError> {
-        set_distributed_option_extension_from_headers::<T>(
-            self.config().get_or_insert_default(),
-            headers,
-        )
+        set_distributed_option_extension_from_headers::<T>(self, headers)
     }
 
     fn set_distributed_user_codec<T: PhysicalExtensionCodec + 'static>(&mut self, codec: T) {
-        set_distributed_user_codec(self.config().get_or_insert_default(), codec)
+        set_distributed_user_codec(self, codec)
     }
 
     fn set_distributed_user_codec_arc(&mut self, codec: Arc<dyn PhysicalExtensionCodec>) {
-        set_distributed_user_codec_arc(self.config().get_or_insert_default(), codec)
+        set_distributed_user_codec_arc(self, codec)
     }
 
     fn set_distributed_channel_resolver<T: ChannelResolver + Send + Sync + 'static>(
         &mut self,
         resolver: T,
     ) {
-        let cfg = self.config().get_or_insert_default();
-        set_distributed_channel_resolver(cfg, resolver);
+        set_distributed_channel_resolver(self, resolver);
     }
 
     fn set_distributed_task_estimator<T: TaskEstimator + Send + Sync + 'static>(
         &mut self,
         estimator: T,
     ) {
-        set_distributed_task_estimator(self.config().get_or_insert_default(), estimator)
+        set_distributed_task_estimator(self, estimator)
     }
 
     fn set_distributed_files_per_task(
         &mut self,
         files_per_task: usize,
     ) -> Result<(), DataFusionError> {
-        let cfg = self.config().get_or_insert_default();
-        let d_cfg = DistributedConfig::from_config_options_mut(cfg.options_mut())?;
+        let d_cfg = DistributedConfig::from_config_options_mut(self.options_mut())?;
         d_cfg.files_per_task = files_per_task;
         Ok(())
     }
@@ -414,8 +410,7 @@ impl DistributedExt for SessionStateBuilder {
         &mut self,
         factor: f64,
     ) -> Result<(), DataFusionError> {
-        let cfg = self.config().get_or_insert_default();
-        let d_cfg = DistributedConfig::from_config_options_mut(cfg.options_mut())?;
+        let d_cfg = DistributedConfig::from_config_options_mut(self.options_mut())?;
         d_cfg.cardinality_task_count_factor = factor;
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


src/distributed_planner/task_estimator.rs
```diff
@@ -0,0 +1,187 @@
...
+/// about how many tasks should be employed in [Stage]s that contain leaf nodes.
+pub struct TaskEstimation {
+    /// The amount of tasks that should be used in the [Stage] containing the leaf node.
+    ///
+    /// Even if implementations get to decide this number, there are situations where it can
+    /// get overridden:
+    /// - If a [Stage] contains multiple leaf nodes, the one that declares the biggest
+    ///   task_count wins.
+    /// - If there are less available workers than this number, the amount of available workers
+    ///   is chosen.
```
jayshrivastava, line 21: Unsure if these two cases are covered by tests.

src/distributed_planner/distributed_config.rs
```diff
@@ -1,61 +1,145 @@
...
+    }
 }
 
 impl DistributedConfig {
-    /// Sets the amount of tasks used in a network shuffle operation.
-    pub fn with_network_shuffle_tasks(mut self, tasks: usize) -> Self {
-        self.network_shuffle_tasks = Some(tasks);
+    /// Appends a [TaskEstimator] to the list. [TaskEstimator] will be executed sequentially in
+    /// order on leaf nodes, and the first one to provide a value is the one that gets to decide
+    /// how many tasks are used for that [Stage].
```
jayshrivastava, line 55: nit: would it make sense to take the one that returns the max tasks? similar to the behavior where there's multiple leaf nodes with different estimates. 

Is this behavior asserted in a test anywhere? I don't think we have a test with multiple estimators.

src/distributed_planner/task_estimator.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn estimate_tasks(
+        &self,
+        plan: &Arc<dyn ExecutionPlan>,
+        cfg: &ConfigOptions,
+    ) -> Option<TaskEstimation> {
+        let d_cfg = cfg.extensions.get::<DistributedConfig>()?;
+        let dse: &DataSourceExec = plan.as_any().downcast_ref()?;
+        let file_scan: &FileScanConfig = dse.data_source().as_any().downcast_ref()?;
+
+        // Count how many distinct files we have in the FileScanConfig.
```
jayshrivastava: We need this?

src/distributed_planner/distributed_config.rs
```diff
@@ -1,61 +1,145 @@
...
-        pub network_coalesce_tasks: Option<usize>, default = None
+        /// Sets the maximum amount of files that will be assigned to each task. Reducing this
+        /// number will spawn more tasks for the same number of files.
+        pub files_per_task: usize, default = files_per_task_default()
+        /// Task multiplying factor for when a node declares that it changes the cardinality
+        /// of the data:
+        /// - If a node is increasing the cardinality of the data, this factor will increase.
+        /// - If a node reduces the cardinality of the data, this factor will decrease.
+        /// - In any other situation, this factor is left intact.
+        pub cardinality_task_count_factor: f64, default = cardinality_task_count_factor_default()
```
jayshrivastava, line 26: Can this stuff be set via session properties in datafusion sql? Asking so I can fuzz this.

src/distributed_planner/task_estimator.rs
jayshrivastava: A comment explaining why would help.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -89,86 +90,174 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+                this_stage_sf: 1.0,
+                next_stage_sf: 1.0,
+                plan,
+            });
+        };
+        return Ok(ApplyNetworkBoundariesCtx {
+            task_count: estimate.task_count,
+            this_stage_sf: 1.0,
+            next_stage_sf: 1.0,
+            plan: estimate.new_plan.unwrap_or(plan),
```
jayshrivastava, line 291: We should probably bubble this error up. It's bad if a user returned a task estimate but the node didn't transform. There might be bad read amplification if there's 10 tasks for a DataSourceExec but the PartitionIsolator was not applied.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -89,86 +90,174 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+    /// factor calculated for the next stage by the current one, resetting the next stage scale
+    /// factor.
+    ///
+    /// This is called whenever a new [NetworkBoundary] is introduced, which marks the end of
+    /// one [Stage], and the beginning of the next one.
+    fn scale_task_count_and_swap(&mut self) -> usize {
+        let task_count = (self.task_count as f64 * self.this_stage_sf).ceil() as usize;
+        self.this_stage_sf = self.next_stage_sf;
+        self.next_stage_sf = 1.0;
+        task_count
```
jayshrivastava: It would be nice if we explicitly ensured the returned value is >= 1 here. I'm nervous about `self.this_stage_sf` being zero. Also, I don't like that the task calculation / scale factor math is distributed so much (a few lines in`_apply_network_boundaries`, `apply_network_boundaries`, `scale_task_count_and_swap`). It would be nice if the final task count was calculated in one small function so it's easier to follow.

src/distributed_planner/task_estimator.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        Some(TaskEstimation {
+            task_count,
+            new_plan: Some(plan),
+        })
+    }
+}
+
+#[derive(Clone, Default)]
+pub(crate) struct CombinedTaskEstimator {
```
jayshrivastava, line 173: A small comment here and on FileScanConfigTaskEstimator would be nice.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -89,86 +90,174 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+    let ctx = _apply_network_boundaries(plan, cfg, urls.len())?;
+    Ok(ctx.plan)
+}
 
-    let result = plan.transform_up(|plan| {
-        // If this node is a DataSourceExec, we need to wrap it with PartitionIsolatorExec so
-        // that not all tasks have access to all partitions of the underlying DataSource.
-        if plan.as_any().is::<DataSourceExec>() {
-            let node = PartitionIsolatorExec::new(plan);
+struct ApplyNetworkBoundariesCtx {
```
jayshrivastava, line 203: A diagram like you have in the PR comment right here would be great :)

src/distributed_planner/distributed_config.rs
```diff
@@ -1,61 +1,145 @@
...
-        ///              ( task 1 )
-        ///                  ▲
-        ///      ┌───────────┴──────────┐
-        ///  ( task 1 )  ( task 2 ) ( task 3 )  N tasks
-        /// ```
-        /// This parameter defines N
-        pub network_coalesce_tasks: Option<usize>, default = None
+        /// Sets the maximum amount of files that will be assigned to each task. Reducing this
+        /// number will spawn more tasks for the same number of files.
+        pub files_per_task: usize, default = files_per_task_default()
```
jayshrivastava, line 20: Since this is closely tied to DataSourceExec, maybe we can specify that here in the name or at least in the comment.

src/distributed_planner/distributed_config.rs
```diff
@@ -1,61 +1,145 @@
...
+        /// Task multiplying factor for when a node declares that it changes the cardinality
+        /// of the data:
+        /// - If a node is increasing the cardinality of the data, this factor will increase.
+        /// - If a node reduces the cardinality of the data, this factor will decrease.
+        /// - In any other situation, this factor is left intact.
+        pub cardinality_task_count_factor: f64, default = cardinality_task_count_factor_default()
+        /// Collection of [TaskEstimator]s that will be applied to leaf nodes in order to
+        /// estimate how many tasks should be spawned for the [Stage] containing the leaf node.
+        pub(crate) __private_task_estimator: CombinedTaskEstimator, default = CombinedTaskEstimator::default()
```
jayshrivastava, line 29: nit: What if we did something like `CombinedTaskEstimator::with_default_estimators(DefaultEstimators)` where `DefaultEstimators` contains the default ones? It would make it clearer to users what the default estimators are.

src/distributed_planner/distributed_config.rs
```diff
@@ -1,61 +1,145 @@
...
+impl ConfigExtension for DistributedConfig {
+    const PREFIX: &'static str = "distributed";
+}
+
+// FIXME: Ideally, both ChannelResolverExtension and TaskEstimators would be passed as
+//  extensions in SessionConfig's AnyMap instead of the ConfigOptions. However, we need
+//  to pass this as ConfigOptions as we need these two fields to be present during
+//  planning in the DistributedPhysicalOptimizerRule, and the signature of the optimize()
+//  method there accepts a ConfigOptions instead of a SessionConfig.
+//  The following PR addresses this: https://github.com/apache/datafusion/pull/18168
```
jayshrivastava, line 92: Should we open a ticket with a list to things to do after upgrading?

src/distributed_planner/task_estimator.rs
gabotechs: Nope, I have definitely not made a good job in this PR for adding tests. I'll add them.

src/distributed_planner/distributed_config.rs
gabotechs: My intention was that task estimators have an "overriding" effect, where people could override the default behavior of the `FileScanConfigTaskEstimator` if they wanted to.

I think we still want to let people completely override what the default `FileScanConfigTaskEstimator` says, but it might make sense to do the max over the user-provided estimators, rather than just taking the first one.

src/distributed_planner/task_estimator.rs
gabotechs: Each entry is a PartitionedFile, meaning that in some cases it is just a chunk (a range) of a file. I don't think it makes a lot of sense to distribute across workers multiple chunks of the same file, here, we are mainly interested in actual different files, therefore calculating that number requires this HashSet.

Updated the comment explaining this.

src/distributed_planner/distributed_config.rs
gabotechs: Yes! it should be:

```
SET distributed.cardinality_task_count_factor = 3.0
```

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: The `unwrap_or` here is not really unwrapping an error (`Result<>`) it's unwrapping an `Option<>`. It's fine if the user returned a `TaskEstimation` with `new_plan == None`, as long as they handle `DistributedTaskContext` in their own custom nodes.

Remember that `PartitionIsolatorExec` here is relevant for `DataSourceExec(FileScanConfig)` nodes, but it is not necessarily relevant for anything else, therefor returning `new_plan == None` is completely fine.

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: 👍 makes sense to return an error if `task_count == 0`, adding it.

gabotechs: > Also, I don't like that the task calculation / scale factor math is distributed so much (a few lines in_apply_network_boundaries, apply_network_boundaries, scale_task_count_and_swap). It would be nice if the final task count was calculated in one small function so it's easier to follow.

It gets challenging as multiple things happen at multiple points in time, making it hard to condense everything into a single function:
- at every single node, we needed to upscale/downscale the factor depending on the cardinality
- just at network boundaries, we need to actually apply this factor to the task count in order to perform the final calculation

One thing I can do though, is try to condense all the functions that do the math in `ApplyNetworkBoundariesCtx`, that way, all this math logic is condensed in that single structure.

gabotechs: Just did that, let me know if it looks better

src/distributed_planner/task_estimator.rs
gabotechs: done

src/distributed_planner/distributed_config.rs
gabotechs: Added! I added it to `apply_network_boundaries`, I found it to be more suitable as that way it's part of the public docs.

src/distributed_planner/distributed_config.rs
gabotechs: The issue with that is that default estimator are treated specially in the sense that we always want to execute them last, as we want to give users a chance to overriding them, so they cannot be just added as normal estimators.

src/distributed_planner/distributed_config.rs
gabotechs: I think for now it's fine, I'd still like to get some input first from the community in https://github.com/apache/datafusion/pull/18168 to see if they have a better suggestion or something

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: Added! I added it to `apply_network_boundaries`, I found it to be more suitable as that way it's part of the public docs.

src/distributed_planner/distributed_config.rs
gabotechs: done!

src/distributed_planner/distributed_config.rs
jayshrivastava: 🔥

src/distributed_planner/distributed_config.rs
jayshrivastava: Ah okay that sounds reasonable.

src/distributed_planner/distributed_physical_optimizer_rule.rs
jayshrivastava: True. This is fine then!

jayshrivastava (approved):
LGTM!

NGA-TRAN:
I am actively reviewing this

PR #224, opened 2025-11-17
fix(ttl_map): initialize time to 1 for correct wrapping sub

### LINKED ISSUE #221: Fix TTL map wrapping sub
I'm pretty sure this is wrong:
https://github.com/datafusion-contrib/datafusion-distributed/blob/dd853a7dcfef1b7c6b6301dda66d965376cad906/src/common/ttl_map.rs#L241-L242

Assume there's 10 buckets. When inserting a key at `time=0`, we want the key to go into bucket 9.

However, it ends up going to bucket 5.
```
(time.wrapping_sub(1)) % self.buckets.len() as u64
= (0 - 1) % 10
= 18,446,744,073,709,551,615 % 10
= 5
```

Note that time is a u64 that we increment by 1 each time and wraps around to 0.

Prematurely evicting entries in this rare case is not a significant problem, but it's a bug nonetheless.

ruchirK:
Fixes #221. Previously, we encountered a bug at the initial gc tick on ttl_map where `time` (denoting the number of ticks since map initialization) starts at 0, but the logic to compute the time to remove the inserted value is

```
free_time = (time - 1) // (number of ticks to ttl)
```

In other words, we maintain a circular buffer of inserts, with a single slot per tick, and we will delete the item once `buffer.len()` ticks have elapsed (ttl = tick * buffer.len()).

This is all fine, except we implement the logic as:

```
free_time = time.wrapping_sub(1) / buffer.len()
```

which is computing for `u64` `time`:

```
free_time = ((time - 1) mod 2^64) mod buffer.len()
```

when we really just want `free_time = (time - 1) mod buffer.len()`.

Luckily this equality holds as long as ` 0 < time < 2^64`, as for those times `time - 1 mod 2&64 = time`. This commit changes our behavior to initialize `time` at 1 instead of 0. We don't need to worry about the overflow case because even if the tick duration was a nanosecond, it would take ~584 years to overflow 64 bits at which point we would surely have other problems besides the momentarily incorrect ttl. For the underflow case, this should primarily help with unit tests above anything else, as the bug only happened at `time = 0`.

### DIFF
--- a/src/common/ttl_map.rs
+++ b/src/common/ttl_map.rs
@@ -184,7 +184,8 @@ where
         Self {
             buckets,
             data: stage_targets,
-            time: Arc::new(AtomicU64::new(0)),
+            // Explicitly initialize `time` to 1 to avoid underflow issues with circular buffer.
+            time: Arc::new(AtomicU64::new(1)),
             gc_scheduler_task: None,
             config,
             #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/common/ttl_map.rs
```diff
@@ -184,7 +184,8 @@ where
         Self {
             buckets,
             data: stage_targets,
-            time: Arc::new(AtomicU64::new(0)),
+            // Explicitly initialize `time` to 1 to avoid underflow issues with circular buffer.
+            time: Arc::new(AtomicU64::new(1)),
```
gabotechs, line 188: 👍 nice. Any chance to reproduce this in a test that fails in `main` and succeeds in this branch?

ruchirK: Ah that's a good q let me think on that for a minute

ruchirK: Done! PTAL

src/common/ttl_map.rs
```diff
@@ -474,6 +447,34 @@ mod tests {
...
+            ttl: Duration::from_millis(70),
+            tick: Duration::from_millis(10),
+        });
+
+        ttl_map.get_or_init("test_key".to_string(), || 999);
+
+        // Advance GC 3 times, which shouldn't free the first key.
+        for _ in 0..3 {
+            TTLMap::<String, i32>::gc(ttl_map.time.clone(), &ttl_map.buckets);
+        }
```
jayshrivastava, line 464: nit: I think we can just truncate everything after this.

PR #229, opened 2025-11-18
added broadcast join operator

### LINKED ISSUE #223: Add Network Broadcast Join

Quick definition and benefit:

Impact:

Some useful links:

gene-bordegaray:
Draft for broadcast join
Closes #223

Benches are volatile, doesn't seem like drastic improvements due to tpch data

Follow up could be some more benching

### DIFF
--- a/src/stage.rs
+++ b/src/stage.rs
@@ -497,10 +497,10 @@ fn display_plan(
                     if i >= isolator_partition_group.as_ref().map_or(0, |v| v.len()) {
                         style = "[style=dotted, label=empty]";
                     }
-                } else if let Some(partition_group) = &isolator_partition_group {
-                    if !partition_group.contains(&i) {
-                        style = "[style=invis]";
-                    }
+                } else if let Some(partition_group) = &isolator_partition_group
+                    && !partition_group.contains(&i)
+                {
+                    style = "[style=invis]";
                 }
 
                 writeln!(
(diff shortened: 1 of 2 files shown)


src/stage.rs
```diff
@@ -493,10 +497,10 @@ fn display_plan(
                     if i >= isolator_partition_group.as_ref().map_or(0, |v| v.len()) {
                         style = "[style=dotted, label=empty]";
                     }
-                } else if let Some(partition_group) = &isolator_partition_group {
-                    if !partition_group.contains(&i) {
-                        style = "[style=invis]";
-                    }
+                } else if let Some(partition_group) = &isolator_partition_group
```
gene-bordegaray, line 500: clippy wouldn't leave me alone 😄

src/stage.rs
```diff
@@ -751,10 +755,10 @@ fn find_input_stages(plan: &dyn ExecutionPlan) -> Vec<&Stage> {
 
 pub(crate) fn find_all_stages(plan: &Arc<dyn ExecutionPlan>) -> Vec<&Stage> {
     let mut result = vec![];
-    if let Some(plan) = plan.as_network_boundary() {
-        if let Some(stage) = plan.input_stage() {
-            result.push(stage);
-        }
+    if let Some(plan) = plan.as_network_boundary()
```
gene-bordegaray, line 758: same here

PR #230, opened 2025-11-18
Improve TPCH gen command

gabotechs:
Some misc improvements:
- Fixes the `--partitions` argument
- Removes all unused stuff

### DIFF
--- a/benchmarks/src/tpch/convert.rs
+++ b/benchmarks/src/tpch/convert.rs
@@ -48,7 +48,7 @@ pub struct ConvertOpt {
     #[structopt(short = "c", long = "compression", default_value = "zstd")]
     compression: String,
 
-    /// Number of partitions to produce
+    /// Number of partitions to produce. By default, uses only 1 partition.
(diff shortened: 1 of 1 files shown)


benchmarks/src/tpch/convert.rs
```diff
@@ -106,9 +107,7 @@ impl ConvertOpt {
             csv = csv.select(selection)?;
             // optionally, repartition the file
             let partitions = self.partitions;
-            if partitions > 1 {
-                csv = csv.repartition(Partitioning::RoundRobinBatch(partitions))?
-            }
+            csv = csv.repartition(Partitioning::RoundRobinBatch(partitions))?;
```
ruchirK: nit: I think we should either keep the conditional repartition OR change the comment to indicate that we always repartition the file now?

gabotechs: Added a doc comment in the CLI so that it shows in the user's terminal

PR #231, opened 2025-11-19
add tpc-ds tests and property-based testing utilities

jayshrivastava:
This change introduces a new `property_based.rs` test utility which lets us evaluate
correctness using properties. These are useful for evaluating correctness when
we do not know the expected output of a test (ex. if we were to fuzz the database
with randomized data or randomzed queries, then we can only verify the output using
properties). The two oracles are
- ResultSetOracle: Compares the result set between single node and distributed datafusion
- OrderingOracle: Uses plan properties to figure out the expected ordering and asserts it

This change does not introduce a fuzz test, but it introduces a TPC-DS test. This test
randomly generates data using the duckdb CLI and runs 99 queries on a distributed cluster.
The query outputs are validated against single-node datafusion using test utils in
`metamorphic.rs`. This test also randomizes the test cluster parameters - there's no harm
in doing so.

Next steps:
- Add fuzzing
  - Now that we have property-based testing utils, we can properly fuzz the project
    using SQLancer
  - SQLancer produces INSERT and SELECT statements which we could point at a datafusion
    distributed cluster and verify against single node datafusion
  - Although it doesn't support nested select statements, 70% of the queries were valid
    datafusion queries, meaning these are good test cases for us
- Add metrics oracle to validate output_rows metric / metrics propagation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@
 /benchmarks/data/
 testdata/tpch/*
 !testdata/tpch/queries
-testdata/tpch/data/
 testdata/tpcds/data/
--- a/testdata/tpcds/generate.sh
+++ b/testdata/tpcds/generate.sh
@@ -2,6 +2,9 @@
 
 set -e
 
+# Get the directory where this script is located
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
 if [ $# -ne 1 ]; then
     echo "Usage: $0 <scale_factor>"
     echo "Scale factor must be greater than or equal to 0"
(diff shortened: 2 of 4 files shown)


src/bin/fuzz.rs
```diff
@@ -0,0 +1,228 @@
...
+        } => {
+            // Validate arguments
+            if let Err(e) = validate_tpcds_args(&scale_factor) {
+                error!("Error: {}", e);
+                process::exit(1);
+            }
+
+            info!("🌀 Starting Distributed DataFusion Fuzz Test");
+
+            let workload = TpcdsWorkload::try_new(scale_factor.clone(), force_regenerate, queries)?;
```
gabotechs: I think we might be missing concepts here:

- Fuzzing consists on autogenerating SQL statements massively in order to bombard the system with weird and complex queries in order to reveal hangs, memory leaks, deadlocks, panics, and equal behavior to single node.
- TPC-DS is a benchmarking dataset similar to TPC-H that consists on specific "non-autogenerated" queries that measures performance of any relational query execution engine. The only thing that can be "fuzzed" here are the two config params I assume, but I'm not sure how much value there is from fuzzing those.

For TPC-DS, The only think we can fuzz are a few config params, like the cardinality task scale factor or the files per task param, and I don't think fuzzing is going to uncover anything that the most extreme configuration wouldn't.

I'd say that if specifically `files_per_task=1` and `cardinality_scale_factor=1.0` do not uncover issues, no other configuration will, and brute forcing it is probably going to be just unnecessary and expensive.

src/bin/fuzz.rs
```diff
@@ -0,0 +1,228 @@
...
+};
+use log::{debug, error, info};
+use rand::Rng;
+use std::process;
+
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+#[command(name = "fuzz")]
+#[command(about = "Fuzz test distributed datafusion with various workloads")]
+struct Cli {
```
gabotechs: My bet is that we probably should not have a CLI for this. Following this pattern we'll end up with one CLI per different way of testing distributed datafusion, and I don't think that will scale very well, or be ergonomic.

I'd just integrate it as a normal test, that runs normally with `cargo test` as any other test. That way, when people contribute to this project, they can just trust `cargo test` to verify that their code is working as expected.

Note how upstream these are just normal integration tests cases:
https://github.com/apache/datafusion/tree/6ab4d216b768c9327982e59376a62a29c69ca436/datafusion/core/tests/fuzz_cases

I think following that pattern will provide a better user experience for running this tests, and will make it easier if we ever want to move this project to upstream.

jayshrivastava: Sure. I can move this to a test file which looks similar to the tpch tests.

src/bin/fuzz.rs
jayshrivastava: I agree. My intention was to add sqlancer as another workload which is a real fuzzer because it generates random SQL statements. It's definitely confusing to call TPC-DS fuzzing. I'll change this PR to run TPC-DS as integration tests, but still use random data and config options. We can use the single node oracle to verify correctness for that.

I'm still hoping to add sqlancer support later. That can be added as a different integration test.

jayshrivastava: A better name for `test-utils/fuzz.rs` is probably metamorphic.rs or property_based.rs since the file contains utils for property-based testing and/or metamorphic testing

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,52 @@
...
+
+concurrency:
+  # Run tests sequentially to avoid using too much disk space
+  group: nightly
+  cancel-in-progress: false
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  SCALE_FACTOR: "1" # use tpcds scale factor 1
```
jayshrivastava: I was running out of disk space with the default scale factor (see https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/19548125913/job/55972184966)

Since this test is slow and requires too much disk (duckdb CLI etc), we can run it as a nightly.

jayshrivastava: Any other fuzz / large scale test can be added in this file as a nightly test as well.

jayshrivastava: Actually - I'm still getting disk space issues. Will debug.

jayshrivastava: Please disregard. This was a caching issue.

.github/actions/setup/action.yml
```diff
@@ -26,3 +26,4 @@ runs:
     - uses: Swatinem/rust-cache@v2
       with:
         key: ${{ runner.os }}-${{ inputs.targets }}-rust-cache
+        version: v2
```
jayshrivastava: My CI jobs kept running out of disk space because I had some huge artifacts cached
```
/dev/root        72G   71G  1.2G  99% /
15G	./target/debug/deps
```
These are the stats from `main` for reference 
```
/dev/root        72G   66G  5.8G  93% /
9.2G	./target/debug/deps
```
I fixed the bad caching issue by bumping the version here.

jayshrivastava:
@gabotechs This is ready for another review 🙇🏽

.github/actions/setup/action.yml
jayshrivastava: See follow up: https://github.com/datafusion-contrib/datafusion-distributed/pull/234

gabotechs: nice! I'll merge that one now

.github/workflows/ci.yml
```diff
@@ -40,6 +40,38 @@ jobs:
...
+        run: cargo test --features tpcds --test tpc_ds_randomized
+        continue-on-error: true
+      - name: Upload test artifacts on failure
+        if: failure() || steps.test.outcome == 'failure'
+        uses: actions/upload-artifact@v4
+        with:
+          name: tpcds-test-artifacts-${{ github.run_id }}
+          path: testdata/tpcds/data/**
+          retention-days: 7
+          if-no-files-found: ignore
```
gabotechs: 🤔 why would we want to upload the data on failure as an artifact?

.github/workflows/ci.yml
```diff
@@ -40,6 +40,38 @@ jobs:
...
+      - name: Clean up test data
+        run: |
+          rm -rf testdata/tpcds/data/*
+          rm -f $HOME/.local/bin/duckdb
+          rm -rf /home/runner/.duckdb
+          df -h
```
gabotechs: GitHub runners are stateless, meaning that each step run is done in a newly spawned machine, so this step is unnecessary, as the machine running this step is destroyed after the run.

.github/workflows/ci.yml
```diff
@@ -40,6 +40,38 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+      - uses: ./.github/actions/setup
+      - name: Install DuckDB CLI
+        run: |
+          curl https://install.duckdb.org | sh
+          mkdir -p $HOME/.local/bin
+          mv /home/runner/.duckdb/cli/latest/duckdb $HOME/.local/bin/
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
```
gabotechs, line 56: How is it not sufficient to just do:
```sh
curl https://install.duckdb.org | sh
```

I imagine that will already install the cli in the appropriate place accessible from $PATH

Cargo.toml
```diff
@@ -34,6 +34,7 @@ object_store = "0.12.3"
 bytes = "1.10.1"
 
 # integration_tests deps
+base64 = { version = "0.22", optional = true }
```
gabotechs: Shouldn't this and `rand_chacha` be also added to `dev-dependencies`?

src/test_utils/property_based.rs
```diff
@@ -0,0 +1,566 @@
...
+    ) -> Result<Self> {
+        let oracles: Vec<Box<dyn Oracle + Send + Sync>> =
+            vec![Box::new(ResultSetOracle {}), Box::new(OrderingOracle {})];
+
+        Ok(Validator {
+            test_ctx,
+            compare_ctx,
+            oracles,
+        })
+    }
```
gabotechs: These two functions are exactly the same, and the second one is unused, so I think it can be removed

tests/tpc_ds_randomized.rs
```diff
@@ -0,0 +1,112 @@
...
+    use rand::Rng;
+
+    async fn setup() -> Result<(Validator, JoinSet<()>)> {
+        let (mut rng, seed_b64) = rng()?;
+        eprintln!("Seed: {}", seed_b64);
+
+        let num_workers = rng.gen_range(3..=8);
+        let files_per_task = rng.gen_range(2..=4);
+        let cardinality_task_count_factor = rng.gen_range(1.1..=3.0);
+
```
gabotechs: I'm still not convinced about this idea, if this has the bad luck of using `num_workers=3` and `cardinality_task_count_factor=3.0` then very little things will get distributed, making this test randomly succeed in situations where it should be failing.

I'm struggling to find value in randomizing this parameters, it seems like we are trying to brute force into finding issues that could perfectly be replicated with appropriate specific numbers, and I'm afraid the results of randomness are just going to be flaky tests instead.

gabotechs: In order words, for the same reason that we are randomizing config values in this test, we should also be randomizing config values in every other test in this project. I don't think the TPC-DS tests have anything special on top of all other tests that makes it more suitable for parameters to be randomized.

tests/tpc_ds_randomized.rs
```diff
@@ -0,0 +1,112 @@
...
+            "Test summary - Success: {} Invalid: {} Failed: {} Valid %: {:.2}%",
+            successful,
+            invalid,
+            failed,
+            if successful + invalid > 0 {
+                (successful as f64 / (successful + invalid) as f64) * 100.0
+            } else {
+                0.0
+            }
+        );
```
gabotechs: Rather than having our print statements for reporting test results, I would arrange tests here so that it's plain `cargo test` the one that performs the reporting.

This is how it's done upstream:

https://github.com/apache/datafusion/blob/7fa2a694bbc18608a46f85974e721fafa2503219/datafusion/core/tests/tpcds_planning.rs#L32-L32

And we also do the same in this project for the TPCH test.

having independent tests per query is also going to allow us to rerun just the query that fails vs running them all even if we only care about one.

Having one test per query is going to result in ~1000 LOC to make, so very good opportunity to learn about VIM macros.

src/test_utils/property_based.rs
```diff
@@ -0,0 +1,566 @@
...
+
+#[async_trait]
+impl Oracle for ResultSetOracle {
+    async fn validate(
+        &self,
+        _test_ctx: &SessionContext,
+        compare_ctx: &SessionContext,
+        query: &str,
+        test_result: &Result<Vec<RecordBatch>>,
+    ) -> Result<()> {
```
gabotechs: Do you think this can be simplified to just use a couple of functions?

Having this `trait` here is:
- forcing you to do some juggling by ignoring certain arguments in certain implementations
- running the query unnecessarily twice in single node, like in the `OrderingOracle` (run first in `Validator::run` and second in `OrderingOracle::validate`)
- having an extra struct that glues all implementations together

I think that unless we start using a fuzzing framework that require us to implement some traits, we are perfectly fine with just simple pure functions, for example something like this:

https://github.com/datafusion-contrib/datafusion-distributed/commit/be2c567f1e15f79bbbd836d7caa355b61905447a

Trims 100 lines of code of extra traits and structs and it runs the same code.

.github/workflows/ci.yml
jayshrivastava: Ah I previously thought that data gen was random, but it's deterministic based on the scale factor. Removed.

.github/workflows/ci.yml
jayshrivastava: Done

.github/workflows/ci.yml
jayshrivastava: Looking at the script, it does not add `/home/runner/.duckdb/cli/latest/duckdb` to the $PATH.

Cargo.toml
jayshrivastava: Removed both dependencies. Since we aren't randomizing anything, we don't need either. Removed the `rand` test util as well.

src/test_utils/property_based.rs
jayshrivastava: Done. Removed as a part of the refactor

src/test_utils/property_based.rs
jayshrivastava: Done. I copied your commit

tests/tpc_ds_randomized.rs
jayshrivastava: Done. I did not learn VIM macros 😅. I asked claude

tests/tpc_ds_randomized.rs
jayshrivastava: Done. I hardcoded this 
```
const NUM_WORKERS: usize = 4;
const FILES_PER_TASK: usize = 2;
const CARDINALITY_TASK_COUNT_FACTOR: f64 = 2.0;
```
let me know if different numbers make more sense.

testdata/tpcds/generate.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+set -e
+
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <scale_factor>"
+    echo "Scale factor must be greater than or equal to 0"
+    exit 1
+fi
+
```
gabotechs, line 13: One trick for ensuring this script can be run from an arbitrary directory, is to do something like this:
```shell
# https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
Like what upstream has here:
https://github.com/apache/datafusion/blob/main/benchmarks/bench.sh#L28-L30

It makes the script a bit more future proof

testdata/tpcds/README.md
```diff
@@ -0,0 +1,8 @@
+This directory contains 99 TPC-DS queries from https://github.com/duckdb/duckdb
+
+## Modifications for DataFusion Compatibility
+
+ - Queries 47 and 57 were modified to add explicit ORDER BY d_moy to avg() window function. DataFusion requires explicit ordering in window functions with PARTITION BY for deterministic results.
+ - Query 72 was modified to support date functions in datafusion
+
```
gabotechs, line 7: :+1: I also faced this in one TPCH query. What I did was to just do a string replacement in the test itself in order to keep the .sql files untouched, but what you did here is also fine

https://github.com/datafusion-contrib/datafusion-distributed/blob/67ce84c7c9752a8687da2c7bdb36718941272acf/tests/tpch_correctness_test.rs#L66-L73

tests/tpcds_test.rs
```diff
@@ -0,0 +1,587 @@
...
+                }
+            })
+            .await;
+    }
+
+    async fn run(
+        ctx: &SessionContext,
+        query_sql: &str,
+    ) -> (Arc<dyn ExecutionPlan>, Result<Vec<RecordBatch>>) {
+        let df = ctx.sql(&query_sql).await.unwrap();
```
gabotechs: nit: you should be fine with:
```suggestion
        let df = ctx.sql(query_sql).await.unwrap();
```

gabotechs (approved):
Great!

.gitignore
```diff
@@ -3,3 +3,5 @@
 /benchmarks/data/
 testdata/tpch/*
 !testdata/tpch/queries
+testdata/tpch/data/
```
gabotechs:
```suggestion
```

This can be removed, it's already ignored on line 4

tests/tpcds_test.rs
```diff
@@ -0,0 +1,587 @@
+#[cfg(all(feature = "integration", feature = "tpcds", test))]
+mod tests {
```
gabotechs, line 2: This tests are failing for me, it created a bunch of empty folders and now it wont try to regenerate them again.

Probably something went wrong along the way, but I did not get any error message or anything at all, just test errors claiming "column not found"

PR #245, opened 2025-11-28
Collect metrics optionally

gabotechs:
While investigating https://github.com/datafusion-contrib/datafusion-distributed/issues/228, I tried disabling metrics collection to see if that was somehow contributing to the deadlock.

The conclusion is that it not only does not affect the deadlock, but also does not affect performance in any measurable way.

As I had this code anyways that enables/disables metrics collection optionally, and I now that you @jayshrivastava is something you were planning to do, I though I might as well just create a separate PR for this.

---

The `collect_metrics` field in `DistributedConfig` now needs to be accessible not only on the head node, but also in any other node involved in the query, which means that we now need to send `DistributedConfig` over the wire as an option extension. While implementing that, I stepped with https://github.com/datafusion-contrib/datafusion-distributed/issues/247, which prevents us from sending a mutated config extension over the wire.

I made a workaround in this PR for addressing the issue manually with a `manually_propagate_distributed_config()`, but once the issue gets fixed, manual propagation is no longer needed.


src/distributed_planner/distributed_config.rs
```diff
@@ -31,6 +31,9 @@ extensions_options! {
         /// batches over the wire.
         /// If set to 0, batch coalescing is disabled on network shuffle operations.
         pub shuffle_batch_size: usize, default = 8192
+        /// Propagate collected metrics from all nodes in the plan across network boundaries
+        /// so that they can be reconstructed on the head node of the plan.
+        pub collect_metrics: bool, default = false
```
jayshrivastava: nit: defaulting to `true` might be better since that's how vanilla df works.

src/execution_plans/network_coalesce.rs
```diff
@@ -273,8 +278,10 @@ impl ExecutionPlan for NetworkCoalesceExec {
         let target_task = partition / partitions_per_task;
         let target_partition = partition % partitions_per_task;
 
+        // TODO: this propagation should be automatic <link to issue>
```
jayshrivastava: I think this is missing a link. Do you want to create an issue in this repo as well?

src/execution_plans/network_coalesce.rs
```diff
@@ -311,13 +318,14 @@ impl ExecutionPlan for NetworkCoalesceExec {
                 .into_inner()
                 .map_err(|err| FlightError::Tonic(Box::new(err)));
 
-            let metrics_collecting_stream =
-                MetricsCollectingStream::new(stream, metrics_collection_capture);
+            let stream = if retrieve_metrics {
+                MetricsCollectingStream::new(stream, metrics_collection_capture).left_stream()
+            } else {
+                stream.right_stream()
```
jayshrivastava, line 325: Do you mind explaining this / leaving a comment? I pulled your branch and I honestly cannot tell where the `Either` type comes from.

jayshrivastava (approved):

src/execution_plans/network_coalesce.rs
gabotechs: https://github.com/datafusion-contrib/datafusion-distributed/issues/247. It's this one! just forgot to update it

src/execution_plans/network_coalesce.rs
gabotechs: This comes from `futures_util::stream::stream::StreamExt`. Whenever you are returning two different streams on a conditional statement, Rust by default will fail to compile because you are trying to assign two different types to the same variable.

The `Either` stream wraps the two streams, so that as long as both are compatible in terms of what they yield, you can assign either one or the other to the same variable.

The other solution is to wrap them in `BoxStream`, but AFAIK that would perform some extra heap allocations, so this might be like 0.0001 nanoseconds faster.

PR #261, opened 2025-12-19
Add Distributed DataFusion cli

gabotechs:
Adds a CLI similar to https://datafusion.apache.org/user-guide/cli/ but with distributed capabilities.

The CLI uses in-memory workers, so the "distribution" aspect is more like a simulation rather than real distribution.


NGA-TRAN (approved):
❤️  this!

PR #262, opened 2025-12-19
Distribute UNION operations

gabotechs:
This PR builds on top of https://github.com/datafusion-contrib/datafusion-distributed/pull/259 and uses its tools for distributing UNION operations.

The way unions are distributed is with a new `ChildrenIsolatorUnionExec` operator, that will selectively execute its children depending on the distributed task index in which its running.

The best way of imagining how this works is by looking at the simplest case, when the union operation runs in the same amount of tasks than children it has. In that case, each tasks executes one child respectively:

```
┌─────────────────────────────┐┌─────────────────────────────┐┌─────────────────────────────┐
│           Task 1            ││           Task 2            ││           Task 3            │
│┌───────────────────────────┐││┌───────────────────────────┐││┌───────────────────────────┐│
││ ChildrenIsolatorUnionExec ││││ ChildrenIsolatorUnionExec ││││ ChildrenIsolatorUnionExec ││
│└───▲─────────▲─────────▲───┘││└───▲─────────▲─────────▲───┘││└───▲─────────▲─────────▲───┘│
│    │                        ││              │              ││                        │    │
│┌───┴───┐ ┌  ─│ ─   ┌  ─│ ─  ││┌  ─│ ─   ┌───┴───┐ ┌  ─│ ─  ││┌  ─│ ─   ┌  ─│ ─   ┌───┴───┐│
││Child 1│  Child 2│  Child 3│││ Child 1│ │Child 2│  Child 3│││ Child 1│  Child 2│ │Child 3││
│└───────┘ └  ─  ─   └  ─  ─  ││└  ─  ─   └───────┘ └  ─  ─  ││└  ─  ─   └  ─  ─   └───────┘│
└─────────────────────────────┘└─────────────────────────────┘└─────────────────────────────┘
```

It can get complicated if there are either more tasks than children, ore more children than tasks.

For example if there are more children than tasks, their execution will be spread as evenly as possible across distributed tasks:

```
┌─────────────────────────────┐┌─────────────────────────────┐
│           Task 1            ││           Task 2            │
│┌───────────────────────────┐││┌───────────────────────────┐│
││ ChildrenIsolatorUnionExec ││││ ChildrenIsolatorUnionExec ││
│└───▲─────────▲─────────▲───┘││└───▲─────────▲─────────▲───┘│
│    │         │              ││                        │    │
[...]

PR #264, opened 2025-12-22
Make queries in TPC-DS get distributed 

gabotechs:
While playing around with this tests, I realized that in current `main` they are not getting distributed, so we are just comparing single-node VS single-node.

In this PR:
- Rather than generating with `duckdb` CLI, we follow upstream approach and we download a fixed TPC-DS dataset from https://github.com/apache/datafusion-benchmarks/archive/refs/heads/main.zip
- The downloaded parquet files are all partitioned in 4 smaller files so that distribution kicks in in almost all of them
- The tests for the few queries that cannot get distributed are ignored
- Adds some tests for asserting plans, useful for qualifying https://github.com/datafusion-contrib/datafusion-distributed/pull/262 changes

The main point about making this PR is so that we can test that https://github.com/datafusion-contrib/datafusion-distributed/pull/262 works correctly with the UNION operations present in TPC-DS (TPC-H queries do not have unions)

### DIFF
--- a/tests/tpcds_correctness_test.rs
+++ b/tests/tpcds_correctness_test.rs
@@ -24,364 +24,364 @@ mod tests {
     const SF: f64 = 1.0;
     const PARQUET_PARTITIONS: usize = 4;
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_1() -> Result<()> {
         test_tpcds_query(1).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_2() -> Result<()> {
         test_tpcds_query(2).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_3() -> Result<()> {
         test_tpcds_query(3).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_4() -> Result<()> {
         test_tpcds_query(4).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_5() -> Result<()> {
         test_tpcds_query(5).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_6() -> Result<()> {
         test_tpcds_query(6).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_7() -> Result<()> {
         test_tpcds_query(7).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_8() -> Result<()> {
         test_tpcds_query(8).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     #[ignore = "expected no error but got: Arrow error: Invalid argument error: must either specify a row count or at least one column"]
     async fn test_tpcds_9() -> Result<()> {
         test_tpcds_query(9).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_10() -> Result<()> {
         test_tpcds_query(10).await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
     async fn test_tpcds_11() -> Result<()> {
         test_tpcds_query(11).await
     }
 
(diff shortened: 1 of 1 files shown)


src/test_utils/property_based.rs
```diff
@@ -12,7 +12,7 @@ use datafusion::{
 use std::sync::Arc;
 
 /// compares the set of record batches for equality
-pub async fn compare_result_set(
+pub fn compare_result_set(
```
gabotechs, line 15: This functions were marked async, but they were not async, so I removed the `async` keyword so that people do not mistakenly `.await` them blocking the tokio runtime.

src/test_utils/tpcds.rs
```diff
@@ -251,53 +228,40 @@ pub async fn register_tpcds_table_with_options(
...
     }
 
     let output = Command::new("bash")
-        .arg(&generate_script)
-        .arg(scale_factor)
-        .current_dir(&tpcds_dir)
+        .arg(&gen_script)
+        .env("SCALE_FACTOR", sf.to_string())
+        .env("FILE_SIZE_BYTES", max_file_size.to_string())
+        .env("OUT_DIR", data_dir)
```
gabotechs: Made the `generate.sh` configurable, as now there are different configurations depending on if we are asserting plans or correctness

tests/tpcds_correctness_test.rs
```diff
@@ -0,0 +1,648 @@
...
+            tokio::task::spawn_blocking(move || async move {
+                compare_result_set(&d_results, &s_results)
+            })
+        };
+        let compare_ordering = {
+            let d_results = d_results.clone();
+            tokio::task::spawn_blocking(move || async move {
+                compare_ordering(d_plan, s_plan, &d_results)
+            })
+        };
```
gabotechs, line 594: GitHub thinks this is a completely new file, but it's instead the old `tpcds_test.rs` adapted and renamed.

Previously, this code looked like this:
```rust
        let compare_result = tokio::try_join!(
            compare_result_set(&distributed_results, &single_node_results),
            compare_ordering(
                distributed_physical_plan,
                single_node_physical_plan,
                &distributed_results
            ),
        );
```
Where `compare_result_set` and `compare_ordering` where marked as async, and were spawned as normal async tasks, but they were indeed computationally heavy operations that were blocking the `tokio` runtime and were triggering errors while communicating with localhost workers.

I moved them to normal functions that get spawned in tokio blocking pools instead.

src/test_utils/tpcds.rs
gabotechs: No longer relevant

src/test_utils/tpcds.rs
```diff
@@ -46,259 +35,222 @@ pub fn get_test_tpcds_query(id: usize) -> Result<String> {
...
 
-            let mut dict_batches = Vec::new();
-            for batch in batches {
-                dict_batches.push(force_dictionary_encoding_for_table(table_name, batch)?);
-            }
+        if let Some(parent) = outpath.parent() {
+            fs::create_dir_all(parent)?;
+        }
+        let mut outfile = fs::File::create(&outpath)?;
+        std::io::copy(&mut file, &mut outfile)?;
```
jayshrivastava: I wonder if `mv` would be faster.

src/test_utils/tpcds.rs
```diff
@@ -46,259 +35,222 @@ pub fn get_test_tpcds_query(id: usize) -> Result<String> {
...
+    }
 
-            // Write dictionary-encoded data to disk
-            if !dict_batches.is_empty() {
-                fs::create_dir_all(&dict_table_path)?;
-                let dict_file_path = dict_table_path.join("data.parquet");
-                let file = fs::File::create(&dict_file_path)?;
-                let props = WriterProperties::builder().build();
-                let mut writer = ArrowWriter::try_new(file, dict_batches[0].schema(), Some(props))?;
+    println!("Extracted to {}", extract_to.display());
```
jayshrivastava: nit: printing to stdout in util library function seems bad

tests/tpcds_correctness_test.rs
jayshrivastava: Could use a comment here then

jayshrivastava (commented):
LGTM. Seems reasonable :)

jayshrivastava (approved):

src/test_utils/tpcds.rs
gabotechs: It probably would, but this is not a filesystem operation were we are relocating files, we are moving a compressed archive entry to an actual file by copying the bytes. 🤔 I don't think there's any system call that can do that for us.

I'm going to choose better names for the variables so that this is clearer.

src/test_utils/tpcds.rs
gabotechs: indeed, removing it

PR #269, opened 2025-12-24
Refactor benchmarks crate and add TPC-DS benchmarks

gabotechs:
There are two main chunks shipped in this PR:
- A refactor to the benchmarks, that remove stuff that we were not using and stuff that we inherited from upstream that is not relevant here. The refactor is mostly about removing things, rather than refactoring current ones.
- Adapt the benchmarks to also hold TPC-DS datasets and queries. That way we can better qualify changes in https://github.com/datafusion-contrib/datafusion-distributed/pull/262 and https://github.com/datafusion-contrib/datafusion-distributed/pull/229


benchmarks/README.md
```diff
@@ -1,81 +1,36 @@
 # Distributed DataFusion Benchmarks
 
-### Generating TPCH data
+### Generating Benchmarking data
 
 Generate TPCH data into the `data/` dir
 
 ```shell
 ./gen-tpch.sh
+./gen-tpcds.sh
```
gene-bordegaray, line 9: I think this is "get-tpcds.sh"

gabotechs: 🤔 it seems like I made a typo with the file name, this was intended to be `gen-tpcds.sh`

PR #271, opened 2025-12-26
Apply TaskEstimator on all nodes

gabotechs:
Allows user-provided `TaskEstimator`s to be applied to all nodes, not only leaf nodes.

This allows nodes other than leaf nodes to provided task count estimations, for example:

```
┌────────────────────────────────────────┐       
│       ChildrenIsolatorUnionExec        │       
└───▲───────────────▲────────────────▲─Maximum(3)
    │               │                │           
┌───┴───┐       ┌───┴───┐        ┌───┴───┐       
│Child 1│       │Child 2│        │Child 3│       
└───▲─Maximum(1)└───▲─Maximum(1) └───▲─Maximum(1)
    │               │                │           
┌───┴───┐       ┌───┴───┐        ┌───┴───┐       
│Child 1│       │Child 2│        │Child 3│       
└────Desired(2) └─────Desired(5) └────Desired(3) 
```

In this example, leaf nodes are "desiring" 2, 5 and 3 tasks, so `ChildrenIsolatorUnionExec` might want to use 10 tasks, but because an intermediate estimation of `Maximum(1)` is being provided, the `ChildrenIsolatorUnionExec` will only use 3 tasks (1+1+1, and not 2+5+3)


src/distributed_planner/task_estimator.rs
```diff
@@ -112,40 +150,20 @@ impl TaskEstimator for usize {
...
-    }
-
-    fn scale_up_leaf_node(
-        &self,
-        plan: &Arc<dyn ExecutionPlan>,
-        task_count: usize,
-        cfg: &ConfigOptions,
-    ) -> Option<Arc<dyn ExecutionPlan>> {
-        self.as_ref().scale_up_leaf_node(plan, task_count, cfg)
+    delegate! {
```
ahmed-mez, line 162: nice use of delegate 💯

PR #273, opened 2025-12-30
Add TLS feature in default ChannelResolver

gabotechs:
Adds a new `tls` feature for enriching the default `ChannelResolver` implementation with TLS features.

This is typically not needed, as communication between workers usually happens within a trusted network, and it's not worth paying the price of encryption/decryption, but adding the option to also account for the most strict setups.


src/networking/channel_resolver.rs
```diff
@@ -253,6 +290,39 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    #[cfg(feature = "tls")]
+    async fn can_establish_connection_https() -> Result<(), Box<dyn Error>> {
+        // Using Grafana's public gRPC demo server for testing HTTPS with valid certificates.
+        // This verifies that our TLS configuration works with real-world endpoints using
+        // system root certificates.
+        const URL: &str = "https://grpc-quickpizza.grafana.com:443";
```
NGA-TRAN, line 299: nice

NGA-TRAN (approved):
LGTM

gabotechs:
Not doing it for now

PR #275, opened 2025-12-30
Improve default task estimator

gabotechs:
Improves the default task estimator by selecting the task count based on the number of file partitions rather than the number of actual files themselves


LiaCastaneda (approved):
This makes sense to me. Just out of curiosity, do you know what's DataFusion thereshold to split a physical file into multiple `PartitionedFiles`?

PR #259, opened 2025-12-18
Rework distributed planning logic

gabotechs:
This is a full rewrite of the distributed planner that aims for two things:
- Solving some bugs/improvement points around how tasks are assigned to each stage and how leaf nodes are scaled up for accounting for greater parallelism.
- Simplifying the code and make it more reliable by rendering invalid states irrepresentable.

---
## How it worked before

This is how the distributed planner worked before:
1. estimates distributed task counts for leaf nodes based on `TaskEstimator`s, scaling them up in parallelism
2. places "Pending" network boundaries in the appropriate places
3. transitions those "Pending" network boundaries to "Ready" assigning them stages
4. rollback unnecessary network boundaries, or re-assign them a different number of tasks

The order in which things happened is problematic, as it can lead to several non-desirable situations:

- A `TaskEstimator` might decide that a leaf node needs to run in 100 tasks, and it will scale it up accordingly. Then for whatever reason, it's decided that a stage containing that leaf node decides to just use 1 task (e.g., it contains a CollectLeft HashJoin). Then we end up with a x100 scaled leaf node running in just 1 task.
- A `TaskEstimator` might decide that a leaf node needs to run in 100 tasks, and the plan has 3 stages, so the planner decides to use something like 100 tasks -> 90 tasks -> 80 tasks. Then the "Pending" to "Ready" transition happens, and it's decided that the middle stage is only going to run in 1 task, so we are left with 100 tasks -> 1 task -> 80 tasks. The middle stage is going to heavily bottleneck the query, so there's no point in using 80 tasks for the head stage.
- Placing network boundaries before knowing the task count of each nodes forces us to represent a network boundary as an enum with two states: "Pending" and "Ready". This means that in a lot of places we need to do runtime checks for the nodes being in the correct state, leading to potential bugs that only happen at runtime.

## How it works now

The distribution step is simplified, and just two things happen:

1. the plan is annotated with the number of tasks that should be assigned for each node, and if a network boundary should be placed or not. At this point, the plan is left untouched, it's just annotated.
```text
 ┌────────────────────┐ task_count: Maximum(1) (because we try to coalesce all partitions into 1)
 │ CoalescePartitions │ network_boundary: Some(Coalesce)
 └──────────▲─────────┘
            │
 ┌──────────┴─────────┐ task_count: Desired(3) (inherited from the child)
 │     Projection     │ network_boundary: None
 └──────────▲─────────┘
            │
[...]

PR #279, opened 2026-01-03
Gene.bordegaray/2025/12/add broadcast exec

### LINKED ISSUE #223: Add Network Broadcast Join

gene-bordegaray:
Closes #223 

last one got closed and I can't reopen

Added a separate exexution plan node `BroadcastExec` to handle the partition caching logic that previously lived in the flight service.

Still get cool speed-ups 😄 

Screenshot 2026-01-02 at 7 53 39 PM


src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -468,28 +538,31 @@ mod tests {
...
+        │           [Stage 4] => NetworkShuffleExec: output_partitions=4, input_tasks=3
         └──────────────────────────────────────────────────
-          ┌───── Stage 2 ── Tasks: t0:[p0..p3] t1:[p0..p3] 
-          │ ProjectionExec: expr=[avg(weather.MinTemp)@1 as MinTemp, RainTomorrow@0 as RainTomorrow]
-          │   AggregateExec: mode=FinalPartitioned, gby=[RainTomorrow@0 as RainTomorrow], aggr=[avg(weather.MinTemp)]
-          │     [Stage 1] => NetworkShuffleExec: output_partitions=4, input_tasks=3
+          ┌───── Stage 3 ── Tasks: t0:[p0] 
+          │ CoalescePartitionsExec
+          │   [Stage 2] => NetworkCoalesceExec: output_partitions=8, input_tasks=2
           └──────────────────────────────────────────────────
```
gabotechs: There is something wrong here. Both the `DistributedExec` stage and `Stage 3` have 1 task, so they should be collapsed into a single stage, but for some reason they are not.

Normally, this piece of code in the planner will kick in and prevent this from happening:

https://github.com/datafusion-contrib/datafusion-distributed/blob/a42d4a2fc6184e3b90d62a9bbff1c62311a85657/src/distributed_planner/distributed_physical_optimizer_rule.rs#L116-L121

src/distributed_planner/plan_annotator.rs
```diff
@@ -410,11 +410,11 @@ mod tests {
...
-        CoalesceBatchesExec: task_count=Maximum(1)
-          HashJoinExec: task_count=Maximum(1)
-            CoalescePartitionsExec: task_count=Maximum(1)
-              DataSourceExec: task_count=Maximum(1)
-            DataSourceExec: task_count=Maximum(1)
+        CoalesceBatchesExec: task_count=Desired(3)
+          HashJoinExec: task_count=Desired(3), required_network_boundary=Broadcast
+            CoalescePartitionsExec: task_count=Desired(3)
+              DataSourceExec: task_count=Desired(3)
+            DataSourceExec: task_count=Desired(3)
```
gabotechs: It would be cool to add a test that demonstrates how having a node that reduces cardinality below the HashJoinExec prompts the task annotator to choose less tasks for the stage containing the join.

Maybe something like doing a JOIN between the results of two aggregations?

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,185 @@
...
+/// │  NetworkBroadcastExec  │      │  NetworkBroadcastExec  │           ...          │  NetworkBroadcastExec  │      │
+/// │        (task 1)        │      │        (task 2)        │                        │        (task M)        │  Stage N+1
+/// └┬─┬─────┬─┬─────────────┘      └─────┬───┬─────┬──┬─────┘                        └─────┬────────┬─────┬──┬┘      │
+///  │1│     │N│                          │N+1│     │2N│                                    │N+1(M-1)│     │NM│       │
+///  └▲┘ ... └▲┘                          └─▲─┘ ... └─▲┘                                    └────▲───┘ ... └─▲┘       ■
+///   │       │     Populates               │         │                                          │           │
+///   │       └───Cache Index N───┐    Cache Hit   Cache Hit                      ┌─Cache Hit────┘           │
+///   │                           │     Index 1     Index N                       │                          │
+///   └────Populates ─────┐       │         │         │                           │           ┌───Cache Hit─-┘
+///      Cache Index 1    │       │         │         │                           │           │
```
gabotechs: Shouldn't this be Index 0?

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,185 @@
...
+/// once by the framework.
+///
+/// Broadcast is used in a 1 to many context, like this:
+/// ```text
+/// ┌────────────────────────┐      ┌────────────────────────┐                        ┌────────────────────────┐      ■
+/// │  NetworkBroadcastExec  │      │  NetworkBroadcastExec  │           ...          │  NetworkBroadcastExec  │      │
+/// │        (task 1)        │      │        (task 2)        │                        │        (task M)        │  Stage N+1
+/// └┬─┬─────┬─┬─────────────┘      └─────┬───┬─────┬──┬─────┘                        └─────┬────────┬─────┬──┬┘      │
+///  │1│     │N│                          │N+1│     │2N│                                    │N+1(M-1)│     │NM│       │
+///  └▲┘ ... └▲┘                          └─▲─┘ ... └─▲┘                                    └────▲───┘ ... └─▲┘       ■
```
gabotechs: Awesome doc! without even reading the text, I can just tell what's happening under the hood. Thanks for taking the time in providing such quality doc comments.

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,185 @@
...
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        if children.len() != 1 {
+            return Err(datafusion::error::DataFusionError::Internal(
+                "BroadcastExec requires exactly one child".to_string(),
+            ));
+        }
+        Ok(Arc::new(Self::new(
+            children.into_iter().next().unwrap(),
+            self.consumer_task_count,
+        )))
```
gabotechs: Luckily, we have a helper for this:

```suggestion
        Ok(Arc::new(Self::new(
            require_one_child(children)?,
            self.consumer_task_count,
        )))
```

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,185 @@
...
+                        futures::TryStreamExt::try_collect(stream).await?;
+                    Ok::<_, datafusion::error::DataFusionError>(Arc::new(batches))
+                })
+                .await?;
+            let batches = Arc::clone(batches);
+            let batches_vec: Vec<RecordBatch> = batches.iter().cloned().collect();
+            Ok::<_, datafusion::error::DataFusionError>(stream::iter(
+                batches_vec.into_iter().map(Ok),
+            ))
+        });
```
gabotechs, line 176: Ideally, instead of collecting all the batches together and then emitting them, we could broadcast them in a streaming fashion. This will allow to early send produced batches over the wire while new ones are being produced, which has good chances of speeding up execution.

I imagine for that to work we would need to use a `tokio::sync::broadcast` channel.

src/execution_plans/network_broadcast.rs
```diff
@@ -0,0 +1,314 @@
...
+///
+/// ```text
+/// ┌────────────────────────┐                        ┌────────────────────────┐           ■
+/// │  NetworkBroadcastExec  │                        │  NetworkBroadcastExec  │           │
+/// │        (task 1)        │           ...          │        (task M)        │           │
+/// │                        │                        │                        │        Stage N
+/// │    Populates Caches    │                        │       Cache Hits       │           │
+/// └────────┬─┬┬─┬┬─┬───────┘                        └────────┬─┬┬─┬┬─┬───────┘           │
+///          │1││2││3│                                         │1││2││3│                   │
+///          └▲┘└▲┘└▲┘                                         └▲┘└▲┘└▲┘                   ■
```
gabotechs, line 40: This docs are pure gold! my only comment is about the 1-based indexing, should it just be 0-based? Unless I missed something the code indexes partitions in a 0-based manner.

gabotechs (commented):
Flushing the first round of comments. This looks really good! pretty clean design.

For now, I don't think I have any major comment, the path forward looks like the correct one.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -88,21 +88,58 @@ fn distribute_plan(
...
     }
 
-    let parent_task_count = annotated_plan.task_count.as_usize();
-    let max_child_task_count = children.iter().map(|v| v.task_count.as_usize()).max();
+    // Broadcast requires different task counts for build vs probe.
+    if annotated_plan.required_network_boundary == Some(RequiredNetworkBoundary::Broadcast) {
+        let mut build = children.remove(0);
+        let mut probe = children.remove(0);
+
+        set_task_count_until_boundary(&mut probe, parent_task_count);
```
gabotechs: Any chance of handling this as a normal `if nb_req == RequiredNetworkBoundary::* { ... }` statement at the end of this function like the other network boundaries?

Also, one improvement that comes to mind for consistency with the other network boundaries, is to wrap all the `NetworkBroadcastExec` instantiation logic inside `NetworkBroadcastExec::try_new` instead of here.

For example, note how in `NetworkShuffleExec::try_new` we handle the logic of transforming the child nodes in order to adapt them to shuffles, rather than doing it in this function:

https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/execution_plans/network_shuffle.rs#L151-L151

Do you think we could reach a point were instantiation one `NetworkBroadcastExec` is no different than instantiating any of the other network boundaries?

src/execution_plans/broadcast.rs
gene-bordegaray: yes, my typo. Will fix, thank you

src/execution_plans/broadcast.rs
gene-bordegaray: Awesome, ya I really enjoy the documentation / explanation side of things, which maybe I am crazy for. I am glad you enjoy them 😄 

PS: you moght already have it, but check out Moondraw @NGA-TRAN introduced it to me and its pretty sweet

src/distributed_planner/plan_annotator.rs
gene-bordegaray: I am a bit confused with this and not understanding. Could you explain a bit more?

src/distributed_planner/distributed_physical_optimizer_rule.rs
gene-bordegaray: totally missed this, thanks

src/distributed_planner/distributed_physical_optimizer_rule.rs
gene-bordegaray: refactored this best I could, let me know if you have more ideas. Some logic still lives here as unlike other operators Broadcast relies on the annotated plan.

src/distributed_planner/distributed_physical_optimizer_rule.rs
gene-bordegaray: After reviewing I found where we are skipping this but opted into not having this check for broadcasts. I added a comment above the check explaining why. I plan to do some follow-up work with this, trying to not cram too much in this PR. Let me know your thoughts 😄

src/execution_plans/broadcast.rs
gene-bordegaray: This makes sense, and I see the upsides, especially for larger datasets.

I did some research on `tokio::sync::broadcast` and think the implementation will be non-trivial (you may have some insight that I lack and feel free to add here). I added a comment noting this with follow-up work.

Hoping to address this and the single stage collapse in follow up work to get this in and avoid more merge conflicts. Of course, unless you think it is necessary for these changes to be included in this PR 😄

src/distributed_planner/distributed_physical_optimizer_rule.rs
gene-bordegaray: I could extract logic into a helper, but seems like unneeded indirection since logic only happens once

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -80,78 +80,157 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+        }
+        // If this is a CollectLeft HashJoinExec, it means that the single-node planner noticed the
+        // build side table is small enough to use a hash-table. To reduce network calls,
+        // distribute the smaller table to each consuming tasks local memory to run the join on
+        // multiple workers.
+        Some(RequiredNetworkBoundary::Broadcast) => {
+            let mut build = children.remove(0);
+            let mut probe = children.remove(0);
+
+            set_task_count_until_boundary(&mut probe, parent_task_count);
```
gabotechs: This should not be happening here, the whole point of separating the planner in an "annotation" + "distribute" steps is so that all task count assignations happen in the "annotation" step so that the "distribute" step can safely assume that task counts are set in stone. This is mainly for:
- organization and separation of concerns
- avoiding bugs where one node is scaled up assuming N task counts, but suddenly a later step overwrites that with M tasks.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -80,78 +80,157 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
+
+            // If there's only one consumer task, use Coalesce instead of Broadcast.
+            let build_child: Arc<dyn ExecutionPlan> = if parent_task_count == 1 {
+                set_task_count_until_boundary(&mut build, 1);
+                let build_side = distribute_plan(build, cfg, query_id, stage_id)?;
+                Arc::new(NetworkCoalesceExec::try_new(
+                    build_side, query_id, *stage_id, 1, 1,
+                )?)
```
gabotechs: Why do we need to special case this here instead of just letting the `BroadcastExec` to naturally broadcast to only 1 consumer task? If `parent_task_count == 1` I think we should be able to just let be without special cases.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -464,39 +555,42 @@ mod tests {
...
         .await;
         assert_snapshot!(plan, @r"
-        ┌───── DistributedExec ── Tasks: t0:[p0] 
+        ┌───── DistributedExec ── Tasks: t0:[p0]
         │ CoalescePartitionsExec
         │   CoalesceBatchesExec: target_batch_size=8192
         │     HashJoinExec: mode=CollectLeft, join_type=Left, on=[(RainTomorrow@1, RainTomorrow@1)], projection=[MinTemp@0, MaxTemp@2]
-        │       CoalescePartitionsExec
-        │         [Stage 2] => NetworkCoalesceExec: output_partitions=8, input_tasks=2
+        │       [Stage 3] => NetworkCoalesceExec: output_partitions=1, input_tasks=1
```
gabotechs: This still looks wrong. The `DistributedExec` stage and `Stage 3` should be collapsing into 1 without any repercussion in correctness. We should be able to accomplish that in this PR.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -80,78 +80,157 @@ impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
...
-        return annotated_plan.plan.with_new_children(new_children);
-    };
-
-    // It would need a network boundary, but on both sides of the boundary there is just 1 task,
-    // so we are fine with not introducing any network boundary.
-    if parent_task_count == 1 && max_child_task_count == Some(1) {
+    // Skip network boundary if both parent and all children have 1 task.
+    // Broadcast is excluded because it's used for CollectLeft join build sides that need explicit
+    // materialization. Without the stage boundary nested CollectLeft joins cause re-execution
+    // (seen in TPC-DS Q99: 29x slow-down) since CoalescePartitionsExec doesn't cache.
```
gabotechs: 🤔 I don't fully understand this comment. If the communication is from 1 producer task to 1 consumer task, no caching mechanism in the broadcast step should come into play.

Serving data from cache in `BroadcastExec` should only be relevant if consumer tasks > 1, otherwise `BroadcastExec` should be acting as pretty much just a pure passthrough node.

src/distributed_planner/plan_annotator.rs
gabotechs: When reaching a network boundary, we are delimiting the end of one stage and the beginning of another one.
This is the moment when we:
1. make sure all nodes in the stage below we just formed agree on a specific task count
2. we chose a starting point for the task count of the new stage we just started forming

For choosing this starting point, we take the final decided task count for the stage below, and multiply it by a scale factor that depends on how much the cardinality of the data was reduced in the stage below (e.g., aggregation/filtering nodes).

As we have a new type of network boundary that delimits stages, it would be nice to add a tests that verifies that this mechanism still applies when crossing those new boundaries.

The docs in https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/distributed_planner/plan_annotator.rs#L67-L67 explains a bit better how this works.

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: Overall, I find it suspicious that this PR needed so little changes to the `plan_annotator.rs` file.

src/execution_plans/broadcast.rs
gabotechs: I think using a `tokio::sync::broadcast` should not be too bad, it should not be much more code than this. There might even be other options using other types of queues that are even simpler.

Overall, if we can avoid shipping intermediate states that will need to be deleted in follow up PRs, that would be ideal, although I do acknowledge that PR conflicts are a real issue that specially you have suffered in this PR.

I would expect this change to be so simple that it doesn't matter much to do it either now or in a follow up, so if you think doing it in a follow up will make your life easier then lets do it, but I would create a GitHub issue for tracking it and leave a TODO comment with a brief explanation and a link to the issue.

src/execution_plans/network_broadcast.rs
```diff
@@ -0,0 +1,338 @@
...
+        use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
+
+        let broadcast_exec = Arc::new(super::BroadcastExec::new(input, consumer_task_count));
+        let network_broadcast = Arc::new(Self::new_inner(
+            broadcast_exec,
+            query_id,
+            stage_num,
+            input_task_count,
+        )?);
+        Ok(Arc::new(CoalescePartitionsExec::new(network_broadcast)))
```
gabotechs: 🤔 I find this constructor a bit weird. It's not returning a `Self` like all the other constructor methods for the other execution plans in this project.

src/distributed_planner/plan_annotator.rs
```diff
@@ -445,8 +445,8 @@ mod tests {
         "#;
         let annotated = sql_to_annotated(query).await;
         assert_snapshot!(annotated, @r"
-        CoalesceBatchesExec: task_count=Maximum(1)
-          HashJoinExec: task_count=Maximum(1)
+        CoalesceBatchesExec: task_count=Desired(1)
+          HashJoinExec: task_count=Desired(1), required_network_boundary=Broadcast
             CoalescePartitionsExec: task_count=Maximum(1), required_network_boundary=Coalesce
```
gabotechs: This looks wrong, we should not need two consecutive different network boundaries here.

gabotechs: 🤔 This might actually be the reason you needed so much special casing and changes in the `distribute_plan` function.

src/distributed_planner/plan_annotator.rs
```diff
@@ -476,11 +476,11 @@ mod tests {
...
         let annotated = sql_to_annotated(query).await;
         assert_snapshot!(annotated, @r"
-        CoalesceBatchesExec: task_count=Maximum(1)
-          HashJoinExec: task_count=Maximum(1)
-            CoalescePartitionsExec: task_count=Maximum(1)
-              DataSourceExec: task_count=Maximum(1)
-            DataSourceExec: task_count=Maximum(1)
+        CoalesceBatchesExec: task_count=Desired(3)
+          HashJoinExec: task_count=Desired(3), required_network_boundary=Broadcast
+            CoalescePartitionsExec: task_count=Desired(3)
```
gabotechs: 🤔 I don't think this is right, it should look more like:
```suggestion
          HashJoinExec: task_count=Desired(3)
            CoalescePartitionsExec: task_count=Desired(3), required_network_boundary=Broadcast
```

The `CoalescePartitionsExec` is still needed right below the HashJoinExec, and the network boundary should only affect the build side, not the HashJoinExec as a whole.

gabotechs (commented):
I think a lot of my suggestions arise from the fact that further changes are needed to the `plan_annotator.rs` file. 

Most of the things I find "weird" in this PR come from the fact that annotations should look differently. The concept of "annotating" a plan with the task count is fairly new, and was introduced very recently in https://github.com/datafusion-contrib/datafusion-distributed/pull/259, so I might be the one to blame for not announcing/documenting it properly. I think if you get familiar with it most of the changes I'm suggesting should feel very natural.

BTW, this is awesome work, the reason why I'm leaving so many comments is because this piece is so fundamental to the project that IMO should be as close to perfection as possible, but this is definitely one of the best (if not the best) contribution this project has had so far.

src/execution_plans/broadcast.rs
gabotechs: This project has been using it from the beginning, all the other charts we have in the docs are built with it, pretty nice tool!

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: I would not worry to much about the size, it's still pretty reasonable, and it would be nice to ship something as polished as possible. I don't want you to deal with more conflicts though, so upon reaching another conflict scenario lets just give priority to this PR.

Moving this discussion to https://github.com/datafusion-contrib/datafusion-distributed/pull/279#discussion_r2659568596

src/execution_plans/network_broadcast.rs
gabotechs: I think this might be one of the artifacts or network boundary annotations in `plan_annotator.rs` not being placed correctly.

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,187 @@
...
+///                                               │                   │                                               ■
+///                                              ┌┴┐       ...       ┌┴┐                                              │
+///                                              │1│                 │N│                                          Stage N-1
+///                                             ┌┴─┴─────────────────┴─┴┐                                             │
+///                                             │Arc<dyn ExecutionPlan> │                                             │
+///                                             └───────────────────────┘                                             ■
+/// ```
+///
+/// Notice that the first consumer task, [NetworkBroadcastExec] task 1, triggers the execution of
+/// the operator below the [BroadCastExec] and populates each cache index with the repective
```
NGA-TRAN:
```suggestion
/// the operator below the [BroadCastExec] and populates each cache index with the respective
```

src/execution_plans/broadcast.rs
```diff
@@ -0,0 +1,187 @@
...
+/// partition. Subsequent consumer tasks, rather than executing the same partitions, read the
+/// data from the cache for each partition.
```
NGA-TRAN, line 66: Questions regarding `Stage N-1`, `Stage N` and `Stage N+1` in the diagram:
  - Does N here have anything to do with the input N number of partitions? I assume not, right? If so, you may want to use different letter for it to avoid confusion.
  - Does the word `Stage` here means a general stage we use so far which means there will be data shuffling between stages. I also assume not. I think what you mean here is 3 steps and 2 stages. The BroadcastExec is in the same stage of its child. If so, you may want to modify the diagram for that, too
  
I am thinking about the name `NetworkBroadcastExec`. Can we make it `NetworkBroadcastCoalesceExec` or can we simply use the available `NetworkCoalesecExec`?


This is a question for @gabotechs. Does the BroadcastExec have to produce N * M output partitions? Is it possible to make a network channel to send one partition to M tasks? Just like send once but we have many receivers. Thus, we only need N output partitions here

src/distributed_planner/distributed_physical_optimizer_rule.rs
NGA-TRAN: I have the same question. Something like this will be easier to understand if you add a specific plan example in the comment.

src/execution_plans/network_broadcast.rs
```diff
@@ -0,0 +1,338 @@
...
+///            └▲┘ └▲┘ └▲┘                                       └▲┘ └▲┘ └▲┘                ■
+///             │   │   │                                         │   │   │
+///             │   │   │                                         │   │   │
+///             │   │   │                                         │   │   │
+///            ┌┴┐ ┌┴┐ ┌┴┐                                       ┌┴┐ ┌┴┐ ┌┴┐                ■
+///            │1│ │2│ │3│                                       │1│ │2│ │3│                │
+///     ┌──────┴─┴─┴─┴─┴─┴──────┐                         ┌──────┴─┴─┴─┴─┴─┴──────┐     Stage N-2
+///     │Arc<dyn ExecutionPlan> │           ...           │Arc<dyn ExecutionPlan> │         │
+///     │       (task 1)        │                         │       (task N)        │         │
+///     └───────────────────────┘                         └───────────────────────┘         ■
```
NGA-TRAN: If the diagram in Broadcast gets modified, these 2 diagrams need to get changed accordingly.

This beautiful diagram makes me think we definitely need examples (and then document) for our broadcast to show best practice because it simply should only be used in certain cases.  For example, we may want to either merge all partitions or split them into different partitions after `NetworkBroadcastExec`.

Let us chat. I will explain this to you better because it is not that simple

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -426,11 +505,23 @@ mod tests {
...
-            DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[MaxTemp, RainToday], file_type=parquet
+        ┌───── DistributedExec ── Tasks: t0:[p0]
+        │ CoalescePartitionsExec
+        │   [Stage 2] => NetworkCoalesceExec: output_partitions=3, input_tasks=3
+        └──────────────────────────────────────────────────
+          ┌───── Stage 2 ── Tasks: t0:[p0] t1:[p1] t2:[p2]
+          │ CoalesceBatchesExec: target_batch_size=8192
+          │   HashJoinExec: mode=CollectLeft, join_type=Left, on=[(RainToday@1, RainToday@1)], projection=[MinTemp@0, MaxTemp@2]
+          │     CoalescePartitionsExec
+          │       [Stage 1] => NetworkBroadcastExec: partitions_per_consumer=1, stage_partitions=3, input_tasks=3
```
NGA-TRAN: This is very cool. I think we will be able to display clearer information here. Let us chat

src/distributed_planner/distributed_physical_optimizer_rule.rs
NGA-TRAN: Agree. We do not want to change available tests. New tests should be added for new bahavior

tests/tpch_explain_analyze.rs
```diff
@@ -86,37 +82,62 @@ mod tests {
...
...ata/tpch/explain_analyze_sf0.1/supplier/5.parquet], [/testdata/tpch/explain_analyze_sf0.1/supplier/6.parquet, /testdata/tpch/explain_analyze_sf0.1/supplier/7.parquet, /testdata/tpch/explain_analyze_sf0.1/supplier/8.parquet], ...]}, projection=[s_suppkey, s_nationkey], file_type=parquet, predicate=DynamicFilter [ s_nationkey@1 >= 0 AND s_nationkey@1 <= 24 ], pruning_predicate=s_nationkey_null_count@1 != row_count@2 AND s_nationkey_max@0 >= 0 AND s_nationkey_null_count@1 != row_count@2 AND s_nationkey_min@3 <= 24, required_guarantees=[], metrics=[output_rows=1000, elapsed_compute=<metric>, batches_split=<metric>, bytes_scanned=<metric>, file_open_errors=<metric>, file_scan_errors=<metric>, num_predicate_creation_errors=<metric>, predicate_cache_inner_records=0, predicate_cache_records=0, predicate_evaluation_errors=<metric>, pushdown_rows_matched=<metric>, pushdown_rows_pruned=<metric>, bloom_filter_eval_time=<metric>, metadata_load_time=<metric>, page_index_eval_time=<metric>, row_pushdown_eval_time=<metric>, statistics_eval_time=<metric>, time_elapsed_opening=<metric>, time_elapsed_processing=<metric>, time_elapsed_scanning_total=<metric>, time_elapsed_scanning_until_data=<metric>]
```
NGA-TRAN: We can pick one or a few of the tpcds queries for our examples and documents. In general, we do not want too much network shuffling (aka too many stages). We will identify use cases for this and also list the cases that we do not want to use  broadcast for us to play with and document

NGA-TRAN (commented):
What a PR. Nice work @gene-bordegaray 🎉 

I have a question for @gabotechs, a few questions for you and some thing for us to chat.

src/distributed_planner/plan_annotator.rs
```diff
@@ -63,88 +65,219 @@ impl Debug for AnnotatedPlan {
...
+///                                   └──────────────────────┘
+/// ```
 ///
-/// 1. Leaf nodes have the opportunity to provide an estimation of how many distributed tasks should
-///    be used for the whole stage that will execute them.
+/// # Pass 1: bottom-to-top Task Estimation and Mark Network Boundaries
 ///
-/// 2. If a stage contains multiple leaf nodes, and all provide a task count estimation, the
-///    biggest is taken.
+/// This pass is a bottom-to-top pass that sets each node's task_count that depends on its children's task_count
```
gene-bordegaray: I could just leave out marking broadcsat at all here and only mark in the top-down, @gabotechs let me know what you think

src/distributed_planner/plan_annotator.rs
```diff
@@ -476,11 +662,11 @@ mod tests {
         "#;
         let annotated = sql_to_annotated(query).await;
         assert_snapshot!(annotated, @r"
-        CoalesceBatchesExec: task_count=Maximum(1)
-          HashJoinExec: task_count=Maximum(1)
-            CoalescePartitionsExec: task_count=Maximum(1)
-              DataSourceExec: task_count=Maximum(1)
-            DataSourceExec: task_count=Maximum(1)
+        CoalesceBatchesExec: task_count=Desired(3)
+          HashJoinExec: task_count=Desired(3)
```
gene-bordegaray: now distributed but doesn't broadcast

benchmarks/cdk/bin/worker.rs
```diff
@@ -38,6 +38,10 @@ struct Cmd {
     /// The bucket name.
     #[structopt(long, default_value = "datafusion-distributed-benchmarks")]
     bucket: String,
+
+    // Turns broadcast joins on.
+    #[structopt(long)]
+    broadcast_joins: bool,
```
gabotechs, line 44: 😄 it's funny how you managed to choose three different arguments for the broadcast joins flags in three different places (`broadcast_joins`, `enable_broadcast_joins`, `broadcast_joins_enabled`).

It's a pretty good idea to add this option, I'd probably just choose one flag name in all places, probably `broadcast_joins` since its shorter.

benchmarks/cdk/bin/worker.rs
```diff
@@ -67,12 +71,15 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-    let state = SessionStateBuilder::new()
+    let mut state = SessionStateBuilder::new()
         .with_default_features()
         .with_runtime_env(Arc::clone(&runtime_env))
         .with_distributed_worker_resolver(Ec2WorkerResolver::new())
         .with_physical_optimizer_rule(Arc::new(DistributedPhysicalOptimizerRule))
         .build();
+    if cmd.broadcast_joins {
+        state = state.with_distributed_broadcast_joins_enabled(true)?;
+    }
```
gabotechs: Nit: even sorter:

```suggestion
        .with_physical_optimizer_rule(Arc::new(DistributedPhysicalOptimizerRule))
        .with_distributed_broadcast_joins_enabled(cmd.broadcast_joins)?
        .build();
```

src/distributed_planner/plan_annotator.rs
gabotechs: I think we should be able to do all this in a simpler way without two annotation passes.

I have a clear idea on how to do this, and the result should be pretty simple, but I might be overlooking something. My two cents are:

1. We should try doing a first pass (no matter if it's top-down or bottom-up, both work) unrelated to plan annotator (probably not even in `plan_annotator.rs`) for placing `BroadcastExec` nodes right below `CoalescePartitionExecs` present in build side of hash joins. This should be pretty straight forward.
2. In `plan_annotator.rs`, we should be able to just treat the `BroadcastExec` nodes as if it was pretty much a normal `RepartitionExec`, which happens to trigger `Broadcast` boundaries rather than `Shuffle`.
3. in `distributed_physical_optimizer_rule.rs`, we would react to a `Broadcast` boundary in the same way we react to `Shuffle`, but with a `NetworkBroadcastExec` instead of a `NetworkShuffleExec`

I think if you do 1 then the rest should come pretty naturally.

gabotechs (commented):
Let me create a PR to this one solving the conflicts with `main`

gabotechs:
Here's a PR for solving the conflicts with `main`:

https://github.com/gene-bordegaray/datafusion-distributed/pull/1

src/distributed_planner/plan_annotator.rs
```diff
@@ -192,13 +210,16 @@ fn _annotate_plan(
...
+        // Only distriubte CollectLeft HashJoins after we broadcast more intelligently or when it
+        // is explicitly enabled.
         task_count = Maximum(1);
-    } else {
-        // The task count for this plan is decided by the biggest task count from the children; unless
+    } else if !has_broadcast_child {
+        // For Shuffle/Coalesce we inherit parallelism from children in the same stage.
+        // For Broadcast we intentionally skip this so the build-side task count does not
+        // drive the probe-side stage size.
+        // The task count for this plan is decided by the biggest task count from the children unless
```
gabotechs: Unfortunately we are not going to be able to get away with this. We do need any node present on the build side above the broadcast boundary to drive the stage size of the probe size, if there's a node in the build-side that says `Maximum(1)` and we don't honor it, we might distribute a node in multiple task that is only mean run in one.

I don't think this is happening today, but we should be fine not special-casing this so that it does not bite us in the future.

src/distributed_planner/plan_annotator.rs
```diff
@@ -629,6 +686,157 @@ mod tests {
...
+        INNER JOIN weather c ON b."RainToday" = c."RainToday"
+        "#;
+        let annotated = sql_to_annotated_broadcast(query, 4, 4, true).await;
+        assert_snapshot!(annotated, @r"
+        HashJoinExec: task_count=Desired(3)
+          CoalescePartitionsExec: task_count=Desired(3)
+            [NetworkBoundary] Broadcast: task_count=Desired(3)
+              BroadcastExec: task_count=Desired(3)
+                HashJoinExec: task_count=Desired(3)
+                  CoalescePartitionsExec: task_count=Desired(1)
```
gabotechs: This is a wrong annotation. By the time a plan is annotated all the task counts in all the nodes within a stage need to agree on one task count, and these two seem to not be agreeing yet.

src/distributed_planner/plan_annotator.rs
```diff
@@ -629,6 +686,157 @@ mod tests {
...
+            CoalescePartitionsExec: task_count=Maximum(1)
+              [NetworkBoundary] Broadcast: task_count=Maximum(1)
+                BroadcastExec: task_count=Desired(1)
+                  DataSourceExec: task_count=Desired(1)
+            DataSourceExec: task_count=Maximum(1)
+          HashJoinExec: task_count=Maximum(2)
+            CoalescePartitionsExec: task_count=Maximum(2)
+              [NetworkBoundary] Broadcast: task_count=Maximum(2)
+                BroadcastExec: task_count=Desired(2)
+                  DataSourceExec: task_count=Desired(2)
```
gabotechs, line 849: 🤔 this also seems wrong. This `DataSourceExec` should have `task_count=Desired(3)` as all the other `DataSourceExec` (in the end, they are all the same). But somehow, it seems like the `Maximum(2)` from the task above is propagating that `2` downstairs?

src/distributed_planner/plan_annotator.rs
```diff
@@ -629,6 +686,157 @@ mod tests {
...
+        FROM weather a INNER JOIN weather b
+        ON a."RainToday" = b."RainToday"
+        "#;
+        let annotated = sql_to_annotated_broadcast(query, 4, 4, true).await;
+        assert_snapshot!(annotated, @r"
+        HashJoinExec: task_count=Desired(3)
+          CoalescePartitionsExec: task_count=Desired(3)
+            [NetworkBoundary] Broadcast: task_count=Desired(3)
+              BroadcastExec: task_count=Desired(3)
+                DataSourceExec: task_count=Desired(3)
```
gabotechs, line 696: 👍

src/distributed_planner/plan_annotator.rs
```diff
@@ -629,6 +686,157 @@ mod tests {
...
+        ON a."RainToday" = b."RainToday"
+        "#;
+        let annotated =
+            sql_to_annotated_broadcast_with_estimator(query, 3, BuildSideOneTaskEstimator).await;
+        assert_snapshot!(annotated, @r"
+        HashJoinExec: task_count=Desired(3)
+          CoalescePartitionsExec: task_count=Desired(3)
+            [NetworkBoundary] Broadcast: task_count=Desired(3)
+              BroadcastExec: task_count=Maximum(1)
+                DataSourceExec: task_count=Maximum(1)
```
gabotechs, line 745: 💯 nice test case!

src/distributed_planner/plan_annotator.rs
```diff
@@ -106,31 +109,40 @@ impl Debug for AnnotatedPlan {
...
-///            │
-/// ┌──────────┴─────────┐ task_count: Desired(4) (inherited from the child)
-/// │    Aggregation     │ network_boundary: None
-/// │     (partial)      │
-/// └──────────▲─────────┘
-///            │
-/// ┌──────────┴─────────┐ task_count: Desired(4) (this was set by a TaskEstimator implementation)
-/// │   DataSourceExec   │ network_boundary: None
-/// └────────────────────┘
+///
```
gabotechs, line 112: Awesome docs! My only suggestion is to rearrange them so that they occupy less width. IDEs use this doc comments for rendering a nicely formatted doc on hover, but it's so wide that its barely readable.

🤔 how about offsetting the two branches vertically so that one is above the other, and you can afford to put all the text at the right side, leaving the drawings very close to the left?

src/distributed_planner/plan_annotator.rs
gabotechs: Ok, I think I understand what you did here. You are force-propagating a `Maximum` annotation across broadcast boundaries so that they do not get unnecessarily distributed.

With the specific plans we see in the tests, it's clear that it's better to limit the broadcast producers to as many broadcast consumers we have, that actually avoids some unnecessary network hops in situations were we are broadcasting to 1 task.

However, I'd love to be in a position where this happens automatically without us explicitly enforcing it, and I think the root problem that is causing this is a suboptimal task assignation mechanism that prompts the distributed planner to tend to assign too many tasks to build sides. What I think should be happening automatically is:
- The build side in a CollectLeft join is supposed to be small
- The probe side in a CollectLeft join is supposed to be big (bigger than the build side at least)
- As the build side is small, it naturally gets assigned very few tasks
- As the probe side is big, it naturally gets assigned more tasks than the build side

I think a proper solution to https://github.com/datafusion-contrib/datafusion-distributed/issues/296 should bring us closer to this situation.

Unfortunately, we are not there yet, so I think you are making the right call by enforcing this limit manually in case of a broadcast boundary 👍.

src/distributed_planner/plan_annotator.rs
```diff
@@ -256,44 +285,68 @@ fn _annotate_plan(
...
-                task_count.as_usize(),
-            )?;
-            for children_and_tasks in c_i_union.task_idx_map.iter() {
-                for (child_i, task_ctx) in children_and_tasks {
-                    if let Some(child) = annotation.children.get_mut(*child_i) {
-                        propagate_task_count(child, &Maximum(task_ctx.task_count), d_cfg)?
-                    };
+        match &annotation.plan_or_nb {
+            PlanOrNetworkBoundary::Plan(plan) => {
+                if d_cfg.children_isolator_unions && plan.as_any().is::<UnionExec>() {
```
gabotechs: I understand the changes in this function, I think you are making the right call here 👍.

My only suggestion is to try to reduce the level of indentation here, as this is complicated code, and the less nesting the clearer it is for readers.

Would you be up to doing something like:

```rust
    fn propagate_task_count(
        annotation: &mut AnnotatedPlan,
        task_count: &TaskCountAnnotation,
        d_cfg: &DistributedConfig,
    ) -> Result<(), DataFusionError> {
        annotation.task_count = task_count.clone();
        let plan = match &annotation.plan_or_nb {
            // If it's a normal plan, continue with the propagation.
            PlanOrNetworkBoundary::Plan(plan) => plan,
            // Broadcast is a stage split only propagate a Maximum cap into the build stage.
            PlanOrNetworkBoundary::Broadcast => {
                if let Maximum(max) = task_count {
                    for child in &mut annotation.children {
                        let child_task_count = child.task_count.clone().limit(*max);
                        propagate_task_count(child, &child_task_count, d_cfg)?;
                    }
                }
                return Ok(());
            }
            // This is a network boundary.
            //
            // Nothing to propagate here, all the nodes below the network boundary were already
            // assigned a task count, we do not want to overwrite it.
            PlanOrNetworkBoundary::Shuffle => return Ok(()),
            PlanOrNetworkBoundary::Coalesce => return Ok(()),
        };

        if d_cfg.children_isolator_unions && plan.as_any().is::<UnionExec>() {
            // Propagating through ChildrenIsolatorUnionExec is not that easy, each child will
            // be executed in its own task, and therefore, they will act as if they were in executing
            // in a non-distributed context. The ChildrenIsolatorUnionExec itself will make sure to
            // determine which children to run and which to exclude depending on the task index in
            // which it's running.
            let c_i_union = ChildrenIsolatorUnionExec::from_children_and_task_counts(
                plan.children().into_iter().cloned(),
                annotation.children.iter().map(|v| v.task_count.as_usize()),
                task_count.as_usize(),
            )?;
            for children_and_tasks in c_i_union.task_idx_map.iter() {
                for (child_i, task_ctx) in children_and_tasks {
                    if let Some(child) = annotation.children.get_mut(*child_i) {
                        propagate_task_count(child, &Maximum(task_ctx.task_count), d_cfg)?
                    };
                }
            }
            annotation.plan_or_nb = PlanOrNetworkBoundary::Plan(Arc::new(c_i_union));
        } else {
            for child in &mut annotation.children {
                propagate_task_count(child, task_count, d_cfg)?;
            }
        }
        Ok(())
    }
```

?

src/distributed_planner/plan_annotator.rs
gabotechs: IMO I'd just leave this chunk of code as it is on `main` right now, we should be fine to not special case here, and it actually solves one issue that is shown in the tests in this file https://github.com/datafusion-contrib/datafusion-distributed/pull/279#discussion_r2702281060

src/distributed_planner/plan_annotator.rs
```diff
@@ -256,44 +285,68 @@ fn _annotate_plan(
...
+        if matches!(annotation.plan_or_nb, PlanOrNetworkBoundary::Broadcast) {
+            // Broadcast splits build/probe stages only cap the build stage if the probe is Maximum.
+            if let Maximum(max) = annotation.task_count {
+                for annotated_child in annotation.children.iter_mut() {
+                    let child_task_count = annotated_child.task_count.clone().limit(max);
+                    propagate_task_count(annotated_child, &child_task_count, d_cfg)?;
+                }
+            }
+            return Ok(annotation);
+        }
```
gabotechs: 🤔 I don't think you need this in both `fn propagate_task_count()` and here. I think with just the one on `fn propagate_task_count()` should be enough.

tests/tpcds_correctness_test.rs
```diff
@@ -572,6 +572,13 @@ mod tests {
             .with_distributed_files_per_task(FILES_PER_TASK)?
             .with_distributed_cardinality_effect_task_scale_factor(CARDINALITY_TASK_COUNT_FACTOR)?;
 
+        // Enable broadcast joins if BROADCAST_JOINS env var is set
+        let d_ctx = if std::env::var("BROADCAST_JOINS").is_ok() {
+            d_ctx.with_distributed_broadcast_joins(true)?
+        } else {
+            d_ctx
+        };
+
```
gabotechs: How about just enabling this for the tests? Ideally, tests should not fail or succeed depending on environment variables, a good practice is to make them hermetic to the outside world.

I'd be fine with just having this enabled for all tests.

gabotechs (commented):
💯 This is looking pretty good! I'd be very surprised if this is not my last round of comments on this PR.

src/distributed_planner/plan_annotator.rs
gabotechs: Just auto answered my question, I'll leave the comments above anyways for your entertainment.

gene-bordegaray: Yes, this was a tricky situation when I ran into it and thought this is the best approach before we have intelligent broadcasting 😄 . With heuristics and a more sophisticated task estimator, I agree, this should not be needed

src/distributed_planner/plan_annotator.rs
gene-bordegaray: much cleaner 💯

src/distributed_planner/plan_annotator.rs
gene-bordegaray: ya it seemed a bit awkward, I think extending and keeping text on right is more legible

benchmarks/cdk/bin/worker.rs
gene-bordegaray: hahaha ya it was, too many names 😅

src/distributed_planner/plan_annotator.rs
gene-bordegaray: ended up going this route 💯

src/distributed_planner/plan_annotator.rs
gene-bordegaray: I believe that you found the answer to this logic in later comments 😄 , this is preventing the build side from influencing the probe side

src/distributed_planner/plan_annotator.rs
gene-bordegaray: Yes they should and can agree. I can relax the condition to only ignore teh build side at the join's child and allow normal propagation elsewhere which will keep the same properties but have stages agree on their task counts

src/distributed_planner/plan_annotator.rs
gene-bordegaray: actually I drew out some tests and concluded incorrectly on this extra loop

yes, the propagate_task_count handles this and not needed

src/distributed_planner/plan_annotator.rs
```diff
@@ -192,17 +229,22 @@ fn _annotate_plan(
...
+        // probe side drives the stage size.
+        // The task count for this plan is decided by the biggest task count from the children unless
         // a child specifies a maximum task count, in that case, the maximum is respected. Some
         // nodes can only run in one task. If there is a subplan with a single node declaring that
         // it can only run in one task, all the rest of the nodes in the stage need to respect it.
-        for annotated_child in annotated_children.iter() {
+        for (idx, annotated_child) in annotated_children.iter().enumerate() {
+            if build_is_broadcast && idx == 0 {
+                continue;
+            }
```
gabotechs: This is the only point left I'm struggling to understand. Why would you not want the build side to take part of the task count decision? it might have legitimate reasons to affect the task count, like if the user has a custom node applying a `Maximum(1)` or if the the build side turns out to be greater than expected and it's even bigger than the probe side (if planning was perfect, this would never happen).

src/distributed_planner/plan_annotator.rs
```diff
@@ -298,6 +361,12 @@ fn _annotate_plan(
             propagate_task_count(annotated_child, &annotation.task_count, d_cfg)?;
         }
 
+        // If the NetworkBoundary boundary below is broadcast then we don't want to scale based on
+        // cardinality, thus return early.
+        if matches!(annotation.plan_or_nb, PlanOrNetworkBoundary::Broadcast) {
+            return Ok(annotation);
+        }
```
gabotechs: Same comment as above

src/distributed_planner/plan_annotator.rs
gene-bordegaray: this check is needed to protect the build side from not influencing the probe side. Without this check if the probe side has say Desired(3) and the build Maximum(1), the probe side would then be changed to Maximum(1).

If you comment this out and run tests `test_broadcast_one_to_many` will show this behavior

gene-bordegaray: I think this is something that should be handled in planning based on statistics. If we decide broadcast a large build side then that's a planning issue

src/distributed_planner/plan_annotator.rs
gene-bordegaray: yes, this is right to question. I had incorrect logic here and think this is essentally a no-op

this is exclusively in a build side and the join's stage size is based on the probe side thus the broadcast never affects the parent stage.

I think this is the right choice since choosing to broadcast should be only when the build side is confidently small and a planning issue.

src/distributed_planner/plan_annotator.rs
gene-bordegaray: This also may be a reason to change where broadcasts are inserted going forward in the new task estimator design. Such as if there are more tasks than expected it should most likely just not broadcast

gabotechs: > If you comment this out and run tests test_broadcast_one_to_many will show this behavior

I can comment that out and `test_broadcast_one_to_many` passes equally. 

That makes sense, because the `Maximum(1)` exists below the `Broadcast` network boundary, not above, and therefore, it does not get propagated as a `Maximum(1)`, it gets propagated as a `Desired(1)`, which in the reconciliation step gets scaled up to `Desired(3)` because the probe side is providing a `Desired(3)`.


> Without this check if the probe side has say Desired(3) and the build Maximum(1), the probe side would then be changed to Maximum(1).

But this is what we want actually, if the build size declares a `Maximum(1)` we need the whole hash join to run in exactly 1 task. 

Lets see this with an example:

Imagine that we implement a custom intermediate results caching system, were you can place intermediate result cache nodes anywhere in the plan, and the results would get cached:

```
                    ┌─────────────────────┐                 
                    │    HashJoinExec     │                 
                    └─────────────────────┘                 
   ┌──────────────────────┐         ┌──────────────────────┐
 + │IntermediateCacheExec │         │      FilterExec      │
   └──────────────────────┘         └──────────────────────┘
   ┌──────────────────────┐         ┌──────────────────────┐
   │CoalescePartitionsExec│         │    DataSourceExec    │
   └──────────────────────┘         └──────────────────────┘
   ┌──────────────────────┐                                 
   │    AggregateExec     │                                 
   └──────────────────────┘                                 
   ┌──────────────────────┐                                 
   │    DataSourceExec    │                                 
   └──────────────────────┘                                 
```

Now, imagine that whoever created `IntermediateCacheExec` does not want the node to run in more than 1 task, because the node is just not prepared to do so, as that would translate in N tasks trying to write to the same cache in a remote storage. In that case, the developer of `IntermediateCacheExec` would provide a `Maximum(1)` task estimation, telling the planner: "no matter what you do, you cannot run my node in more than 1 task, as that would yield incorrect results"

With the changes in this PR, the task estimation would look like this:

```
                 ┌─────────────────────┐                          
                 │    HashJoinExec     Desired(3)                 
                 └─────────────────────┘                          
┌──────────────────────┐         ┌──────────────────────┐         
│IntermediateCacheExec Desired(3)│      FilterExec      Desired(3)
└──────────────────────┘         └──────────────────────┘         
┌──────────────────────┐         ┌──────────────────────┐         
│CoalescePartitionsExecDesired(3)│    DataSourceExec    Desired(3)
└──────────────────────┘         └──────────────────────┘         
┌──────────────────────┐                                          
│ NetworkBroadcastExec Desired(3)                                 
└──────────────────────┘                                          
┌──────────────────────┐                                          
│    AggregateExec     Desired(1)                                 
└──────────────────────┘                                          
┌──────────────────────┐                                          
│    DataSourceExec    Desired(1)                                 
└──────────────────────┘                                          
```

But that will yield incorrect results, because `IntermediateCacheExec` is not prepared to run in more than 1 task.

The correct thing to do in this case is:

```
                 ┌─────────────────────┐                          
                 │    HashJoinExec     Maximum(1)                 
                 └─────────────────────┘                          
┌──────────────────────┐         ┌──────────────────────┐         
│IntermediateCacheExec Maximum(1)│      FilterExec      Maximum(1)
└──────────────────────┘         └──────────────────────┘         
┌──────────────────────┐         ┌──────────────────────┐         
│CoalescePartitionsExecMaximum(1)│    DataSourceExec    Maximum(1)
└──────────────────────┘         └──────────────────────┘         
┌──────────────────────┐                                          
│    AggregateExec     Desired(1)                                 
└──────────────────────┘                                          
┌──────────────────────┐                                          
│    DataSourceExec    Desired(1)                                 
└──────────────────────┘                                          
```

And not distribute the query at all.

gene-bordegaray: discussed online and resolved, added tests for this by forcing with a Maximum(1) on coalesce

tests/tpch_plans_test.rs
```diff
@@ -748,22 +748,22 @@ mod tests {
...
+          ┌───── Stage 6 ── Tasks: t0:[p0..p5] t1:[p0..p5] t2:[p0..p5] t3:[p0..p5] 
           │ SortExec: expr=[s_suppkey@0 ASC NULLS LAST], preserve_partitioning=[true]
           │   HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(max(revenue0.total_revenue)@0, total_revenue@4)], projection=[s_suppkey@1, s_name@2, s_address@3, s_phone@4, total_revenue@5]
           │     CoalescePartitionsExec
-          │       [Stage 3] => NetworkBroadcastExec: partitions_per_consumer=1, stage_partitions=2, input_tasks=1
+          │       [Stage 3] => NetworkBroadcastExec: partitions_per_consumer=1, stage_partitions=4, input_tasks=1
           │     HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(s_suppkey@0, supplier_no@0)], projection=[s_suppkey@0, s_name@1, s_address@2, s_phone@3, total_revenue@5]
           │       CoalescePartitionsExec
-          │         [Stage 4] => NetworkBroadcastExec: partitions_per_consumer=4, stage_partitions=8, input_tasks=4
+          │         [Stage 4] => NetworkBroadcastExec: partitions_per_consumer=4, stage_partitions=16, input_tasks=4
```
gene-bordegaray, line 760: this build side is actually larger than the probe here 4 > 2

tests/tpch_plans_test.rs
```diff
@@ -748,22 +748,22 @@ mod tests {
         assert_snapshot!(plan, @r"
         ┌───── DistributedExec ── Tasks: t0:[p0] 
         │ SortPreservingMergeExec: [s_suppkey@0 ASC NULLS LAST]
-        │   [Stage 6] => NetworkCoalesceExec: output_partitions=12, input_tasks=2
+        │   [Stage 6] => NetworkCoalesceExec: output_partitions=24, input_tasks=4
         └──────────────────────────────────────────────────
-          ┌───── Stage 6 ── Tasks: t0:[p0..p5] t1:[p0..p5] 
+          ┌───── Stage 6 ── Tasks: t0:[p0..p5] t1:[p0..p5] t2:[p0..p5] t3:[p0..p5]
```
gene-bordegaray, line 753: this gets scaled up

src/distributed_planner/plan_annotator.rs
```diff
@@ -886,6 +890,33 @@ mod tests {
...
+    struct BroadcastBuildCoalesceMaxEstimator;
+
+    impl TaskEstimator for BroadcastBuildCoalesceMaxEstimator {
+        fn task_estimation(
+            &self,
+            plan: &Arc<dyn ExecutionPlan>,
+            _: &ConfigOptions,
+        ) -> Option<TaskEstimation> {
+            let coalesce = plan.as_any().downcast_ref::<CoalescePartitionsExec>()?;
+            if coalesce.input().as_any().is::<BroadcastExec>() {
```
gene-bordegaray, line 903: don't know if this is what you had in mind to force this behavior, lmk

gabotechs: Yes! this is exactly it

src/distributed_ext.rs
```diff
@@ -549,55 +566,59 @@ impl DistributedExt for SessionConfig {
     }
 
     delegate! {
-        to self {
```
gabotechs, line 552: This seems to be funny indented now?

tests/tpch_plans_test.rs
gabotechs: It's likely that here the issue is with DataFusion itself, as it's placing the big table on the left.

gabotechs (approved):
💯 Absolutely epic work

src/distributed_ext.rs
gene-bordegaray: I can check this

Issue #260, opened 2025-12-18
Invalid plans with metrics collection


gabotechs:
https://github.com/datafusion-contrib/datafusion-distributed/pull/259 is uncovering a bug with the metrics collection that can be very well seen in `main` by just adding some debug statements in the metrics collection tests.

The plans below are gathered by just placing a dbg! statement in each test:

```rust
        let display = display_plan_ascii(plan.as_ref(), true);
        dbg!(&display);
```

https://github.com/datafusion-contrib/datafusion-distributed/blob/gabrielmusat%2Frework-distributed-planning-logic/src/metrics/task_metrics_rewriter.rs#L470-L470

The plan shown here is:
```
┌───── DistributedExec ── Tasks: t0:[p0] 
│ SortPreservingMergeExec: [id@0 ASC NULLS LAST], fetch=10, metrics=[output_rows=2, elapsed_compute=12.667µs, output_bytes=24.0 B]
│   SortExec: TopK(fetch=10), expr=[id@0 ASC NULLS LAST], preserve_partitioning=[true], metrics=[output_rows=2, elapsed_compute=326.918µs, output_bytes=0.0 B, row_replacements=2]
│     ProjectionExec: expr=[id@0 as id, count(Int64(1))@1 as count], metrics=[output_rows=2, elapsed_compute=4.169µs, output_bytes=1088.0 B]
│       AggregateExec: mode=FinalPartitioned, gby=[id@0 as id], aggr=[count(Int64(1))], metrics=[output_rows=2, elapsed_compute=233.002µs, output_bytes=1088.0 B, spill_count=0, spilled_bytes=0.0 B, spilled_rows=0, peak_mem_used=8576, aggregate_arguments_time=4.377µs, aggregation_time=48.002µs, emitting_time=4.043µs, time_calculating_group_ids=5.294µs]
│         [Stage 1] => NetworkShuffleExec: output_partitions=4, input_tasks=2, metrics=[]
└──────────────────────────────────────────────────
  ┌───── Stage 1 ── Tasks: t0:[p0..p3] t1:[p0..p3] 
  │ CoalesceBatchesExec: target_batch_size=8192, metrics=[output_rows=4, elapsed_compute=34µs]
[...]

PR #281, opened 2026-01-08
Fix metrics collection bug: preserve PartitionIsolatorExec in plan

sundinesh1:
- Fix stage_metrics_rewriter to correctly match metrics to nodes in pre-order traversal
- Update test assertions to allow empty metrics for PartitionIsolatorExec nodes
- Fix assert_metrics_present_in_plan to handle PartitionIsolatorExec wrapped in MetricsWrapperExec

Fixes issue where PartitionIsolatorExec nodes were missing from displayed plans after metrics rewriting. All metrics collection tests now pass.

this is a fix for : #260

### DIFF
--- a/src/metrics/task_metrics_collector.rs
+++ b/src/metrics/task_metrics_collector.rs
@@ -124,9 +124,7 @@ mod tests {
     use crate::test_utils::parquet::register_parquet_tables;
     use crate::test_utils::plans::{count_plan_nodes, get_stages_and_stage_keys};
     use crate::test_utils::session_context::register_temp_parquet_table;
-    use crate::{
-        DistributedExt, DistributedPhysicalOptimizerRule, NetworkBoundaryExt, PartitionIsolatorExec,
-    };
+    use crate::{DistributedExt, DistributedPhysicalOptimizerRule};
     use datafusion::execution::{SessionStateBuilder, context::SessionContext};
     use datafusion::prelude::SessionConfig;
     use datafusion::{
@@ -281,33 +279,10 @@ mod tests {
                 "Mismatch between collected metrics and actual nodes for {expected_stage_key:?}"
             );
 
-            // Collect which nodes are expected to have no metrics (like PartitionIsolatorExec)
-            // We need to traverse in the same order as TaskMetricsCollector (DFS/pre-order, excluding network boundaries)
-            let mut node_idx = 0;
-            let mut nodes_without_metrics = std::collections::HashSet::new();
-            stage_plan
-                .clone()
-                .transform_down(|plan| {
-                    if plan.is_network_boundary() {
-                        return Ok(Transformed::new(plan, false, TreeNodeRecursion::Jump));
-                    }
-                    // PartitionIsolatorExec nodes typically don't have metrics
-                    if plan
-                        .as_any()
-                        .downcast_ref::<PartitionIsolatorExec>()
-                        .is_some()
-                    {
-                        nodes_without_metrics.insert(node_idx);
-                    }
-                    node_idx += 1;
-                    Ok(Transformed::no(plan))
-                })
-                .unwrap();
-
-            // Ensure metrics were collected for all nodes. Some nodes (like PartitionIsolatorExec)
-            // may legitimately have empty metrics, which is fine - we just verify the metrics set exists.
-            // The actual metrics content is verified by checking that the count matches the plan nodes.
-            // Note: Empty metrics sets are allowed for certain nodes like PartitionIsolatorExec.
+            // Verify that metrics were collected for all nodes. Some nodes may legitimately have
+            // empty metrics (e.g., custom execution plans without metrics), which is fine - we
+            // just verify that a metrics set exists for each node. The count assertion above
+            // ensures all nodes are included in the metrics collection.
         }
     }
 
@@ -404,47 +379,19 @@ mod tests {
         let collector = TaskMetricsCollector::new();
         let result = collector.collect(dist_exec.plan.clone()).unwrap();
 
-        // Verify PartitionIsolatorExec nodes are preserved in metrics collection
+        // Verify all nodes (including PartitionIsolatorExec) are preserved in metrics collection
         for expected_stage_key in expected_stage_keys {
             let actual_metrics = result.input_task_metrics.get(&expected_stage_key).unwrap();
             let stage = stages.get(&(expected_stage_key.stage_id as usize)).unwrap();
             let stage_plan = stage.plan.decoded().unwrap();
 
-            // Count PartitionIsolatorExec nodes and verify they're included in metrics
-            let mut partition_isolator_indices = Vec::new();
-            let mut node_idx = 0;
-            stage_plan
-                .clone()
-                .transform_down(|plan| {
-                    if plan.is_network_boundary() {
-                        return Ok(Transformed::new(plan, false, TreeNodeRecursion::Jump));
-                    }
-                    if plan
-                        .as_any()
-                        .downcast_ref::<PartitionIsolatorExec>()
-                        .is_some()
-                    {
-                        partition_isolator_indices.push(node_idx);
-                    }
-                    node_idx += 1;
-                    Ok(Transformed::no(plan))
-                })
-                .unwrap();
-
-            // Verify metrics count matches (this ensures PartitionIsolatorExec nodes are included)
+            // Verify metrics count matches - this ensures all nodes are included in metrics collection
+            // regardless of whether they have metrics or not (some nodes may have empty metrics sets)
             assert_eq!(
                 actual_metrics.len(),
                 count_plan_nodes(stage_plan),
-                "Metrics count must match plan nodes (including PartitionIsolatorExec) for stage {expected_stage_key:?}"
+                "Metrics count must match plan nodes for stage {expected_stage_key:?}"
             );
-
-            // Verify PartitionIsolatorExec nodes have empty metrics (corner case)
-            for &idx in &partition_isolator_indices {
-                assert!(
-                    actual_metrics[idx].metrics.is_empty(),
-                    "PartitionIsolatorExec at index {idx} should have empty metrics"
-                );
-            }
         }
     }
 }


src/common/ttl_map.rs
```diff
@@ -156,7 +156,7 @@ impl TTLMapConfig {
             ));
         }
 
-        if self.ttl.as_nanos() % self.tick.as_nanos() != 0 {
+        if !self.ttl.as_nanos().is_multiple_of(self.tick.as_nanos()) {
```
gabotechs: I'd avoid making changes unrelated to the purpose of this PR for the sake of traceability of changes.

Feel free to open a separate PR for these small improvements though.

src/distributed_planner/plan_annotator.rs
```diff
@@ -319,10 +319,10 @@ fn required_network_boundary_below(parent: &dyn ExecutionPlan) -> Option<Require
...
     let first_child = children.first()?;
 
-    if let Some(r_exec) = first_child.as_any().downcast_ref::<RepartitionExec>() {
-        if matches!(r_exec.partitioning(), Partitioning::Hash(_, _)) {
-            return Some(RequiredNetworkBoundary::Shuffle);
-        }
+    if let Some(r_exec) = first_child.as_any().downcast_ref::<RepartitionExec>()
+        && matches!(r_exec.partitioning(), Partitioning::Hash(_, _))
+    {
+        return Some(RequiredNetworkBoundary::Shuffle);
```
gabotechs: Same as above, even if they are nice changes, they can unnecessarily result in conflicts with other people's work, so if it's not related to the purpose of the PR, I'd try to leave them aside.

src/distributed_planner/plan_annotator.rs
```diff
@@ -632,7 +632,7 @@ mod tests {
 
     #[allow(clippy::type_complexity)]
     struct CallbackEstimator {
-        f: Arc<dyn Fn(&(dyn ExecutionPlan)) -> Option<TaskEstimation> + Send + Sync>,
+        f: Arc<dyn Fn(&dyn ExecutionPlan) -> Option<TaskEstimation> + Send + Sync>,
```
gabotechs: Same as above

src/execution_plans/network_shuffle.rs
```diff
@@ -162,10 +162,10 @@ impl NetworkShuffleExec {
                 // This is fine, we can let the node be here.
                 Ok(Transformed::no(plan))
             } else {
-                return plan_err!(
+                plan_err!(
                     "NetworkShuffleExec input must be hash partitioned, but {} is not",
                     plan.name()
-                );
+                )
```
gabotechs: Same as above

src/metrics/proto.rs
```diff
@@ -159,11 +159,11 @@ pub fn df_metrics_set_to_proto(
...
-                if let DataFusionError::Internal(msg) = &err {
-                    if msg == CUSTOM_METRICS_NOT_SUPPORTED || msg == UNSUPPORTED_METRICS {
-                        // Filter out custom/unsupported metrics error - continue processing other metrics
-                        continue;
-                    }
+                if let DataFusionError::Internal(msg) = &err
+                    && (msg == CUSTOM_METRICS_NOT_SUPPORTED || msg == UNSUPPORTED_METRICS)
+                {
+                    // Filter out custom/unsupported metrics error - continue processing other metrics
+                    continue;
```
gabotechs: Same as above

src/metrics/task_metrics_collector.rs
```diff
@@ -272,32 +274,54 @@ mod tests {
...
-                );
+            // Collect which nodes are expected to have no metrics (like PartitionIsolatorExec)
+            // We need to traverse in the same order as TaskMetricsCollector (DFS/pre-order, excluding network boundaries)
+            let mut node_idx = 0;
+            let mut nodes_without_metrics = std::collections::HashSet::new();
+            fn traverse_plan(
+                plan: &Arc<dyn ExecutionPlan>,
+                node_idx: &mut usize,
+                nodes_without_metrics: &mut std::collections::HashSet<usize>,
+            ) {
```
gabotechs: DataFusion already has methods for plan traversal, maybe we can use them instead? I imagine `plan.transform_down` will do what you want with less code.

src/metrics/task_metrics_collector.rs
```diff
@@ -335,8 +358,158 @@ mod tests {
...
     async fn test_metrics_collection_e2e_4() {
         run_metrics_collection_e2e_test("SELECT distinct company from table2").await;
     }
+
+    /// Test that verifies metrics collection works correctly with PartitionIsolatorExec nodes.
+    /// PartitionIsolatorExec nodes are created when using children_isolator_unions feature.
+    /// This test ensures that PartitionIsolatorExec nodes are preserved in the plan during
+    /// metrics collection and that empty metrics are collected for them.
+    #[tokio::test]
+    async fn test_metrics_collection_with_partition_isolator() {
```
gabotechs, line 340: This test seems unreasonably long, I think we should be able to scope it a bit more and make it simpler.

What do you think about using one one of those predefined parquet files like the `weather` data set for testing this?

Also, I'd say that we should be well covered with the existing tests, I don't see this one really testing anything that the others are not testing, so I think we should even be fine without this test.

sundinesh1: ok, reduced the test size and added it for the weather dataset. what im testing

- PartitionIsolatorExec nodes are included in metrics collection (not skipped)
- Metrics count matches plan node count (ensures all nodes, including PartitionIsolatorExec, are counted)
- PartitionIsolatorExec nodes have empty metrics (the expected behavior)

src/metrics/task_metrics_collector.rs
```diff
@@ -272,32 +274,50 @@ mod tests {
...
+                        return Ok(Transformed::new(plan, false, TreeNodeRecursion::Jump));
+                    }
+                    // PartitionIsolatorExec nodes typically don't have metrics
+                    if plan
+                        .as_any()
+                        .downcast_ref::<PartitionIsolatorExec>()
+                        .is_some()
+                    {
+                        nodes_without_metrics.insert(node_idx);
+                    }
```
gabotechs: 🤔 I'm not sure if this is the right approach. `PartitionIsolatorExec` should not be special here, in the same way that today `PartitionIsolatorExec` does not have metrics, there might be some other nodes that also do not have metrics (like if a user has their own custom plan without metrics), and that should not be a problem for the task metrics collector.

From another point of view:

We do want to add metrics to `PartitionIsolatorExec`, and these tests can wrongly start failing when we do that, requiring the developer who contributed `PartitionIsolatorExec` metrics to fix it.

gabotechs: In short: If you find yourself needing to special case certain nodes (having logic conditionally executed `if node == ParittionIsolatorExec`) then it's probably not the right approach.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -445,18 +469,20 @@ mod tests {
 
     // Assert every plan node has at least one metric except partition isolators, network boundary nodes, and the root DistributedExec node.
     fn assert_metrics_present_in_plan(plan: &Arc<dyn ExecutionPlan>) {
+        // Check if this is a PartitionIsolatorExec (possibly wrapped in MetricsWrapperExec)
+        // For MetricsWrapperExec, we check the name() which delegates to the inner plan
+        let is_partition_isolator = plan.name() == "PartitionIsolatorExec"
+            || plan
+                .as_any()
+                .downcast_ref::<PartitionIsolatorExec>()
+                .is_some();
```
gabotechs, line 478: Even simpler:

```suggestion
        let is_partition_isolator = plan.name() == PartitionIsolatorExec::static_name();
```

src/metrics/task_metrics_rewriter.rs
```diff
@@ -170,12 +174,32 @@ pub fn stage_metrics_rewriter(
...
+        node_metrics_vec.push(metrics_set_proto_to_df(&stage_metrics)?);
         node_idx += 1;
+        Ok(Transformed::no(plan))
+    })?;
+
+    // Now rewrite the plan with the collected metrics
+    let mut node_idx = 0;
+    plan.clone().transform_down(|plan| {
+        // Stop at network boundaries.
+        if plan.as_network_boundary().is_some() {
```
gabotechs, line 186:
```suggestion
        if plan.is_network_boundary() {
```

src/metrics/task_metrics_collector.rs
```diff
@@ -272,32 +272,27 @@ mod tests {
...
-                let metrics_set = metrics_set_proto_to_df(metrics_set).unwrap();
-                assert!(
-                    metrics_set.iter().count() > 0,
-                    "Did not found metrics for Stage {expected_stage_key:?}"
-                );
-            }
+            // Verify that metrics were collected for all nodes. Some nodes may legitimately have
+            // empty metrics (e.g., custom execution plans without metrics), which is fine - we
+            // just verify that a metrics set exists for each node. The count assertion above
+            // ensures all nodes are included in the metrics collection.
```
gabotechs, line 285: 🤔 there's a comment here... but no code honor the comment?

jayshrivastava: I think this comment was meant to go above the `assert_eq!` instead of below it.

jayshrivastava (changes requested):
Thanks for the contribution! Looks good but I think we can revert a part of this change.

It seems that `test_metrics_collection_e2e*` tests were failing because they were not accounting for certain nodes (ex. partition isolator) which have no metrics.

Separately, it seems that `test_executed_distributed_plan_has_metrics` was failing because `plan.as_any().downcast_ref::<MetricsWrapperExec>()` can never happen since `as_any()` on `MetricsWrapperExec` is delegated to the inner node [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/a89de5f94f026c463bb0b04d93ec7ba081753380/src/execution_plans/metrics.rs?plain=1#L52).

Both of these problems are fixed by fixing the tests.

It seems that the changes to the `stage_metrics_rewriter` function can be reverted as the function is still logically equivalent to the previous one. The new implementation is more clear, but it allocates a vec and traverses the plan twice. It should be fine to keep the previous one.

PR #282, opened 2026-01-09
Identify benchmark queries as arbitrary strings

gabotechs:
This allows us to put any extra query we want in the `queries/` folder and use it for benchmarking distributed DataFusion.

My use case is that I wanted to create a query that unions 300+ children, and while it might not be worth it to permanently commit all the queries we use for benchmarks, it's fine to still have the option to create temporary ones just for 1-shot scenarios.

### DIFF
--- a/tests/tpch_explain_analyze.rs
+++ b/tests/tpch_explain_analyze.rs
@@ -1069,15 +1069,12 @@ mod tests {
         Ok(())
     }
 
-    // test_non_distributed_consistency runs each tpc-h query twice - once in a distributed manner
-    // and once in a non-distributed manner. For each query, it asserts that the results are identical.
+    // test_tpch_query runs each TPC-H query in a distributed manner while collecting metrics.
+    // This allows us to call `explain_analyze` on the resulting executed plan.
     async fn test_tpch_query(query_id: &str) -> Result<String, Box<dyn Error>> {
         let (mut ctx, _guard) = start_localhost_context(4, DefaultSessionBuilder).await;
         ctx.set_distributed_metrics_collection(true)?;
-        run_tpch_query(ctx, query_id).await
-    }
 
-    async fn run_tpch_query(ctx: SessionContext, query_id: &str) -> Result<String, Box<dyn Error>> {
         let data_dir = ensure_tpch_data(TPCH_SCALE_FACTOR, TPCH_DATA_PARTS).await;
         let sql = tpch::get_query(query_id)?;
         ctx.state_ref()
--- a/tests/tpch_plans_test.rs
+++ b/tests/tpch_plans_test.rs
@@ -994,12 +994,6 @@ mod tests {
     // test_tpch_query generates and displays a distributed plan for each TPC-H query.
     async fn test_tpch_query(query_id: &str) -> Result<String, Box<dyn Error>> {
         let (ctx, _guard) = start_localhost_context(4, DefaultSessionBuilder).await;
-        run_tpch_query(ctx, query_id).await
-    }
-
-    // test_non_distributed_consistency runs each TPC-H query twice - once in a distributed manner
-    // and once in a non-distributed manner. For each query, it asserts that the results are identical.
-    async fn run_tpch_query(ctx: SessionContext, query_id: &str) -> Result<String, Box<dyn Error>> {
         let data_dir = ensure_tpch_data(TPCH_SCALE_FACTOR, TPCH_DATA_PARTS).await;
         let sql = tpch::get_query(query_id)?;
         ctx.state_ref()


benchmarks/src/prepare_tpch.rs
```diff
@@ -113,3 +113,102 @@ impl PrepareTpchOpt {
         Ok(())
     }
 }
+
+const TPCH_TABLES: &[&str] = &[
+    "part", "supplier", "partsupp", "customer", "orders", "lineitem", "nation", "region",
+];
+
+pub fn get_tpch_table_schema(table: &str) -> datafusion::arrow::datatypes::Schema {
```
gabotechs, line 121: Moved here as it was only used in this file

tests/tpch_plans_test.rs
```diff
@@ -992,38 +992,29 @@ mod tests {
...
     // test_tpch_query generates and displays a distributed plan for each TPC-H query.
-    async fn test_tpch_query(query_id: u8) -> Result<String, Box<dyn Error>> {
+    async fn test_tpch_query(query_id: &str) -> Result<String, Box<dyn Error>> {
         let (ctx, _guard) = start_localhost_context(4, DefaultSessionBuilder).await;
         run_tpch_query(ctx, query_id).await
     }
 
-    async fn run_tpch_query(ctx: SessionContext, query_id: u8) -> Result<String, Box<dyn Error>> {
+    // test_non_distributed_consistency runs each TPC-H query twice - once in a distributed manner
+    // and once in a non-distributed manner. For each query, it asserts that the results are identical.
```
JSOD11: I think this comment is left over from earlier work and is outdated.

tests/tpch_plans_test.rs
```diff
@@ -992,38 +992,29 @@ mod tests {
     }
 
     // test_tpch_query generates and displays a distributed plan for each TPC-H query.
-    async fn test_tpch_query(query_id: u8) -> Result<String, Box<dyn Error>> {
+    async fn test_tpch_query(query_id: &str) -> Result<String, Box<dyn Error>> {
```
JSOD11, line 995: Looks like the only thing that this function does is start the context and then call `run_tpch_query`, in which case maybe it's unnecessary to have two functions. We could take the body of `run_tpch_query` and move it here?

Also side note, I liked how in `tpch_correctness_test.rs` we have the `test_tpch_query` function take in the `sql` string, so we can just call `tpch::get_query` right there in line and start from the `sql`. Just a minor stylistic thing, I'm also happy to take a pass and standardize these in that way in a future PR.

tests/tpch_explain_analyze.rs
```diff
@@ -1069,41 +1069,30 @@ mod tests {
         Ok(())
     }
 
-    // test_tpch_query runs each TPC-H query in a distributed manner while collecting metrics.
-    // This allows us to call `explain_analyze` on the resulting executed plan.
-    async fn test_tpch_query(query_id: u8) -> Result<String, Box<dyn Error>> {
+    // test_non_distributed_consistency runs each tpc-h query twice - once in a distributed manner
+    // and once in a non-distributed manner. For each query, it asserts that the results are identical.
```
JSOD11: I think this comment is from earlier work and is outdated, I believe the original comment there was correct.

tests/tpch_correctness_test.rs
```diff
@@ -74,62 +74,62 @@ mod tests {
...
     #[tokio::test]
     async fn test_tpch_21() -> Result<(), Box<dyn Error>> {
-        test_tpch_query(tpch::get_test_tpch_query(21)?).await
+        test_tpch_query(tpch::get_query("q21")?).await
     }
 
     #[tokio::test]
     async fn test_tpch_22() -> Result<(), Box<dyn Error>> {
-        test_tpch_query(tpch::get_test_tpch_query(22)?).await
+        test_tpch_query(tpch::get_query("q22")?).await
```
JSOD11, line 132: I like the pattern here where `test_tpch_query` takes a `sql` string and we use `tpch::get_query` inline.

JSOD11 (approved):
LGTM. Made a couple small comments about outdated comments that I believe we can remove and a possible small refactor for standardization, but the changes make sense to me.

tests/tpch_plans_test.rs
gabotechs: Yes! missed that while resolving conflicts

tests/tpch_plans_test.rs
gabotechs: Agree! this file has evolved a lot and I imagine there was a time where having two functions made sense. It doesn't now.

tests/tpch_explain_analyze.rs
gabotechs: Thanks for pointing this out

PR #285, opened 2026-01-09
Coalesce partitions before and after NetworkCoalesceExec

gabotechs:
Coalesces partitions before and after a network boundary, instead of just after it:

This:
```
        │     CoalescePartitionsExec
        │       [Stage 1] => NetworkCoalesceExec: output_partitions=6, input_tasks=3
          ┌───── Stage 1 ── Tasks: t0:[p0..p1] t1:[p2..p3] t2:[p4..p5] 
          │ AggregateExec: mode=Partial, gby=[], aggr=[count(Int64(1))]
          │   ProjectionExec: expr=[]
```
Becomes this:

```
        │     CoalescePartitionsExec
        │       [Stage 1] => NetworkCoalesceExec: output_partitions=3, input_tasks=3
          ┌───── Stage 1 ── Tasks: t0:[p0] t1:[p1] t2:[p2] 
          │ CoalescePartitionsExec
          │   AggregateExec: mode=Partial, gby=[], aggr=[count(Int64(1))]
          │     ProjectionExec: expr=[]
```

This allows us to:
1. Simplify the network communication between the boundaries
2. Do some early work in a distributed manner for when coalescing implies ordering with `SortMergePreservingExec`


src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -254,13 +261,14 @@ mod tests {
...
-        │     [Stage 2] => NetworkCoalesceExec: output_partitions=8, input_tasks=2
+        │     [Stage 2] => NetworkCoalesceExec: output_partitions=2, input_tasks=2
         └──────────────────────────────────────────────────
-          ┌───── Stage 2 ── Tasks: t0:[p0..p3] t1:[p0..p3] 
-          │ SortExec: expr=[count(*)@0 ASC NULLS LAST], preserve_partitioning=[true]
-          │   ProjectionExec: expr=[count(Int64(1))@1 as count(*), RainToday@0 as RainToday, count(Int64(1))@1 as count(Int64(1))]
-          │     AggregateExec: mode=FinalPartitioned, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))]
-          │       [Stage 1] => NetworkShuffleExec: output_partitions=4, input_tasks=3
+          ┌───── Stage 2 ── Tasks: t0:[p0] t1:[p1] 
+          │ SortPreservingMergeExec: [count(Int64(1))@2 ASC NULLS LAST]
```
gabotechs, line 267: By doing an early `SortPreservingMergeExec` in the distributed tasks, we make the top level `SortPreservingMergeExec`'s life easier, as it only needs to merge 2 partitions instead of 8

src/distributed_planner/plan_annotator.rs
```diff
@@ -324,15 +324,17 @@ fn required_network_boundary_below(parent: &dyn ExecutionPlan) -> Option<Require
...
         // If the next node is a leaf node, distributing this is going to be a bit wasteful, so
         // we don't want to do it.
         if first_child.children().is_empty() {
             return None;
         }
-        return Some(RequiredNetworkBoundary::Coalesce);
+        return Some(RequiredNetworkBoundary::Coalesce(None));
+    }
+
+    if let Some(node) = parent.as_any().downcast_ref::<SortPreservingMergeExec>() {
```
gene-bordegaray, line 336: would we want the same first_child.children().is_empty() check here like the coalesce has to prevent unnecessary coalescing on leaf?

gene-bordegaray (commented):
Overall looks good and logic makes sense, had one question about leaf node case

PR #289, opened 2026-01-12
feat: add binary crate for ddf observability tool

EdsonPetry:


console/.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
+name: CI
```
JSOD11: I think we can likely delete this entire `.github` directory including everything within it.

console/.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
JSOD11: I think we can get rid of this file.

console/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
JSOD11: Let's also delete this file for now.

Also, rather than "Edson Petry" the license should be something like "Apache Software Foundation" I believe, you can find the exact wording in other files throughout the project.

console/README.md
```diff
@@ -0,0 +1,14 @@
+# console
```
JSOD11: Let's delete this as well, we can create documentation later.

Cargo.toml
```diff
@@ -1,58 +1,11 @@
-[workspace]
```
JSOD11, line 1: I suspect something might be wrong here — this is the outer `Cargo.toml` responsible for the entire project, so a huge diff being correct feels unlikely. Something closer to a 1-line change adding the "console" package is more along the lines of what I might expect.

PR #291, opened 2026-01-13
Move forward to datafusion 52

marc-pydantic:
This PR moves everything to Datafusion 52, which has been released yesterday/today (depending where you are!). All dependencies that are shared with `datafusion` are upgraded to the same version the Datafusion 52 release is using to minimize friction. Other that that, it was just a minor patch so far.

The big issue is with `ballista`: It depends transitively on a different version of the native `lzma`, due to a different `apache-avro` version being amongst its dependencies. There's no way to reconcile this due to C's lack of namespaces, so I commented out ballista for now. That's why this PR is a draft for now, please let me know how to proceed!

**Update**: I reviewed all `cargo insta` snapshots, but a second pair of eyes can't hurt. From what I understand these are exclusively changes in the optimizer causing less `CoalesceBatchesExec` and `RoundRobinBatch` nodes to be emitted.


gabotechs:
I think the fact that the `CoalesceBatchesExec` has been embedded into `RepartitionExec` is giving problems...

Previously, we were marking the `CoalesceBatchesExec` node as the place were a network boundary should be placed below, but this now disappeared... solving that might be a bit complex, as it will require changes in the planning logic.

If you are down for doing it, I'm more than happy to guide you, otherwise I can make a PR to this one with the necessary changes.

PR #236, opened 2025-11-21
Upgrade to DF v51

ahmed-mez:
- Bump datafusion version to v51
- Adapt deps and code to datafusion v51.

PR #294, opened 2026-01-15
Replace explain_analyze.rs tests with more scoped metrics_collection.rs tests

gabotechs:
The `explain_analyze.rs` tests break on each DataFusion upgrade, and ideally that should not be the case, as our current solution for propagating metrics is supposed to be valid for any DataFusion version, so in this PR we remove the `explain_analyze.rs` tests in favor of a more scoped `metrics_collection.rs` test that should be valid across DataFusion upgrades.
- [How they broke in the 51 upgrade](https://github.com/datafusion-contrib/datafusion-distributed/pull/236/files#diff-ea45675cb8df1aa21e37909eed2086b358874899ccb052c3ef94d65ac22c7aea)
- [How they are breaking in the 52 upgrade](https://github.com/datafusion-contrib/datafusion-distributed/pull/293)

### DIFF
--- a/src/test_utils/parquet.rs
+++ b/src/test_utils/parquet.rs
@@ -1,6 +1,11 @@
 use datafusion::error::DataFusionError;
 use datafusion::prelude::{ParquetReadOptions, SessionContext};
 
+/// Adds two test tables to the provided [SessionContext]:
+/// - `flights_1m`: 1M rows of flight data.
+/// - `weather`: smaller dataset with weather forecast.
+///
+/// Useful for testing queries using SQL directly.
 pub async fn register_parquet_tables(ctx: &SessionContext) -> Result<(), DataFusionError> {
     ctx.register_parquet(
         "flights_1m",


tests/metrics_collection.rs
```diff
@@ -0,0 +1,193 @@
...
+    async fn test_metrics_collection_in_aggregation() -> Result<(), Box<dyn std::error::Error>> {
+        let (d_ctx, _guard) = start_localhost_context(3, DefaultSessionBuilder).await;
+
+        let query =
+            r#"SELECT count(*), "RainToday" FROM weather GROUP BY "RainToday" ORDER BY count(*)"#;
+
+        let s_ctx = SessionContext::default();
+        let (s_physical, d_physical) = execute(&s_ctx, &d_ctx, query).await?;
+        println!("{}", display_plan_ascii(s_physical.as_ref(), true));
+        println!("{}", display_plan_ascii(d_physical.as_ref(), true));
```
jayshrivastava: nit: use `eprintln`?

tests/metrics_collection.rs
```diff
@@ -0,0 +1,193 @@
...
+                &s_physical,
+                &d_physical,
+                data_source_index,
+            );
+        }
+
+        Ok(())
+    }
+
+    fn assert_metrics_equal<T: ExecutionPlan + 'static>(
```
jayshrivastava, line 120: Very nice. If you want, this can live in `test_utils` instead of here.

tests/metrics_collection.rs
```diff
@@ -0,0 +1,193 @@
...
+            assert_eq!(one_metric, other_metric);
+        }
+    }
+
+    async fn execute(
+        s_ctx: &SessionContext,
+        d_ctx: &SessionContext,
+        query: &str,
+    ) -> Result<(Arc<dyn ExecutionPlan>, Arc<dyn ExecutionPlan>), Box<dyn std::error::Error>> {
+        register_parquet_tables(s_ctx).await?;
```
jayshrivastava, line 138: Mind adding a doc comment to register_parquet_tables while you are there? ex. how it adds weather and flight tables by default.

tests/metrics_collection.rs
jayshrivastava: If so, then a doc comment would be useful.

tests/metrics_collection.rs
```diff
@@ -0,0 +1,193 @@
...
+        let d_physical = d_df.create_physical_plan().await?;
+        execute_stream(d_physical.clone(), d_ctx.task_ctx())?
+            .try_collect::<Vec<_>>()
+            .await?;
+        let d_physical = rewrite_distributed_plan_with_metrics(d_physical.clone())?;
+
+        Ok((s_physical, d_physical))
+    }
+
+    fn node_metrics<T: ExecutionPlan + 'static>(
```
jayshrivastava, line 157: Cool.

tests/metrics_collection.rs
gabotechs: It can even be removed I think

tests/metrics_collection.rs
gabotechs: 🤔 It's a very good candidate indeed.

I'll keep in eye in case this can be used in more places. If at any point we do, this function will deserve the "promotion" of being in test_utils

gabotechs: I'll add the doc comment anyways though

tests/metrics_collection.rs
gabotechs: done!

PR #297, opened 2026-01-15
Add functionality for worker-console communication

EdsonPetry:
Added tonic-based observability gRPC codegen for console process. Updated worker/server wiring to include observability service, updated console to use client gRPC and render ping status in the TUI (to be extended to stats and metrics for cluster/workers later).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,6 @@ members = ["benchmarks", "cli", "console"]
 [workspace.dependencies]
 datafusion = { version = "52.0.0", default-features = false }
 datafusion-proto = { version = "52.0.0" }
-datafusion-distributed = { path = "." }
 
 [package]
 name = "datafusion-distributed"
@@ -84,6 +83,3 @@ hyper-util = "0.1.16"
 pretty_assertions = "1.4"
 reqwest = "0.12"
 zip = "6.0"
-
-[build-dependencies]
-tonic-prost-build = "0.14.1"
--- a/console/Cargo.toml
+++ b/console/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "console"
+name = "datafusion-distributed-console"
 version = "0.1.0"
 edition = "2024"
 
@@ -9,6 +9,6 @@ crossterm = "0.29.0"
 ratatui = "0.30.0"
 tokio = { version = "1.49.0", features = ["full"] }
 tonic = "0.14.2"
-datafusion-distributed = { workspace = true }
+datafusion-distributed = { path = "..", features = ["integration"] }
 arrow-flight = "57.2.0"
 structopt = "0.3.26"
--- a/src/flight_service/worker.rs
+++ b/src/flight_service/worker.rs
@@ -3,7 +3,6 @@ use crate::flight_service::WorkerSessionBuilder;
 use crate::flight_service::do_get::TaskData;
 use crate::protobuf::StageKey;
 use crate::{DefaultSessionBuilder, ObservabilityServiceImpl};
-use crate::{ObservabilityService, PingRequest, PingResponse};
 use arrow_flight::flight_service_server::{FlightService, FlightServiceServer};
 use arrow_flight::{
     Action, ActionType, Criteria, Empty, FlightData, FlightDescriptor, FlightInfo,
@@ -214,15 +213,5 @@ impl FlightService for Worker {
         _: Request<Empty>,
     ) -> Result<Response<Self::ListActionsStream>, Status> {
         Err(Status::unimplemented("Not yet implemented"))
-    }
-}
-
-#[tonic::async_trait]
-impl ObservabilityService for Worker {
-    async fn ping(
-        &self,
-        _request: tonic::Request<PingRequest>,
-    ) -> Result<tonic::Response<PingResponse>, tonic::Status> {
-        Ok(tonic::Response::new(PingResponse { value: 1 }))
     }
 }
--- a/src/observability/README.md
+++ b/src/observability/README.md
@@ -2,7 +2,7 @@
 
 ---
 This module provides gRPC-based observability services using Protocol Buffers.
-The setup is inspired by Apache DataFusion's proto crate structure but adapted
+The setup is inspired by [Apache DataFusion's proto crate](https://github.com/apache/datafusion/tree/main/datafusion/proto) structure but adapted
 for tonic/gRPC services.
 
 In the root of the datafusion-distribued repo, run:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1407,20 +1407,6 @@ checksum = "75984efb6ed102a0d42db99afb6c1948f0380d1d91808d5529916e6c08b49d8d"
 
 [[package]]
 name = "console"
-version = "0.1.0"
-dependencies = [
- "arrow-flight",
- "color-eyre",
- "crossterm",
- "datafusion-distributed",
- "ratatui",
- "structopt",
- "tokio",
- "tonic",
-]
-
-[[package]]
-name = "console"
 version = "0.15.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "054ccb5b10f9f2cbf51eb355ca1d05c2d279ce1804688d0db74b4733a5aeafd8"
@@ -2033,7 +2019,6 @@ dependencies = [
  "tokio-stream",
  "tonic",
  "tonic-prost",
- "tonic-prost-build",
  "tower",
  "tpchgen",
  "tpchgen-arrow",
@@ -2090,6 +2075,20 @@ dependencies = [
  "tonic",
  "tower",
  "url",
+]
+
+[[package]]
+name = "datafusion-distributed-console"
+version = "0.1.0"
+dependencies = [
+ "arrow-flight",
+ "color-eyre",
+ "crossterm",
+ "datafusion-distributed",
+ "ratatui",
+ "structopt",
+ "tokio",
+ "tonic",
 ]
 
 [[package]]
@@ -3504,7 +3503,7 @@ version = "1.46.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "248b42847813a1550dafd15296fd9748c651d0c32194559dbc05d804d54b21e8"
 dependencies = [
- "console 0.15.11",
+ "console",
  "once_cell",
  "regex",
  "similar",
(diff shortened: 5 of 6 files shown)


src/protobuf/mod.rs
```diff
@@ -1,11 +1,18 @@
...
 mod errors;
+pub mod observability;
 mod user_codec;
 
 pub(crate) use app_metadata::{AppMetadata, FlightAppMetadata, MetricsCollection, TaskMetrics};
 pub(crate) use distributed_codec::{DistributedCodec, StageKey};
-pub(crate) use errors::{datafusion_error_to_tonic_status, map_flight_to_datafusion_error};
+pub(crate) use errors::{
+    datafusion_error_to_tonic_status, map_flight_to_datafusion_error,
+    map_status_to_datafusion_error,
```
JSOD11: Can we remove this unused import?

build.rs
```diff
@@ -0,0 +1,10 @@
+use std::io::Result;
```
JSOD11: Could you provide some documentation / explanation of what this file does? This is compiling protos at build time?

.github/actions/setup/action.yml
```diff
@@ -20,6 +20,12 @@ runs:
         echo $HOME
         echo ${{ runner.os }}
 
+    - name: Install protoc
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y protobuf-compiler
```
JSOD11: I'm a little concerned about these sudo commands. I think `apt-get` may only be available on linux and return an error on mac or windows machines. Can we gate this to linux and double check that this change is safe / works on all machines?

src/flight_service/worker.rs
```diff
@@ -123,6 +126,31 @@ impl Worker {
             .max_decoding_message_size(usize::MAX)
             .max_encoding_message_size(usize::MAX)
     }
+
+    /// Convenience method that converts [Worker] into a [`ObservabilityServiceServer`]. This allows
```
JSOD11: Does it convert the worker into an `ObservabilityServiceServer`, consuming it? Will we be unable to then register the same worker for flight?

console/src/main.rs
```diff
@@ -1,20 +1,29 @@
+use datafusion_distributed::PingRequest;
+use datafusion_distributed::protobuf::observability::observability_service_client::ObservabilityServiceClient;
+use ratatui::widgets::Paragraph;
 use ratatui::{DefaultTerminal, Frame};
 
-fn main() -> color_eyre::Result<()> {
+#[tokio::main]
```
JSOD11, line 6: This looks like it's headed in the right direction, though I'm thinking it would be nice to have some instructions on how to run this example which we can build on as we continue with the project. Maybe it's time to add an example folder with a README? E.g., a `console/examples/README.md` containing commands describing how to start worker(s) with observability service, then run the console and see an interaction.

.github/actions/setup/action.yml
EdsonPetry: Thanks for pointing this out, it was a quick fix when I did it to just get the CI checks passing, but I think a better way to do it would be to use a setup action that installs `protoc` and is cross platform w/ a specific version.

src/flight_service/worker.rs
EdsonPetry: Yes, similarly to how `.into_flight_server()` converts the worker instance into a `FlightServiceServer`, using `into_observability_server()` will also require a separate worker instance and consumes that instance. This was the shortest path to just get the ping to work but I realize the issue is that there's no way for shared state since it's not the same worker handling both flight and observability. I can make a small API tweak for the worker, where I implement `ObservabilityServiceServer` for the `Arc<Worker>` and that should let us use the worker as so:

```
let worker = Arc::new(Worker::default());
Server::builder()
  .add_service(FlightServiceServer::new(worker.clone())
  .add_service(ObservabilityServiceServer::new(worker.clone()))
```

Let me know what you think.

JSOD11: This seems like the right approach to me if we can get it to work.

src/flight_service/worker.rs
```diff
@@ -31,6 +33,17 @@ pub struct Worker {
     pub(super) max_message_size: Option<usize>,
 }
 
+/// A wrapper for [Arc<Worker>] useful for implementing multiple services
+/// such as [with_flight_service()] and [with_observability_service()].
```
JSOD11: Nit: Fix comment now that we dropped these methods.

src/flight_service/worker.rs
```diff
@@ -211,3 +251,109 @@ impl FlightService for Worker {
...
+    async fn ping(
+        &self,
+        _request: tonic::Request<PingRequest>,
+    ) -> Result<tonic::Response<PingResponse>, tonic::Status> {
+        Ok(tonic::Response::new(PingResponse { value: 1 }))
+    }
+}
+
+#[tonic::async_trait]
+impl ObservabilityService for SharedWorker {
```
gabotechs: In order to satisfy the `ObservabilityService` methods, I don't think we need a full worker for this.

A `Worker` has several things that are unrelated to observability, like:
- the `session_builder`: only relevant for building DataFusion sessions, but I don't think we'd want to build any new DataFusion session for observing the cluster.
- the `WorkerHooks`: these are fired on receiving distributed query requests, but we are not receiving those in the ObservabilityService.
- the `max_message_size`: this is the message for inter-worker communication, so it's irrelevant during observability.

This leads me to think that we might be better off creating a new struct that, for now, just shares the `Arc<TTLMap<StageKey, Arc<OnceCell<TaskData>>>>` with a worker.

That should also translate into less code, as we do not need a `SharedWorker` struct.

src/proto/observability.proto
gabotechs, line 1: It seems like now we have a folder called `proto/` and another one called `protobuf/`. I want to propose a different organization for this, so that we can scope things a bit better for building the console:

How about having a new folder called `src/observability/`, and we scope everything regarding observability there? we'd put things like:
- The observability.proto file
- The `ObservabilityService` struct implementation
- The generated server stubs
- etc...

build.rs
```diff
@@ -0,0 +1,14 @@
...
+// The output is compiled into the crate, so no generated files are checked in.
+
+fn main() -> Result<()> {
+    tonic_prost_build::configure()
+        .build_server(true)
+        .build_client(true)
+        .compile_protos(&["src/proto/observability.proto"], &["src/proto"])?;
+
+    Ok(())
+}
```
gabotechs: This is usually the way to go for having an autogenerated protobuf server, however... there might be reasons for doing it differently in this project.

The main challenge comes because of the fact that this is a library that a lot of people might use, and by generating a Tonic sever out of protobuf the classic way, we are forcing anyone importing the library to have `protoc` installed so that they can compile their protos. This will likely:
- break people's codebase if they don't have `protoc` installed (I see this is even seen in the CI of this project, hence the need of an extra GitHub action)
- slow down compilation times, as everyone will need go through this proto generation pass

DataFusion upstream has solved this in a smart way:
- https://github.com/apache/datafusion/tree/main/datafusion/proto

Rather than generating protos at compile-time, they just generate them with a script and commit the generated protos. That way, users do not need to have `protoc` installed in their machines or in their CI in order to compile this project.

I think it might be worth generating the Tonic server stubs manually, commit them to the project, and remove the `build.rs` file.

src/observability/gen/regen.sh
gabotechs, line 1: 👍  nice

src/observability/README.md
```diff
@@ -0,0 +1,12 @@
+# Distributed DataFusion Observability Proto Generation
+
+---
+This module provides gRPC-based observability services using Protocol Buffers.
+The setup is inspired by Apache DataFusion's proto crate structure but adapted
```
gabotechs: 👍 awesome! how about adding a link to Apache DataFusion's proto crate?

Cargo.toml
```diff
@@ -82,3 +84,6 @@ hyper-util = "0.1.16"
 pretty_assertions = "1.4"
 reqwest = "0.12"
 zip = "6.0"
+
+[build-dependencies]
+tonic-prost-build = "0.14.1"
```
gabotechs: This should no longer be needed right? we don't have any `build.rs` in which this dependency is used.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = ["benchmarks", "cli", "console"]
 [workspace.dependencies]
 datafusion = { version = "52.0.0", default-features = false }
 datafusion-proto = { version = "52.0.0" }
+datafusion-distributed = { path = "." }
```
gabotechs: Rather than adding this here, I'd add the import in `console/` like this https://github.com/datafusion-contrib/datafusion-distributed/blob/main/benchmarks/Cargo.toml#L10-L10.

Same way it's getting imported in the benchmarks crate. That way we can leave this file untouched.

console/Cargo.toml
```diff
@@ -7,3 +7,8 @@ edition = "2024"
 color-eyre = "0.6.5"
 crossterm = "0.29.0"
 ratatui = "0.30.0"
+tokio = { version = "1.49.0", features = ["full"] }
+tonic = "0.14.2"
+datafusion-distributed = { workspace = true }
```
gabotechs: In order to be consistent with the rest of the project, I'd import `datafusion-distributed` like this:

https://github.com/datafusion-contrib/datafusion-distributed/blob/main/benchmarks/Cargo.toml#L10-L10

src/flight_service/worker.rs
```diff
@@ -211,3 +216,13 @@ impl FlightService for Worker {
...
+
+#[tonic::async_trait]
+impl ObservabilityService for Worker {
+    async fn ping(
+        &self,
+        _request: tonic::Request<PingRequest>,
+    ) -> Result<tonic::Response<PingResponse>, tonic::Status> {
+        Ok(tonic::Response::new(PingResponse { value: 1 }))
+    }
+}
```
gabotechs: I want to double down on my comment here: https://github.com/datafusion-contrib/datafusion-distributed/pull/297#discussion_r2715709743. The `ObservabilityService` and the `Worker` should be decoupled. The `Worker` should not know about an "ObservabilityService".

I see that in this PR there's already a `impl ObservabilityService for ObservabilityServiceImpl` so maybe this is just a leftover that you forgot to delete?

EdsonPetry: Yes, sorry! Didn't realize I left that in there.

Issue #247, opened 2025-11-28
Option extensions not propagating if mutated


gabotechs:
There's currently one bug with the current implementation of the option extension propagation:
1. The user adds a new option extension with `with_distributed_option_extension`
2. The option extension gets mutated either programaticaly or by a `SET my_extension.foo = 'bar'` estatement
3. The option extension that gets propagated is the initial version, not the updated version

PR #298, opened 2026-01-19
Fix header extension propagation

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/247

Now config extensions can be mutated and it will still be correctly sent to downstream workers through headers.

There is one test that replicated the issue that was previously ignored, but it now runs fine.

---

This PR contains a breaking change on the public API, where `with_distributed_option_extension()` no longer returns a `Result<Self>`. It just returns `Self` now.

### DIFF
--- a/src/distributed_planner/task_estimator.rs
+++ b/src/distributed_planner/task_estimator.rs
@@ -179,10 +179,13 @@ pub(crate) fn set_distributed_task_estimator(
     } else {
         let mut estimators = CombinedTaskEstimator::default();
         estimators.user_provided.push(Arc::new(estimator));
-        set_distributed_option_extension(cfg, DistributedConfig {
-            __private_task_estimator: estimators,
-            ..Default::default()
-        }).expect("Calling set_distributed_option_extension with a default DistributedConfig should never fail");
+        set_distributed_option_extension(
+            cfg,
+            DistributedConfig {
+                __private_task_estimator: estimators,
+                ..Default::default()
+            },
+        )
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/distributed_planner/task_estimator.rs
```diff
@@ -179,10 +179,13 @@ pub(crate) fn set_distributed_task_estimator(
     } else {
         let mut estimators = CombinedTaskEstimator::default();
         estimators.user_provided.push(Arc::new(estimator));
-        set_distributed_option_extension(cfg, DistributedConfig {
-            __private_task_estimator: estimators,
-            ..Default::default()
-        }).expect("Calling set_distributed_option_extension with a default DistributedConfig should never fail");
+        set_distributed_option_extension(
+            cfg,
+            DistributedConfig {
```
jayshrivastava, line 184: It seems that files_per_task, children_isolator_unions, cardinality_task_count_factor are only used during planning, so we don't need to propagate them over the network for execution.

There's also stuff like `__private_task_estimator` which does not get propagated.

Wonder if there's a way to make it even more explicit what gets propagated and what doesnt.

jayshrivastava (approved):

src/distributed_planner/task_estimator.rs
gabotechs: > There's also stuff like __private_task_estimator which does not get propagated.

This is very annoying, I'd love to use something other than these config extensions, but the PR that enabled that upstream was not accepted.

gabotechs: > It seems that files_per_task, children_isolator_unions, cardinality_task_count_factor are only used during planning, so we don't need to propagate them over the network for execution.

It should not bee too bad to propagate them, it will require more code to explicitly avoid the propagation, and it does not have an impact in performance, so unless we see it becoming a problem and think we can just let everything get put into headers.

Issue #243, opened 2025-11-24
Use integrated `RepartitionExec` batching instead of adding a `CoalesceBatchesExec`


gabotechs:
Once https://github.com/apache/datafusion/issues/18782 is shipped, we should no longer manually add a CoalesceBatchesExec on top of repartitions manually

PR #299, opened 2026-01-19
Add new `batch_coalescing_below_network_boundaries` optimization pass

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/243

After https://github.com/apache/datafusion/issues/18782 and other series of PRs, DataFusion plans no longer have `CoalesceBatchesExec` nodes, as the batch coalescing logic has been embedded in other nodes themselves.

This PR accounts for that, and it allows users of this library to apply more aggressive batching right above network boundaries if they desire.

### DIFF
--- a/src/distributed_planner/distributed_physical_optimizer_rule.rs
+++ b/src/distributed_planner/distributed_physical_optimizer_rule.rs
@@ -1,16 +1,14 @@
 use crate::common::require_one_child;
+use crate::distributed_planner::batch_coalescing_below_network_boundaries;
 use crate::distributed_planner::plan_annotator::{
     AnnotatedPlan, PlanOrNetworkBoundary, annotate_plan,
 };
 use crate::{
-    DistributedConfig, DistributedExec, NetworkBoundaryExt, NetworkCoalesceExec,
-    NetworkShuffleExec, TaskEstimator,
+    DistributedConfig, DistributedExec, NetworkCoalesceExec, NetworkShuffleExec, TaskEstimator,
 };
-use datafusion::common::tree_node::{Transformed, TreeNode};
 use datafusion::config::ConfigOptions;
 use datafusion::error::DataFusionError;
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
-use datafusion::physical_plan::coalesce_batches::CoalesceBatchesExec;
 use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
 use datafusion::physical_plan::{ExecutionPlan, ExecutionPlanProperties};
 use std::fmt::Debug;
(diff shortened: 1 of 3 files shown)


src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -150,37 +151,47 @@ fn distribute_plan(
...
+            // There was already a CoalesceBatchesExec below...
+            if existing_coalesce.target_batch_size() == d_cfg.shuffle_batch_size {
+                // ...so either leave it alone if the batch size is correctly set...
+                Ok(Transformed::no(plan))
+            } else {
+                // ... or replace it with one with the correct batch size.
+                let coalesce_input = existing_coalesce.input();
+                let new_coalesce =
+                    CoalesceBatchesExec::new(Arc::clone(coalesce_input), d_cfg.shuffle_batch_size)
+                        .with_fetch(existing_coalesce.fetch());
```
jayshrivastava: nit: small risk. 

If someone adds a field to `CoalesceBatchesExec` in the future, we may reset that here because we we using `new` instead of `with_new_children`.

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -150,37 +151,47 @@ fn distribute_plan(
...
+                let new_plan = plan.with_new_children(vec![Arc::new(new_coalesce)])?;
+                Ok(Transformed::yes(new_plan))
+            }
+        } else {
+            // No CoalesceBatchesExec below, need to put one.
+            let coalesce_input = input;
+            let new_coalesce = CoalesceBatchesExec::new(coalesce_input, d_cfg.shuffle_batch_size);
+            let new_plan = plan.with_new_children(vec![Arc::new(new_coalesce)])?;
+            Ok(Transformed::yes(new_plan))
+        }
```
jayshrivastava: This should probably have a test

src/distributed_planner/distributed_physical_optimizer_rule.rs
```diff
@@ -150,37 +151,47 @@ fn distribute_plan(
...
+                // ... or replace it with one with the correct batch size.
+                let coalesce_input = existing_coalesce.input();
+                let new_coalesce =
+                    CoalesceBatchesExec::new(Arc::clone(coalesce_input), d_cfg.shuffle_batch_size)
+                        .with_fetch(existing_coalesce.fetch());
+                let new_plan = plan.with_new_children(vec![Arc::new(new_coalesce)])?;
+                Ok(Transformed::yes(new_plan))
+            }
+        } else {
+            // No CoalesceBatchesExec below, need to put one.
```
jayshrivastava: You said in DF 52, the plans are missing `CoalesceBatchesExec`. Does that mean we hit this case always?

jayshrivastava (approved):

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: 🤔 good call. But how could we workaround it? we need the `CoalesceBatchesExec` to have a different batch size, and we cannot modify the batch size of one that was already built

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: At least with the plans we see in the benchmarks yes, we are likely always going to hit this.

However, `CoalesceBatchesExec` is still there, and it's still exposed in DataFusion's API, so I assume there might still be a use case for it.

src/distributed_planner/distributed_physical_optimizer_rule.rs
gabotechs: yeah, you are right here, I'll work on this

gabotechs: Done!

PR #300, opened 2026-01-19
Misc improvements for benchmarks

gabotechs:
This PR improves DX for working with benchmarks in this project. It does several things:

- Allows storing individual query results in individual files, so that running just a partial set of queries stores some results that can co-live with previous runs for other queries.
- Adds comparison commands, so that we can compare with previous runs without necessarily rerunning the benchmarks.
- Stores results per branch (or per engine in the remote benchmarks). That way, we do not need to rerun the base branch just so that we can compare to it. This is essential when running big benchmarks in remote mode, as they take a lot of time, and we don't want to be running them over an over just so that we can compare them.
- Refactors the Trino deployment to be more reliable and configurable.
- Adds a `built` dependency so that we have compile time information that can be served by a remote worker (commit hash, build time, etc...)

All of this will make it easier to work with benchmarks in the future.

### DIFF
--- a/benchmarks/cdk/bin/datafusion-bench.ts
+++ b/benchmarks/cdk/bin/datafusion-bench.ts
@@ -17,6 +17,7 @@ async function main() {
         .option('--batch-size <number>', 'Standard Batch coalescing size (number of rows)', '32768')
         .option('--shuffle-batch-size <number>', 'Shuffle batch coalescing size (number of rows)', '32768')
         .option('--children-isolator-unions <number>', 'Use children isolator unions', 'true')
+        .option('--broadcast-joins <number>', 'Use broadcast joins', 'false')
         .option('--collect-metrics <boolean>', 'Propagates metric collection', 'true')
         .option('--compression <string>', 'Compression algo to use within workers (lz4, zstd, none)', 'lz4')
         .option('--queries <string>', 'Specific queries to run', undefined)
@@ -34,6 +35,7 @@ async function main() {
     const queries = options.queries?.split(",") ?? []
     const collectMetrics = options.collectMetrics === 'true' || options.collectMetrics === 1
     const childrenIsolatorUnions = options.childrenIsolatorUnions === 'true' || options.childrenIsolatorUnions === 1
+    const broadcastJoins = options.broadcastJoins === 'true' || options.broadcastJoins === 1
 
     const runner = new DataFusionRunner({
         filesPerTask,
@@ -42,7 +44,8 @@ async function main() {
         shuffleBatchSize,
         collectMetrics,
         childrenIsolatorUnions,
-        compression
+        compression,
+        broadcastJoins
     });
 
     await runBenchmark(runner, {
@@ -70,6 +73,7 @@ class DataFusionRunner implements BenchmarkRunner {
         collectMetrics: boolean;
         compression: string;
         childrenIsolatorUnions: boolean;
+        broadcastJoins: boolean;
     }) {
     }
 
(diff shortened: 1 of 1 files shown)


benchmarks/cdk/bin/@results.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    compare(prevQuery: BenchResult): void {
+        const prevErr = prevQuery.iterations.find(v => v.error)?.error;
+        const newErr = this.iterations.find(v => v.error)?.error;
+
+        if (prevErr && !newErr) {
+            console.log(`${this.id}: Previously failed, but now succeeded 🟠`);
+            return;
+        }
```
gene-bordegaray, line 125: Like these additions 👍

benchmarks/cdk/bin/cdk.ts
```diff
@@ -3,14 +3,16 @@ import * as cdk from 'aws-cdk-lib/core';
 import {CdkStack} from '../lib/cdk-stack';
 import {DATAFUSION_DISTRIBUTED_ENGINE} from "../lib/datafusion-distributed";
 import {TRINO_ENGINE} from "../lib/trino";
+import {SPARK_ENGINE} from "../lib/spark";
 
 const app = new cdk.App();
 
 const config = {
-    instanceType: 't3.xlarge',
+    instanceType: 't3.2xlarge',
```
gene-bordegaray, line 11: what was motivation for the upgrade? reducing unpredictable results from benches?

found "AWS Spot Instances" https://aws.amazon.com/ec2/spot/. I don't know much about ec2 but might be useful to save cost if that's a concern

benchmarks/cdk/bin/datafusion-bench.ts
```diff
@@ -10,11 +10,13 @@ async function main() {
...
-        .option('--cardinality-task-sf <number>', 'Cardinality task scale factor', '2')
-        .option('--shuffle-batch-size <number>', 'Shuffle batch coalescing size (number of rows)', '8192')
+        .option('--files-per-task <number>', 'Files per task', '8')
+        .option('--cardinality-task-sf <number>', 'Cardinality task scale factor', '1')
+        .option('--batch-size <number>', 'Standard Batch coalescing size (number of rows)', '32768')
+        .option('--shuffle-batch-size <number>', 'Shuffle batch coalescing size (number of rows)', '32768')
+        .option('--children-isolator-unions <number>', 'Use children isolator unions', 'true')
         .option('--collect-metrics <boolean>', 'Propagates metric collection', 'true')
         .option('--compression <string>', 'Compression algo to use within workers (lz4, zstd, none)', 'lz4')
         .option('--queries <string>', 'Specific queries to run', undefined)
```
gene-bordegaray, line 23: option for broadcasting? I see it in the worker but think this should be option here.

benchmarks/cdk/bin/cdk.ts
gabotechs: The main reason for upgrading is that otherwise Trino just OOMs and crashes pretty badly with the `t3.xlarge` machines on TPC-H 100.

With these new benchmarks, I'm starting to come up with some numbers for Trino on TPC-H, but it's too resource hungry, and this is the bare minimum (some queries do not even work still, like TPCH-H Q21, Trino still just crashes on this one)

benchmarks/cdk/bin/datafusion-bench.ts
gabotechs: yes, let me add that

gabotechs: Done!

PR #301, opened 2026-01-19
Allow sending compressed data over the wire

gabotechs:
This PR uses the option in `arrow-ipc` to compress the data in either LZ4 or ZSTD format for a more efficient data transfer between workers.

This is configurable by the `distributed.compression` config option, that can take the following values:
- `lz4`
- `zstd`
- `none`

The new default is `lz4` as it has almost 0 impact in decompression performance, while `zstd` does a x1.5 in latency in the local benchmarks.

---

TODO: Add some benchmark results

### DIFF
--- a/src/config_extension_ext.rs
+++ b/src/config_extension_ext.rs
@@ -1,6 +1,5 @@
 use datafusion::common::{DataFusionError, internal_datafusion_err};
 use datafusion::config::ConfigExtension;
-use datafusion::execution::TaskContext;
 use datafusion::prelude::SessionConfig;
 use http::{HeaderMap, HeaderName};
 use std::error::Error;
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -21,6 +21,7 @@ datafusion = { workspace = true, features = [
 datafusion-proto = { workspace = true }
 arrow-flight = "57.1.0"
 arrow-select = "57.1.0"
+arrow-ipc = { version = "57.1.0", features = ["zstd"] }
```
gabotechs, line 24: Need to double check if it's fine to always enable this feature, VS hiding it behind a feature flag.

jayshrivastava (approved):
LGTM. Let's file an issue to run benchmarks.

Sadly compression means we miss out on the zero copy nature of arrow flight, right? I wonder how useful it even is given that we have other forms of compression like dictionary encoding etc.

gabotechs:
>  In the old metrics stack, we use lz4 between workers, so I would be interested in running df metrics queries with this setting enabled.

LZ4 for inter-worker comms is what I've empirically found to be the best, so +1 to that.

PR #256, opened 2025-12-17
Distributed hive-style join testing with superset satisfaction

JSOD11:
Add new test data and testing showing distributed datafusion's ability to leverage [recent additions to upstream datafusion](https://github.com/gene-bordegaray/datafusion/pull/3) to register and utilize hive-style partitioned data sources to generate optimal distributed plans avoiding data-shuffling repartitions and using partitioned hash joins.

Also added a tool for using the datafusion cli to regenerate parquet files from easy-to-read csv files.

This PR is based off of the single node superset satisfaction datafusion PR linked above. Once that PR is merged upstream, I'll switch the datafusion base in this PR's Cargo.toml to reference the newer datafusion version.

PR #303, opened 2026-01-22
Clippy fixes

JSOD11:
In https://github.com/datafusion-contrib/datafusion-distributed/pull/256, we noticed that some clippy warnings were not caught by CI. I've taken a look, and I believe the relevant line is [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/deb53e95379d47d91a4ce934a40a1795f77ab338/.github/workflows/ci.yml#L33):

```
cargo clippy --all-targets --features integration
```

If we want the CI to fail on a clippy warning like for [this print statement](https://github.com/datafusion-contrib/datafusion-distributed/pull/256#discussion_r2715868357), seems like we have to change it to

```
cargo clippy --all-targets --features integration -- -D warnings
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: ./.github/actions/setup
         with:
           components: clippy
-      - run: cargo clippy --all-targets --features integration -- -D warnings
+      - run: cargo clippy --all-targets --all-features -- -D warnings
 
   unit-test:
     runs-on: ubuntu-latest


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: ./.github/actions/setup
         with:
           components: clippy
-      - run: cargo clippy --all-targets --features integration
+      - run: cargo clippy --all-targets --features integration -- -D warnings
```
gabotechs: 🤔 I'm realizing something... I think this might not be getting applied to any of the tests behind the `tpch`, `tpcds` and `clickbench` features.

gabotechs: Maybe we need to say something like `--all-features`?

JSOD11: Yeah, from what I can tell clippy right now is only being applied to the `integration` feature.

Upstream, clippy in CI [here](https://github.com/apache/datafusion/blob/f0e38df39e13921ae19e79d26760c6554466955c/.github/workflows/rust.yml#L637) looks like it runs [this script](https://github.com/apache/datafusion/blob/main/ci/scripts/rust_clippy.sh), which ends up running

```
cargo clippy --all-targets --workspace --features avro,integration-tests,extended_tests -- -D warnings
```

so the `-- -D warnings` is definitely there, and it looks like they manually list out the features rather than using `--all-features`, presumably because there are so many features it could slow down CI. For us, I think `--all-features` could be a good idea. Thoughts?

gabotechs: 👍 yeap! lets go with that

PR #305, opened 2026-01-22
Support OutputBytes, OutputBatches, PruningMetrics, and Ratio metrics

### LINKED ISSUE #237: Support transferring OutputByte metrics
Address this todo https://github.com/datafusion-contrib/datafusion-distributed/pull/236/files#diff-9be3fccf02d9d99c9431fc94e1bf32e70c96676a11ca28e2ee6567701a842c77R292

jayshrivastava:
Add proto serialization support for four additional DataFusion metric types that were previously filtered out during metrics collection. This enables these metrics to be transferred across network boundaries in distributed query execution and displayed in explain analyze output.

Closes #237

### DIFF
--- a/src/metrics/proto.rs
+++ b/src/metrics/proto.rs
@@ -3,18 +3,22 @@ use datafusion::common::internal_err;
 use datafusion::error::DataFusionError;
 use datafusion::physical_plan::metrics::{Count, Gauge, Label, Time, Timestamp};
 use datafusion::physical_plan::metrics::{Metric, MetricValue, MetricsSet};
+use datafusion::physical_plan::metrics::{PruningMetrics as DfPruningMetrics, RatioMetrics};
 use std::borrow::Cow;
 use std::sync::Arc;
 
 /// A MetricProto is a protobuf mirror of [datafusion::physical_plan::metrics::Metric].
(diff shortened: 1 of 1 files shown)


src/metrics/proto.rs
```diff
@@ -3,18 +3,22 @@ use datafusion::common::internal_err;
...
+        oneof = "MetricValueProto",
+        tags = "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15"
+    )]
     // This field is *always* set. It is marked optional due to protobuf "oneof" requirements.
     pub metric: Option<MetricValueProto>,
-    #[prost(message, repeated, tag = "12")]
+    #[prost(message, repeated, tag = "16")]
     pub labels: Vec<ProtoLabel>,
-    #[prost(uint64, optional, tag = "13")]
+    #[prost(uint64, optional, tag = "17")]
```
gabotechs, line 21: For now it's completely fine, but at some point we'll need to keep in mind backwards compatibility. Completely fine now as we are not guaranteeing that changes are backwards compatible.

jayshrivastava: 👍🏽

PR #312, opened 2026-01-26
Coalesce partitions above and below network coalesce

gabotechs:
Rework of https://github.com/datafusion-contrib/datafusion-distributed/pull/285.

---

Coalesces partitions before and after a network boundary, instead of just after it:

This:
```
        │     CoalescePartitionsExec
        │       [Stage 1] => NetworkCoalesceExec: output_partitions=6, input_tasks=3
          ┌───── Stage 1 ── Tasks: t0:[p0..p1] t1:[p2..p3] t2:[p4..p5] 
          │ AggregateExec: mode=Partial, gby=[], aggr=[count(Int64(1))]
          │   ProjectionExec: expr=[]
```
Becomes this:

```
        │     CoalescePartitionsExec
        │       [Stage 1] => NetworkCoalesceExec: output_partitions=3, input_tasks=3
          ┌───── Stage 1 ── Tasks: t0:[p0] t1:[p1] t2:[p2] 
          │ CoalescePartitionsExec
          │   AggregateExec: mode=Partial, gby=[], aggr=[count(Int64(1))]
          │     ProjectionExec: expr=[]
```

This allows us to:
1. Simplify the network communication between the boundaries
2. Do some early work in a distributed manner for when coalescing implies ordering with `SortMergePreservingExec`

### DIFF
--- a/src/distributed_planner/coalesce_partitions_below_network_coalesce.rs
+++ b/src/distributed_planner/coalesce_partitions_below_network_coalesce.rs
@@ -1,0 +1,165 @@
+use crate::NetworkCoalesceExec;
+use crate::common::require_one_child;
+use datafusion::common::DataFusionError;
+use datafusion::common::tree_node::{Transformed, TreeNode};
+use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
+use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+use datafusion::physical_plan::{ExecutionPlan, ExecutionPlanProperties};
+use std::sync::Arc;
+
+pub(crate) fn coalesce_partitions_below_network_coalesce(
+    plan: Arc<dyn ExecutionPlan>,
+) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    let result = plan.transform_down(|parent| {
+        let Some(child) = parent.children().pop() else {
+            return Ok(Transformed::no(parent));
+        };
+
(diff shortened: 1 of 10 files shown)


src/distributed_planner/coalesce_partitions_below_network_coalesce.rs
```diff
@@ -0,0 +1,165 @@
...
+            assert!(!grand_child.as_any().is::<SortPreservingMergeExec>());
+
+            Ok(Transformed::no(plan))
+        })
+        .unwrap();
+
+        assert!(at_least_one_coalesce);
+    }
+
+    async fn sql_to_plan(query: &str) -> Arc<dyn ExecutionPlan> {
```
gene-bordegaray, line 147: I think `sql_to_physical_plan` in test_utils/plans.rs can be reused or refactored here

gabotechs: It's a bit different as that one returns the serialized plan, doesn't execute multiple statements, etc...

I think it's worth consolidating all these functions in something a bit more ergonomic. I think using a builder pattern has very good chances of providing a better experience for these kind of functions, I think we should do this, probably in another PR.

I'll open an issue for this

gene-bordegaray: Sounds good, ya I also noticed a little clunkiness too 😅

gabotechs: https://github.com/datafusion-contrib/datafusion-distributed/issues/313

PR #315, opened 2026-01-26
Add fast deploys

gabotechs:
Adds fast deploys for the benchmarking cluster, this takes the iteration time from ~3mins to 30 seconds for deploying changes to all ec2 machines in the cluster.

### DIFF
--- a/benchmarks/cdk/lib/fastDeploy.ts
+++ b/benchmarks/cdk/lib/fastDeploy.ts
@@ -27,6 +27,10 @@ function sleep(ms: number): Promise<void> {
 async function waitForCommand(commandId: string, instanceId: string): Promise<boolean> {
     const maxAttempts = 60;
     const pollInterval = 2000;
+    const SUCCESS = 'Success'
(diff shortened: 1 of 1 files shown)


benchmarks/cdk/lib/fastDeploy.ts
```diff
@@ -0,0 +1,129 @@
...
+    for (let attempt = 0; attempt < maxAttempts; attempt++) {
+        try {
+            const { stdout } = await execAsync(
+                `aws ssm get-command-invocation --command-id "${commandId}" --instance-id "${instanceId}" --query "[Status, StatusDetails]" --output text`
+            );
+            const [status, statusDetails] = stdout.trim().split(/\s+/);
+
+            if (status === 'Success') {
+                return true;
+            } else if (status === 'Failed' || status === 'Cancelled' || status === 'TimedOut') {
```
jayshrivastava: nit: use constants?

gabotechs: Done!

PR #320, opened 2026-01-27
Handle stream cancellations gracefully

gabotechs:
Before, if a stream was abandoned, we were leaking the `TaskData` until the TTLMap decides to clean it up due to a timeout.

In this PR:
- We also remove the `StageKey` from the TTLMap in the event of a drop (not just a graceful completion)
- We are faster at cancelling the background threads that poll data

### DIFF
--- a/src/flight_service/spawn_select_all.rs
+++ b/src/flight_service/spawn_select_all.rs
@@ -115,6 +115,54 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn memory_reservation_backpressure() -> Result<(), Box<dyn Error>> {
+        let pool: Arc<dyn MemoryPool> = Arc::new(UnboundedMemoryPool::default());
+
+        let mut stream = spawn_select_all(
+            vec![futures::stream::iter(vec![
+                Ok::<_, String>(1),
+                Ok(2),
+                Ok(3),
+            ])],
+            Arc::clone(&pool),
+            1,
+        );
+        // First two messages are buffered (1+2)
+        tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
+        let reserved = pool.reserved();
+        assert_eq!(reserved, 3);
+
+        // First message is pulled
+        let n = stream.next().await.unwrap()?;
+        assert_eq!(n, 1);
+
+        // The third message is buffered, but the first one came out (2+3)
+        tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
+        let reserved = pool.reserved();
+        assert_eq!(reserved, 5);
+
+        // Second message is pulled
+        let n = stream.next().await.unwrap()?;
+        assert_eq!(n, 2);
+
+        // Only the third message is buffered
+        tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
+        let reserved = pool.reserved();
+        assert_eq!(reserved, 3);
+
+        // The third message is pulled
+        let n = stream.next().await.unwrap()?;
+        assert_eq!(n, 3);
+
+        // Nothing remains in the pool
+        tokio::time::sleep(tokio::time::Duration::from_millis(1)).await;
+        let reserved = pool.reserved();
+        assert_eq!(reserved, 0);
+
+        Ok(())
+    }
+
     impl MemoryFootPrint for usize {
         fn get_memory_size(&self) -> usize {
             *self


tests/stateful_data_cleanup.rs
```diff
@@ -0,0 +1,103 @@
...
+                tasks_running, 0,
+                "Expected Worker {i} to have 0 tasks running, but got {tasks_running}"
+            )
+        }
+
+        Ok(())
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn no_pending_tasks_if_query_aborts() -> Result<()> {
```
gabotechs, line 40: This test reproduces the issue on `main`

src/flight_service/spawn_select_all.rs
```diff
@@ -26,20 +30,29 @@ where
...
+            loop {
+                // Capture the closed() event as soon as possible. We don't want to do
+                // extra work if we know nobody is going to listen to it.
+                let msg = tokio::select! {
+                    biased;
+                    _ = tx.closed() => return,
+                    msg = t.next() => msg
+                };
+                let Some(msg) = msg else { return };
+
```
gabotechs, line 42: This is not strictly necessary, but it makes the task get dropped earlier, as it no longer needs to wait for an extra message just so that the task can get aborted on `tx.send()`

src/flight_service/worker_connection_pool.rs
```diff
@@ -199,11 +208,20 @@ impl WorkerConnection {
...
+                let msg = tokio::select! {
+                    biased;
+                    _ = cancel.cancelled() => return,
+                    msg = interleaved_stream.next() => {
+                        match msg {
+                            Some(Ok(v)) => v,
+                            Some(Err(err)) => return fanout(&per_partition_tx, err),
+                            None => return, // Stream exhausted
+                        }
+                    }
```
gabotechs, line 222: Same as above. Now the task can be aborted before waiting an extra message that will never see the light.

src/flight_service/do_get.rs
```diff
@@ -182,17 +188,31 @@ impl Worker {
...
+            let stream = on_drop_stream(stream, move || {
+                if !fully_finished_cloned.load(Ordering::SeqCst) {
+                    // If the stream was not fully consumed, but it ws dropped (abandoned), we
+                    // still need to remove the entry from `task_data_entries`, otherwise we
+                    // might leak memory until it gets automatically released due to a TTL.
+                    if num_partitions_remaining.fetch_sub(1, Ordering::SeqCst) == 1 {
+                        task_data_entries.remove(key_clone);
+                    }
+                }
+            });
```
jayshrivastava, line 208: Sounds like we don't need the TTL map anymore

src/flight_service/spawn_select_all.rs
```diff
@@ -4,20 +4,24 @@ use datafusion::common::runtime::SpawnedTask;
...
     pool: Arc<dyn MemoryPool>,
+    queue_size: usize,
 ) -> impl Stream<Item = Result<El, Err>>
 where
     T: Stream<Item = Result<El, Err>> + Send + Unpin + 'static,
     El: MemoryFootPrint + Send + 'static,
     Err: Send + 'static,
 {
-    let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+    let (tx, rx) = tokio::sync::mpsc::channel(queue_size);
```
jayshrivastava, line 24: Worth adding a test which tests backpressure from the queue size IMO. Adding a test to this file I mean

jayshrivastava, line 24: I remember seeing a comment about how we need to use `unbounded_channel(), or else some obscure deadlock would happen. Is that still relevant?

src/flight_service/spawn_select_all.rs
```diff
@@ -26,20 +30,29 @@ where
...
         tasks.push(SpawnedTask::spawn(async move {
-            while let Some(msg) = t.next().await {
+            loop {
+                // Capture the closed() event as soon as possible. We don't want to do
+                // extra work if we know nobody is going to listen to it.
+                let msg = tokio::select! {
+                    biased;
+                    _ = tx.closed() => return,
+                    msg = t.next() => msg
+                };
```
jayshrivastava, line 40: This is so nice. I wish go had that.

src/flight_service/worker_connection_pool.rs
jayshrivastava: 👍🏽

src/flight_service/do_get.rs
gabotechs: Well, I would not be surprised if there are still situations where tasks could leak. For example, if some partition never starts on the first place.

At least with this PR we handle all cases that are in our hands.

src/flight_service/spawn_select_all.rs
gabotechs: good suggestion, doing that now

gabotechs: Yes, that's still relevant, although the code that handles that now is the `worker_connection_pool.rs`, which buffers unboundedly.

That's the reason why we are fine with a bounded queue now.

PR #326, opened 2026-01-28
Gene.bordegaray/2026/01/fix data benchmark data generators

gene-bordegaray:
**Problem 1:**
The rpcds benchmark would only let you generate data if you were in the "./benchmarks" dir. Changed so that you can do it from anywhere

before:
Screenshot 2026-01-28 at 6 49 44 AM

after:
Screenshot 2026-01-28 at 6 51 03 AM

**Problem 2:**
gen-tpch.sh was fialing to convert to parquet because the csv reader expected less columns than there actually was. This was because it did not properly take into account its empty trailing field.

before:
Screenshot 2026-01-28 at 6 57 03 AM

after:
Screenshot 2026-01-28 at 6 55 32 AM

**Problem 3:**
Docs were out of date and added some more info

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -2,10 +2,13 @@
 
 ### Generating Benchmarking data
 
-Generate TPCH data into the `data/` dir
+Generate datasets into `benchmarks/data/`.
 
 ```shell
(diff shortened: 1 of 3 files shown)


gene-bordegaray:
this included my streaming cache work

PR #306, opened 2026-01-22
Add metrics to network boundaries

gabotechs:
Adds some metrics to the network boundaries so that they can be displayed in the plans. They look like this:

```
┌───── DistributedExec ── Tasks: t0:[p0] 
│ ProjectionExec: expr=[count(*)@0 as count(*), RainToday@1 as RainToday], metrics=[output_rows=2, elapsed_compute=1.67µs, output_bytes=48.0 B, output_batches=1, expr_0_eval_time=167ns, expr_1_eval_time=84ns]
│   SortPreservingMergeExec: [count(Int64(1))@2 ASC NULLS LAST], metrics=[output_rows=2, elapsed_compute=32.29µs, output_bytes=64.0 B, output_batches=1]
│     [Stage 2] => NetworkCoalesceExec: output_partitions=6, input_tasks=2, metrics=[elapsed_compute=1.34ms, bytes_transferred=19.28 K, max_mem_used=1.96 K]
└──────────────────────────────────────────────────
  ┌───── Stage 2 ── Tasks: t0:[p0..p2] t1:[p0..p2] 
  │ SortExec: expr=[count(*)@0 ASC NULLS LAST], preserve_partitioning=[true], metrics=[output_rows=2, elapsed_compute=18.71µs, spill_count=0, spilled_bytes=0.0 B, spilled_rows=0]
  │   ProjectionExec: expr=[count(Int64(1))@1 as count(*), RainToday@0 as RainToday, count(Int64(1))@1 as count(Int64(1))], metrics=[output_rows=2, elapsed_compute=9.55µs, expr_0_eval_time=421ns, expr_1_eval_time=129ns, expr_2_eval_time=128ns]
  │     AggregateExec: mode=FinalPartitioned, gby=[RainToday@0 as RainToday], aggr=[count(Int64(1))], metrics=[output_rows=2, elapsed_compute=208.66µs, spill_count=0, spilled_bytes=0.0 B, spilled_rows=0, peak_mem_used=100.9 K, aggregate_arguments_time=21.21µs, aggregation_time=16.34µs, emitting_time=5.17µs, time_calculating_group_ids=30.05µs]
  │       [Stage 1] => NetworkShuffleExec: output_partitions=3, input_tasks=3, metrics=[elapsed_compute=2.08ms, bytes_transferred=19.72 K, max_mem_used=5.19 K]
  └──────────────────────────────────────────────────
[...]

PR #314, opened 2026-01-26
Fix metrics collection bug: preserve PartitionIsolatorExec in plan

jayshrivastava:
Dupe of https://github.com/datafusion-contrib/datafusion-distributed/pull/281 with minor fixes

- Update test assertions to allow empty metrics for `PartitionIsolatorExec` nodes
- Fix `assert_metrics_present_in_plan` test helper to handle `PartitionIsolatorExec` wrapped in `MetricsWrapperExec`
- Fix `count_plan_nodes_up_to_network_boundary` test helper now that network boundary nodes have metrics

Fixes issue where PartitionIsolatorExec nodes were missing from displayed plans after metrics rewriting. All metrics collection tests now pass.

PR #336, opened 2026-02-02
Fix metrics display on leaf nodes

gabotechs:
I think I messed something up in https://github.com/datafusion-contrib/datafusion-distributed/pull/306 and I see some leaf stages are missing their metrics.

This PR solves it and adds a new test that fails on main but succeeds in this PR.

### DIFF
--- a/src/metrics/task_metrics_rewriter.rs
+++ b/src/metrics/task_metrics_rewriter.rs
@@ -264,7 +264,7 @@ pub fn stage_metrics_rewriter(
 
 #[cfg(test)]
 mod tests {
-    use crate::PartitionIsolatorExec;
+    use crate::Stage;
     use crate::metrics::DISTRIBUTED_DATAFUSION_TASK_ID_LABEL;
     use crate::metrics::proto::{
         MetricsSetProto, df_metrics_set_to_proto, metrics_set_proto_to_df,
@@ -281,7 +281,6 @@ mod tests {
     use crate::test_utils::plans::count_plan_nodes_up_to_network_boundary;
     use crate::test_utils::session_context::register_temp_parquet_table;
     use crate::{DistributedExec, DistributedPhysicalOptimizerRule};
-    use crate::{NetworkBoundaryExt, Stage};
     use bytes::Bytes;
     use datafusion::arrow::array::{Int32Array, StringArray};
     use datafusion::arrow::datatypes::{DataType, Field, Schema};
@@ -556,21 +555,10 @@ mod tests {
 
     // Assert every plan node has at least one metric except partition isolators, network boundary nodes, and the root DistributedExec node.
     fn assert_metrics_present_in_plan(plan: &Arc<dyn ExecutionPlan>) {
-        // Check if this is a PartitionIsolatorExec (possibly wrapped in MetricsWrapperExec)
-        // For MetricsWrapperExec, we check the name() which delegates to the inner plan
-        let is_partition_isolator = plan.name() == PartitionIsolatorExec::static_name();
-
         if let Some(metrics) = plan.metrics() {
-            // PartitionIsolatorExec nodes are allowed to have empty metrics
-            if !is_partition_isolator {
-                assert!(metrics.iter().count() > 0);
-            }
+            assert!(metrics.iter().count() > 0);
         } else {
-            assert!(
-                plan.is_network_boundary()
-                    || is_partition_isolator
-                    || plan.as_any().is::<DistributedExec>()
-            );
+            assert!(plan.as_any().is::<DistributedExec>());
         }
         for child in plan.children() {
             assert_metrics_present_in_plan(child);


src/metrics/task_metrics_rewriter.rs
```diff
@@ -246,18 +246,39 @@ pub fn stage_metrics_rewriter(
...
+            ))?;
+            // Wrap the new network boundary with metrics
+            let wrapped_plan_node: Arc<dyn ExecutionPlan> =
+                Arc::new(MetricsWrapperExec::new(new_network_boundary, stage_metrics));
+            return Ok(Transformed::new(
+                wrapped_plan_node,
+                true,
+                TreeNodeRecursion::Jump,
+            ));
+        }
```
jayshrivastava: Hmm I don't think this is right - we shouldn't have to traverse into the stage plan. `rewrite_distributed_plan_with_metrics` already does that...
```
let stage = network_boundary.input_stage();
// This transform is a bit inefficient because we traverse the plan nodes twice
// For now, we are okay with trading off performance for simplicity.
let plan_with_metrics =
    stage_metrics_rewriter(stage, metrics_collection.clone(), format)?;
let network_boundary = network_boundary.with_input_stage(Stage::new(
    stage.query_id,
    stage.num,
    plan_with_metrics,
    stage.tasks.len(),
))?;
let network_boundary =
    MetricsWrapperExec::new(network_boundary, plan.metrics().unwrap_or_default());
return Ok(Transformed::yes(Arc::new(network_boundary)));
```

Maybe something else is going on. Like `MetricsWrapperExec` doesn't delegate `as_network_boundary()` or something.

I'm also more suspicious about this metrics collection system now that the tests asserting the actually output rows are gone. It's very easy to get wrong. The rewriter traversals code should align 1:1 with the collection traversals. They have diverged it seems. Thoughts on adding back the tests? IIRC, they only asserted output_rows.

gabotechs: damn... you are right

gabotechs: Ok, let me try again

src/execution_plans/metrics.rs
```diff
@@ -53,22 +46,17 @@ impl ExecutionPlan for MetricsWrapperExec {
...
 
     fn with_new_children(
         self: Arc<Self>,
         children: Vec<Arc<dyn ExecutionPlan>>,
     ) -> Result<Arc<dyn ExecutionPlan>> {
         Ok(Arc::new(MetricsWrapperExec {
-            inner: self.inner.clone(),
+            inner: Arc::clone(&self.inner).with_new_children(children.clone())?,
             metrics: self.metrics.clone(),
-            children: Some(children),
```
gabotechs, line 71: The problem I found was that, when this method was called, the `self.children` where being updated, but not the actual children from `self.inner`.

This is fine as long as you access `MetricsWrapperExec` with the `children()` method, as then, the ones just inject will be returned, but it's not fine if you access the children through `input_stage()`, which is what was happening in `display_plan_ascii`.

I think it should be fine with just removing this `self.children` in favor of just applying the new children to `self.inner` directly, but I may be missing something.

WDYT @jayshrivastava?

src/metrics/task_metrics_rewriter.rs
gabotechs: Last time I let Claude do the homework for me. Thanks for catching this!

tests/metrics_collection.rs
```diff
@@ -163,6 +165,31 @@ mod tests {
...
+        d_physical = rewrite_distributed_plan_with_metrics(d_physical.clone(), format)?;
+
+        let display =
+            DisplayableExecutionPlan::with_metrics(d_physical.children().swap_remove(0).as_ref())
+                .indent(true)
+                .to_string();
+        assert_not_contains!(display, "metrics=[]");
+
+        let display = display_plan_ascii(d_physical.as_ref(), true);
+        assert_not_contains!(display, "metrics=[]");
```
jayshrivastava, line 188: Might as well use the node_metrics helper. You can also assert the exact output rows since that's very  deterministic (or at least, assert they are equal).

src/execution_plans/network_broadcast.rs
```diff
@@ -263,4 +264,8 @@ impl ExecutionPlan for NetworkBroadcastExec {
             futures::stream::select_all(streams),
         )))
     }
+
+    fn metrics(&self) -> Option<MetricsSet> {
+        Some(self.worker_connections.metrics.clone_inner())
+    }
```
jayshrivastava, line 270: Update the other network boundary nodes too? Like shuffle, coalese.

src/execution_plans/partition_isolator.rs
```diff
@@ -169,13 +176,22 @@ impl ExecutionPlan for PartitionIsolatorExec {
                     Ok(Box::pin(EmptyRecordBatchStream::new(self.input.schema()))
                         as SendableRecordBatchStream)
                 } else {
-                    self.input.execute(*actual_partition_number, context)
+                    Ok(Box::pin(RecordBatchStreamAdapter::new(
+                        self.schema(),
+                        self.input
+                            .execute(*actual_partition_number, context)?
+                            .inspect_ok(move |v| metric.add(v.num_rows())),
+                    )))
```
jayshrivastava, line 184: There's some assertions in https://github.com/datafusion-contrib/datafusion-distributed/pull/314 which are made weaker to skip over PartitionIsolatorExec nodes. Do you think you can revisit those assertions?

tests/metrics_collection.rs
jayshrivastava, line 234: It sounds like we missed this bug because we only call this for DataSourceExec. Want to update the tests in this file to call this for `Network*Exec`?

jayshrivastava (changes requested):

src/execution_plans/metrics.rs
jayshrivastava: This sounds okay. 

I forget why `with_new_children` is even called on `MetricsWrapperExec`. It means we're transforming a node after its wrapped...

I made this change intentionally though - we used to error in this method and I remember updating it so we don't error. Might be in the git blame.

src/execution_plans/network_broadcast.rs
gabotechs: It's already in the other ones, it was missing in this one because this two PRs where merged roughly at the same time:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/306
- https://github.com/datafusion-contrib/datafusion-distributed/pull/279

tests/metrics_collection.rs
gabotechs: Not really, that would also not have caught the bug. The bug was only appearing in the `display_plan_ascii` (it was purely a visualization bug) because there is the only place we access a network boundary's children through `input_stage()` rather than through `children()`.

gabotechs: In fact, even if you try to render the metrics with `DisplayablePlanWithMetrics::new(plan).indent(true)`, it worked correctly, so the metrics have always been there, it was really only during visualization through `display_plan_ascii`

tests/metrics_collection.rs
gabotechs: The bug was specific to the `display_plan_ascii` visualization, the metrics have always been placed correctly, so accessing the metrics programatically was succeeding. This is why we were not able to catch the bug before.

src/execution_plans/partition_isolator.rs
gabotechs: Yes! done

src/execution_plans/metrics.rs
gabotechs: yeah, but if we want `input_stage()` from a network boundary to return the correct thing... we need to transform it if we are expecting it to return the wrapped children rather than the original one.

If we don't transform the original node with the new wrapped children, anytime we call input_stage() we will be returning the original children, not the wrapped ones.

gabotechs: In https://github.com/datafusion-contrib/datafusion-distributed/blob/838bf7afa08e30bd076879a10d4f3316af559280/src/metrics/task_metrics_rewriter.rs#L67-L67, we call `transform_down`, which means that if we return `Transformed::yes` inside the function (which we do), then DataFusion will call `with_new_children` under the hood pretty much on every node.

But this should be fine and expected, why would we error in this situation? I see that just removing `children` from `MetricsWrarpperExec` has no negative implications.

PR #337, opened 2026-02-02
Claude Skill for benchmarks

gabotechs:
This PR adds several improvements to the benchmarks:
- Claude Skills: adds some Claude Code skills for dealing with remote benchmarks. Typically these benchmarks take a while to execute, and analyzing the results can be a bit time consuming. This PR allows Claude to autonomously do all the tedious tasks of running benchmarks, waiting for the result, analyzing them, performing small changes, and redeploy for benchmark again.
- Adds ballista back: adds ballista benchmarks back so that we have another engine to compare to. This time, ballista is added in a cargo crate detached from the workspace, so users do not need to care about importing it. All the ballista code is pretty much what was removed from https://github.com/datafusion-contrib/datafusion-distributed/pull/292
- Stores the execution plans with metrics attached to the benchmarks run. This way, Claude can run benchmarks and analyze the execution plans along with the results in order to identify bottlenecks and improvement points.


.claude/skills/ec2-cluster-provision/SKILL.md
```diff
@@ -0,0 +1,62 @@
+---
+name: ec2-cluster-provision
+description: uses the code present in this repository for provision an EC2 cluster for benchmarking purposes
+---
+
+This project uses a remote benchmarks EC2 cluster constructed with AWS CDK located at `benchmarks/cdk`.
+
+There's a package.json file in `benchmarks/cdk/package.json` with relevant commands about deploying.
+
+All the commands in this skill need to be prefixed with whatever the user declared in `./claude/settings.local.json`
```
fmonjalet, line 10: Did you mean this?
```suggestion
All the commands in this skill need to be prefixed with whatever the user declared in `~/.claude/settings.local.json`
```

.claude/skills/remote-benchmark/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+The --dataset command is mandatory, and its value can be any of the folder names in `benchmarks/data`, for example:
+clickbench_0-100, tpcds_sf1, tpch_sf1, tpch_sf10 or tpch_sf100.
+
+Also, the --queries argument can be used for executing just a partial subset of queries, for example:
+```shell
+--queries q1,q2,q3
+```
+
+When benchmarking a very specific feature, it's convenient to choose wisely a relevant query and just execute that one.
```
fmonjalet, line 64: minor: it may be helpful to explain what "wisely" is. Something like:

> When benchmarking a specific feature, select one or a few query that relies on this feature and execute just this one.

fmonjalet (commented):
Sorry for the very late review. I only reviewed `SKILL.md` files and only had minor comments.
I was wondering: since these are very scoped tasks where you're only interested by the output, would they make sense as agents? I am still not 100% clear on skill vs agent.

.claude/skills/ec2-cluster-provision/SKILL.md
gabotechs: It should actually be .claude/settings.local.json (the one local to this project). For some reason has always been smart enough to go the the appropriate place

Issue #187, opened 2025-10-15
Determine if limit streams impact metrics propagation


jayshrivastava:
Right now, a task will send metrics from the flight endpoint to the client when the last partition's stream has completed.

Does the stream necessarily complete?
Is it possible that the stream completes but the client has moved on from consuming it?

Metrics propagation relies on the stream completing and the client waiting for the last message which contains metrics.

We should look into if adding a limit on the client side causes issues with this scheme.

PR #341, opened 2026-02-06
test: reproduce metrics loss on early stream termination

gferrate:
Adds a test reproducing the issue described in [#187](https://github.com/datafusion-contrib/datafusion-distributed/issues/187).

Since the test fails by design, it is marked as `#[ignore]`.

The underlying issue is that metrics are attached to the last `FlightData` message of the last partition stream on the worker side (in `do_get.rs`). When a `LIMIT` is present, the client-side network boundary node (e.g. `NetworkShuffleExec`) drops the stream once it has enough rows, before the worker finishes all partitions. The final message carrying the metrics is never sent or never consumed, so metrics for that stage are lost.

The test runs a `GROUP BY ... LIMIT 1` query against the `flights_1m` dataset (1M rows). The large dataset ensures the worker is still producing data when the LIMIT causes the client to drop the stream.

Note: This don't actually solve the issue, just showcases that it is really an issue!

### DIFF
--- a/src/metrics/task_metrics_collector.rs
+++ b/src/metrics/task_metrics_collector.rs
@@ -344,6 +344,71 @@ mod tests {
         run_metrics_collection_e2e_test("SELECT distinct company from table2").await;
     }
 
+    /// Tests whether metrics are lost when a LIMIT causes early stream termination.
+    ///
+    /// Issue: https://github.com/datafusion-contrib/datafusion-distributed/issues/187
+    ///
+    /// Metrics are piggybacked on the last FlightData message of the last partition stream
+    /// (see `do_get.rs`). If a LIMIT causes the client-side stream to be dropped before the
+    /// worker finishes, the last message (carrying metrics) is never received.
+    ///
+    /// This uses the `flights_1m` dataset (1M rows) so the worker is still producing data
+    /// when the LIMIT causes the client to drop the stream.
+    ///
+    /// This test is ignored because it demonstrates a known issue: metrics are lost when the
+    /// client drops the stream early. Un-ignore it to reproduce the issue or to verify a fix.
+    #[tokio::test]
+    #[ignore]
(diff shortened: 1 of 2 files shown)


src/metrics/task_metrics_collector.rs
```diff
@@ -344,6 +344,71 @@ mod tests {
...
+        for expected_stage_key in expected_stage_keys {
+            let actual_metrics = result
+                .input_task_metrics
+                .get(&expected_stage_key)
+                .unwrap_or_else(|| {
+                    panic!(
+                        "Missing metrics for stage key {expected_stage_key:?}. \
+                         The LIMIT caused the stream to be dropped before the worker \
+                         sent the last FlightData message with metrics."
+                    )
```
gferrate, line 399: ℹ️ This is the bug. Uncomment `#[ignore]` to reproduce.

src/metrics/task_metrics_collector.rs
```diff
@@ -344,6 +344,71 @@ mod tests {
...
+    /// worker finishes, the last message (carrying metrics) is never received.
+    ///
+    /// This uses the `flights_1m` dataset (1M rows) so the worker is still producing data
+    /// when the LIMIT causes the client to drop the stream.
+    ///
+    /// This test is ignored because it demonstrates a known issue: metrics are lost when the
+    /// client drops the stream early. Un-ignore it to reproduce the issue or to verify a fix.
+    #[tokio::test]
+    #[ignore]
+    async fn test_metrics_collection_with_limit_causing_early_stream_termination() {
```
gabotechs, line 362: I've seen that another consequence of this is that calling `rewrite_distributed_plan_with_metrics` on the plan fails with an error that says something like "not enough metrics provided to rewrite task".

This is expected an another way of reproducing the issue. I wonder if `rewrite_distributed_plan_with_metrics` should be error proof instead?

cc @jayshrivastava

gabotechs (approved):
💯 nice

PR #342, opened 2026-02-06
Broadcast NestedLoopJoinExec

gabotechs:
Also broadcasts NestedLoopJoinExec


src/distributed_planner/insert_broadcast.rs
```diff
@@ -148,30 +149,15 @@ pub(super) fn insert_broadcast_execs(
...
-        // Otherwise, use the build child directly (DataSourceExec)
-        let broadcast_input = if let Some(coalesce) = build_child
-            .as_any()
-            .downcast_ref::<CoalescePartitionsExec>()
-        {
-            Arc::clone(coalesce.input())
-        } else {
-            Arc::clone(build_child)
-        };
-
```
gabotechs, line 161: There is one challenge with this kind of joins. There can be situations where what's immediately below the join node is not a `CoalescePartitionsExec`, it might be something like:

https://github.com/datafusion-contrib/datafusion-distributed/blob/692fa473abfe9d362fc6e7378da3270e99405c3b/tests/tpcds_plans_test.rs#L1465-L1469

This means that there might be a whole chain of nodes above the `CoalescePartitionExec` that need to be accounted for, and that ideally, they run below the `BroadcastExec` so that the compute happens only once.

src/distributed_planner/plan_annotator.rs
```diff
@@ -264,14 +272,10 @@ fn _annotate_plan(
...
-                task_count,
-            };
-        } else {
+        // If the node above the CoalescePartitionsExec or SortPreservingMergeExec is a
+        // BroadcastExec, then we don't need to insert a network boundary, because we are in the
+        // situation where a broadcast join will happen.
+        if !parent_stack.iter().rev().take(2).any(|plan| plan.as_any().is::<BroadcastExec>()) {
             annotation = AnnotatedPlan {
                 plan_or_nb: PlanOrNetworkBoundary::Coalesce,
                 children: vec![annotation],
```
gabotechs: Following the logic from the above comment, it's only fine to skip the coalesce boundary if the parent of the CoalescePartitionExec is a BroadcastExec, but if there is any compute in between the `BroadcastExec` and the `CoalescePartitionExec`, we'll need to follow the rules and inject a Coalesce boundary as normal.

src/distributed_planner/plan_annotator.rs
```diff
@@ -264,14 +272,15 @@ fn _annotate_plan(
...
     {
-        // A BroadcastExec underneath a coalesce parent means the build side will cross stages.
-        if plan.as_any().is::<BroadcastExec>() {
-            annotation = AnnotatedPlan {
-                plan_or_nb: PlanOrNetworkBoundary::Broadcast,
-                children: vec![annotation],
-                task_count,
-            };
-        } else {
+        // If the node above the CoalescePartitionsExec or SortPreservingMergeExec is a
```
gene-bordegaray, line 273: Isnt this problematic for a SPM?

If we don't coalesce them then we have tasks with sorted partitions on the task level but not across all tasks before broadcasting

gene-bordegaray (commented):
I think with the SPM logic we would be broadcasting each task's local sorted stream, not the global

src/distributed_planner/plan_annotator.rs
gabotechs: 🤔 could be? I think it's safer to just not do this at all if this is a SPM

gabotechs: Done

gabotechs:
No longer working on this. Closing for now

PR #344, opened 2026-02-13
Measure latency server-side in benchmarks

gabotechs:
Latency measured was taking into account client-side retrieval through the port forward to the benchmark cluster.

This change reports the latency measured server-side for all the engines in the benchmarks.

Specially for smaller datasets, the latency introduced by the port-forward network hop to our local machines is significant, and also a bit unreliable, so not taking it into account gives more stable benchmarks.

### DIFF
--- a/benchmarks/cdk/ballista/src/ballista_http.rs
+++ b/benchmarks/cdk/ballista/src/ballista_http.rs
@@ -22,7 +22,7 @@ use url::Url;
 struct QueryResult {
     plan: String,
     count: usize,
-    elapsed_ms: usize,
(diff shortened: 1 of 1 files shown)


benchmarks/cdk/ballista/src/ballista_http.rs
```diff
@@ -105,7 +106,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
                     let ms = elapsed.as_secs_f64() * 1000.0;
```
gene-bordegaray, line 106: this is f64 but elapsed_ms is usize

PR #345, opened 2026-02-13
Send subplans in the coordinator stage rather than in NetworkBoundary nodes

gabotechs:
This PR optimizes the way plan is feed to the different workers.

## Before

The plan was propagated recursively through the network boundaries, and each stage is responsible for feeding the plan to the next one.

While this makes the code simple, it puts pressure on the network boundaries for dealing with slices of the plan they do not care about, and introduces a cascading problem in which stages depend on the previous one to send the plan for starting.

## Now

The "coordinator stage" is responsible for contacting in parallel all the workers involved in the query, feeding them just the slice of the plan they care about.

This puts more pressure in the coordinating stage, but this is not too bad, because that stage typically does very little work.

## Benchmarks

There does not seem to be any impact in benchmarks. This is the overhead the head `DistributedExec` shows in TPC-H benchmarks:

```
q1: plan_bytes_sent_0=88.79 K, plan_send_latency_avg_0=3.07ms, plan_send_latency_max_0=4.22ms
q2: plan_bytes_sent_0=111.6 K, plan_send_latency_avg_0=2.74ms, plan_send_latency_max_0=4.63ms
q3: plan_bytes_sent_0=155.9 K, plan_send_latency_avg_0=2.81ms, plan_send_latency_max_0=4.35ms
q4: plan_bytes_sent_0=128.1 K, plan_send_latency_avg_0=2.95ms, plan_send_latency_max_0=4.27ms
q5: plan_bytes_sent_0=161.4 K, plan_send_latency_avg_0=2.83ms, plan_send_latency_max_0=5.30ms
q6: plan_bytes_sent_0=78.43 K, plan_send_latency_avg_0=3.03ms, plan_send_latency_max_0=3.83ms
q7: plan_bytes_sent_0=164.1 K, plan_send_latency_avg_0=3.29ms, plan_send_latency_max_0=6.44ms
q8: plan_bytes_sent_0=191.7 K, plan_send_latency_avg_0=3.75ms, plan_send_latency_max_0=7.09ms
q9: plan_bytes_sent_0=196.0 K, plan_send_latency_avg_0=2.71ms, plan_send_latency_max_0=3.94ms
q10: plan_bytes_sent_0=162.0 K, plan_send_latency_avg_0=3.17ms, plan_send_latency_max_0=4.25ms
q11 plan_bytes_sent_0=79.11 K, plan_send_latency_avg_0=3.05ms, plan_send_latency_max_0=4.31ms
q12 plan_bytes_sent_0=135.4 K, plan_send_latency_avg_0=3.36ms, plan_send_latency_max_0=5.02ms
q13 plan_bytes_sent_0=76.59 K, plan_send_latency_avg_0=2.32ms, plan_send_latency_max_0=3.02ms
q14 plan_bytes_sent_0=104.2 K, plan_send_latency_avg_0=2.47ms, plan_send_latency_max_0=3.56ms
q15 plan_bytes_sent_0=162.9 K, plan_send_latency_avg_0=3.12ms, plan_send_latency_max_0=5.95ms
q16 plan_bytes_sent_0=67.38 K, plan_send_latency_avg_0=2.45ms, plan_send_latency_max_0=3.47ms
q17 plan_bytes_sent_0=181.3 K, plan_send_latency_avg_0=4.19ms, plan_send_latency_max_0=6.19ms
q18 plan_bytes_sent_0=230.7 K, plan_send_latency_avg_0=4.42ms, plan_send_latency_max_0=6.69ms
q19 plan_bytes_sent_0=110.9 K, plan_send_latency_avg_0=3.09ms, plan_send_latency_max_0=4.03ms
q20 plan_bytes_sent_0=147.0 K, plan_send_latency_avg_0=2.74ms, plan_send_latency_max_0=3.79ms
q21 plan_bytes_sent_0=129.9 K, plan_send_latency_avg_0=3.12ms, plan_send_latency_max_0=4.37ms
q22 plan_bytes_sent_0=51.39 K, plan_send_latency_avg_0=2.35ms, plan_send_latency_max_0=2.66ms
```

Knowing that the actual data moved in TPC-H is 1Gb for SF1 and 10Gb for SF2, it makes sense that the performance impact of this is negligible.

For properly benchmarking this, there should be some benchmarks for which the size of the plans is more or less the same as the size of the data returned.

<details><summary>TPC-H SF1 TOTAL: prev=20275 ms, new=20759 ms, diff=1.02 slower</summary>

```
      q1: prev= 343 ms, new= 234 ms, diff=1.47 faster ✅
      q2: prev= 323 ms, new= 378 ms, diff=1.17 slower ✖
      q3: prev= 532 ms, new= 523 ms, diff=1.02 faster ✔
      q4: prev= 320 ms, new= 346 ms, diff=1.08 slower ✖
      q5: prev= 397 ms, new= 246 ms, diff=1.61 faster ✅
      q6: prev= 297 ms, new= 156 ms, diff=1.90 faster ✅
      q7: prev= 381 ms, new= 352 ms, diff=1.08 faster ✔
      q8: prev= 339 ms, new= 398 ms, diff=1.17 slower ✖
      q9: prev= 318 ms, new= 403 ms, diff=1.27 slower ❌
     q10: prev= 227 ms, new= 289 ms, diff=1.27 slower ❌
     q11: prev= 256 ms, new= 266 ms, diff=1.04 slower ✖
     q12: prev= 192 ms, new= 257 ms, diff=1.34 slower ❌
     q13: prev= 162 ms, new= 202 ms, diff=1.25 slower ❌
     q14: prev= 198 ms, new= 206 ms, diff=1.04 slower ✖
     q15: prev= 185 ms, new= 232 ms, diff=1.25 slower ❌
     q16: prev= 294 ms, new= 319 ms, diff=1.09 slower ✖
     q17: prev= 326 ms, new= 368 ms, diff=1.13 slower ✖
     q18: prev= 270 ms, new= 320 ms, diff=1.19 slower ✖
     q19: prev= 339 ms, new= 355 ms, diff=1.05 slower ✖
     q20: prev= 290 ms, new= 294 ms, diff=1.01 slower ✖
     q21: prev= 544 ms, new= 524 ms, diff=1.04 faster ✔
     q22: prev= 216 ms, new= 240 ms, diff=1.11 slower ✖
   TOTAL: prev=20275.287493999997 ms, new=20759.101284999997 ms, diff=1.02 slower ✖
```

</details>

<details><summary>TPC-H SF10 TOTAL: prev=73574 ms, new=74445 ms, diff=1.01 slower</summary>

```
      q1: prev=1290 ms, new=1320 ms, diff=1.02 slower ✖
      q2: prev= 441 ms, new= 437 ms, diff=1.01 faster ✔
      q3: prev=1040 ms, new= 981 ms, diff=1.06 faster ✔
      q4: prev= 547 ms, new= 499 ms, diff=1.10 faster ✔
      q5: prev=1459 ms, new=1477 ms, diff=1.01 slower ✖
      q6: prev= 612 ms, new= 738 ms, diff=1.21 slower ❌
      q7: prev=1593 ms, new=1640 ms, diff=1.03 slower ✖
      q8: prev=1814 ms, new=1805 ms, diff=1.00 faster ✔
      q9: prev=2286 ms, new=2299 ms, diff=1.01 slower ✖
     q10: prev=1094 ms, new=1155 ms, diff=1.06 slower ✖
     q11: prev= 392 ms, new= 395 ms, diff=1.01 slower ✖
     q12: prev= 858 ms, new= 834 ms, diff=1.03 faster ✔
     q13: prev= 684 ms, new= 718 ms, diff=1.05 slower ✖
     q14: prev= 693 ms, new= 740 ms, diff=1.07 slower ✖
     q15: prev= 953 ms, new= 937 ms, diff=1.02 faster ✔
     q16: prev= 352 ms, new= 333 ms, diff=1.06 faster ✔
     q17: prev=1942 ms, new=1907 ms, diff=1.02 faster ✔
     q18: prev=2058 ms, new=2127 ms, diff=1.03 slower ✖
     q19: prev= 854 ms, new= 956 ms, diff=1.12 slower ✖
     q20: prev= 844 ms, new= 847 ms, diff=1.00 slower ✖
     q21: prev=2377 ms, new=2339 ms, diff=1.02 faster ✔
     q22: prev= 330 ms, new= 322 ms, diff=1.02 faster ✔
   TOTAL: prev=73574.787017 ms, new=74445.69362600002 ms, diff=1.01 slower ✖
```

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,7 +80,7 @@ insta = { version = "1.46.0", features = ["filters"] }
 tpchgen = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "e83365a5a9101906eb9f78c5607b83bc59849acf" }
 tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "e83365a5a9101906eb9f78c5607b83bc59849acf" }
 parquet = "57.1.0"
-arrow = "57.1.0"
+arrow = { version = "57.1.0", features = ["test_utils"] }
 tokio-stream = { version = "0.1.17", features = ["sync"] }
 hyper-util = "0.1.16"
 pretty_assertions = "1.4"


src/flight_service/worker_connection_pool.rs
```diff
@@ -149,14 +149,20 @@ impl WorkerConnection {
         MetricBuilder::new(metrics).build(MetricValue::ElapsedCompute(elapsed_compute.clone()));
 
         // Building the actual request that will be sent to the worker.
-        let mut headers = get_config_extension_propagation_headers(ctx.session_config())?;
-        headers.extend(get_passthrough_headers(ctx.session_config()));
+        let cfg = ctx.session_config();
+        let mut headers = get_passthrough_headers(cfg);
+        let mut plan_proto = Bytes::new();
+        if target_partition_range.start == 0 {
```
gene-bordegaray: if a request that has non zero arrives first, it isn't wrong but may fail. Is there a cse this can happen. I would assume so

src/flight_service/single_write_multi_read.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<T: Clone> SingleWriteMultiRead<T> {
+    pub(crate) fn new() -> Self {
+        let (tx, rx) = watch::channel(None);
+        Self { tx, rx }
+    }
+
+    /// Write the value. Only the first call is meaningful;
+    /// subsequent calls overwrite silently.
+    pub(crate) fn write(&self, item: T) {
+        self.tx.send_replace(Some(item));
```
gene-bordegaray: I know we are assuming that we really only have one writer so this shouldn't be an issue. But we could be defensive and reject attempts to overwrite, thoughts?

src/flight_service/do_get.rs
```diff
@@ -82,49 +92,62 @@ impl Worker {
...
+            let task_data = TaskData {
+                headers: Arc::new(headers),
+                plan,
+                task_ctx,
+                num_partitions_remaining: Arc::new(AtomicUsize::new(total_partitions)),
+            };
+            entry.write(task_data.clone());
+            task_data
+        } else {
+            entry
```
gene-bordegaray: If I am understanding correctly, this will wait 10 seconds to timeout even if the writer errors much before this. Like say line number 118 fails, this would not propagate immediately to the reader for initialization errors.

Is there some way we could handle this?

tests/tpcds_correctness_test.rs
```diff
@@ -29,37 +29,37 @@ mod tests {
         test_tpcds_query("q1").await
     }
 
-    #[tokio::test(flavor = "multi_thread")]
+    #[tokio::test]
```
gabotechs, line 32: Multithread seems to be too aggressive for all these tests, and I've seen some times this flaking because of the hardcoded 10s timeout while some network calls wait for another network call to provide the plan.

In principle, all the calls should happen almost instantly, but these tests clutter the CPU so much that this can flake. It has also flaked for other reasons before, all related to the huge load.

Something that could alleviate this is to use the InMemoryChannelResolver rather than the LocalhostChannelResolver.

src/flight_service/worker_connection_pool.rs
gene-bordegaray: If this happens and a writer is super delayed or cancelled is is desire the readers will fail by timeout, or is there a better fallback?

gabotechs: > if a request that has non zero arrives first

Unless I messed up, this PR is prepared so that requests can arrive at their own pace.
- If the query for partition 0 arrives first, it sets the plan and any subsequent requests pick it up
- If the query for partition 0 arrives last, all the other request will block until it arrives, waiting for the plan

src/flight_service/single_write_multi_read.rs
gabotechs: Yeah, This might be better of enforced at runtime, I can do that

gabotechs: done!

src/flight_service/do_get.rs
gabotechs: 👍 nice catch, this is indeed a problem. I've addressed it in my last commit.

src/flight_service/worker_connection_pool.rs
gene-bordegaray: Yes but thinking in the extreme cases:

There are 4 tasks:
- Task 2 arrives first and has non 0 start so it waits
- Task 0 gets cancelled or somehow super delayed
- Task 2 would timeout

I know we aren't fault tolerant but maybe a better way to handle rather than 10 second timeout?

gabotechs: 🤔 My bet is that if this happens, the fate of the query was set long before this mechanism kicks in.

In the scenario you are describing, the query would fail either way, the difference is that:
- before this PR, we'll do a bunch of useless work before realizing that Task 0 died and the whole progress needs to be discarded, probably longer than the 10s timeout we have hardcoded in this PR.
- with this PR, Task 2 will just not start work that would otherwise need to be discarded.

src/flight_service/do_get.rs
```diff
@@ -357,139 +300,3 @@ fn garbage_collect_arrays(batch: RecordBatch) -> Result<RecordBatch, DataFusionE
...
-    use datafusion::physical_plan::empty::EmptyExec;
-    use datafusion::physical_plan::repartition::RepartitionExec;
-    use datafusion_proto::physical_plan::DefaultPhysicalExtensionCodec;
-    use prost::{Message, bytes::Bytes};
-    use tonic::Request;
-    use uuid::Uuid;
-
-    #[tokio::test]
-    async fn test_task_data_partition_counting() {
-        let mut endpoint = Worker::default();
```
gabotechs, line 378: This test is no longer relevant given the split between the `do_get` action in two, and we are already covered by the integration tests, so removing it.

src/flight_service/worker_connection_pool.rs
```diff
@@ -148,20 +147,13 @@ impl WorkerConnection {
         let elapsed_compute_clone = elapsed_compute.clone();
         MetricBuilder::new(metrics).build(MetricValue::ElapsedCompute(elapsed_compute.clone()));
 
-        // Track the size of the serialized plan sent over the wire.
-        let plan_bytes = MetricBuilder::new(metrics).global_counter("plan_bytes_sent");
-        let plan_proto = input_stage.plan.encoded()?;
-        plan_bytes.add(plan_proto.len());
```
gabotechs, line 157: Now the plan is sent in `DistributedExec`

src/execution_plans/distributed.rs
```diff
@@ -49,38 +80,99 @@ impl DistributedExec {
...
+                return internal_err!("Plan is not set for stage {}", stage.num);
+            };
 
-            let ready_stage = Stage {
+            // Right now, we assign random workers to tasks. This might change in the future.
+            let start_idx = rand::rng().random_range(0..urls.len());
+
+            // This assumes the plan is the same for all the tasks within a stage. This is fine for
+            // now, but it should be possible to send different versions of the subplan to the
+            // different tasks.
```
gene-bordegaray, line 126: I don't fully understand this comment, when would this be needed?

src/flight_service/do_action.rs
```diff
@@ -0,0 +1,124 @@
...
+            // Initialize partition count to the number of partitions in the stage
+            let total_partitions = plan.properties().partitioning.partition_count();
+            Ok::<_, DataFusionError>(TaskData {
+                plan,
+                task_ctx,
+                num_partitions_remaining: Arc::new(AtomicUsize::new(total_partitions)),
+            })
+        };
+
+        entry.write(task_data().await.map_err(datafusion_error_to_tonic_status)).map_err(|_| {
```
gene-bordegaray: this is a little odd to me, could we just fail fast then to write this to the cache?

src/flight_service/do_get.rs
```diff
@@ -59,86 +53,34 @@ pub struct DoGet {
...
-                })
-            })
+        // Other request is responsible for writing the plan that belongs to this StageKey, so
+        // we'll resolve immediately if it was already there, or wait until it's ready.
+        let task_data = entry
+            .read(Duration::from_secs(WAIT_PLAN_TIMEOUT_SECS))
             .await
-            .map_err(|err| Status::invalid_argument(format!("Cannot decode stage proto: {err}")))?;
-        let plan = Arc::clone(&stage_data.plan);
+            .map_err(|v| Status::invalid_argument(v.to_string()))??;
```
gene-bordegaray: I think this should be timeout or internal err

src/execution_plans/distributed.rs
```diff
@@ -49,38 +80,99 @@ impl DistributedExec {
...
+    ///    URL per task.
+    /// 2. Sending the sliced subplans to the assigned URLs. For each URL assigned to a task, a
+    ///    network call feeding the subplan is necessary.
+    /// 3. In each network boundary, set the input plan to `None`. That way, network boundaries
+    ///    become nodes without children and traversing them will not go further down in.
+    fn prepare_plan(&self, ctx: &Arc<TaskContext>) -> Result<PreparedPlan> {
+        let worker_resolver = get_distributed_worker_resolver(ctx.session_config())?;
+        let codec = DistributedCodec::new_combined_with_user(ctx.session_config());
+
+        let mut headers = get_config_extension_propagation_headers(ctx.session_config())?;
```
gene-bordegaray: I believe this is dead code

src/execution_plans/distributed.rs
gabotechs: If, for example, you have 10 leaf tasks, and a DataSourceExec that contains 10 FileGroups, it should be fine to send 1 FileGroup to each worker (just the one that's needed) rather than sending the 10 FileGroups to the 10 tasks, given that 9/10 in each task are going to end up being ignored.

src/flight_service/do_action.rs
gabotechs: 🤔 by cache I imagine you mean the SingleWriteMultiRead structure. We need to write to the SingleWriteMultiRead so that any other task that is expecting a `TaskData` on the read side aborts immediately.

If we didn't write to the SingleWriteMultiRead, then the other tasks will need to wait for a timeout to abort.

Writing here is pretty much free also, it's a sync operation, so there's no wait on it thanks to how tokio watch works.

src/flight_service/do_get.rs
gabotechs: Or even better, we could be propagating the original DataFusionError. I'll do that.

src/execution_plans/distributed.rs
gabotechs: Indeed it is, thanks!

src/flight_service/do_get.rs
gabotechs: Done! also moved to an internal err in case the error is a timeout or something like that.

PR #355, opened 2026-02-23
Add max_tasks_per_sate config parameter

gabotechs:
Previously, the number of tasks assigned to each stage was capped based on the number of workers available.

Now, this is still true, unless the user explicitly specifies the `max_tasks_per_stage` parameter.

This is useful for:
- Limiting the amount of distribution in case not all the machines in the cluster need to be used.
- For benchmarks, adding a higher limit, in order to reproduce issues that only appear in very wide plans.

### DIFF
--- a/benchmarks/src/run.rs
+++ b/benchmarks/src/run.rs
@@ -99,7 +99,7 @@ pub struct RunOpt {
     #[structopt(long, default_value = "lz4")]
     compression: String,
 
-    /// Collects metrics across network boundaries
+    /// Sets the limits of tasks for each stage
     #[structopt(long, default_value = "0")]
     max_tasks_per_stage: usize,
 
--- a/src/distributed_planner/distributed_physical_optimizer_rule.rs
+++ b/src/distributed_planner/distributed_physical_optimizer_rule.rs
@@ -528,7 +528,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore] // FIXME: fix this test
     async fn test_limited_by_worker() {
         let query = r#"
         SET datafusion.execution.target_partitions=2;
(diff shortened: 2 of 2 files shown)


benchmarks/cdk/bin/datafusion-bench.ts
```diff
@@ -17,9 +17,12 @@ async function main() {
         .option('--batch-size <number>', 'Standard Batch coalescing size (number of rows)', '32768')
         .option('--shuffle-batch-size <number>', 'Shuffle batch coalescing size (number of rows)', '32768')
         .option('--children-isolator-unions <number>', 'Use children isolator unions', 'true')
-        .option('--broadcast-joins <boolean>', 'Use broadcast joins', 'false')
+        .option('--broadcast-joins <boolean>', 'Use broadcast joins', 'true')
```
gabotechs, line 20: Not related to this PR, but I think it might be a better default for benchmarks to enable broadcast joins

benchmarks/cdk/bin/datafusion-bench.ts
```diff
@@ -17,9 +17,12 @@ async function main() {
...
         .option('--collect-metrics <boolean>', 'Propagates metric collection', 'true')
         .option('--compression <string>', 'Compression algo to use within workers (lz4, zstd, none)', 'lz4')
+        .option('--max-tasks-per-stage <number>', 'Max tasks per stage', '0')
+        .option('--repartition-file-min-size <number>', 'repartition_file_min_size DF option', '10485760' /* upstream default */)
+        .option('--target-partitions <number>', 'target_partitions DF option', '8')
```
gabotechs, line 25: Same as above, not related to the PR itself, but I want to use this in a follow up, so would like to add them.

benchmarks/src/run.rs
```diff
@@ -99,6 +99,10 @@ pub struct RunOpt {
     #[structopt(long, default_value = "lz4")]
     compression: String,
 
+    /// Collects metrics across network boundaries
```
jayshrivastava: comment looks wrong

src/distributed_planner/plan_annotator.rs
```diff
@@ -174,7 +174,10 @@ fn _annotate_plan(
     let d_cfg = DistributedConfig::from_config_options(cfg)?;
     let broadcast_joins = d_cfg.broadcast_joins;
     let estimator = &d_cfg.__private_task_estimator;
-    let n_workers = d_cfg.__private_worker_resolver.0.get_urls()?.len().max(1);
+    let max_tasks = match d_cfg.max_tasks_per_stage {
+        0 => d_cfg.__private_worker_resolver.0.get_urls()?.len().max(1),
+        v => v,
+    };
```
jayshrivastava, line 180: probably worth adding a small test if possible

jayshrivastava (approved):

PR #357, opened 2026-02-24
Split DistributedPhysicalOptimizerRule into multiple, more scoped ones.

### LINKED ISSUE #347: Need for Customizable Plan Annotation and Network Boundary Logic in DFD
**Summary:**  
This issue aims to start a discussion around improving extensibility in `datafusion-distributed`, especially for custom plan annotations and network boundaries. I would appreciate insights from the DataFusion community on potential design directions and best practices.

***

## **Challenge and Motivation**

I believe there is significant value in expanding the extensibility of `datafusion-distributed` (DFD). The project’s core strengths—plan annotation, insertion of network boundaries, and distribution of sub‑plans to workers—make it a natural place for more flexible customization.

My colleagues and I have been working toward implementing custom network boundaries and plan annotations in a fork of DFD. The use case involves inserting multiple `ExecutionPlan` nodes instead of relying solely on `NetworkShuffleExec`, `NetworkCoalesceExec`, or `NetworkBroadcastExec`. In practice, this requires a mechanism to introduce custom plan annotations and network boundaries beyond what DFD currently supports.

An initial attempt at introducing this extensibility can be found in this draft PR by @kurtvolmar:  
<https://github.com/kurtvolmar/datafusion-distributed/pull/1>.

However, given that DataFusion itself already provides many extension points, it may not be ideal for DFD to introduce a separate, parallel extensibility framework. Instead, it seems likely that a DataFusion‑native mechanism could be leveraged to support more flexible behavior within DFD.

***

## **Proposal and Call for Collaboration**

The existing `DistributedPhysicalOptimizerRule` encapsulates a substantial amount of logic. One possible direction would be to decompose this rule into smaller, more focused components—such as a `PlanAnnotationRule` and `NetworkBoundaryRule`—and expose configuration or hooks that allow users to implement custom logic where needed.

Community input would be highly valuable, particularly around:

*   Whether splitting `DistributedPhysicalOptimizerRule` into smaller, pluggable rules aligns with the project’s direction.
*   Alternative approaches in DataFusion that could enable the desired extensibility without modifying DFD directly.
*   Prior art or patterns in the DataFusion ecosystem that could help inform a clean design.

Feedback, suggestions, or discussion from maintainers and contributors would be greatly appreciated. The goal is to collaborate on a design that increases flexibility without adding unnecessary complexity to DFD.

cc: @gabotechs

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/347

This PR ships two main things:

## Split the `DistributedPhysicalOptimizerRule` into multiple, more scoped ones

Previously, all the planning in the project was done in one rule. Now, it's split in multiple more composable ones, so that users are free to insert their own rules in the middle:

```rust
StartDistributedContext
AddCoalesceOnTop
InsertBroadcast
InjectNetworkBoundaryPlaceholders
ApplyNetworkBoundaries
BatchCoalesceBelowBoundaries
EndDistributedContext
```

The big bulk of the work happens in:
- `InjectNetworkBoundaryPlaceholders` which injects network boundary placeholders in the middle of the plan, that just contain information about the network boundary type and the number of input tasks.
- `ApplyNetworkBoundaries` which reads the network boundary placeholders and inject the actual network boundaries.

## Expose the `NetworkBoundaryPlaceholder` as part of the public API

In https://github.com/datafusion-contrib/datafusion-distributed/issues/347, we discussed about exposing the annotations system as part of the public API, and https://github.com/apache/datafusion/issues/20396 upstream aims to provide such a public API in vanilla DataFusion.

While that's powerful, after some consideration, this project would have a hard time exposing a stable API if the full power of the annotations is exposed, so instead, another approach is attempted:
- All the annotation system (`AnnotatedPan` struct, etc...) is made private, is not even exposed outside of the `rule_4_inject_network_boundary_placeholders.rs` file, so it's private even to this project itself.
- The new API for customizing distributed plan is given in the form of `NetworkBoundaryPlaceholder`s, which are planning-time `ExecutionPlan` implementations with a very narrow API:
  - The NetworkBoundary kind to inject (coalesce, shuffle, or broadcast)
  - The input task count of the stage below

The chain of optimizer rules use this same `NetworkBoundaryPlaceholder` API for injecting its network boundaries, which is what allows making `AnnotatedPlan` private even within the project, and at the same time, allows users to inject their own rules in between `InjectNetworkBoundaryPlaceholders` and `ApplyNetworkBoundaries`, or even replace `InjectNetworkBoundaryPlaceholders` all together with their own custom implementation.

This test demonstrates how to use the `NetworkBoundaryPlaceholder` API for customizing a distributed plan:
https://github.com/datafusion-contrib/datafusion-distributed/blob/2dd97539bf736e83c1697b763c5ba1c724a1e69d/tests/custom_network_boundary_placeholders.rs

### DIFF
--- a/src/distributed_planner/network_boundary_placeholder.rs
+++ b/src/distributed_planner/network_boundary_placeholder.rs
@@ -1,0 +1,87 @@
+use crate::common::require_one_child;
+use datafusion::common::{Result, not_impl_err};
+use datafusion::execution::{SendableRecordBatchStream, TaskContext};
+use datafusion::physical_plan::{DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties};
+use std::any::Any;
+use std::fmt::Formatter;
+use std::sync::Arc;
+
+/// [ExecutionPlan] implementation that acts as a simple placeholder for the distributed planner
+/// to know when a `NetworkBoundary` should be injected.
+///
+/// This structure is part of the public API, and it's what allows users of this library to decide
+/// where should the network boundaries be placed, and what task count should be used for the
+/// stage below.
+///
+/// Note that there are restrictions around where can the [NetworkBoundaryPlaceholder]s be placed,
+/// for example:
+/// - A [NetworkBoundaryKind::Broadcast] needs to be placed right above a `BroadcastExec` node.
+/// - A [NetworkBoundaryKind::Shuffle] needs to be placed right about a `RepartitionExec` node.
+///
+/// Failure to do so will result in planning errors.
+#[derive(Debug)]
+pub struct NetworkBoundaryPlaceholder {
+    /// The kind of network boundary that should be injected.
+    pub kind: NetworkBoundaryKind,
+    /// The task count for the input stage of this network boundary.
+    ///
+    /// Note that the task count for this network boundary is decided by the other network boundary
+    /// immediately above, and not this one.
+    pub input_task_count: usize,
+    /// The input [ExecutionPlan] that will run remotely on the stage below.
+    pub input: Arc<dyn ExecutionPlan>,
+}
+
+/// The type of network boundary that should be injected:
+/// - [NetworkBoundaryKind::Shuffle] -> `NetworkShuffleExec`
+/// - [NetworkBoundaryKind::Coalesce] -> `NetworkCoalesceExec`
+/// - [NetworkBoundaryKind::Broadcast] -> `NetworkBroadcastExec`
+#[derive(Debug, Clone)]
+pub enum NetworkBoundaryKind {
+    Shuffle,
+    Coalesce,
+    Broadcast,
+}
+
+impl DisplayAs for NetworkBoundaryPlaceholder {
+    fn fmt_as(&self, _t: DisplayFormatType, f: &mut Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "NetworkBoundaryPlaceholder: kind={:?}, input_tasks={}",
+            self.kind, self.input_task_count
+        )
+    }
+}
+
+impl ExecutionPlan for NetworkBoundaryPlaceholder {
+    fn name(&self) -> &str {
+        "NetworkBoundaryPlaceholder"
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn properties(&self) -> &PlanProperties {
+        self.input.properties()
+    }
+
+    fn children(&self) -> Vec<&Arc<dyn ExecutionPlan>> {
+        vec![&self.input]
+    }
+
+    fn with_new_children(
+        self: Arc<Self>,
+        children: Vec<Arc<dyn ExecutionPlan>>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        Ok(Arc::new(Self {
+            kind: self.kind.clone(),
+            input_task_count: self.input_task_count,
+            input: require_one_child(children)?,
+        }))
+    }
+
+    fn execute(&self, _: usize, _: Arc<TaskContext>) -> Result<SendableRecordBatchStream> {
+        not_impl_err!("NetworkBoundaryPlaceholder does not support execution.")
+    }
+}
--- a/src/distributed_planner/rules/mod.rs
+++ b/src/distributed_planner/rules/mod.rs
@@ -1,0 +1,17 @@
+mod rule_1_start_distributed_context;
+mod rule_2_add_coalesce_on_top;
+mod rule_3_insert_broadcast;
+mod rule_4_annotate_plan;
+mod rule_5_apply_network_boundaries;
+mod rule_6_batch_coalesce_below_boundaries;
+mod rule_7_end_distributed_context;
+
+pub use rule_1_start_distributed_context::StartDistributedContext;
+pub use rule_2_add_coalesce_on_top::AddCoalesceOnTop;
+pub use rule_3_insert_broadcast::InsertBroadcast;
+pub use rule_4_annotate_plan::AnnotatePlan;
+pub use rule_5_apply_network_boundaries::ApplyNetworkBoundaries;
+pub use rule_6_batch_coalesce_below_boundaries::BatchCoalesceBelowBoundaries;
+pub use rule_7_end_distributed_context::EndDistributedContext;
+
+pub(crate) use rule_4_annotate_plan::AnnotatedPlan;
--- a/src/distributed_planner/rules/mod.rs
+++ b/src/distributed_planner/rules/mod.rs
@@ -1,7 +1,7 @@
 mod rule_1_start_distributed_context;
 mod rule_2_add_coalesce_on_top;
 mod rule_3_insert_broadcast;
-mod rule_4_annotate_plan;
+mod rule_4_inject_network_boundary_placeholders;
 mod rule_5_apply_network_boundaries;
 mod rule_6_batch_coalesce_below_boundaries;
 mod rule_7_end_distributed_context;
@@ -9,9 +9,7 @@ mod rule_7_end_distributed_context;
 pub use rule_1_start_distributed_context::StartDistributedContext;
 pub use rule_2_add_coalesce_on_top::AddCoalesceOnTop;
 pub use rule_3_insert_broadcast::InsertBroadcast;
-pub use rule_4_annotate_plan::AnnotatePlan;
+pub use rule_4_inject_network_boundary_placeholders::InjectNetworkBoundaryPlaceholders;
 pub use rule_5_apply_network_boundaries::ApplyNetworkBoundaries;
 pub use rule_6_batch_coalesce_below_boundaries::BatchCoalesceBelowBoundaries;
 pub use rule_7_end_distributed_context::EndDistributedContext;
-
-pub(crate) use rule_4_annotate_plan::AnnotatedPlan;
--- a/src/distributed_planner/rules/rule_4_inject_network_boundary_placeholders.rs
+++ b/src/distributed_planner/rules/rule_4_inject_network_boundary_placeholders.rs
@@ -1,8 +1,13 @@
+use crate::NetworkBoundaryKind::{Broadcast, Coalesce, Shuffle};
 use crate::TaskCountAnnotation::{Desired, Maximum};
 use crate::distributed_planner::distributed_context::DistributedContext;
+use crate::distributed_planner::network_boundary_placeholder::NetworkBoundaryKind;
 use crate::execution_plans::ChildrenIsolatorUnionExec;
-use crate::{BroadcastExec, DistributedConfig, TaskCountAnnotation, TaskEstimator};
-use datafusion::common::{DataFusionError, Result, plan_datafusion_err};
+use crate::{
+    BroadcastExec, DistributedConfig, NetworkBoundaryPlaceholder, TaskCountAnnotation,
+    TaskEstimator,
+};
+use datafusion::common::{DataFusionError, Result, plan_datafusion_err, plan_err};
 use datafusion::config::ConfigOptions;
 use datafusion::physical_expr::Partitioning;
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
@@ -16,10 +21,14 @@ use datafusion::physical_plan::union::UnionExec;
 use std::fmt::{Debug, Formatter};
 use std::sync::Arc;
 
+/// Injects the appropriate [NetworkBoundaryPlaceholder]s in the plan. These placeholders tell the
+/// next step where to put the actual network boundaries.
+///
+/// Additionally, it replaces [UnionExec]s with [ChildrenIsolatorUnionExec]s.
 #[derive(Debug)]
-pub struct AnnotatePlan;
+pub struct InjectNetworkBoundaryPlaceholders;
 
-impl PhysicalOptimizerRule for AnnotatePlan {
+impl PhysicalOptimizerRule for InjectNetworkBoundaryPlaceholders {
     fn optimize(
         &self,
         plan: Arc<dyn ExecutionPlan>,
@@ -27,17 +36,37 @@ impl PhysicalOptimizerRule for AnnotatePlan {
     ) -> Result<Arc<dyn ExecutionPlan>> {
         let d_ctx = DistributedContext::ensure(self, &plan)?;
 
-        d_ctx.annotated_plan.lock().unwrap().replace(annotate_plan(
-            Arc::clone(&d_ctx.plan),
-            None,
-            config,
-            true,
-        )?);
-        Ok(plan)
+        let annotated = annotate_plan(Arc::clone(&d_ctx.plan), None, config, true)?;
+
+        fn inject_network_boundary_placeholders(
+            mut annotation: AnnotatedPlan,
+        ) -> Result<Arc<dyn ExecutionPlan>> {
+            match annotation.plan_or_nb {
+                PlanOrNetworkBoundary::Plan(plan) => {
+                    let mut children = Vec::with_capacity(annotation.children.len());
+                    for child in annotation.children {
+                        children.push(inject_network_boundary_placeholders(child)?);
+                    }
+                    plan.with_new_children(children)
+                }
+                PlanOrNetworkBoundary::NetworkBoundary(kind) => {
+                    let Some(child) = annotation.children.pop() else {
+                        return plan_err!("Expected NetworkBoundary annotation to have one child");
+                    };
+                    Ok(Arc::new(NetworkBoundaryPlaceholder {
+                        kind,
+                        input_task_count: child.task_count.as_usize(),
+                        input: inject_network_boundary_placeholders(child)?,
+                    }))
+                }
+            }
+        }
+
+        plan.with_new_children(vec![inject_network_boundary_placeholders(annotated)?])
     }
 
     fn name(&self) -> &str {
-        "AnnotatePlan"
+        "InjectNetworkBoundaryPlaceholders"
     }
 
     fn schema_check(&self) -> bool {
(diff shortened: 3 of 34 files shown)


src/distributed_planner/network_boundary_placeholder.rs
```diff
@@ -0,0 +1,87 @@
...
+/// to know when a `NetworkBoundary` should be injected.
+///
+/// This structure is part of the public API, and it's what allows users of this library to decide
+/// where should the network boundaries be placed, and what task count should be used for the
+/// stage below.
+///
+/// Note that there are restrictions around where can the [NetworkBoundaryPlaceholder]s be placed,
+/// for example:
+/// - A [NetworkBoundaryKind::Broadcast] needs to be placed right above a `BroadcastExec` node.
+/// - A [NetworkBoundaryKind::Shuffle] needs to be placed right about a `RepartitionExec` node.
```
gabrielkerr, line 19:
```suggestion
/// - A [NetworkBoundaryKind::Shuffle] needs to be placed right above a `RepartitionExec` node.
```

src/distributed_planner/network_boundary_placeholder.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::sync::Arc;
+
+/// [ExecutionPlan] implementation that acts as a simple placeholder for the distributed planner
+/// to know when a `NetworkBoundary` should be injected.
+///
+/// This structure is part of the public API, and it's what allows users of this library to decide
+/// where should the network boundaries be placed, and what task count should be used for the
+/// stage below.
+///
+/// Note that there are restrictions around where can the [NetworkBoundaryPlaceholder]s be placed,
```
gabrielkerr, line 16:
```suggestion
/// Note that there are restrictions around where the [NetworkBoundaryPlaceholder]s can be placed,
```

src/distributed_planner/network_boundary_placeholder.rs
```diff
@@ -0,0 +1,87 @@
...
+/// - A [NetworkBoundaryKind::Shuffle] needs to be placed right about a `RepartitionExec` node.
+///
+/// Failure to do so will result in planning errors.
+#[derive(Debug)]
+pub struct NetworkBoundaryPlaceholder {
+    /// The kind of network boundary that should be injected.
+    pub kind: NetworkBoundaryKind,
+    /// The task count for the input stage of this network boundary.
+    ///
+    /// Note that the task count for this network boundary is decided by the other network boundary
```
gabrielkerr, line 28: If this placeholder sits between Stage N (below) and Stage N+1 (above):

  | Field                                  | Refers to                              |
  |----------------------------------------|----------------------------------------|
  | input                                  | The ExecutionPlan for Stage N          |
  | input_task_count                       | Task count for Stage N                 |
  | "task count for this network boundary" | Task count for Stage N+1 (output side) |
  | "boundary immediately above"           | The boundary between Stage N+1 and N+2 |

Could we clarify the comments with explicit stage references?
  - `input_task_count`: The number of tasks in Stage N (the stage below this boundary that executes `input`)
  - The task count for Stage N+1 (this boundary's output/consumer stage) is determined by the `input_task_count` of the
  NetworkBoundaryPlaceholder above this one, if any.

tests/custom_network_boundary_placeholders.rs
```diff
@@ -0,0 +1,133 @@
...
+            ) -> Result<Arc<dyn ExecutionPlan>> {
+                plan.transform_up(|plan| {
+                    // upon finding a NetworkBoundaryPlaceholder of Coalesce kind, stack another
+                    //  one on top with half the input tasks.
+                    if let Some(nb) = plan.as_any().downcast_ref::<NetworkBoundaryPlaceholder>()
+                        && matches!(nb.kind, NetworkBoundaryKind::Coalesce)
+                    {
+                        return Ok(Transformed::yes(Arc::new(NetworkBoundaryPlaceholder {
+                            kind: NetworkBoundaryKind::Coalesce,
+                            input_task_count: nb.input_task_count.div_ceil(2),
```
gabrielkerr, line 45: Related to my comment on `NetworkBoundaryPlaceholder` above, `task_count/2` is the number of tasks the `Arc::new(NetworkBoundaryPlaceholder {...}` is expecting to receive from the (now) child coalesce boundary? If so, maybe it makes more sense to rename `input_task_count` to `output_task_count`?

Please correct me I misunderstand 😄

tests/custom_network_boundary_placeholders.rs
```diff
@@ -0,0 +1,133 @@
...
+        ApplyNetworkBoundaries, DefaultSessionBuilder, DistributedExt, NetworkBoundaryKind,
+        NetworkBoundaryPlaceholder, SessionStateBuilderExt, assert_snapshot,
+    };
+    use futures::TryStreamExt;
+    use itertools::Itertools;
+    use std::error::Error;
+    use std::sync::Arc;
+
+    #[tokio::test]
+    async fn custom_network_boundary_placeholder() -> Result<(), Box<dyn Error>> {
```
gabrielkerr, line 23: This is very helpful, thank you for adding this!

src/distributed_planner/session_state_builder_ext.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn set_distributed_physical_optimizer_rules(&mut self);
+}
+
+impl SessionStateBuilderExt for SessionStateBuilder {
+    fn with_distributed_physical_optimizer_rules(mut self) -> Self {
+        self.set_distributed_physical_optimizer_rules();
+        self
+    }
+
+    fn set_distributed_physical_optimizer_rules(&mut self) {
```
gabrielkerr, line 24: It would be nice to have a method for inserting custom rules similarly to how you've done `insert_before` in the test. Could we have something like below? I imagine that it would automatically add rules between rules 4 and 5 since that is the intended order described in your comments.
```rust
fn with_custom_network_boundary_rules(
      self,
      rules: impl IntoIterator<Item = Arc<dyn PhysicalOptimizerRule + Send + Sync>>
  ) -> Self;
```

Used like 
```rust
  let builder = SessionStateBuilder::new()
      .with_distributed_physical_optimizer_rules()
      .with_custom_network_boundary_rules([Arc::new(TreeReductionRule::new(type_checker))]);
```

Thoughts?

tests/custom_network_boundary_placeholders.rs
```diff
@@ -0,0 +1,133 @@
...
+        #[derive(Debug)]
+        struct HierarchicalCoalesce;
+
+        impl PhysicalOptimizerRule for HierarchicalCoalesce {
+            fn optimize(
+                &self,
+                plan: Arc<dyn ExecutionPlan>,
+                _: &ConfigOptions,
+            ) -> Result<Arc<dyn ExecutionPlan>> {
+                plan.transform_up(|plan| {
```
gabrielkerr, line 37: nit: If this is in the context of custom distributed rules, should this call `Distributed::ensure`? Even a comment describing that it should be called here might be enough to show the intended pattern if that is difficult to do in this test.

src/distributed_planner/network_boundary_placeholder.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// immediately above, and not this one.
+    pub input_task_count: usize,
+    /// The input [ExecutionPlan] that will run remotely on the stage below.
+    pub input: Arc<dyn ExecutionPlan>,
```
kurtvolmar, line 32: If I understand correctly, will this allow a custom `PhysicalOptimizerRule` using the `NetworkBoundaryPlaceholder` to rewrite its children? 

For our extensibility use case are looking to replace the shuffle with a hierarchical aggregation, which will require us to rewrite the children below the placeholder. I'm not sure if this is what you're intending here, but we will likely use this field to this.

src/distributed_planner/rules/rule_4_inject_network_boundary_placeholders.rs
```diff
@@ -14,44 +21,93 @@ use datafusion::physical_plan::union::UnionExec;
...
 use std::sync::Arc;
 
+/// Injects the appropriate [NetworkBoundaryPlaceholder]s in the plan. These placeholders tell the
+/// next step where to put the actual network boundaries.
+///
+/// Additionally, it replaces [UnionExec]s with [ChildrenIsolatorUnionExec]s.
+#[derive(Debug)]
+pub struct InjectNetworkBoundaryPlaceholders;
+
+impl PhysicalOptimizerRule for InjectNetworkBoundaryPlaceholders {
```
kurtvolmar, line 31: With this model, if I create a `PhysicalOptimizerRule` with a `NetworkBoundaryPlaceholder` targeting a hash `RepartitionExec` and insert it before `InjectNetworkBoundaryPlaceholders`, this will cause a "collision" on on the custom `NetworkBoundaryPlaceholder` and the the Shuffle logic in this rule. Both rules will be targeting the same node and attempting to inject `NetworkBoundaryPlaceholders`. 

I'd be interested in seeing some detection whether a `NetworkBoundaryPlaceholder` already exists for a node, so that custom rules can play nicely with `InjectNetworkBoundaryPlaceholders`.

src/distributed_planner/rules/rule_5_apply_network_boundaries.rs
```diff
@@ -1,170 +1,165 @@
...
-#[derive(Debug, Default)]
-pub struct DistributedPhysicalOptimizerRule;
-
-impl PhysicalOptimizerRule for DistributedPhysicalOptimizerRule {
+/// The previous [PhysicalOptimizerRule] to this one (`InjectNetworkBoundaryPlaceholders`) is in
+/// charge of placing the [NetworkBoundaryPlaceholder]s in the appropriate places, but users are
+/// free to place their own, either by having more rules in between, or straight await replacing the
+/// `InjectNetworkBoundaryPlaceholders` with a custom one.
+#[derive(Debug)]
+pub struct ApplyNetworkBoundaries;
```
gabrielkerr, line 27: Checking my understanding here:
Before `rule_5`, the user can insert custom `PhysicalOptimizerRules`. These custom rules can insert `NetworkBoundaryPlaceholders` and any `ExecutionPlan` nodes it likes. Then, `rule_5` will convert the network boundary placeholders to the appropriate network boundary exec and leave all other `ExecutionPlan` nodes untouched. Do I have that right?

src/distributed_planner/rules/mod.rs
```diff
@@ -0,0 +1,15 @@
+mod rule_1_start_distributed_context;
+mod rule_2_add_coalesce_on_top;
+mod rule_3_insert_broadcast;
+mod rule_4_inject_network_boundary_placeholders;
+mod rule_5_apply_network_boundaries;
+mod rule_6_batch_coalesce_below_boundaries;
+mod rule_7_end_distributed_context;
+
+pub use rule_1_start_distributed_context::StartDistributedContext;
```
gabrielkerr, line 9: I like that you've split up the rule into smaller rules with clear responsibilities. However, I feel like splitting this up into 7 different rules has introduced a lot of complexity when there is really a single extension point between rules 4 and 5. I think swinging back towards a single optimizer rule makes more sense. I'll try to explain my thoughts below.

The current interface places the responsibility on the user to add custom physical optimizer rules between rules 4 and 5. Users should only worry about creating their custom rule, with the library inserting it in the correct position between rules 4 and 5.

I think we can keep the seven rules, but hide them (make them private to the crate) from the user. I think a cleaner interface would be to wrap these 7 rules in a single `DistributedPhysicalOptimizerRule` and add a single method for adding custom rules. This abstracts away the non-customizable details from the user and delegates inserting the custom rules in the correct place to the library.

Maybe something like this:
```rust
pub struct DistributedPhysicalOptimizerRule {
    custom_boundary_rules: Vec<Arc<dyn PhysicalOptimizerRule + Send + Sync>>,
}

impl DistributedPhysicalOptimizerRule {
    pub fn new() -> Self {
        Self { custom_boundary_rules: vec![] }
    }

    pub fn with_custom_boundary_rule(
        mut self,
        rule: impl PhysicalOptimizerRule + Send + Sync + 'static
    ) -> Self {
        self.custom_boundary_rules.push(Arc::new(rule));
        self
    }
}
```

The internal `optimize()` would then:
1. Wrap in DistributedContext (current rule 1)
2. Add coalesce on top (current rule 2)
3. Insert broadcast (current rule 3)
4. **Run each custom boundary rule in order**
5. Apply network boundaries (current rule 5)
6. Batch coalesce below boundaries (current rule 6)
7. Unwrap DistributedContext (current rule 7)

You'll notice I've replaced rule 4 with the custom rules. The current rule 4 adds some placeholders that the user may not want, introducing complexity for removing these boundaries if the user wants to completely customize the network boundary placeholders from the start. In this case I think rule 4 is a reasonable default, but should be able to be replaced (or added to I suppose) by the users custom rules.

gabrielkerr (changes requested):
I like where this is headed! Looking forward to your thoughts on the feedback.

kurtvolmar (commented):
Thank you for tackling this. A concern I have is that a custom `PhysicalOptimizerRule` using the `NetworkBoundaryPlaceholder` paradigm will likely collide with the `InjectNetworkBoundaryPlaceholders`. In the way this is set up now, we would need to at plan-time decide whether to use the custom rule or the default `InjectNetworkBoundaryPlaceholders` (this would probably be yet another `PhysicalOptimizerRule` we need to implement). I would love to explore either further extensibility _into_ rule 4 or some checks to whether a previous rule has inserted `NetworkBoundaryPlaceholder`s around a node before adding more.

gabotechs:
Thanks for the feedback guys! I'll get back to this one shortly.

src/distributed_planner/network_boundary_placeholder.rs
gabotechs: In this case `input_task_count` is the task count for Stage N+1, not Stage N. Network boundaries have no say in the amount of tasks they run on, they just control the amount of tasks below them.

tests/custom_network_boundary_placeholders.rs
gabotechs: 🤔 lets discuss first the approach and see if it makes sense to have a `DistributedContext` at all.

tests/custom_network_boundary_placeholders.rs
gabotechs: 🤔 I think it really is `input_task_count` the name we are looking for. It's the amount of tasks that will input data to the network boundary.

gabotechs: Maybe with the comment above is more clear?

src/distributed_planner/session_state_builder_ext.rs
gabotechs: Yeah, I think this might be worth it

src/distributed_planner/rules/rule_5_apply_network_boundaries.rs
gabotechs: > Do I have that right?

Mostly yes, but converting to the appropriate network boundary exec does perform further modifications to the plans.

For example, NetworkShuffleExec will scale up the output partitions of its RepartitionExec below.

src/distributed_planner/rules/mod.rs
gabotechs: > However, I feel like splitting this up into 7 different rules has introduced a lot of complexity when there is really a single extension point between rules 4 and 5

Yes... realistically speaking, I'd only expect users to either replace rule 4 with their own, or inject an additional one between 4 and 5.

> I think we can keep the seven rules, but hide them (make them private to the crate) from the user. I think a cleaner interface would be to wrap these 7 rules in a single DistributedPhysicalOptimizerRule and add a single method for adding custom rules

👍 this sounds reasonable. It would significantly simplify the code, as we wouldn't need to do that `DistributedContext` workaround for propagating the original single-node plan.

src/distributed_planner/network_boundary_placeholder.rs
gabotechs: > If I understand correctly, will this allow a custom PhysicalOptimizerRule using the NetworkBoundaryPlaceholder to rewrite its children?

The scope of the `NetworkBoundaryPlaceholder` is just to inject network boundaries in the plan so that the `ApplyNetworkBoundaries` rule can then transform them to the equivalent `Network*Exec`, but nothing else.

Taking execution plans unrelated to Distributed DataFusion, like the classical Aggregate(partial) + RepartitionExec + Aggregate(final) and replacing them with a hierarchical pattern sounds like something that is better done in a previous rule, and at that point it should be a matter of just using the existing DataFusion tools for rewriting plans. I don't think this project has a say about that.

src/distributed_planner/rules/rule_4_inject_network_boundary_placeholders.rs
gabotechs: Yeap, this would be a nice addition, although i'd argue that if you have a custom rule that is injecting its own shuffles in the plan... you probably don't want the default `InjectNetworkBoundaryPlaceholders` to kick it at all.

A valid stance that comes to mind is that it's the users responsibility to ensure the plan is correct once it reaches `ApplyNetworkBoundaries`

src/distributed_planner/rules/rule_4_inject_network_boundary_placeholders.rs
```diff
@@ -14,44 +21,93 @@ use datafusion::physical_plan::union::UnionExec;
...
+struct AnnotatedPlan {
     /// The annotated [ExecutionPlan].
-    pub(super) plan_or_nb: PlanOrNetworkBoundary,
+    plan_or_nb: PlanOrNetworkBoundary,
     /// The annotated children of this [ExecutionPlan]. This will always hold the same nodes as
     /// `self.plan.children()` but annotated.
-    pub(super) children: Vec<AnnotatedPlan>,
+    children: Vec<AnnotatedPlan>,
 
     // annotation fields
```
gene-bordegaray, line 108: leftover comment I believe

kurtvolmar (approved):
@gabotechs we've built some prototypes based on this PR branch. Overall, I'm happy with the direction of this PR. Composing our custom optimizers with this looks like it will work well. Our team may have a few suggestions in follow up PRs, but I think I'd rather see this get merged and iterate as needed. Thank you for digging into this and working on the feedback.

src/distributed_planner/rules/mod.rs
kurtvolmar: > I think we can keep the seven rules, but hide them (make them private to the crate) from the user. I think a cleaner interface would be to wrap these 7 rules in a single DistributedPhysicalOptimizerRule

I would much prefer to keep each of the rules public and compose them based on implementation. If we can only inject rules into a single spot, we limit ourselves a bit. For example, users may want to disable rules 4/5 or put a rule in front of this optimizer chain. I'm fine with a standard DistributedOptimizerRule, but we would need each rule in the chain to stay public.

gabrielkerr (approved):

gabrielkerr:
Hey @gabotechs! We've got a POC build on our side using the changes as is. I know you wanted to make more updates. Would you like us to look at anything to help get this into main? Thanks again for all this work!

gabotechs:
The party continues in https://github.com/datafusion-contrib/datafusion-distributed/pull/416. There's a nicer API incoming for building custom plans.

PR #362, opened 2026-03-02
Add shuffle benchmark

gabotechs:
Adds a microbenchmark for shuffles.

The intention is to scope down as much as possible the way performance is measured specifically for shuffle operations in order to qualify further improvements.

The benchmark can be run with:

```
cargo bench -p datafusion-distributed-benchmarks --bench shuffle
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ insta = { version = "1.46.0", features = ["filters"], optional = true }
 tpchgen = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "e83365a5a9101906eb9f78c5607b83bc59849acf", optional = true }
 tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs", rev = "e83365a5a9101906eb9f78c5607b83bc59849acf", optional = true }
 parquet = { version = "57.1.0", optional = true }
-arrow = { version = "57.1.0", optional = true }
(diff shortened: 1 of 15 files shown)


src/flight_service/test_utils/memory_worker.rs
```diff
@@ -0,0 +1,88 @@
...
+                TaskData {
+                    plan: plan.clone(),
+                    num_partitions_remaining: Arc::new(AtomicUsize::new(
+                        self.partitions_batches.len(),
+                    )),
+                }
+            })
+            .await;
+
+        let (client, server) = tokio::io::duplex(1024 * 1024);
```
gene-bordegaray, line 68: you learn something new every day. This is so cool.

I see this is used in the `InMemoryChannelResolver` too but first time seeing 😄

Cargo.toml
gene-bordegaray: this guy needs the same change as well I believe

src/flight_service/test_utils/memory_worker.rs
gabotechs: This is super handy for testing. We should probably even move our integration tests to this pattern, rather than spawning local hosts

gene-bordegaray: Ya this was very fun to read about. Haha nice, another tool in the bag 💯

PR #363, opened 2026-03-04
feat(console): rewrite TUI with worker-centric dashboard and live metrics

EdsonPetry:
## Summary

- **Rewrite the console TUI** with a worker-centric dashboard featuring a cluster overview and per-worker detail views, replacing the previous global task view
- **Extend the observability service** to collect worker-level metrics including CPU usage, RSS memory, output rows, unique query IDs, and task counts via updated protobuf definitions
- **Add CPU/RSS sparklines and sortable columns** to the worker table, with queries-in-flight metric in the cluster metrics panel
- **Add new examples**: a cluster spawner/killer example and concurrent TPC-DS query runner with explain analyze support

### Console TUI (`console/`)
- Split monolithic `ui.rs` into modular components: `cluster.rs`, `worker.rs`, `header.rs`, `footer.rs`, `help.rs`
- New `input.rs` for keyboard event handling and `state.rs` for shared UI state
- Worker table with sortable columns (Name, Stage, Tasks, Queries, CPU, Memory)
- Sparkline charts for CPU and RSS per worker
- Cluster-level metrics: throughput, active/total workers, completed queries, avg duration
- Help overlay (toggle with `?`)

### Observability (`src/observability/`)
- Extended protobuf schema with `cpu_usage`, `rss_bytes`, `output_rows`, `query_ids`, `num_tasks` fields
- Updated service to generate and collect the new per-worker metrics
- Worker now reports richer metrics through the progress API

### Flight service (`src/flight_service/`)
- Changed `observability_service()` generator method on `Worker` to `with_observability_service()`, users now no longer need to wrap the `ObservbilityServiceImpl` that the old observability service generator created with a `ObservabilityServiceServer`.
old:
```rust
    let worker = Worker::default();
   let observability_service = worker.observability_service();

    Server::builder()
        .add_service(ObservabilityServiceServer::new(observability_service))
        .add_service(worker.into_flight_server())
        .serve(SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), args.port))
        .await?;
```

new
```rust
    let worker = Worker::default();

    Server::builder()
        .add_service(worker.with_observability_service())
        .add_service(worker.into_flight_server())
        .serve(SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), args.port))
        .await?;
```

### Examples
- `cluster.rs`: spawns and manages a local cluster of workers with observability
- `tpcds_runner.rs`: supports concurrent query execution and explain analyze

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2158,7 +2158,7 @@ dependencies = [
  "reqwest",
  "sketches-ddsketch",
  "structopt",
- "sysinfo 0.34.2",
+ "sysinfo",
  "test-case",
  "tokio",
  "tokio-stream",
@@ -2201,7 +2201,7 @@ dependencies = [
  "serde",
  "serde_json",
  "structopt",
- "sysinfo 0.30.13",
+ "sysinfo",
  "tokio",
  "tonic",
  "url",
@@ -4341,15 +4341,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "objc2-core-foundation"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a180dd8642fa45cdb7dd721cd4c11b1cadd4929ce112ebd8b9f5803cc79d536"
-dependencies = [
- "bitflags 2.11.0",
-]
-
-[[package]]
 name = "object"
 version = "0.37.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5900,20 +5891,7 @@ dependencies = [
  "ntapi",
  "once_cell",
  "rayon",
- "windows 0.52.0",
-]
-
-[[package]]
-name = "sysinfo"
-version = "0.34.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4b93974b3d3aeaa036504b8eefd4c039dced109171c1ae973f1dc63b2c7e4b2"
-dependencies = [
- "libc",
- "memchr",
- "ntapi",
- "objc2-core-foundation",
- "windows 0.57.0",
+ "windows",
 ]
 
 [[package]]
@@ -6873,16 +6851,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "windows"
-version = "0.57.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "12342cb4d8e3b046f3d80effd474a7a02447231330ef77d71daa6fbc40681143"
-dependencies = [
- "windows-core 0.57.0",
- "windows-targets 0.52.6",
-]
-
-[[package]]
 name = "windows-core"
 version = "0.52.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


src/flight_service/do_get.rs
```diff
@@ -63,7 +63,7 @@ pub struct DoGet {
 /// TaskData stores state for a single task being executed by this Endpoint. It may be shared
 /// by concurrent requests for the same task which execute separate partitions.
 pub struct TaskData {
-    pub(super) plan: Arc<dyn ExecutionPlan>,
+    pub(crate) plan: Arc<dyn ExecutionPlan>,
     /// `num_partitions_remaining` is initialized to the total number of partitions in the task (not
```
gabotechs: You might need to update with `main`, there might be a conflict here, but it should be pretty easy to solve.

console/Cargo.toml
```diff
@@ -5,17 +5,17 @@ edition = "2024"
 
 [dependencies]
 datafusion = { workspace = true }
+clap = { version = "4", features = ["derive"] }
```
gabotechs: Any chance to keep using `structopt` like in the other parts of the project (cli and benchmarks)?

Cargo.lock
```diff
@@ -148,9 +148,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
```
gabotechs: Are all this version bumps necessary? if we could keep them as minimal as possible that would be great

src/observability/service.rs
```diff
@@ -40,17 +47,60 @@ impl ObservabilityService for ObservabilityServiceImpl {
...
+        {
+            use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate};
+
+            let pid = Pid::from_u32(std::process::id());
+            let mut sys = self.system.lock().unwrap_or_else(|e| e.into_inner());
+            sys.refresh_processes_specifics(
+                ProcessesToUpdate::Some(&[pid]),
+                true,
+                ProcessRefreshKind::nothing().with_cpu().with_memory(),
+            );
```
gabotechs: 🤔 It might not be the best idea to perform syscalls over a locked mutex. A syscall might take a decent amount of time, stalling the tokio thread handling this task, and keeping other tasks from making progress in that same thread.

gabotechs: One thing that comes to mind is: collect metrics always from just one thread, and communicate through tokio channels with other threads the most updated information about the sys info.

console/Cargo.toml
EdsonPetry: We could, but would we want to move to clap in the near future since `structopt` has been depreciated?

src/observability/service.rs
EdsonPetry: Got it, I just pushed a change that moves sys metrics collection on a spawned background thread

src/observability/service.rs
```diff
@@ -1,26 +1,80 @@
...
+                            cpu_usage_percent: process.cpu_usage() as f64 / num_cpus,
+                        };
+                        if tx.send(metrics).is_err() {
+                            break;
+                        }
+                    } else if tx.send(WorkerMetrics::default()).is_err() {
+                        break;
+                    };
+
+                    thread::sleep(Duration::from_millis(100))
```
gabotechs: :thinking: I'm not sure if a thread sleep here might be dangerous. This might be blocking a whole thread from making progress, stalling all the tokio tasks that happen to be on that thread.

It might be better to have a tokio::time::sleep here instead.

Cargo.lock
```diff
@@ -2158,6 +2158,7 @@ dependencies = [
  "reqwest",
  "sketches-ddsketch",
  "structopt",
+ "sysinfo 0.34.2",
```
gabotechs: This is weird, I see that the lock here is referring to multiple versions of `sysinfo`:
- 0.34.2
- 0.30.13

Ideally there should be only one version of the `sysinfo` dep in this project.

EdsonPetry: Looks like this might be an issue with the `sysinfo` version defined in `benchmarks/Cargo.toml`. I'll downgrade the version used in the root project to match the one used in the benchmarks.

src/observability/service.rs
```diff
@@ -1,26 +1,79 @@
...
+
+        #[cfg(feature = "system-metrics")]
+        {
+            let pid = Pid::from_u32(std::process::id());
+            let mut sys = sysinfo::System::new_all();
+
+            // Spawn background thread to send system metrics.
+            // This is done to prevent stalling the tokio thread
+            // due to the sys call, leading to task pool starvation.
+            thread::spawn(async move || {
```
EdsonPetry, line 45: @gabotechs I didn't catch this before, but this code here isn't correct. I should use `move` here instead of `async move` because the body of the closure is never executed within the spawned thread. The spawned thread receives a future here that is goes unpolled. Also because this is a separate background thread that isn't in tokio's thread pool thus I believe calling `thread::sleep` is correct.

EdsonPetry: I'll push a fix for this in the PR #372

gabotechs: 🤔 I don't understand why this should be incorrect. Can you push the fix as a separate PR so that we can review it in isolation?

EdsonPetry: Sounds good, the issue I was running into is that because the Future within that async closure was going unpolled the system metrics collection loop was never executing and thus the console was only displaying 0 values, this compiles fine because what the compiler sees is just `thread::spawn(async move || { Future<...> })`. 

Opening a separate PR for this now.

PR #365, opened 2026-03-06
gRPC timeout configs

gene-bordegaray:
Added the option to set some tonic / gRPC configs regarding timeouts. I see this being useful for new adopters to easily configure DFD to their liking.

## Added Configs

Docs can be found here: 
[tonic](https://docs.rs/tonic/latest/tonic/transport/struct.Endpoint.html)
[gRPC](https://grpc.io/docs/guides/)

  - `grpc_connect_timeout_ms`
      - can I establish a channel to the worker?
  - `grpc_request_timeout_ms`
      - can the full do_get RPC finish in time?
  - `wait_plan_timeout_ms`
      - once I reached the worker, how long do I wait for task data?
  - `grpc_tcp_keepalive_ms`
      - how do we notice the underlying TCP connection failed?

### DIFF
--- a/src/distributed_planner/distributed_config.rs
+++ b/src/distributed_planner/distributed_config.rs
@@ -48,6 +48,16 @@ extensions_options! {
         /// The compression used for sending data over the network between workers.
         /// It can be set to either `zstd`, `lz4` or `none`.
         pub compression: String, default = "lz4".to_string()
+        /// Maximum time to wait while establishing a gRPC connection to a worker.
+        /// This is intended to bound connection setup to unreachable workers.
+        pub grpc_connect_timeout_ms: usize, default = grpc_connect_timeout_ms_default()
+        /// Total timeout for an outbound `do_get` request, in milliseconds.
+        /// This is a full-stream deadline for the whole RPC, not an idle timeout.
+        pub grpc_request_timeout_ms: usize, default = grpc_request_timeout_ms_default()
+        /// Maximum time a worker waits for task data to become available before failing the request.
+        pub wait_plan_timeout_ms: usize, default = wait_plan_timeout_ms_default()
+        /// TCP keepalive period used for worker-to-worker connections.
+        pub grpc_tcp_keepalive_ms: usize, default = grpc_tcp_keepalive_ms_default()
         /// Maximum tasks that will be assigned per stage during distributed planning.
         /// If set to 0, this value is the number of workers returned by the provided `WorkerResolver`.
         /// It defaults to 0.
@@ -78,6 +88,22 @@ fn cardinality_task_count_factor_default() -> f64 {
     } else {
         1.0
     }
+}
+
+fn grpc_connect_timeout_ms_default() -> usize {
+    5_000
+}
+
+fn grpc_request_timeout_ms_default() -> usize {
+    30_000
+}
+
+fn wait_plan_timeout_ms_default() -> usize {
+    10_000
+}
+
+fn grpc_tcp_keepalive_ms_default() -> usize {
+    60_000
 }
 
 impl DistributedConfig {
--- a/src/distributed_planner/distributed_config.rs
+++ b/src/distributed_planner/distributed_config.rs
@@ -95,7 +95,7 @@ fn grpc_connect_timeout_ms_default() -> usize {
 }
 
 fn grpc_request_timeout_ms_default() -> usize {
-    30_000
+    300_000
 }
 
 fn wait_plan_timeout_ms_default() -> usize {
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -319,7 +319,6 @@ mod tests {
     };
     use crate::flight_service::single_write_multi_read::SingleWriteMultiRead;
     use bytes::Bytes;
-    use datafusion::common::assert_contains;
     use http::Extensions;
     use std::sync::Arc;
     use tonic::metadata::MetadataMap;
@@ -360,6 +359,7 @@ mod tests {
             Ok(_) => panic!("expected wait-plan timeout error"),
             Err(err) => err,
         };
-        assert_contains!(err.message(), "timed out waiting for value");
+        assert_eq!(err.code(), tonic::Code::Internal);
+        assert_eq!(err.message(), "timed out waiting for value");
     }
 }
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -326,7 +326,7 @@ mod tests {
     use uuid::Uuid;
 
     #[tokio::test]
-    async fn do_get_wait_for_task_data_honors_propagated_timeout() {
+    async fn do_get_honors_propagated_wait_timeout() {
         let worker = Worker::default();
 
         let stage_key = StageKey::new(Bytes::from(Uuid::from_u128(1).as_bytes().to_vec()), 0, 0);
--- a/src/flight_service/do_get.rs
+++ b/src/flight_service/do_get.rs
@@ -1,4 +1,5 @@
 use crate::common::{map_last_stream, on_drop_stream, task_ctx_with_extension};
+use crate::config_extension_ext::set_distributed_option_extension_from_headers;
 use crate::flight_service::worker::Worker;
 use crate::metrics::TaskMetricsCollector;
 use crate::metrics::proto::df_metrics_set_to_proto;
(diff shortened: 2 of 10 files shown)


src/flight_service/do_get.rs
```diff
@@ -301,3 +308,58 @@ fn garbage_collect_arrays(batch: RecordBatch) -> Result<RecordBatch, DataFusionE
...
+    };
+    use crate::flight_service::single_write_multi_read::SingleWriteMultiRead;
+    use bytes::Bytes;
+    use http::Extensions;
+    use std::sync::Arc;
+    use tonic::metadata::MetadataMap;
+    use uuid::Uuid;
+
+    #[tokio::test]
+    async fn do_get_honors_propagated_wait_timeout() {
```
gene-bordegaray, line 328: seems do_get was most tested indirectly before this. Seems fitting to add this here as it's easy to isolate the failure

src/distributed_planner/distributed_config.rs
```diff
@@ -48,6 +48,16 @@ extensions_options! {
...
+        /// Maximum time to wait while establishing a gRPC connection to a worker.
+        /// This is intended to bound connection setup to unreachable workers.
+        pub grpc_connect_timeout_ms: usize, default = grpc_connect_timeout_ms_default()
+        /// Total timeout for an outbound `do_get` request, in milliseconds.
+        /// This is a full-stream deadline for the whole RPC, not an idle timeout.
+        pub grpc_request_timeout_ms: usize, default = grpc_request_timeout_ms_default()
+        /// Maximum time a worker waits for task data to become available before failing the request.
+        pub wait_plan_timeout_ms: usize, default = wait_plan_timeout_ms_default()
+        /// TCP keepalive period used for worker-to-worker connections.
+        pub grpc_tcp_keepalive_ms: usize, default = grpc_tcp_keepalive_ms_default()
```
gabotechs, line 60: I think the numbers can be inlined directly here, we don't need intermediate functions

src/distributed_planner/distributed_config.rs
```diff
@@ -48,6 +48,16 @@ extensions_options! {
         /// The compression used for sending data over the network between workers.
         /// It can be set to either `zstd`, `lz4` or `none`.
         pub compression: String, default = "lz4".to_string()
+        /// Maximum time to wait while establishing a gRPC connection to a worker.
+        /// This is intended to bound connection setup to unreachable workers.
+        pub grpc_connect_timeout_ms: usize, default = grpc_connect_timeout_ms_default()
```
gabotechs, line 53: 🤔 This one is not really per-request. The connection happens only once, but once a worker to worker connection is stablished, is never broken. This means that this parameter will be respected just once on the first query after a deployment, and ignored until a restart of the worker process.

src/networking/channel_resolver.rs
```diff
@@ -84,14 +86,30 @@ pub fn get_distributed_channel_resolver(
...
+        .map(|cfg| cfg.grpc_connect_timeout_ms)
+        .unwrap_or(default_cfg.grpc_connect_timeout_ms);
+    let grpc_tcp_keepalive_ms = distributed_cfg
+        .map(|cfg| cfg.grpc_tcp_keepalive_ms)
+        .unwrap_or(default_cfg.grpc_tcp_keepalive_ms);
+    let key = DefaultChannelResolverKey {
+        runtime_addr: Arc::as_ptr(&task_ctx.runtime_env()) as usize,
+        grpc_connect_timeout_ms,
+        grpc_tcp_keepalive_ms,
+    };
```
gabotechs, line 106: This starts to get a bit hacky, it was hacky before already, but I don't think we should be doubling down on this pattern.

gabotechs: I think this is a limitation that could be better solved if the method in `ChannelResolver` accepted the `Arc<TaskContext>` in which they are running on as arguments.

src/flight_service/do_get.rs
```diff
@@ -58,10 +58,17 @@ impl Worker {
...
     ) -> Result<Response<<Worker as FlightService>::DoGetStream>, Status> {
-        let body = request.into_inner();
+        let (metadata, _ext, body) = request.into_parts();
         let doget = DoGet::decode(body.ticket).map_err(|err| {
             Status::invalid_argument(format!("Cannot decode DoGet message: {err}"))
         })?;
+        let headers = metadata.into_headers();
+        let mut cfg = SessionConfig::default();
+        set_distributed_option_extension_from_headers::<DistributedConfig>(&mut cfg, &headers)
+            .map_err(datafusion_error_to_tonic_status)?;
```
gabotechs, line 68: It's a shame we need to pass the full `DistributedConfig` just for the `wait_plan_timeout_ms`.

If you ask me, I think I would not let the users configure the `wait_plan_timeout_ms`, this is an internal detail of how Distributed DataFusion works, and ideally users should not care about this.

src/flight_service/do_get.rs
gabotechs: It had some tests before, but we deleted them on purpose, as `do_get` is an internal detail that is susceptible to change, making this test break.

I can see how this test will break in the future even if the functionality remains the same, maybe this is better represented as an integration test?

src/networking/channel_resolver.rs
```diff
@@ -127,21 +164,26 @@ impl Default for DefaultChannelResolver {
...
+            grpc_tcp_keepalive_ms,
         }
     }
-}
 
-impl DefaultChannelResolver {
     /// Gets the cached [BoxCloneSyncChannel] for the given URL, or builds a new one.
     pub async fn get_channel(&self, url: &Url) -> Result<BoxCloneSyncChannel, DataFusionError> {
+        let grpc_connect_timeout_ms = self.grpc_connect_timeout_ms;
+        let grpc_tcp_keepalive_ms = self.grpc_tcp_keepalive_ms;
```
gabotechs, line 175: These two parameters are only relevant for if users decide to live with the `DefaultChannelResolver`, but I do would expect the majority of users to have their own `ChannelResolver`, as they will likely need to adapt this project to their own networking infrastructure.

This means that, for any user that does not rely on `DefaultChannelResolver`, these two `grpc_connect_timeout_ms` and `grpc_tcp_keepalive_ms` will pretty much be dead parameters that do nothing to their queries.

From the two scenarios that I imagine:
- People do not want to customize their networking setup: they probably also do not care about tweaking gRPC configs, and they expect this project to ship good defaults
- People that do want to customize their networking setup: these parameters are not useful for them, as they build their gRPC clients on their own

src/networking/channel_resolver.rs
gabotechs: Because of https://github.com/datafusion-contrib/datafusion-distributed/pull/365/changes#r2976789668, I'm not sure if it's worth the complexity...

src/flight_service/do_get.rs
gabotechs: Or if you strongly think it's necessary, I'd probably just let it be configurable at the `Worker` level, so that we don't need to thread the `DistributedConfig` across inter-worker calls just for this single field.

src/distributed_planner/distributed_config.rs
```diff
@@ -48,6 +48,16 @@ extensions_options! {
...
+        /// Total timeout for an outbound `do_get` request, in milliseconds.
+        /// This is a full-stream deadline for the whole RPC, not an idle timeout.
+        pub grpc_request_timeout_ms: usize, default = grpc_request_timeout_ms_default()
```
gabotechs, line 56: 🤔 I imagine that users will have an API for serving their DataFusion queries, and they would already have a timeout on their API, and reaching the timeout in their API will automatically probably a cancellation recursively to all the workers.

What comes to mind is that, probably the timeout should not even be imposed by Distributed DataFusion, and should just be whatever timeout users naturally have in their APIs.

src/distributed_planner/distributed_config.rs
gabotechs, line 60: Same as https://github.com/datafusion-contrib/datafusion-distributed/pull/365/changes#r2976715396, the TCP keep alive will be taken into account once, and ignored until a worker restart.

gabotechs (commented):
😅 ended up leaving a bunch of comments.

The TL;DR is that I think there's different reasons for each individual timeout config to not have it in this project, left individual comments for the different configs, let me know what you think.

PR #367, opened 2026-03-08
Fix metrics display `bytes_transferred` as human-readable bytes

### LINKED ISSUE #361: Improve network boundary metrics display
There's some weirdness in how metrics are displayed in network boundaries:

```
    │           [Stage 2] => NetworkShuffleExec: output_partitions=12, input_tasks=3, metrics=[elapsed_compute_0=12.02s, elapsed_compute_1=6.45s, elapsed_compute_2=9.61s, bytes_transferred_0=1.91 B, bytes_transferred_1=1.16 B, bytes_transferred_2=1.27 B, max_mem_used_0=103.4 M, max_mem_used_1=8.31 M, max_mem_used_2=4.52 M, network_latency_avg_0=name:network_latency_avg 352.47ms, network_latency_avg_1=name:network_latency_avg 5.61ms, network_latency_avg_2=name:network_latency_avg 134.63ms, network_latency_first_0=name:network_latency_first 937.76µs, network_latency_first_1=name:network_latency_first 1.96ms, network_latency_first_2=name:network_latency_first 794.53µs, network_latency_max_0=name:network_latency_max 1.99s, network_latency_max_1=name:network_latency_max 39.55ms, network_latency_max_2=name:network_latency_max 792.54ms, network_latency_min_0=name:network_latency_min 617.75µs, network_latency_min_1=name:network_latency_min 347.36µs, network_latency_min_2=name:network_latency_min 274.88µs]
```

1. The bytes_transferred should be displaying Gb, and not Billions.
2. The network latencies value are serialized as `name:network_latency_avg 134.63ms` instead of `134.63ms`

EdsonPetry:
## Summary
- Introduces `BytesCounterMetric`, a custom metric type that formats values using `human_readable_size` (e.g., "16.0 KB") instead of `human_readable_count` (e.g., "1.91 B" where B means billions not bytes)
- Adds proto serialization support for the new metric type (`CustomBytesCount` variant, tag 29)
- Updates `worker_connection_pool.rs` to use `bytes_counter()` instead of `global_counter()` for the `bytes_transferred` metric
 
Closes #361

### DIFF
--- a/src/metrics/bytes_metric.rs
+++ b/src/metrics/bytes_metric.rs
@@ -11,6 +11,8 @@ use datafusion::{
 };
 use std::sync::atomic::Ordering::Relaxed;
 
+/// Extension trait for DataFusion's metric system that adds support for byte count metrics
+/// that display using human-readable byte sizes (KB, MB, GB) instead of plain count notation.
 pub trait BytesMetricExt {
     fn bytes_counter(self, name: impl Into<Cow<'static, str>>) -> BytesCounterMetric;
 }
@@ -25,6 +27,15 @@ impl BytesMetricExt for MetricBuilder<'_> {
         value
     }
 }
+/// A cumulative counter metric for tracking byte counts.
+///
+/// Unlike DataFusion's built-in [`Count`](datafusion::physical_plan::metrics::Count) which formats
+/// large values using plain count notation (e.g., "1.91 B" meaning 1.91 billion), this metric
+/// formats values using [`human_readable_size`] (e.g., "1.91 GB").
+///
+/// This avoids the confusing display where "B" (billions) looks like "bytes".
+///
+/// Aggregation sums values across partitions/tasks.
 #[derive(Debug, Clone)]
 pub struct BytesCounterMetric {
     bytes: Arc<AtomicUsize>,


src/metrics/bytes_metric.rs
```diff
@@ -0,0 +1,144 @@
...
+
+use datafusion::{
+    common::human_readable_size,
+    physical_plan::metrics::{CustomMetricValue, MetricBuilder, MetricValue},
+};
+use std::sync::atomic::Ordering::Relaxed;
+
+/// Extension trait for DataFusion's metric system that adds support for byte count metrics
+/// that display using human-readable byte sizes (KB, MB, GB) instead of plain count notation.
+pub trait BytesMetricExt {
```
gabotechs, line 16: There's already a bytes metric upstream, how about using that one instead of building a custom one?

EdsonPetry: I only see `OutputBytes` and `SpilledBytes` as the only byte aware metrics in upstream datafusion, if we want to have a custom name like `"bytes_transferred"` then I think we need to have our own custom metric. Let me know if I'm missing something.

gabotechs: You are right, actually, output_bytes is something different, so we cannot use that.

Then we are good, what you have here is perfect!

PR #368, opened 2026-03-11
Percentile latencies

gabotechs:
Adds some percentile latency metrics that are more suitable for observing network latencies.

### DIFF
--- a/src/flight_service/worker_connection_pool.rs
+++ b/src/flight_service/worker_connection_pool.rs
@@ -144,8 +144,8 @@ impl WorkerConnection {
         // Track end-to-end network latency distribution for all messages.
         let min_latency = MetricBuilder::new(metrics).min_latency("network_latency_min");
         let max_latency = MetricBuilder::new(metrics).max_latency("network_latency_max");
-        let p50_latency = MetricBuilder::new(metrics).avg_latency("network_latency_p50");
-        let p95_latency = MetricBuilder::new(metrics).avg_latency("network_latency_p95");
+        let p50_latency = MetricBuilder::new(metrics).p50_latency("network_latency_p50");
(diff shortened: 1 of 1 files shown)


src/metrics/latency_metric.rs
```diff
@@ -328,6 +370,107 @@ impl CustomMetricValue for FirstLatencyMetric {
     }
 }
 
+macro_rules! percentile_latency_metric {
+    ($name:ident, $percentile:expr) => {
+        #[derive(Clone)]
+        pub struct $name {
+            inner: Arc<Mutex<DDSketch>>,
```
gabotechs, line 377: I don't love having a mutex here, but I've created some microbenchmarks in a separate branch (that I don't think are worth committing to the repo) to make sure this is the fastest approach vs other options like Atomic-based time buckets, and it actually is, so keeping this.

src/flight_service/worker_connection_pool.rs
```diff
@@ -144,7 +144,8 @@ impl WorkerConnection {
         // Track end-to-end network latency distribution for all messages.
         let min_latency = MetricBuilder::new(metrics).min_latency("network_latency_min");
         let max_latency = MetricBuilder::new(metrics).max_latency("network_latency_max");
-        let avg_latency = MetricBuilder::new(metrics).avg_latency("network_latency_avg");
+        let p50_latency = MetricBuilder::new(metrics).avg_latency("network_latency_p50");
+        let p95_latency = MetricBuilder::new(metrics).avg_latency("network_latency_p95");
```
jayshrivastava: Typo. These should be p50_latency and p95_latency

gabotechs: Shit

gabotechs: Fixed now, thanks!

PR #369, opened 2026-03-16
Upgrade datafusion version to v52.3.0

jayshrivastava:


gene-bordegaray (approved):
Nice 💯

Cargo.toml
```diff
@@ -2,8 +2,8 @@
 members = ["benchmarks", "cli", "console"]
 
 [workspace.dependencies]
-datafusion = { version = "52.0.0", default-features = false }
-datafusion-proto = { version = "52.0.0" }
+datafusion = { version = "52.3.0", default-features = false }
+datafusion-proto = { version = "52.3.0" }
```
gabotechs, line 6: I don't think we should have merged this. What this is doing is restricting unnecessarily the number of compatible versions:
- Before: this project was compatible with `52.0.0`, `52.1.0`, `52.2.0`, `52.3.0`, etc...
- Now: this project is only compatible with `52.3.0`, forcing people that still depend on a previous version of DataFusion to either upgrade or not use the project.

It's a bit complicated how all this Cargo dependency resolution works, but here, we should adhere to minor and patch version numbers (not major though) as low as possible.

PR #372, opened 2026-03-17
Add automatic worker discovery to console

EdsonPetry:
The console now always discovers workers via `GetClusterWorkers` RPC. This change enables the console to automatically discover workers over the network. Instrumentation now works like so:

```rust
// console/examples/console_worker.rs
let localhost_resolver = Arc::new(LocalhostWorkerResolver { ports: ..., });
let worker = Worker::default();

Server::builder()
    .add_service(worker.with_observability_service(resolver)) // users now pass in their WorkerResolver impl
    .add_service(worker.into_flight_server())
    .serve(SocketAddr::new(
        IpAddr::V4(Ipv4Addr::LOCALHOST),
        args.port.unwrap_or(DEFAULT_WORKER_PORT),
    )).await?;
```

# Summary
- Add a new `GetClusterWorkers` gRPC to the observability service that returns all known worker URLs by delegating to the existing `WorkerResolver` trait
- Replace the console's manual `--cluster-ports` CLI flag with a `--connect` seed URL (defaults to port 9001)
- Wire up the observability service in the benchmarks EC2 worker and all console examples with a user-provided `WorkerResolver`
- Export `DEFAULT_WORKER_PORT` (9001) as the canonical default across the codebase
- Add a `console/README.md` and update all example docs for the new auto-discovery workflow

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -7,7 +7,10 @@ default-run = "dfbench"
 [dependencies]
 datafusion = { workspace = true }
 datafusion-proto = { workspace = true }
-datafusion-distributed = { path = "..", features = ["integration"] }
+datafusion-distributed = { path = "..", features = [
+  "integration",
+  "system-metrics",
+] }
 tokio = { version = "1.48", features = ["full"] }
 parquet = { version = "57.1.0" }
 structopt = { version = "0.3.26" }


console/src/main.rs
```diff
@@ -17,9 +18,11 @@ use url::Url;
...
 )]
 struct Args {
-    /// Comma-delimited list of worker ports (assumed localhost)
-    #[structopt(long = "cluster-ports", use_delimiter = true)]
-    cluster_ports: Vec<u16>,
+    /// URL of a worker to connect to for auto-discovery (e.g. http://localhost:9001).
+    /// The console calls GetClusterWorkers on this worker to discover the full cluster.
+    /// Defaults to http://localhost:9001.
+    #[structopt(long = "connect")]
+    connect: Option<Url>,
```
gabotechs: We can even make this a positional argument instead, I'd imagine this to be required almost all the time

src/lib.rs
```diff
@@ -46,12 +46,15 @@ pub use stage::{
...
+    ObservabilityServiceImpl, ObservabilityServiceServer, PingRequest, PingResponse,
+    StageKey as ObservabilityStageKey, TaskProgress, TaskStatus, WorkerMetrics,
 };
 
 pub use protobuf::StageKey;
 
 #[cfg(any(feature = "integration", test))]
 pub use execution_plans::benchmarks::ShuffleBench;
+
+pub const DEFAULT_WORKER_PORT: u16 = 9001;
```
gabotechs: I think it's better to not publicly expose this as a "default port". The port will be set by whatever the user decides to spawn its Tonic server, which can be any arbitrary port, so I don't think we can impose a default.

Thinking about it, this might need to be a required argument for starting the console binary.

PR #379, opened 2026-03-18
Fix system metrics never collecting due to async closure in `thread::spawn`

EdsonPetry:
This change fixes the background system metrics collection thread that silently never ran, causing CPU and memory to always report as zero in the console. Instead of having a dedicated background thread we have a dedicated tokio task handle system metrics collection on a worker's observability service.

### DIFF
--- a/src/observability/service.rs
+++ b/src/observability/service.rs
@@ -5,8 +5,6 @@ use datafusion::physical_plan::ExecutionPlan;
 use moka::future::Cache;
 use std::sync::Arc;
 #[cfg(feature = "system-metrics")]
-use std::thread;
-#[cfg(feature = "system-metrics")]
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


src/observability/service.rs
```diff
@@ -64,7 +64,7 @@ impl ObservabilityServiceImpl {
                         break;
                     };
 
-                    tokio::time::sleep(Duration::from_millis(100)).await
+                    thread::sleep(Duration::from_millis(100));
```
gabotechs: I'm still not super convinced about doing a thread::sleep here... I do think this has chances to block a thread in which tokio is expecting to schedule tasks.

What about spawning the task like this instead?

```rust
tokio::task::spawn(async move || { ... })
```

EdsonPetry: This works!

PR #375, opened 2026-03-18
Move to dedicated Worker gRPC specification instead of Arrow Flight

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/373

This PR moves away from using the Arrow Flight protobuf specification as application protocol for communicating workers in favor of using a dedicated protobuf specification. The rationale behind that can be seen in  https://github.com/datafusion-contrib/datafusion-distributed/issues/373.

This is a breaking API change that will prevent old workers work with new ones, so it should be carefully rolled out.

Part of the reason of moving to a dedicated protobuf specification is about commitment to API stability: we want to reduce as much as possible future breaking API changes, and we expect that maintaining a proper gRPC protobuf specification will help with that.

The migration should be simple:
- People will just need to rename some methods in their codebase (e.g., `into_flight_server` -> `into_worker_server`). Just following the Rust compiler here is enough
- During rollout, old workers should not talk to new ones, and viceversa, otherwise queries will fail. There's ongoing work for making this easier for people https://github.com/datafusion-contrib/datafusion-distributed/issues/380

Besides just moving to a dedicated protocol, this PR does two additional things:
- Rename `StageKey` to `TaskKey`, as it was uniquely identifying tasks and not stages https://github.com/datafusion-contrib/datafusion-distributed/pull/375/changes/882bec293ca95f392b49a4dd59ed2031509efd4b
- Rearrange the protobuf messages so that the do not send unnecessary stuff https://github.com/datafusion-contrib/datafusion-distributed/pull/375/changes/0f42b5fc114511db4b75662bbfe33451f69b15d2

---

If we take into account that ~710 LOC are now automatically generated out of a .proto file, we are looking at a `+724-924` net change, implying a total net reduction of lines of ~200, which sounds pretty good and expected.

---

Remote benchmarks against `main`.
TL;DR: the same

```
  ---
  TPC-H SF1 Results

  ┌───────┬───────────┬────────────────────────────┬──────────────┐
  │ Query │ main (ms) │ custom-worker-service (ms) │     Diff     │
  ├───────┼───────────┼────────────────────────────┼──────────────┤
  │ q1    │ 288       │ 249                        │ 1.16x faster │
  ├───────┼───────────┼────────────────────────────┼──────────────┤
  │ q2    │ 345       │ 412                        │ 1.19x slower │
  ├───────┼───────────┼────────────────────────────┼──────────────┤
  │ q3    │ 592       │ 612                        │ 1.03x slower │
[...]

PR #381, opened 2026-03-19
Add support for worker versioning via `GetWorkerInfo` RPC

### LINKED ISSUE #380: Worker Versioning
Today, users are expected to implement their own safeguard for workers with different incompatible versions to interact with each other. This can very easily happen during rollout: old workers get decommissioned while new workers get added to the cluster. If users do nothing, old workers will account the new ones as part of the cluster, and new ones will see old workers as part of the cluster.

This is problematic, because the interaction of several workers with different versions can result either in errors or wrong results.

The scope of this problem goes even beyond this project, there are several reasons why users might want to control the worker versioning themselves:
- Workers running the same version of Distributed DataFusion but a different versions of DataFusion itself
- Workers where new versions have custom UDFs registered but old ones still don't
- Any other reason this project can't control.

Because of this, users should have control of worker versioning, but the actual protocol for advertising worker versions can be perfectly served from this project.

Based on the protocol proposed in https://github.com/datafusion-contrib/datafusion-distributed/pull/375.

We could have something like this in the proto specification:

```proto
service WorkerService {
+ rpc WorkerInfo(WorkerInfoRequest) returns (WorkerInfoResponse);
  rpc SetPlan(SetPlanRequest) returns (SetPlanResponse);
}

message WorkerInfoRequest {
   // empty message, but declared anyways for retrocompatibility purposes, in case we want to add something in the future
}

message WorkerInfoResponse {
  // user-defined version of the worker
  string version = 1;
}
```

And the actual worker structure could be something like:

```rust
#[derive(Clone)]
pub struct Worker {
    pub(crate) version: String // defaults to an empty string
    ...
}

impl Worker {
    // Sets an arbitrary user-defined version for the worker
    pub fn with_version(mut self, version: String) -> Self;
}
```

As users are expected to implement their own `WorkerResovler::get_urls()` method, they probably would like to perform the version filtering there, in their own code. If there's any helper tool we can provide for that, it would be really nice, it can be as simple as a function that says something like:

```rust
async fn worker_has_version(url: Url, expected_version: &str) -> bool;
```

EdsonPetry:
This change implements worker versioning built on top of #375. 
- Adds `GetWorkerInfo` RPC to `worker.proto`
- Adds a version field to the `Worker` struct and a `with_version()` builder method
- Implements server-side `get_worker_info` handler that returns the worker's version
- Adds a `worker_has_version(channel_resolver, url, expected_version)` helper function for use inside `WorkerResolver` implementation to filter out incompatible workers during rolling deployments.

Closes #380 
Depends on #375

### DIFF
--- a/docs/source/user-guide/worker.md
+++ b/docs/source/user-guide/worker.md
@@ -162,8 +162,6 @@ use datafusion_distributed::{
     DefaultChannelResolver, GetWorkerInfoRequest, WorkerResolver, create_worker_client,
 };
 
-let worker_version = std::env::var("COMMIT_HASH").unwrap_or_default();
-
 struct VersionAwareWorkerResolver {
     compatible_urls: Arc<RwLock<Vec<Url>>>,
 }
@@ -214,10 +212,12 @@ With the resolver in place, wire it into the session and tag each worker with a 
 use datafusion::execution::SessionStateBuilder;
 use datafusion_distributed::{DistributedExt, DistributedPhysicalOptimizerRule, Worker};
 
+let worker_version = std::env::var("COMMIT_HASH").unwrap_or_default();
+
 // `known_urls` comes from your service discovery.
 let resolver = VersionAwareWorkerResolver::start_version_filtering(
     known_urls,
-    worker_version.to_string(),
+    worker_version.clone(),
 );
 
 let state = SessionStateBuilder::new()


gene-bordegaray:
can you base this off of #375  to shrink diff

src/networking/version.rs
```diff
@@ -0,0 +1,45 @@
...
+    let Ok(mut client) = channel_resolver.get_worker_client_for_url(url).await else {
+        return false;
+    };
+
+    let Ok(response) = client.get_worker_info(GetWorkerInfoRequest {}).await else {
+        return false;
+    };
+
+    response.into_inner().version_number == expected_version
+}
```
gabotechs: Wow, this is simple, I was imagining this to be more complicated.

This function is so simple, that I'm not sure if it's worth it to have it here at all, we probably can just not provide a function ourselves, and educate people in how to use this.

One thing that comes to mind is:
- Don't ship the function.
- Add an example in `examples/` that just educate people about how the `get_worker_info` endpoint is expected to be used.
- Add a section in the docs about production rollouts, and how to maintain zero-downtime using worker versioning.

I have the feeling that proper documentation is a better approach than the function, as people will be more free to do whatever they want.

EdsonPetry: I agree with this, adding in this function did feel almost *too* trivial.

gene-bordegaray: Question about the approach if we don't implement a function ourselves. Could we somehow warn the user or have some way of notifying a user if they say upgrade distributed datafusion and do not update their version.

Or will this just be the responsibility of the user?

I think it would be a nice feature. And in the worst case have a best practices guide in the docs

EdsonPetry: I think the cleanest way to do this in `examples/` is to just add a `--version` flag to both `localhost_run.rs` and `localhost_worker.rs` and handle versioning in there. I've also added a `worker_versioning.md` example to explain how to run the examples with worker versioning.

gabotechs: > Or will this just be the responsibility of the user?

It's likely going to be responsibility of the user to ensure correct versioning. There are multiple events that should trigger a version bump, for example: 
- The users are bumping `datafusion` version, but not `datafusion-distributed`

In that case, even if they are not touching `datafusion-distributed`, that still needs a version bump on their side, as the `datafusion` upgrade might have changed something in how the plan is executed (e.g. different intermediate schema between nodes)

So I don't think we are ever going to be able to decide on behalf of the users what's worth a "version upgrade".

gene-bordegaray: Talked with @gabotechs and he brought up a great approach/implementation for users to set this up and not have to worry about forgetting to bump their version (which was my main concern)

Doing this:
```rust
Worker::default().with_version(os::env("COMMIT_HASH"))
```

And explicitly stating this setup as preferred in onboarding documentation.

docs/source/user-guide/worker-resolver.md
```diff
@@ -70,4 +70,4 @@ It's up to you to decide how the URLs should be resolved. One important implemen
 
 A good example can be found
 in [benchmarks/cdk/bin/worker.rs](https://github.com/datafusion-contrib/datafusion-distributed/blob/main/benchmarks/cdk/bin/worker.rs),
-where a cluster of AWS EC2 machines is discovered identified by tags with the AWS Rust SDK.
\ No newline at end of file
+where a cluster of AWS EC2 machines is discovered identified by tags with the AWS Rust SDK.
```
EdsonPetry, line 73: newline removed by my IDEs markdown formatter

docs/source/user-guide/worker.md
```diff
@@ -95,3 +95,153 @@ async fn main() {
 ```
 
 The `into_worker_server()` method builds a `WorkerServiceServer` ready to be added as a Tonic service.
+
+## Worker Versioning
+
+Workers expose a `GetWorkerInfo` gRPC endpoint that reports metadata about the running worker,
```
EdsonPetry, line 101: Decided this wording was best since it's not unlikely that in the future `GetWorkerInfo` may send more than just a version string.

PR #389, opened 2026-03-24
Add support for worker versioning.

### LINKED ISSUE #380: Worker Versioning

EdsonPetry:
Closes #380

# Summary
- Add `GetWorkerInfo` gRPC endpoint to `worker.proto` so workers can report a user-defined version string, enabling coordinators to filter by version during rolling deployments
- Add `Worker::with_version()` builder method and export `GetWorkerInfoRequest/GetWorkerInfoResponse` from the crate root
- Document the full versioning workflow in `docs/`
- Update `localhost_worker` and `localhost_run` examples with `--version` flags

### DIFF
--- a/docs/source/user-guide/worker.md
+++ b/docs/source/user-guide/worker.md
@@ -116,11 +116,8 @@ let worker = Worker::default()
 ```
 
 One way to avoid forgetting to bump the version on each deploy, is to derive it from an environment variable
-set by your CI/CD pipeline:
+set by your infrastructure at runtime;
 
-```rust
-let worker = Worker::default()
-    .with_version(std::env::var("COMMIT_HASH").unwrap_or_default());
 ```
 
 ### Querying a worker's version
--- a/docs/source/user-guide/worker.md
+++ b/docs/source/user-guide/worker.md
@@ -116,8 +116,11 @@ let worker = Worker::default()
 ```
 
 One way to avoid forgetting to bump the version on each deploy, is to derive it from an environment variable
-set by your infrastructure at runtime;
+set by your infrastructure at runtime:
 
+```rust
+let worker = Worker::default()
+    .with_version(std::env::var("COMMIT_HASH").unwrap_or_default());
 ```
 
 ### Querying a worker's version
@@ -133,7 +136,7 @@ let channel = channel_resolver.get_channel(&worker_url).await?;
 let mut client = create_worker_client(channel);
 
 let response = client.get_worker_info(GetWorkerInfoRequest {}).await?;
-println!("version: {}", response.into_inner().version_number);
+println!("version: {}", response.into_inner().version);
 ```
 
 ### Zero-downtime rolling deployments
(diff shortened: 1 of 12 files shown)


docs/source/user-guide/worker.md
```diff
@@ -95,3 +95,191 @@ async fn main() {
...
+    .with_version("2.0.0");
+```
+
+One way to avoid forgetting to bump the version on each deploy, is to derive it from an environment variable
+set by your CI/CD pipeline:
+
+```rust
+let worker = Worker::default()
+    .with_version(std::env::var("COMMIT_HASH").unwrap_or_default());
+```
```
gabotechs, line 124: It'd be more correct to say an "environment variable set by your infrastructure at runtime". `std::env::var` reads environment variables during program execution, not during compilation or CI, and this is fine, as I don't think we want the env variable to be set at compile time.

docs/source/user-guide/worker.md
```diff
@@ -95,3 +95,191 @@ async fn main() {
...
+
+        if unresolved.is_empty() {
+            break;
+        }
+
+        // Fire all unresolved version checks concurrently.
+        let futures: Vec<_> = unresolved
+            .into_iter()
+            .map(|url| {
+                let channel_resolver = &channel_resolver;
```
gabotechs: 🤔 Does this compile on a normal scenario? I'd imagine that you'd need to do something like this here:

```rust
                let channel_resolver = Arc::clone(&channel_resolver);
```

docs/source/user-guide/worker.md
```diff
@@ -95,3 +95,191 @@ async fn main() {
...
+};
+
+struct VersionAwareWorkerResolver {
+    compatible_urls: Arc<RwLock<Vec<Url>>>,
+}
+
+/// Polls only unresolved workers and caches their versions permanently.
+/// Once every worker has responded, the loop exits, no further network calls.
+async fn background_version_resolver(
+    known_urls: Vec<Url>,
```
gabotechs: 🤔 I'm not sure if this setup is normal enough to be shown as part of the docs.

Typically, because of `WorkerResolver::get_urls()` being sync, people already need to spawn a background thread that polls for new URLs, and this function is suggesting another background thread on top of that for filtering the compatible URLs.

docs/source/user-guide/worker.md
```diff
@@ -95,3 +95,191 @@ async fn main() {
...
+    expected_version: String,
+    compatible_urls: Arc<RwLock<Vec<Url>>>,
+) {
+    let Some(capacity) = NonZeroUsize::new(known_urls.len()) else {
+        return; // No workers to resolve.
+    };
+
+    let channel_resolver = DefaultChannelResolver::default();
+    // Cache of URL : version for workers that have already responded.
+    let mut resolved = LruCache::new(capacity);
```
gabotechs: 🤔 There's another issue that I can imagine with this:

1. A pod with URL "http://1.1.1.1" gets decommissioned
2. Some time passes, 3 hours or so
3. the LruCache still did not reached capacity on those 3 hours
4. A brand new pod comes in, reusing the "http://1.1.1.1" url, as it's available
5. We miss out on resolving it again, because it was still stored in `resolved`.

I'm not sure if we can rely on a cache for this, as we'll hit these kind of problems. Probably we can just do a deep compassion of `prev_workers` VS `new_workers`, and just resolve the ones that are new.

docs/source/user-guide/worker.md
gabotechs: Here's an idea of how I would integrate the URL filtering in an existing worker resolver loop https://github.com/datafusion-contrib/datafusion-distributed/commit/e3f998e8d39660356fe08585cfeb858015b047b7.

docs/source/user-guide/worker.md
gabotechs: This is a problem that people might have, depending on their setup. For example, in my attempt in the EC2-based workers from the benchmarks: https://github.com/datafusion-contrib/datafusion-distributed/commit/e3f998e8d39660356fe08585cfeb858015b047b7.

The IPs there do not change between deploys, it's just the worker process the one that gets redeployed and restarted, so it can perfectly happen that even if the IP is exactly the same as before, we still need to perform a filtering step again...

gabotechs: Thinking about it, probably it's not a "new IP" event the one we are looking for towards evaluating if a new worker entered the cluster or not. It's probably a reconnection of the underlaying gRPC channel, which will happen even if a worker process restarts without changing IP.

This means that a general solution to this might be even more complicated than anticipated.

gabotechs: Here is an idea on how to detect if a worker disconnected (because the process restarted, or whatever):
https://github.com/hyperium/tonic/discussions/1243#discussioncomment-4959539

gabotechs: https://github.com/hyperium/tonic/discussions/1030#discussioncomment-3093529

gabotechs: This... gets a bit complicated. If users have the cluster setup in a way so that new workers always come with new IPs (which is true for most k8s based deployments), then this is not a problem, and people can move with their lives assuming that "new IP" == "new worker".

We probably can start with just that assumption, mention it in the docs, and elaborate in future PRs how this will work in more complicated setup where new workers can enter the cluster without changing IPs.

docs/source/user-guide/worker.md
EdsonPetry: I implemented `background_version_resolver` almost identically in the `benchmarks/cdk/worker.rs` and it does compile, `let channel_resolver = &channel_resolver;` creates a shared reference which is then moved in the following `async move` thereafter so each future is just holding a copied pointer to the original channel resolver on line 182.

docs/source/user-guide/worker.md
EdsonPetry: I made a note in the docs explaining that our example makes the assumption that a new worker comes with a new IP. I think for the time being this is fine but let me know if we prefer something different.

src/worker/worker_service.rs
```diff
@@ -196,4 +208,27 @@ impl WorkerService for Worker {
...
+pub async fn get_worker_version(
+    channel_resolver: Arc<dyn ChannelResolver + Send + Sync>,
+    url: &url::Url,
+) -> Result<String, DataFusionError> {
+    let mut client = channel_resolver.get_worker_client_for_url(url).await?;
+    let fut = client.get_worker_info(tonic::Request::new(GetWorkerInfoRequest {}));
+    let response = tokio::time::timeout(Duration::from_secs(2), fut)
+        .await
+        .map_err(|_| exec_datafusion_err!("Timeout getting worker info from {url}"))?
+        .map_err(|e| exec_datafusion_err!("Error getting worker info from {url}: {e}"))?;
```
gabotechs: There's a standard way of setting timeouts in `tonic` requests:

```rust
    let mut request = tonic::Request::new(GetWorkerInfoRequest {});
    request.set_timeout(Duration::from_secs(2));
    let fut = client.get_worker_info(request);
```

Although I still think that this function is so simple that it's not worth having here. If you ask me, I'd just delete it.

EdsonPetry: Sounds good, I'll opt to delete it for simplicity.

gabotechs: I mean, deleting the whole function. I think it's so simple that people can just code the content themselves for now.

examples/localhost_run.rs
gabotechs, line 1: It'd be nice if we can keep the existing `localhost_run.rs` and `localhost_worker.rs` simple as before, and have either a specific example dedicated to versioning. What do you think?

examples/localhost_versioned_run.rs
gabotechs, line 1: 👍 nice!

gabotechs (approved):
💯 Awesome, thanks @EdsonPetry!

Issue #384, opened 2026-03-23
Upgrade`datafusion` to 53.0.0


gabotechs:
DataFusion version 53 will be released soon:
- https://github.com/apache/datafusion/issues/19692

I made an attempt to upgrade on https://github.com/datafusion-contrib/datafusion-distributed/pull/366, which revealed some issues upstream https://github.com/apache/datafusion/issues/20779 that are now solved.

PR #391, opened 2026-03-28
Upgrade to datafusion 53.0.0

sandugood:
Upgraded and refactored several parts to be compatible with datafusion v53.0.0


sandugood:
Tackles #384

Cargo.toml
```diff
@@ -2,26 +2,26 @@
 members = ["benchmarks", "cli", "console"]
 
 [workspace.dependencies]
-datafusion = { version = "52.0.0", default-features = false }
-datafusion-proto = { version = "52.0.0" }
+datafusion = { git = "https://github.com/apache/datafusion", branch = "branch-53", default-features = false }
+datafusion-proto = { git = "https://github.com/apache/datafusion", branch = "branch-53" }
```
gabotechs: Nice! This looks good. I think it should be fine to go ahead now and add do:

```rust
datafusion = { version = "53.0.0", default-features = false }
datafusion-proto = { version = "53.0.0" }
```

sandugood: Fixed, thanks!

Cargo.lock
```diff
@@ -217,19 +217,40 @@ version = "57.3.0"
...
+ "arrow-cast 58.1.0",
+ "arrow-csv 58.1.0",
+ "arrow-data 58.1.0",
+ "arrow-ipc 58.1.0",
+ "arrow-json 58.1.0",
+ "arrow-ord 58.1.0",
+ "arrow-row 58.1.0",
+ "arrow-schema 58.1.0",
+ "arrow-select 58.1.0",
+ "arrow-string 58.1.0",
```
gabotechs: 🤔 There's something wrong here, it looks like we have two simulatenous versions of Arrow crates. Maybe there's a dependency missing an update?

sandugood: Yeah, fixed both arrow and parquet versions. Thanks!

benchmarks/Cargo.toml
```diff
@@ -24,10 +24,10 @@ futures = "0.3.31"
 dashmap = "6.0.1"
 prost = "0.14.1"
 url = "2.5.7"
-arrow-flight = "57.1.0"
+arrow-flight = "58.1.0"
```
gabotechs: Here, I'd probably just put `arrow-flight = "58"`, so that if people want to use arrow-flight 58.0.X they also can

benchmarks/Cargo.toml
```diff
@@ -24,10 +24,10 @@ futures = "0.3.31"
...
 tonic = { version = "0.14.1", features = ["transport"] }
 axum = "0.7"
-object_store = { version = "0.12.4", features = ["aws"] }
+object_store = { version = "0.13.2", features = ["aws"] }
```
gabotechs: Same as above, I think we can leave this as just `0.13`, same as what you did in the top level Cargo.toml

gabotechs (approved):
👍 awesome. Thanks @sandugood! as soon as CI is green, I'll merge this

sandugood:
Looks great, thank you!

PR #396, opened 2026-04-07
Leverage PartialReduce AggregationExec mode to reduce shuffle size

### LINKED ISSUE #360: Leverage PartialReduce AggregationExec mode to drastically reduce shuffle size
# Summary

When DataFusion 53 is released, we should use the `PartialReduce` mode of `AggregateExec` (https://github.com/apache/datafusion/pull/20019, kudos to @njsmith) for maximum local data reduction before shuffling.

The expectation is reducing the amount of data exchanged by up to a factor of `n_partition_per_task` (for example 16x).

# Details

Very schematically, here is how a shuffle happens:

Image

My understanding is that we will send duplicate aggregate keys on the network, where we could actually have reduced them to one line locally before the shuffle.

We should leverage `PartialReduce` to further reduce data before shuffling:

Image

In this example, we'd reduce data exchanged by 4x (because there are 4 partitions per task).

No strong opinion on whether we should add another `RepartitionExec` layer like in this graph, or if we should run the `AggregateExec(PartialReduce)` on the "ready to shuffle" partitions. Since shuffles can have hundreds of partitions, I don't know what is the comparative cost of:
- `RepartitionExec(Hash, output_partitions=16)` + `AggregateExec(PartialReduce)` on 16 partitions
- vs `AggregateExec(PartialReduce)` on potentially hundreds of partitions

nchu05:
Closes #360 

### Summary
Adds AggregationExec(PartialReduce) from Datafusion 53 to reduce the amount of data exchanged over the network. This PR implements approach 2 from the linked [issue](https://github.com/datafusion-contrib/datafusion-distributed/issues/360), which adds the Agg(PartialReduce) above RepartitionExec without the addition of another RepartitionExec layer.

### Implementation

The optimizer pass (`partial_reduce_below_network_shuffles`) runs after `distribute_plan` and before batch coalescing. It transforms the producer stage plan from:

```
NetworkShuffleExec
  RepartitionExec(Hash(group_keys, N×T))
    AggregateExec(Partial)
```

to:

```
NetworkShuffleExec
  AggregateExec(PartialReduce)              ← NEW
    RepartitionExec(Hash(group_keys, N×T))  ← unchanged
      AggregateExec(Partial)
```

`PartialReduce` runs on the already hash-partitioned data, where all rows with the same group key are in the same partition. It merges duplicate partial aggregates (one per input partition) into a single row per key, reducing the data volume before the network shuffle.

### Configuration

The optimization is gated behind the `distributed.partial_reduce` flag (default `false`), since it benefits high-cardinality group-bys but adds overhead on low-cardinality ones. Enable it via:

```sql
SET distributed.partial_reduce = true;
```

or in Rust:

```rust
SessionConfig::new().with_distributed_partial_reduce(true)?
```

### Benchmarks (TPC-H SF50, 4 workers, 22 queries)

| Config | Total time | vs baseline |
|--------|-----------|-------------|
| `partial_reduce=false` (baseline) | 116,413ms | — |
| `partial_reduce=true` | 115,541ms | **1.01x faster** |
| `partial_reduce=true, max_tasks_per_stage=16` | 114,195ms | **1.02x faster** |
| `partial_reduce=true, shuffle_batch_size=131072` | 135,983ms | 1.17x slower |

The overall speedup is modest at the suite level because several queries have low-cardinality group-bys (q1, q4, q18, q20) that regress, offsetting wins on high-cardinality ones (q3: +6%, q17: +15%). Larger shuffle batches are uniformly harmful. Higher task fanout (`max_tasks_per_stage=16`) compounds the benefit on aggregation-heavy queries and gives the best total result.


src/distributed_planner/partial_reduce_below_network_shuffles.rs
```diff
@@ -0,0 +1,195 @@
...
+            │       PartitionIsolatorExec: tasks=3 partitions=3
+            │         DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[RainToday], file_type=parquet
+            └──────────────────────────────────────────────────
+        ");
+    }
+
+    #[tokio::test]
+    async fn non_aggregation() {
+        let explain = sql_to_explain(r#"SELECT * FROM weather LIMIT 10"#).await;
+        assert!(
```
gabotechs: DataFusion has the `assert_contains!` macro that you can use here.

src/distributed_ext.rs
```diff
@@ -647,6 +656,22 @@ impl DistributedExt for SessionConfig {
...
+            set_distributed_option_extension(
+                self,
+                DistributedConfig {
+                    partial_reduce: enabled,
+                    ..Default::default()
+                },
+            );
+        }
+        Ok(())
+    }
```
gabotechs, line 663: I'd look at the pattern on how other flags are set and just copy the same code here, unless there's a good reason for diverging.

src/distributed_planner/partial_reduce_below_network_shuffles.rs
```diff
@@ -0,0 +1,195 @@
...
+            .with_distributed_planner()
+            .with_distributed_worker_resolver(InMemoryWorkerResolver::new(3))
+            .build();
+
+        let ctx = SessionContext::new_with_state(state);
+        register_parquet_tables(&ctx).await.unwrap();
+        let df = ctx.sql(query).await.unwrap();
+        let physical_plan = df.create_physical_plan().await.unwrap();
+        display_plan_ascii(physical_plan.as_ref(), false)
+    }
```
gabotechs: Note how in the rest of the project we place the test helper functions at the bottom. This is because readers review code top-down, and it's more interesting to see first the actual tests than the helpers.

gabotechs (commented):
Looking good! Looks like there's some conflicts that need solving

src/distributed_planner/partial_reduce_below_network_shuffles.rs
nchu05: Wasn't aware of this - thanks!

src/distributed_planner/partial_reduce_below_network_shuffles.rs
nchu05: Sounds good, thanks

src/distributed_planner/partial_reduce_below_network_shuffles.rs
```diff
@@ -0,0 +1,184 @@
...
+pub(crate) fn partial_reduce_below_network_shuffles(
+    plan: Arc<dyn ExecutionPlan>,
+    cfg: &ConfigOptions,
+) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    let partial_reduce_enabled = DistributedConfig::from_config_options(cfg)?.partial_reduce;
+
+    let transformed = plan.transform_up(|plan| {
+        if !partial_reduce_enabled {
+            return Ok(Transformed::no(plan));
+        }
```
gabotechs: If partial reduce is not enabled, there's no reason for traversing the plan, we can return the `partial_reduce_below_network_shuffles` early without entering the `transform_up` block.

src/distributed_ext.rs
```diff
@@ -526,6 +526,13 @@ pub trait DistributedExt: Sized {
         &mut self,
         max_tasks_per_stage: usize,
     ) -> Result<(), DataFusionError>;
+
+    /// Enables or disables the PartialReduce optimization, which inserts an extra aggregation
+    /// pass above hash RepartitionExec before network shuffles to reduce shuffle data size.
+    fn with_distributed_partial_reduce(self, enabled: bool) -> Result<Self, DataFusionError>;
```
gabotechs, line 534: I'd add a small comment here and in `distributed_config.rs` stating why this is not always enabled, pretty much saying that depending on the work load this can either be beneficial or not.

gabotechs (approved):
Nice work @nchu05!

gabotechs:
I'd also make sure CI passes before merging

Issue #378, opened 2026-03-18
Assign specific worker URLs to specific tasks


gabotechs:
Today, worker URL assignation is random, and users have no say over which URL is assigned to each worker.

### How it works today

The process since there's a valid single-node plan until there's a distributed plan with worker URLs assigned is the following:
1. Users tell the distributed planner what is an appropriate amount of tasks to execute each node (typically only done for leaf nodes)

```
          ┌──────────────────┐          
          │CoalescePartitions│          
          └──────────────────┘          
          ┌──────────────────┐          
          │ Aggregate(final) │          
          └──────────────────┘          
          ┌──────────────────┐          
          │   Repartition    │          
          └──────────────────┘          
          ┌──────────────────┐          
          │Aggregate(partial)│          
          └──────────────────┘          
          ┌──────────────────┐          
          │      Filter      │          
          └──────────────────┘          
          ┌──────────────────┐          
          │      Union       │          
          └──────────────────┘          
┌──────────────────┐┌──────────────────┐
│   DataSource1    ││   DataSource2    │
└──────────────────┘└──────────────────┘
    TaskEstimator:       TaskEstimator: 
    I want 2 tasks       I want 3 tasks 
```

2. The distributed planner performs some task count reconciliation in each stage, as there might be multiple nodes contributing an estimated task count for the same stage:

```                   
                                            
                   ...                      
┌ ─ ─ ─ ─ ─                     ─Same stage 
           ┌──────────────────┐            │
│          │   Repartition    │ 3 tasks     
           └──────────────────┘            │
│          ┌──────────────────┐             
           │Aggregate(partial)│ 3 tasks    │
│          └──────────────────┘             
           ┌──────────────────┐            │
│          │      Filter      │ 3 tasks     
           └──────────────────┘            │
│          ┌──────────────────┐             
           │      Union       │ 3 tasks    │
│          └──────────────────┘             
 ┌──────────────────┐┌──────────────────┐  │
││   DataSource1    ││   DataSource2    │   
 └──────────────────┘└──────────────────┘  │
│       3 tasks            3 tasks          
 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
```

3. Network boundaries with specific input task counts are injected, but with empty slots for worker URLs

```
                               ┏━━━━━━━━━━━━━━━━━━┓                                   
[...]

PR #397, opened 2026-04-08
TaskRouter

JSOD11:
Creates the `TaskRouter`, an interface for users to define a routing function for mapping worker URLs to tasks. See 
 https://github.com/datafusion-contrib/datafusion-distributed/issues/378 for the full rationale behind this change.

Allowing users of distributed datafusion to map urls to tasks will allow for consistent routing of queries. This is essential if users want to e.g. cache files locally worker disks — consistent routing will cause a given worker machine to be routed to read the same files, meaning that the cache is taken advantage of.

### DIFF
--- a/src/distributed_ext.rs
+++ b/src/distributed_ext.rs
@@ -229,14 +229,6 @@ pub trait DistributedExt: Sized {
         resolver: T,
     );
 
-    /// Allows users to select the worker that should execute each task when a distributed query
-    /// is about to run.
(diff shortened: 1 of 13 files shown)


JSOD11:
Closed in favor of https://github.com/datafusion-contrib/datafusion-distributed/pull/409, for the reasons described in https://github.com/datafusion-contrib/datafusion-distributed/issues/378.

PR #399, opened 2026-04-14
Introduce WorkUnit concept  for streaming units of work to leaf nodes at runtime.

### LINKED ISSUE #374: [draft] stream metadata from DistributedExec to leaf nodes
There is a common pattern is some large-scale APIs that this project currently does not account for: a two step API contract where:
- The first step discovers metadata related to API calls or files that should be retrieved from somewhere else.
- The second step issues queries or reads files based on the metadata collected by the first API.

Currently, the first step can happen in two places:
1) At planning time, blocking the start of the second step until the first step has fully completed
2) At runtime, which allows kickstarting the second step as the first discovery step makes progress in a "streaming fashion"

Let's assume 1) is not possible due to performance limitations. 2) is still a valid option in single-node DataFusion, however, it implies an additional challenge in Distributed DataFusion:
- There is no shared memory between leaf stage tasks, and therefore, the metadata stream from step one needs to be executed in it's own "silo" (a distributed task), missing out on the opportunity to distribute metadata evenly across leaf stage tasks.
- The task count estimation most likely needs to take into account this metadata to provide a relevant number of tasks. In single-node DataFusion the answer to this is very straight forward: "let's just use all the CPUs in the machine, it's free sending data across threads". However, in Distributed DataFusion there do is a high price to pay in distributing unnecessarily.

Trino handles this pretty well with https://trino.io/docs/current/develop/connectors.html#connectorsplitmanager. There, the coordinator is capable of dynamically generating splits, that are then picked up by workers and executed as they become available.

If we assume that a DataFusion Distributed task is the equivalent to a Trino split, the added value in Trino approach is that it's able to discover splits dynamically at runtime, rather than statically at planning time.

---

How this would look like in Distributed DataFusion? my guess is that it would be something like this:

```
             ┌─────────────────────────────────────────┐                
             │             DistributedExec             │                
             │                                         │                
             │                                         │                
             │ ┌──────────────────────────────────────┐│                
             │ │  DataSourceExec Metadata Discovery   ││                
             │ └──────────────────────────────────────┘│                
             │ ┌────────┐┌────────┐┌───┐┌───┐┌───┐     │                
             │ │Metadata││Metadata││   ││   ││   │ ... │                
             │ └───┬────┘└────┬───┘└─┬─┘└─┬─┘└─┬─┘     │                
             └─────┼──────────┼──────┼────┼────┼───────┘                
                   │          │      │    │    │                        
         ┌─────────┘          │      └────┼────┼──────────────┐         
         │ ┌──────────────────┼───────────┘    │              │         
         │ │                  │ ┌──────────────┘              │         
         │ │                  │ │                             │         
         ▼ ▼                  ▼ ▼                             ▼         
┌─────────────────┐  ┌─────────────────┐             ┌─────────────────┐
│     Task 1      │  │     Task 2      │             │     Task N      │
│                 │  │                 │             │                 │
│                 │  │                 │             │                 │
│                 │  │                 │             │                 │
│                 │  │                 │             │                 │
│                 │  │                 │             │                 │
│ ┌──────────────┐│  │┌──────────────┐ │             │ ┌──────────────┐│
│ │DataSourceExec││  ││DataSourceExec│ │             │ │DataSourceExec││
│ └──────────────┘│  │└──────────────┘ │             │ └──────────────┘│
└─────────────────┘  └─────────────────┘             └─────────────────┘
```

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/374

---

TODO:
- [x] Fix the interaction between `WorkUnitFeedExec` and `ChildrenIsolatorUnionExec`
- [x] Create documentation: Besides the documentation in the code, I think it's worth not documenting in the web page yet in this PR, and instead do a soft launch of the feature. As we experiment with it in our systems, we might want to tweak it a bit before documenting it properly.

---

This PR introduces the concept of a `WorkUnit`, which is pretty much anything the user wants to that is susceptible too be fed into leaf nodes at runtime for performing execution.

A `WorkUnit` is just anything representable in protobuf format:
- A Parquet file address in S3 that must be read.
- An HTTP query that should be issued to an external API.
- An external database query that should be executed externally.
- Etc...

## Disclaimer

This PR is only relevant for cases when all the work units cannot be known at planning time. If a project can afford to resolve all work units at planning time and bake them into the execution plans themselves (e.g., the classical `FileScanConfig`), it's better to move on without this feature, as it's just opt-in.

## Why is this new concept introduced?

In order to understand what this new concept brings to the table, it's important to understand what we have today in plain DataFusion:

Let's look for example at a Parquet file source. When reading Parquet from a remote storage, all the Parquet files are resolved at planning time, and they are backed into the data source node as `FileGroup`s, which looks something like this:

```rust
┌─────────────────────────────────────────────────────┐
│                                                     │
│                                                     │
│              ... rest of the plan ...               │
│                                                     │
│                                                     │
└───────▲────────────▲────────────▲────────────▲──────┘
        │            │            │            │
        │            │            │            │
        │            │            │            │
┌───────┼────────────┼────────────┼────────────┼──────┐
│       │            │MyDataSource│            │      │
│ ┌─────┴────┐ ┌─────┴────┐ ┌─────┴────┐ ┌─────┴────┐ │
│ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │ ( ) Parquet file
│ │( )( )    │ │( )       │ │( )       │ │( )( )    │ │     
│ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │     
└─────────────────────────────────────────────────────┘
```

This works fine when the number of files to read is in the order of magnitude suitable for a single machine, but on a distributed context, there can be millions of files, and discovering all of them might take time, time in which already discovered files can start executing.

This gets worst with other patterns different from files, for example, when there is a two step API where:
- The first API resolves addresses or query fragments that should be issued to a second API
- The second API returns for each query the actual data that should be computed within DataFusion

## How does this new concept solve the issue

What a `WorkUnit` brings to the table is that it's designed to be discovered and streamed at execution time to nodes, rather than fully discovered and stored at planning time.

Feeding `WorkUnit`s into nodes is handled by `WorkUnitFeedExec`, and it works like this:

```rust
┌─────────────────────────────────────────────────────┐
│                                                     │
│                                                     │
│              ... rest of the plan ...               │
│                                                     │
│                                                     │
└─────────▲────────────▲────────────▲────────────▲────┘
          │            │            │            │
          │            │            │            │
┌─────────┼────────────┼────────────┼────────────┼────┐
│         │        WorkUnitFeedExec │            │    │
│ ┌───────┴──┐ ┌───────┴──┐ ┌───────┴──┐ ┌───────┴──┐ │
│ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │
│ │          │ │          │ │          │ │          │ │
│ └──┬────▲──┘ └──┬────▲──┘ └──┬────▲──┘ └──┬────▲──┘ │
└────┼────┼──────( )───┼───────┼────┼──────( )───┼────┘
    ( )   │       │    │       │    │       │    │      ( ) WorkUnit (e.g., file address)
     │    │       │    │      ( )   │       │    │          Discovered at runtime and fed
┌───( )───┼───────┼────┼───────┼────┼──────( )───┼────┐     into the data source
│    │    │       │   MyDataSource  │       │    │    │
│ ┌──▼────┴──┐ ┌──▼────┴──┐ ┌──▼────┴──┐ ┌──▼────┴──┐ │
│ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │
│ │          │ │          │ │          │ │          │ │
│ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │
└─────────────────────────────────────────────────────┘
```

But the challenge does not end here. As we are executing in a distributed context, `WorkUnit`s need be suitable to be streamed over the wire to the worker nodes that are actually going to handle them. 

This works out of the box with no extra action item required from the user's side, the `DistributedExec` head node (which at this point we should probably start calling it `CoordinatorExec`), is responsible for capturing all the streams locally, and fanning them out over the wire to the appropriate workers. Something like this:

```rust
                                                                                  ┌─────────────────────┐
                                                                                  │ Coordinating stage  │
                     ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─└─────────────────────┘
                      ┌────────────────────────────────────────────────────────────────────────────────┐│
                     ││                                DistributedExec                                 │
                      │                                                                                ││
                     ││┌────────┐┌────────┐┌────────┐┌────────┐┌────────┐┌────────┐┌────────┐┌────────┐│
                      ││.feed(0)││.feed(1)││.feed(2)││.feed(3)││.feed(4)││.feed(5)││.feed(6)││.feed(7)│││
                     ││└────┬───┘└───┬────┘└────┬───┘└────┬───┘└───┬────┘└────┬───┘└───┬────┘└───┬────┘│
                      └─────┼────────┼──────────┼─────────┼────────┼──────────┼────────┼─────────┼─────┘│
                     └ ─ ─ ─│─ ─ ─ ─ ┼ ─ ─ ─ ─ ─│─ ─ ─ ─ ─│─ ─ ─ ─ ┼ ─ ─ ─ ─ ─│─ ─ ─ ─ ┼ ─ ─ ─ ─ ┼ ─ ─ ─
              ┌─────────────┘        │          │         │        │          │        │         └─────────────┐
              │           ┌──────────┘          │         │        │          │        └──────────┐            │
              │           │             ┌───────┘         │        │          └───────┐           │            │
              │           │             │           ┌─────┘        └─────┐            │           │            │
              │           │             │           │┌────────┐          │            │           │            │┌────────┐
              │           │             │           ││Worker 1│          │            │           │            ││Worker 2│
     ┌ ─ ─ ─ ─│─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─│┴────────┘ ┌ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ─│┴────────┘
      ┌───────┼───────────┼─────────────┼───────────┼───────┐ │  ┌───────┼────────────┼───────────┼────────────┼───────┐ │
     ││       │  WorkUnitFeedExec(RemoteProvider)   │       │   ││       │  WorkUnitFeedExec(RemoteProvider)   │       │
      │ ┌─────▼────┐ ┌────▼─────┐ ┌─────▼────┐ ┌────▼─────┐ │ │  │ ┌─────▼────┐ ┌─────▼────┐ ┌────▼─────┐ ┌────▼─────┐ │ │
     ││ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │   ││ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │
      │ │          │ │          │ │          │ │          │ │ │  │ │          │ │          │ │          │ │          │ │ │
     ││ └────┬─────┘ └────┬─────┘ └────┬─────┘ └────┬─────┘ │   ││ └────┬─────┘ └────┬─────┘ └────┬─────┘ └────┬─────┘ │
      └──────┼────────────┼────────────┼────────────┼───────┘ │  └──────┼────────────┼────────────┼────────────┼───────┘ │
     │       │            │            │            │           │       │            │            │            │
         .feed(0)     .feed(1)     .feed(2)     .feed(3)      │     .feed(0)     .feed(1)     .feed(2)     .feed(3)      │
     │┌──────┼────────────┼────────────┼────────────┼───────┐   │┌──────┼────────────┼────────────┼────────────┼───────┐
      │      │            │ MyPlanExec │            │       │ │  │      │            │ MyPlanExec │            │       │ │
     ││ ┌────▼─────┐ ┌────▼─────┐ ┌────▼─────┐ ┌────▼─────┐ │   ││ ┌────▼─────┐ ┌────▼─────┐ ┌────▼─────┐ ┌────▼─────┐ │
      │ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │ │  │ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │ │
     ││ │          │ │          │ │          │ │          │ │   ││ │          │ │          │ │          │ │          │ │
      │ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │ │  │ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │ │
     │└─────────────────────────────────────────────────────┘   │└─────────────────────────────────────────────────────┘
      ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘  ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
```

This allows the users to just have this code snippet in their custom leaf node implementations:

```rust
    fn execute(
        &self,
        partition: usize,
        context: Arc<TaskContext>,
    ) -> Result<SendableRecordBatchStream> {
        let Some(work_unit_feed) = work_unit_feed::<MyCustomWorkUnit>(&context) else {
            return exec_err!("Missing MyCustomWorkUnit feed");
        };
    ...
    }
```

And the `WorkUnit` stream will be magically available to user code regardless of whether the node is getting executed in a distributed context or not.


src/worker/worker.proto
```diff
@@ -37,6 +41,28 @@ message SetPlanRequest {
...
+  // channels for the incoming work unit feeds need to be constructed.
+  //
+  // If no WorkUnitFeedExec nodes are present in the plan, this should be empty.
+  repeated WorkUnitFeedDeclaration work_unit_feed_declarations = 4;
+}
+
+message WorkUnit {
+  // Identifier of the node to which this work unit feed belongs to.
+  bytes id = 1;
+  // The partition index withing the node to which the work unit feed belongs to.
```
jayshrivastava: typo. withing

src/execution_plans/work_unit_feed.rs
```diff
@@ -0,0 +1,547 @@
...
+        self
+    }
+    fn encode_to_bytes(&self) -> Vec<u8> {
+        self.encode_to_vec()
+    }
+}
+
+pub(crate) type WorkUnitTx = UnboundedSender<Result<Box<dyn WorkUnit>>>;
+pub(crate) type WorkUnitRx = UnboundedReceiver<Result<Box<dyn WorkUnit>>>;
+pub(crate) type RemoteWorkUnitFeedRxs = HashMap<(Uuid, usize), Mutex<Option<WorkUnitRx>>>;
```
jayshrivastava, line 74: A global type alias in this crate for NodeID: Uuid might be a good idea. Justin's work with caching seems to need that too.

src/execution_plans/work_unit_feed.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// In a distributed context, this method is never called remotely inside of workers, it's
+    /// always called inside the coordinating stage (inside [crate::DistributedExec]), which
+    /// reaches out to the appropriate workers and streams the [WorkUnit] feed over the network.
+    ///
+    /// In a distributed context this method is called `P*T` times, where `P` is the number of
+    /// partitions of the node below, and `T` is the number of distributed tasks.
+    ///
+    /// The [crate::TaskEstimator::scale_up_leaf_node] method can be used to intercept a
+    /// [WorkUnitFeedExec] and tweak its [WorkUnitFeedProvider] implementation so that it returns
+    /// `P*T` [WorkUnit] streams.
```
jayshrivastava, line 264: > The [crate::TaskEstimator::scale_up_leaf_node] method can be used to intercept a
    /// [WorkUnitFeedExec] and tweak its [WorkUnitFeedProvider] implementation so that it returns
    /// `P*T` [WorkUnit] streams.

Still reviewing. Do we have an example or implementation of a `TaskEstimator` that does this?

src/execution_plans/work_unit_feed.rs
```diff
@@ -0,0 +1,547 @@
...
+    ) -> Result<SortOrderPushdownResult<Arc<dyn ExecutionPlan>>> {
+        let child = self.input();
+
+        match child.try_pushdown_sort(order)? {
+            SortOrderPushdownResult::Exact { inner } => {
+                let new_exec = Arc::new(self.clone()).with_new_children(vec![inner])?;
+                Ok(SortOrderPushdownResult::Exact { inner: new_exec })
+            }
+            SortOrderPushdownResult::Inexact { inner } => {
+                let new_exec = Arc::new(self.clone()).with_new_children(vec![inner])?;
```
jayshrivastava, line 541: Does `self.new_with_children` not work here?

src/execution_plans/work_unit_feed.rs
jayshrivastava: Ah it's in the `test_work_unit_feed.rs`

src/test_utils/test_work_unit_feed.rs
```diff
@@ -0,0 +1,428 @@
...
+/// where the first argument is a tag string, the second is the task count (integer),
+/// and the remaining arguments are comma-separated row counts for each partition's feed
+/// messages. An empty string means an empty partition (no messages). The number of
+/// partition arguments must be divisible by the task count — they are distributed evenly
+/// across tasks.
+///
+/// String encoding is used for partitions because DataFusion 52.x has a bug where array
+/// literal arguments are silently dropped by the table-function SQL planner.
+#[derive(Debug)]
+pub struct TestWorkUnitFeedFunction;
```
jayshrivastava, line 192: Very neat

src/test_utils/test_work_unit_feed.rs
```diff
@@ -0,0 +1,428 @@
...
+        Field::new("tag", DataType::Utf8, false),
+        Field::new("task", DataType::Int64, false),
+        Field::new("partition", DataType::Int64, false),
+        Field::new("letter", DataType::Utf8, false),
+    ]))
+}
+
+/// Table function that creates a `TestWorkUnitFeedExec`.
+///
+/// Called in SQL as: `SELECT * FROM test_work_unit('my_tag', 2, '3,1', '5', '2', '')`
```
jayshrivastava: I would add some example output. 

It look be a while to figure out what '3,1' means specifically.

tests/work_unit_feed.rs
```diff
@@ -0,0 +1,546 @@
...
+
+        Ok((plan_display, formatted.to_string()))
+    }
+
+    async fn build_state(ctx: WorkerQueryContext) -> Result<SessionState, DataFusionError> {
+        Ok(ctx
+            .builder
+            .with_distributed_user_codec(TestWorkUnitFeedExecCodec)
+            .build())
+    }
```
jayshrivastava, line 623: nit: a network broadcast test would be nice.

jayshrivastava: For follow up work, should we find a way to hook this into tpch tests?

jayshrivastava (approved):
LGTM

No major concerns. 

`TaskEstimator` has a lot of responsibility now.

src/execution_plans/work_unit_feed.rs
gabotechs: yeap, that's the only example right now.

I think it should be possible to come up with an implementation that replaces a classical DataSourceExec + FileScanConfig combo. It would be interesting to try to implement a work stealing mechanism where free partitions from free workers are the ones who most quickly pickup new work, that would produce a very balance distribute of compute resources.

src/execution_plans/work_unit_feed.rs
gabotechs: Nope, because it has the following signature:
```rust
    fn with_new_children(
        self: Arc<Self>,
```

So you need an owned Arc<Self> reference, but here we only have &self. I pretty much copied this from `CooperativeExec`.

src/test_utils/test_work_unit_feed.rs
gabotechs: Makes sense. Added an example about this

tests/work_unit_feed.rs
gabotechs: Related to https://github.com/datafusion-contrib/datafusion-distributed/pull/399#discussion_r3112536061. If we had an implementation that replaces `DataSourceExec` + `FileScanConfig` combo, wiring this up to tpch would be trivial.

gabotechs: It might be worth adding an issue about this actually.

gabotechs: > nit: a network broadcast test would be nice.

Adding one now

gabotechs:
Before merging this, I want to test internally in Datadog's that these are the correct APIs. Things that I want to verify:

#### **Whether the right API for `WorkUnitFeedProvider` is this:**

```rust
    fn feed(
        &self,
        partition: usize,
        ctx: Arc<TaskContext>,
    ) -> Result<BoxStream<'static, Result<Box<dyn WorkUnit>>>>;
```

And not this

```rust
    fn feed(
        &self,
        feed_count: usize,
        ctx: Arc<TaskContext>,
    ) -> Result<Vec<BoxStream<'static, Result<Box<dyn WorkUnit>>>>>;
```

I can imagine how the second version can be more ergonomic, as there's guaranties that it will only be called once, avoiding patterns with `OnceCell` or `LazyLock` in downstream code for spawning a producer task only once.

#### **Whether `WorkUnitFeedExec` is better represented as it's own node in the plan (what is proposed in this PR), or it's better represented as a wrapper node that replaces the one that is supposed to feed.**

Instead of having this:

```rust
WorkUnitFeedExec: RowGeneratorFeedProvider: tasks=2, rows_per_partition=[[3], [2], [1], [2]]
  RowGeneratorExec
```

To have this:

```rust
WorkUnitFeedExec(RowGeneratorFeedProvider, RowGeneratorExec): tasks=2, rows_per_partition=[[3], [2], [1], [2]]
```

This will prevent weird patterns like `SortExec` being pushed through the provider, but not through the DataSource, for example, what we see in the tests here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/040b8e27fcf008ed581a18c3db884395b7eb98ac/tests/work_unit_feed.rs#L30-L33

gabotechs:
I'm trying this in our codebase, and there's several things that do not seem right regarding representing a work unit feed as an `ExecutionPlan` implementation:
- As the `WorkUnitFeedExec` now is responsible for telling the node below what should be executed, it's now also responsible to answer for a lot questions during planning: implementing `with_fetch()`, `fetch()`, `try_pushdown_filters`, `with_new_state`, etc...
- The `partition_statistics` needs information joined from both `WorkUnitFeedExec` and the `ExecutionPlan` below in order to provide something meaningful.
- The schema needs to be duplicated across both `WorkUnitFeedExec` and the `ExecutionPlan` below, as both likely need to deal with logic related to it.

I'm inclined towards thinking that representing a `WorkUnitFeedExec` as it's own node is not a good decision.

gabotechs:
Closing in favor of https://github.com/datafusion-contrib/datafusion-distributed/pull/411

PR #400, opened 2026-04-14
Factor out UUID functions into common module

gabotechs:
Just some preparatory refactor for keeping future PRs clean:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/399

This refactor, even if small, is not really related to that PR, so I factored it out.

### DIFF
--- a/src/common/mod.rs
+++ b/src/common/mod.rs
@@ -2,8 +2,10 @@ mod children_helpers;
 mod map_last_stream;
 mod on_drop_stream;
 mod task_context_helpers;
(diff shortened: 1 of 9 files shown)


jayshrivastava (approved):
LGTM but you're missing one call to `serialize_uuid` here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/709e6de46adac502b457f043fa6641f293e5887d/src/worker/test_utils/memory_worker.rs?plain=1#L17

PR #401, opened 2026-04-14
Refactor send plan task to a struct

gabotechs:
This is a preparatory PR for:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/399

It isolates a small refactor in order to contain the diff in the PR above. The logic should be the same as before.

The main purpose is to de-bloat the `DistributedExec::prepare_plan()` method.

### DIFF
--- a/src/execution_plans/distributed.rs
+++ b/src/execution_plans/distributed.rs
@@ -20,13 +20,14 @@ use datafusion::error::DataFusionError;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::physical_expr_common::metrics::MetricsSet;
 use datafusion::physical_plan::metrics::{
-    ExecutionPlanMetricsSet, Label, MetricBuilder, MetricValue, Time,
+    Count, ExecutionPlanMetricsSet, Label, MetricBuilder, MetricValue, Time,
 };
 use datafusion::physical_plan::stream::RecordBatchReceiverStreamBuilder;
 use datafusion::physical_plan::{DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties};
-use datafusion_proto::physical_plan::AsExecutionPlan;
+use datafusion_proto::physical_plan::{AsExecutionPlan, PhysicalExtensionCodec};
 use datafusion_proto::protobuf::PhysicalPlanNode;
 use futures::StreamExt;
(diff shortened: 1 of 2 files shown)


src/execution_plans/distributed.rs
```diff
@@ -92,18 +94,18 @@ impl DistributedExec {
...
-            .global_counter("plan_bytes_sent");
-
-        // Latency statistics about the network calls issued to the workers for feeding subplans.
-        let start = Instant::now();
-        let plan_send_latency = Arc::new(LatencyMetric::new(
-            "plan_send_latency",
-            |b| b.with_label(Label::new(DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, "0")),
-            &self.metrics,
-        ));
+        let metrics = CoordinatorToWorkerMetrics {
```
jayshrivastava, line 96: If you want to debloat more, you can move some of this to a constructor.

`CoordinatorToWorkerMetrics::new(&ExecutionPlanMetricsSet) {...}`

src/execution_plans/distributed.rs
```diff
@@ -256,27 +232,100 @@ impl ExecutionPlan for DistributedExec {
...
+        })
+    }
+
+    /// Instantiates and returns the task sends a serialized plan to specific worker. The returned
+    /// task is just a future that does nothing unless polled.
+    fn send_plan_task(
+        &self,
+        ctx: Arc<TaskContext>,
+        task_i: usize,
+        url: Url,
```
jayshrivastava: Consider passing `&mut JoinSet` here so this method spawns them.

jayshrivastava (approved):

src/execution_plans/distributed.rs
gabotechs: That can be a bit more complicated, because `CoordinatorToWorkerMetrics` are constructed once per plan, but `CoordinatorToWorkerTaskBuilder` is constructed once per network-boundary in the plan.

That being said, I bet there's ways to debloat it even more, but at least for now it's in a better state. I bet this will be worth revisiting before implementing the adaptative query execution work, as a lot of the responsibilities will fall in this code here.

src/execution_plans/distributed.rs
gabotechs: Good suggestion 👍

PR #404, opened 2026-04-15
fix: restore Stage plan during deserialization round-trip

### LINKED ISSUE #402: DistributedCodec loses Stage plan during serialization round-trip
## Summary

After #345, the `DistributedCodec` unconditionally sets `stage.plan = None` during deserialization, even when a plan was present before serialization. This breaks any architecture where the distributed plan goes through a serialize/deserialize round-trip before `DistributedExec::prepare_plan()` runs — for example, Arrow Flight SQL's two-phase `GetFlightInfo`/`DoGet` protocol.

We hit this in our system and wanted to share what we found along with a minimal proposed fix.

## Context

PR #345 changed plan distribution so that `DistributedExec::prepare_plan()` sends subplans to workers in parallel via `DoAction("init")`, then sets `stage.plan = None` on network boundary nodes. As part of this change, `parse_stage_proto` in the `DistributedCodec` was updated to always return `plan: None`, ignoring the deserialized children:

```rust
fn parse_stage_proto(
    proto: Option<StageProto>,
    _inputs: &[Arc<dyn ExecutionPlan>],  // ← children ignored
) -> Result<Stage, DataFusionError> {
    // ...
    Ok(Stage {
        // ...
        plan: None,  // ← always None
    })
}
```

This works fine when the plan is only ever held in memory and executed directly — which is the case for DFD's own test suite. But it breaks the codec's round-trip correctness.

> [!NOTE]  
> To use DFD in our system, we needed to make `DistributedCodec` (and the `protobuf` module) public — it's currently `pub(crate)` — so we could include it in our composite `PhysicalExtensionCodec` chain alongside our own custom codecs. This is a natural requirement for any Flight SQL integration: the server needs to serialize and deserialize the full physical plan (including DFD's network boundary nodes) between the planning and execution phases. We'd suggest making `DistributedCodec` public as part of DFD's API surface, since any Flight SQL (or similar stateless protocol) integration will need it.

## How we hit this

Our system uses Arrow Flight SQL, which has a two-phase protocol:

1. **`GetFlightInfo`** — the coordinator optimizes the query and produces a physical plan (including `DistributedExec` wrapping network boundary nodes with `plan: Some(...)`). The plan is serialized via the codec and stored in a ticket.

2. **`DoGet`** — the coordinator deserializes the plan from the ticket, then calls `DistributedExec::execute()` → `prepare_plan()`.

The problem is that between steps 1 and 2, the codec discards all stage plans. By the time `prepare_plan()` runs, every network boundary has `plan: None`, producing:

```
Internal error: Plan is not set for stage N.
```

Here's a diagram of the flow:

```
GetFlightInfo (Phase 1)
├─ Optimize SQL → physical plan
│  (network boundaries have plan = Some(subplan))
├─ Serialize full plan via DistributedCodec
│  (children ARE serialized by DataFusion's framework)
└─ Store serialized bytes in FlightInfo ticket
            │
            ▼
DoGet (Phase 2)
├─ Deserialize plan from ticket
│  ├─ DataFusion framework deserializes children correctly
│  └─ parse_stage_proto() ignores children, sets plan = None  ← BUG
├─ DistributedExec::execute()
│  └─ prepare_plan() finds plan = None
└─ ERROR: "Plan is not set for stage N"
```

Worth noting: the children bytes are fully serialized and transmitted during the round-trip — they're just discarded on the receiving end. The `children()` method on network boundary nodes already encodes plan presence correctly (returns `[plan]` when `Some`, `[]` when `None`), so DataFusion's serialization framework handles this naturally.

## Proposed fix

Since `children()` returns an empty vec when `plan` is `None` and a single-element vec when `plan` is `Some`, the deserialized `inputs` already carry the information needed to restore the plan. The fix is a one-liner with no proto schema changes:

```rust
fn parse_stage_proto(
    proto: Option<StageProto>,
    inputs: &[Arc<dyn ExecutionPlan>],
) -> Result<Stage, DataFusionError> {
    let Some(proto) = proto else {
        return Err(proto_error("Empty StageProto"));
    };
    let plan = inputs.first().cloned();  // ← infer from children
    Ok(Stage {
        query_id: uuid::Uuid::from_slice(proto.query_id.as_ref())
            .map_err(|_| proto_error("Invalid query_id in StageProto"))?,
        num: proto.num as usize,
        plan,
        tasks: decode_tasks(proto.tasks)?,
    })
}
```

This leverages DataFusion's existing serialization contract: children() controls what gets serialized, and the framework passes those deserialized children back as inputs during decoding. When plan is None, children() returns [], so inputs is empty — and vice versa.

We've validated this fix against our full integration test suite (including Flight SQL round-trips and tree reduction tests) and all tests pass. Two existing unit tests in `distributed_codec.rs` (`test_roundtrip_single_flight` and `test_roundtrip_single_flight_coalesce`) need a small update — they were passing `&[empty_exec()]` as inputs for stages with `plan: None`, which should be `&[]` to accurately reflect the serialization framework's behavior.

Happy to open a PR with this fix if it sounds good or discuss further!

gabrielkerr:
Closes #402

  ### Problem

  After #345, `parse_stage_proto` unconditionally sets `stage.plan = None` during deserialization, discarding children that DataFusion's framework correctly deserialized. This breaks any architecture
  where the distributed plan goes through a serialize/deserialize round-trip before `DistributedExec::prepare_plan()` runs — such as Arrow Flight SQL's two-phase `GetFlightInfo`/`DoGet` protocol.

  ### Fix

  `children()` on network boundary nodes already encodes plan presence: it returns `[plan]` when `Some`, `[]` when `None`. DataFusion's serialization framework preserves this through the round-trip and
   passes the deserialized children as `inputs`. The fix infers plan presence from `inputs` rather than ignoring them:

  ```rust
  plan: inputs.first().cloned(),
  ```

  ### Tests

  - Updated two existing tests (`test_roundtrip_single_flight`, `test_roundtrip_single_flight_coalesce`) to pass correct inputs for stages with `plan: None`
  - Added two new tests (`test_roundtrip_single_flight_with_plan`, `test_roundtrip_single_flight_coalesce_with_plan`) that validate the round-trip when a stage has `plan: Some(...)`
  
 ### Visibility changes
In a Flight SQL architecture, the coordinator server operates in two distinct phases separated by a network boundary:

  1. GetFlightInfo — optimizes the query and serializes the full physical plan (including DFD's network boundary nodes) into a ticket
  2. DoGet — deserializes the plan from the ticket and executes it

  This means the server must be able to serialize and deserialize DFD's plan nodes independently of DFD's internal machinery. To do this, it needs DistributedCodec in its PhysicalExtensionCodec chain — otherwise the server has no way to round-trip NetworkShuffleExec, NetworkCoalesceExec, etc. through the ticket.

  Similarly, a Flight SQL server that also acts as a DFD worker needs to implement the WorkerService gRPC trait. This requires WorkerServiceServer to register the service with tonic, FlightAppMetadata to read passthrough headers from requests, and Worker::impl_execute_task to delegate task execution. The error conversion functions (datafusion_error_to_tonic_status, etc.) are needed to translate between DFD's error types and gRPC status codes in the server's request handlers.

Currently these types are all pub(crate), which forces downstream Flight SQL integrations to fork the crate just to access them. Making them public allows DFD to be used as a library in these architectures without modification.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,8 @@ mod worker;
 mod distributed_planner;
 mod networking;
 mod observability;
-mod protobuf;
+pub mod protobuf;
+pub use protobuf::DistributedCodec;
 #[cfg(any(feature = "integration", test))]
 pub mod test_utils;
 
@@ -40,7 +41,9 @@ pub use stage::{
     DistributedTaskContext, ExecutionTask, Stage, display_plan_ascii, display_plan_graphviz,
     explain_analyze,
 };
+pub use worker::generated::worker::FlightAppMetadata;
 pub use worker::generated::worker::worker_service_client::WorkerServiceClient;
+pub use worker::generated::worker::worker_service_server::WorkerServiceServer;
 pub use worker::generated::worker::{GetWorkerInfoRequest, GetWorkerInfoResponse, TaskKey};
 pub use worker::{
     DefaultSessionBuilder, MappedWorkerSessionBuilder, MappedWorkerSessionBuilderExt, TaskData,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,7 @@ mod worker;
 mod distributed_planner;
 mod networking;
 mod observability;
-pub mod protobuf;
+mod protobuf;
 pub use protobuf::DistributedCodec;
 #[cfg(any(feature = "integration", test))]
 pub mod test_utils;
@@ -41,9 +41,7 @@ pub use stage::{
     DistributedTaskContext, ExecutionTask, Stage, display_plan_ascii, display_plan_graphviz,
     explain_analyze,
 };
-pub use worker::generated::worker::FlightAppMetadata;
 pub use worker::generated::worker::worker_service_client::WorkerServiceClient;
-pub use worker::generated::worker::worker_service_server::WorkerServiceServer;
 pub use worker::generated::worker::{GetWorkerInfoRequest, GetWorkerInfoResponse, TaskKey};
 pub use worker::{
     DefaultSessionBuilder, MappedWorkerSessionBuilder, MappedWorkerSessionBuilderExt, TaskData,
--- a/src/protobuf/distributed_codec.rs
+++ b/src/protobuf/distributed_codec.rs
@@ -60,7 +60,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
 
         fn parse_stage_proto(
             proto: Option<StageProto>,
-            _inputs: &[Arc<dyn ExecutionPlan>],
+            inputs: &[Arc<dyn ExecutionPlan>],
         ) -> Result<Stage, DataFusionError> {
             let Some(proto) = proto else {
                 return Err(proto_error("Empty StageProto"));
@@ -70,7 +70,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
                 query_id: uuid::Uuid::from_slice(proto.query_id.as_ref())
                     .map_err(|_| proto_error("Invalid query_id in StageProto"))?,
                 num: proto.num as usize,
-                plan: None,
+                plan: inputs.first().cloned(),
                 tasks: decode_tasks(proto.tasks)?,
             })
         }
@@ -556,6 +556,15 @@ mod tests {
         }
     }
 
+    fn dummy_stage_with_plan() -> Stage {
+        Stage {
+            query_id: Default::default(),
+            num: 0,
+            plan: Some(empty_exec()),
+            tasks: vec![],
+        }
+    }
+
     fn schema_i32(name: &str) -> Arc<Schema> {
         Arc::new(Schema::new(vec![Field::new(name, DataType::Int32, false)]))
     }
@@ -581,7 +590,7 @@ mod tests {
         let mut buf = Vec::new();
         codec.try_encode(plan.clone(), &mut buf)?;
 
-        let decoded = codec.try_decode(&buf, &[empty_exec()], &ctx)?;
+        let decoded = codec.try_decode(&buf, &[], &ctx)?;
         assert_eq!(repr(&plan), repr(&decoded));
 
         Ok(())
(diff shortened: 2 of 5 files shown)


src/protobuf/distributed_codec.rs
```diff
@@ -70,7 +70,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
                 query_id: uuid::Uuid::from_slice(proto.query_id.as_ref())
                     .map_err(|_| proto_error("Invalid query_id in StageProto"))?,
                 num: proto.num as usize,
-                plan: None,
+                plan: inputs.first().cloned(),
```
gabotechs, line 73: Pretty simple 👍

src/worker/impl_execute_task.rs
```diff
@@ -33,7 +33,7 @@ const RECORD_BATCH_BUFFER_SIZE: usize = 2;
 const WAIT_PLAN_TIMEOUT_SECS: u64 = 10;
 
 impl Worker {
-    pub(crate) async fn impl_execute_task(
+    pub async fn impl_execute_task(
```
gabotechs: 🤔 I don't think this should be public. This is an internal detail that can (and most likely will) significantly change over time.

src/lib.rs
```diff
@@ -12,7 +12,8 @@ mod worker;
 mod distributed_planner;
 mod networking;
 mod observability;
-mod protobuf;
+pub mod protobuf;
```
gabotechs: This seems to be exposing a big quantity of internal code to the public. Note how this project never exposes modules, and just cherry picks the specific structs, traits or functions that should be public.

gabotechs (commented):
The fix looks good 👍. 

It looks like there are several other changes about visibility unrelated to the fix in the diff. I recommend to open a separate PR for those instead so that we can evaluate them in isolation.

gabrielkerr:
> It looks like there are several other changes about visibility unrelated to the fix in the diff. I recommend to open a separate PR for those instead so that we can evaluate them in isolation.

Understood! I would still like to push for `DistributedCodec` to be public. Without this, downstream consumers can't include it in their codec chain to round-trip DFD's network boundary nodes.

gabotechs:
> Understood! I would still like to push for DistributedCodec to be public. Without this, downstream consumers can't include it in their codec chain to round-trip DFD's network boundary nodes.

Yeap! this one makes sense definitely, that one and `WorkerServiceServer` I think. `WorkerServiceServer` is indeed indirectly public already do through `Worker::into_worker_server()`, so I think it makes sense to expose it.

src/protobuf/mod.rs
```diff
@@ -2,7 +2,7 @@ mod distributed_codec;
 mod errors;
 mod user_codec;
 
-pub(crate) use distributed_codec::DistributedCodec;
+pub use distributed_codec::DistributedCodec;
```
gabotechs, line 5: 👍 makes sense

gabotechs (approved):

gabotechs:
Thanks @gabrielkerr! I'm merging this one for now. If you still find the need of making more things public feel free to open another PR and we can evaluate there.

PR #407, opened 2026-04-16
Remove batch_coalescing_below_network_boundaries rule

EdsonPetry:
## Summary

Removes the `batch_coalescing_below_network_boundaries` optimizer rule that inserted the DataFusion-53-deprecated `CoalesceBatchesExec` below every network boundary. Closes #386.

Repurposes `distributed.shuffle_batch_size` as a stage-level override for `datafusion.execution.batch_size` on worker-executed stages (it previously configured the deleted rule). When non-zero, `impl_set_plan.rs` rebuilds the stage's `SessionConfig` via `.with_batch_size(shuffle_batch_size)` before the `TaskContext` is constructed; `RepartitionExec` then picks it up at execute time via its internal `LimitedBatchCoalescer`. Default is `0` (no override). The `DistributedExt` setters are kept under the new semantics.

## Why
- DataFusion 53 deprecated `CoalesceBatchesExec`. The rule was carrying `#[expect(deprecated)]` to keep using it.
- Issue #386 hypothesized that the rule's data-copy overhead outweighed the wire-efficiency gains. **Confirmed empirically**, see benchmarks below.
- At default configuration the rule was already a no-op (`shuffle_batch_size=8192` ≤ `execution.batch_size=32768` -> rule early-returned). Real plans never got `CoalesceBatchesExec` inserted in production callers.
- Keeping the knob (per review feedback) lets users tune shuffle-path batch sizing independently of the global `batch_size`, since `RepartitionExec` still has its own `LimitedBatchCoalescer` driven by `session_config().batch_size()`.

## Benchmark findings

Ran on a 4-worker EC2 cluster in us-east-1. Three configurations exercised:

| Config | `batch_size` | `shuffle_batch_size` | `CoalesceBatchesExec` in TPC-H plans | `CoalesceBatchesExec` in TPC-DS plans |
|---|---|---|---:|---:|
| main @ defaults | 32,768 | 32,768 | **0** | **0** |
| main with rule firing | 32,768 | **65,536** | **137** | **1,091** |
| this PR @ defaults | 32,768 | 0 (no override) | **0** | **0** |
| this PR @ `batch_size=65,536` | **65,536** | 0 (no override) | 0 | 0 |

### Fair-comparison: matched effective wire batch size

Both sides produce ~65K-row batches over the wire. Main via the rule's post-hoc coalescing (`batch_size=32K, shuffle_batch_size=65K`); this PR via native batching (`batch_size=65K`).

| Suite | Rule-active main | This PR (`batch_size=65K`) | Δ |
|---|---:|---:|---|
| TPC-H SF10 | 74,138 ms | **66,037 ms** | **1.12× faster** |
| TPC-DS SF1 | 144,019 ms | **129,162 ms** | **1.12× faster** |

Top TPC-H wins: q18 −23%, q7 −19%, q17 −18%, q9 −16%, q22 −20%, q5 −13%.
Top TPC-DS wins: q34 −48%, q15 −43%, q32 −40%, q14 −24%, q72 −13% (absolute: −1s on a 7.7s query).

Confirms the hypothesis from #386: native-large-batch path beats coalesce-after-the-fact by ~12% across both suites. The rule was costing more in data copies than it saved in wire efficiency.


gene-bordegaray (commented):
I think we might want to keep a know still to allow users to tune the batch sizes for shuffles. It is still hapepning in the `RepartitionExec`

src/distributed_planner/distributed_config.rs
```diff
@@ -21,13 +21,6 @@ extensions_options! {
         /// - If a node reduces the cardinality of the data, this factor will decrease.
         /// - In any other situation, this factor is left intact.
         pub cardinality_task_count_factor: f64, default = cardinality_task_count_factor_default()
-        /// Upon shuffling over the network, data streams need to be disassembled in a lot of output
```
gene-bordegaray, line 24: Specifically, we could keep this config still to give a separate know for shuffle batch sizing along with the global one

src/worker/impl_set_plan.rs
```diff
@@ -64,6 +64,13 @@ impl Worker {
                     task_count: request.task_count as usize,
                 }));
             set_distributed_option_extension_from_headers::<DistributedConfig>(&mut cfg, &headers)?;
+
+            let shuffle_batch_size =
+                DistributedConfig::from_config_options(cfg.options())?.shuffle_batch_size;
+            if shuffle_batch_size != 0 {
+                cfg = cfg.with_batch_size(shuffle_batch_size);
+            }
+
```
gabotechs, line 73: 👍 Nice! this is a very good place to configure the batch size indeed

gene-bordegaray: ya noice job 💯

EdsonPetry: Thanks guys!

gabotechs (approved):
This looks good to me! 

It would be nice to get some more measurements with even bigger batch sizes. I'd be specially interested in knowing at witch `shuffle_batch_size` configuration the performance starts decreasing, so that we have a rough estimate about what's the local maxima for our current benchmarks.

We just need to solve some conflicts, and if @gene-bordegaray is good with this, we can ship it.

gene-bordegaray (approved):
This is looking good to me now @gabotechs ! 💯 

This will be also cool to play with in my shuffle investigations.

Nice work @EdsonPetry

gabotechs:
Merging now. Thanks @EdsonPetry for the PR and @gene-bordegaray for the review!

Issue #385, opened 2026-03-23
Improve shuffling performance


gabotechs:
The following PR added support for micro benchmarking shuffle operations:

- https://github.com/datafusion-contrib/datafusion-distributed/pull/362

The benchmarks shipped in that PR have several cases, and all of them shuffle the same amount of data, but the output tasks and partitioning is different. I'd expect all cases to take the same time, as they shuffle the same amount of data, but there is some variations:
```
Benchmarking shuffle/stream/1producer_1consumer_8partitions_1000000rows_1024batch_Nonecompression: 
                        time:   [90.704 ms 91.729 ms 94.261 ms]
shuffle/stream/1producer_1consumer_8partitions_1000000rows_1024batch_Some(LZ4_FRAME)compression: 
                        time:   [115.04 ms 125.10 ms 139.33 ms]
Benchmarking shuffle/stream/8producer_1consumer_8partitions_1000000rows_1024batch_Nonecompression: 
                        time:   [60.487 ms 60.780 ms 61.057 ms]
Benchmarking shuffle/stream/1producer_8consumer_8partitions_1000000rows_1024batch_Nonecompression: 
                        time:   [197.23 ms 197.69 ms 198.10 ms]
Benchmarking shuffle/stream/8producer_8consumer_8partitions_1000000rows_1024batch_Nonecompression: 
                        time:   [76.776 ms 77.825 ms 80.544 ms]
```

I think there should be some low hanging fruits regarding performance, and some other more complicated improvements:
- I'm suspicious about `arrow-flight` cloning data unnecessarily
- RepartitionExec might be doing something funny depending on the amount of output tasks

Rich-T-kid:
does #408 close this ticket? 
cc @gabotechs  @gene-bordegaray

gene-bordegaray:
> does #408 close this ticket? 
> cc @gabotechs  @gene-bordegaray 

No still working on this

gabotechs:
One idea comes to mind:

During network shuffles, perform two levels of `RepartitionExec`, with two different salts in the hashing expressions:

```
     ▲        ▲       ▲              ▲       ▲        ▲           ▲        ▲        ▲     
     │        │       │              │       │        │           │        │        │     
     │        │       │              │       │        │           │        │        │     
┌────┴────────┴───────┴─────┐  ┌─────┴───────┴────────┴────┐ ┌────┴────────┴────────┴────┐
[...]

PR #408, opened 2026-04-17
refactor / add benches for shuffle and transport shuffling

gene-bordegaray:
Builds off previous shuffle bench introduced in #362 and introduces new benches to help unveil some bottlenecks we are seeing in the `NetworkShuffleExec` as reported in #385 . The benches that are related include:

---

### `local_repartition`

Measures upstream `RepartitionExec` cost without any network boundary. This tells us how much fanout hurts before transport. Decided to add this here to have easy iteration in this repo as compared to upstream.

**Results:**
```text
- hash 1 -> 1: ~70.0 ms
- hash 1 -> 16: ~261.5 ms
- hash 8 -> 1: ~67.7 ms
- hash 8 -> 8: ~43.9 ms
- round_robin 1 -> 16: ~39.8 ms
- round_robin 8 -> 8: ~11.7 ms
```

---

### `transport`

Measures exchange cost with repartition removed. Compares in-memory vs TCP transport with different producer/consumer topologies. Using TCP vs in-memory helps us show if mains costs are coming from the excahnce without TCP.

**Results:**
```text
- in_memory 1 -> 1: ~116.5 ms
- tcp 1 -> 1: ~326.9 ms
- in_memory 8 -> 1: ~70.8 ms
- tcp 8 -> 1: ~149.8 ms
- in_memory 1 -> 16: ~406.9 ms
- tcp 1 -> 16: ~488.9 ms
- in_memory 8 -> 8: ~69.7 ms
- tcp 8 -> 8: ~265.1 ms
- tcp 1 -> 16 + lz4: ~580.2 ms
 ```

---

### `shuffle`

Measures the full `RepartitionExec` -> `NetworkShuffleExec` pipeline. This shows the full fanout cost.

**Results:**
```text
- 1 -> 1: ~158.5 ms
- 1 -> 1 + lz4: ~116.8 ms
- 8 -> 1: ~99.4 ms
- 1 -> 8: ~693.4 ms
- 1 -> 16: ~1.954 s
- 1 -> 16 + lz4: ~1.986 s
- 8 -> 8: ~215.9 ms
 ```

---

### Conclusions:
- upstream repartition fanout expensive cost
- the exchange / transport path also regresses under fanout even when upstream repartition is removed
- on the transport boundary, consumer fanout is the multiplier not producers
- the end-to-end `shuffle` path regresses much more drastically as both repartition and exchange costs together compound

### CODE REVIEW

File: src/worker/test_utils/worker_handles.rs [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+
+#[derive(Clone)]
+pub struct MemoryWorkerHandle {
+    task_index: usize,
+    worker: Worker,
+    schema: SchemaRef,
+    partitions_batches: Vec<Vec<RecordBatch>>,
+    compression: Option<CompressionType>,
+    channel: BoxCloneSyncChannel,
+}
```
gene-bordegaray, line 35: This replaces previous: `MemoryWorker` since it was a one-shot builder and the new handles give a reusable endpoint so we dont have to rebuild in each run

gabotechs (approved):
I have not given a very in-depth review, but rather than doing the classical review cycle, I'm more inclined in merging this, using it myself, and then suggest follow-ups if I see anything relevant.

Overall, it looks like a very high quality contribution! everything is pretty clear and easy to understand, and the abstractions and tools you've built for the different kind of benchmarks look very well thought.

So +1 for me 👍!

PR #409, opened 2026-04-17
Task Routing with the Task Estimator: Allow users to map tasks to URLs

### LINKED ISSUE #331: Allow task estimator to return errors
Ex. in `task_estimation`, I may want something like so
```
if let Some(my_custom_exec) = plan.as_any().downcast_ref::<MyCustomExec>() {
     ....
} else {
     internal_err!()
}
```
In case we want to error instead of falling back to single node mode.

JSOD11:
### Task Routing with the Task Estimator

This PR addresses https://github.com/datafusion-contrib/datafusion-distributed/issues/378.

We modify the task estimator interface here by adding a new `route_tasks` method which enables flexible routing of tasks to workers for e.g. consistent routing / cache locality.

If `route_tasks` returns `Ok(Some(Vec<Url>))`, then this vector of returned URLs represents the user's intended routing decision, and tasks are sent to the URLs in that vector in that specified order. If `route_tasks` returns `Ok(None)`, then tasks are routed to URLs according to a default round-robin implementation across the URLs with a random starting index. An error here indicates that the user's routing implementation ran into some error, and for that reason the routing failed.

### Design Process

Throughout the course of this design, multiple interfaces were proposed and @gabotechs and I discussed each one before landing on this one. To ensure full documentation, I explain how we came to this conclusion in the rest of this description.

Upon receiving a query, any query engine first goes through some process of creating a plan of attack for executing the query, and then after that plan is formulated and put into terms that are understandable for execution, the query is handed over for execution and executed. With respect to routing, there are various points in that process at which we could allow a user to inject routing decisions, each with its own set of tradeoffs. My original design was https://github.com/datafusion-contrib/datafusion-distributed/pull/397. One decision this design made was to route tasks to URLs at execution time. See the issue for the full discussion, but one point leading us away from that direction was the thought that integration into existing extensions rather than adding a new one here might be preferable for simplicity, and one option would be to simply expand the existing task estimator `scale_up_leaf_node` method's responsibilities with the ability to also route URLs.

As I implemented this solution, I noticed an issue. All of these steps occur through time, and the task estimator's estimations occur at planning time, at the point in time in which the single node query plan is originally being observed and morphed into the distributed plan. If we poll the cluster at planning time to determine the set of existing workers and then at execution time route tasks to those workers, for this to succeed we rely on an implicit assumption that the period of time between planning and execution is short and that therefore the state of the cluster should be the same at both points. This may not be true in practice, increasing the probability that the set of worker URLs might have changed in between planning and execution time, due to e.g. a machine going offline for some reason. So, to handle this case, we would then need to make design decisions about how to handle this situation. If a task should be routed to a URL, but that URL is no longer in the cluster, should we fail the query? Assign the task somewhere else? Assign them all round robin? Let the user decide?

To avoid this issue, @gabotechs and I discussed and determined that a good path to take would instead be to add `route_tasks` to the task estimator interface as a separate new method and apply the routing decision at execution time, with the set of received URLs `available_urls` coming from a poll to the method made just before execution. This way, we both minimize the new API surface area and make the routing decision at execution time, avoiding the issues described above.

### Future

We know of multiple use cases for this capability and anticipate that it will be important for users, so a follow-up could be to add an example for users, building on the snapshot tests I have provided.


gabrielkerr:
FWIW I'm using this proposed interface in my own fork and it's enabled us to do some cool stuff. Looking forward to any changes you make and getting it in!

JSOD11:
> FWIW I'm using this proposed interface in my own fork and it's enabled us to do some cool stuff. Looking forward to any changes you make and getting it in!

Great to hear! Before merging, I'm planning to add some integration testing which should help us determine whether the interface is precisely how we want it. But this is a great signal, thanks!

src/distributed_planner/distribute_plan.rs
```diff
@@ -104,67 +105,130 @@ fn _distribute_plan(
...
         }
         // This is a normal intermediate plan, just pass it through with the mapped children.
-        PlanOrNetworkBoundary::Plan(plan) => plan.with_new_children(new_children),
+        PlanOrNetworkBoundary::Plan(plan) => {
+            let input_urls = if new_children.len() == 1 {
+                new_children.first().and_then(|child| child.urls())
+            } else {
+                None
+            };
+            let new_child_plans = new_children.into_iter().map(|child| child.plan).collect();
```
gabotechs: A plan is allowed to have multiple children, as this means that as soon as there's a middle node that has two children, we lose URL information.

This will happen pretty often, and we should make sure that URLs are still correctly propagated even if there are UNIONs, JOINs, etc...

One thing that comes to mind is to do some reconciliation:
- If one child provided URLs but the others didn't, then that's fine, pick the URLs of that child
- If more than one child provided the same set of URLs, then that's also fine, propagate those URLs
- If two different children provided a different set of URLs, throw an informative `plan_err!`.

src/distributed_planner/distributed_config.rs
```diff
@@ -94,6 +95,10 @@ impl DistributedConfig {
         self
     }
 
+    pub fn get_urls(&self) -> Result<Vec<Url>, DataFusionError> {
```
gabotechs: Does this function need to be public? 

- If you think it should be part of this project's public API, then some docs are needed
- If not, we can keep it pub(crate)

src/distributed_planner/plan_annotator.rs
```diff
@@ -191,7 +191,7 @@ async fn _annotate_plan(
         // This is a leaf node, maybe a DataSourceExec, or maybe something else custom from the
         // user. We need to estimate how many tasks are needed for this leaf node, and we'll take
         // this decision into account when deciding how many tasks will be actually used.
-        return if let Some(estimate) = estimator.task_estimation(&plan, cfg) {
+        return if let Ok(Some(estimate)) = estimator.task_estimation(&plan, cfg) {
```
gabotechs: This is swallowing the error returned by `estimator.task_estimation`, it should propagate it instead.

src/execution_plans/distributed.rs
```diff
@@ -118,12 +119,32 @@ impl DistributedExec {
...
+                // default to round-robin task assignation with a randomized starting point.
+                //
+                // If the user has defined custom routing through the plan_leaf_node interface,
+                // we attempt to route the task to the specified URL.
+                //
+                // Notably, it is possible that a task is assigned to a machine that goes offline
+                // in between the moment that the task URL is assigned and the moment the task is
+                // routed. In that case, we route the task to another URL that has not yet been
+                // claimed. If all of the URLs are claimed, we begin assigning round-robin.
+                let url = match task.url.clone() {
```
gabotechs: :thinking: I think it's worth erroring here. If the user is expecting a task to be executed in a specific URL, and we cannot execute it in that URL, it might be better to surface that rather than silently move forward with a random URL.

src/execution_plans/network_broadcast.rs
```diff
@@ -134,6 +135,7 @@ impl NetworkBroadcastExec {
     /// the correct consumer_task_count.
     pub fn try_new(
         input: Arc<dyn ExecutionPlan>,
+        _input_urls: Option<Vec<Url>>,
```
gabotechs: :thinking: how is this unused? did you miss to wire it up with the `Stage` maybe?

tests/task_estimator_test.rs
```diff
@@ -0,0 +1,132 @@
...
+        DefaultSessionBuilder, DistributedConfig, DistributedExt, NetworkBoundary,
+        NetworkBroadcastExec, NetworkCoalesceExec, NetworkShuffleExec, PartitionIsolatorExec,
+        PlannedLeafNode, TaskEstimation, TaskEstimator, assert_snapshot,
+        test_utils::localhost::start_localhost_context,
+    };
+    use futures::TryStreamExt;
+    use url::Url;
+
+    #[tokio::test]
+    async fn custom_task_estimator_with_routing() -> Result<(), Box<dyn std::error::Error>> {
```
gabotechs: Nice! we are missing way more tests though.

It would be nice to test how this behaves in more complex cases, like UNIONs, JOINs, etc...

I'd recommend building a custom leaf node that is capable of displaying on the `DisplayAs` trait the URLs in which it's assigned, and perform some insta assertions over that.

I'd use https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/test_utils/test_work_unit_feed.rs as inspiration. The result should be several test cases with a very narrowed body, leveraging the testing tooling to handle the common stuff. You have some example of how the tests should look like in https://github.com/datafusion-contrib/datafusion-distributed/blob/main/tests/work_unit_feed.rs.

src/distributed_planner/task_estimator.rs
```diff
@@ -388,17 +434,17 @@ mod tests {
...
             plan: &Arc<dyn ExecutionPlan>,
             cfg: &ConfigOptions,
-        ) -> Option<TaskEstimation> {
-            self(plan, cfg)
+        ) -> Result<Option<TaskEstimation>> {
+            Ok(self(plan, cfg))
         }
 
-        fn scale_up_leaf_node(
+        fn plan_leaf_node(
```
gabotechs: I think it might be worth calling this something more like `distribute_plan`. There are no fundamental limitations that would prevent calling this on any intermediate node, so it might be worth not closing that door.

In the same fashion, I'd rename `PlannedLeafNode` to `DistributedPlan`, as there's also nothing to it that prevents it being something other than a leaf.

This naming will fit well with using this struct in the `_distribute_plan()` function as return type.

gabotechs (commented):
This is looking very good!

src/distributed_planner/plan_annotator.rs
JSOD11: Good catch 👍

src/distributed_planner/distribute_plan.rs
JSOD11: Moving forward with other API for now, so this problem goes away.

src/distributed_planner/task_estimator.rs
JSOD11: Sounds good, I like that rename.

src/distributed_planner/distributed_config.rs
JSOD11: In the previous version of the interface, this would have been useful for a user externally if they wanted to route tasks to URLs through what is now called the `distribute_plan` method at planning time. But since we are moving towards routing at execution time, I think you're right that this is unneeded additional surface area.

src/distributed_planner/distribute_plan.rs
```diff
@@ -86,80 +86,82 @@ fn _distribute_plan(
     cfg: &ConfigOptions,
     query_id: Uuid,
     stage_id: &mut usize,
-) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+) -> Result<DistributedPlan, DataFusionError> {
```
gabotechs: I think this change would be difficult to reconcile with https://github.com/datafusion-contrib/datafusion-distributed/pull/416. 

As this is not strictly necessary for this PR, how about just leaving it as:

```rust
```

and reevaluate after all the upcoming refactors is this `Arc<dyn ExecutionPlan>` -> `DistributedPlan` is still relevant?

JSOD11: I think we will down the line head in this direction if we want to be able to send different plans to each task in a stage, but I can go back to the `ExecutionPlan` version for now for this PR.

I have been thinking that this `DistributedPlan` could become, for example, an `enum`, where you have something like:

```
enum DistributedPlan {
     Shared(Arc<dyn ExecutionPlan>)
     Unique(Arc<dyn ExecutionPlan>)
}
```

where `Shared` indicates that all the tasks have the same plan and `Unique` indicates that there are unique plans per-task.

Since I had already made the scaffolding changes here and we are changing the API anyway I was thinking I'd leave it this way for extension, but it's also fine with me if we leave that to a future PR if it conflicts with the way we are removing the 2-pass planner

gabotechs: > I think we will down the line head in this direction if we want to be able to send different plans to each task in a stage, but I can go back to the ExecutionPlan version for now for this PR.

I actually think we might not. If we want to dynamically send the stages for execution as the plan makes progress, we cannot return something other than a plain `Arc<dyn ExecutionPlan>` as intermediate struct.

As it's unclear how this should look like in the future, it will conflict with further work, and this PR is now unrelated to that change, I'd advocate towards just not introducing this change for now, and track it properly in its own PR when the time comes.

JSOD11: Yes agreed, I migrated it back to `ExecutionPlan` for this PR. We'll handle https://github.com/datafusion-contrib/datafusion-distributed/pull/410 in the future as a separate issue

tests/task_estimator_test.rs
JSOD11: Thanks this was a helpful pointer, I took this direction with `URLEmitterExec` below.

src/distributed_planner/task_estimator.rs
```diff
@@ -110,58 +114,67 @@ pub trait TaskEstimator {
...
+    fn distribute_plan(
         &self,
         plan: &Arc<dyn ExecutionPlan>,
         task_count: usize,
         cfg: &ConfigOptions,
-    ) -> Option<Arc<dyn ExecutionPlan>>;
+    ) -> Result<Option<Arc<dyn ExecutionPlan>>>;
+
+    // Defines a custom protocol for routing tasks to specific URLs / physical machines.
+    fn route_tasks(&self, tasks: Vec<ExecutionTask>, urls: &[Url]) -> Result<Option<Vec<Url>>>;
```
gabotechs: For people to make a good decision about what URL to route to, I think we'll need to provide them with more info in this signature so that they can make a decision. Things that come to mind that would be useful to expose as part of the arguments:
- `Arc<TaskContext>`: users will likely have here their own DataFusion extensions injected, and they may want to consult them in order to choose the URLs.
- `Arc<dyn ExecutionPlan>`: the head node of the subplan they are going to be sending over the wire. Different `TaskEstimator` implementations will need to recurse this looking for their own leaf nodes. There might even be multiple `TaskEstimator` implementations providing multiple `Vec<Url>` assignations, that they will likely need to be reconciled
- ...Maybe there are more things?

It's probably worth encapsulating this in struct like a:

```rust
struct TaskRoutingContext<'a> {
    task_ctx: Arc<TaskContext>,
    available_urls: &'a [Url],
    input_tasks: usize,
    head_plan: &'a Arc<dyn ExecutionPlan>
    ...
}
```

examples/custom_execution_plan.rs
```diff
@@ -314,22 +314,28 @@ impl TaskEstimator for NumbersTaskEstimator {
...
         &self,
         plan: &Arc<dyn ExecutionPlan>,
         task_count: usize,
         _cfg: &datafusion::config::ConfigOptions,
-    ) -> Option<Arc<dyn ExecutionPlan>> {
-        let plan = plan.as_any().downcast_ref::<NumbersExec>()?;
+    ) -> Result<Option<Arc<dyn ExecutionPlan>>> {
+        let Some(plan) = plan.as_any().downcast_ref::<NumbersExec>() else {
+            return Ok(None);
+        };
```
gabotechs: I know that we've discussed adapting the signature and name of this functions in https://github.com/datafusion-contrib/datafusion-distributed/pull/409#discussion_r3153641033. 

However, as we are finally no longer be using these as extension points for the routing API, we no longer have a need for modifying them in this PR, so it might be better to leave both `task_estimation` and `scale_up_leaf_node` untouched here, and change the signatures in a follow up instead. That will better help tracking changes to the public API of the project.

Imagine for example that you want to show someone this new feature that you are shipping, and for that, you share them a link to this PR. They will see a bunch of unrelated changes and they will have a hard time filtering what changes are actually related to task routing, and what other changes are unrelated.

gabotechs: btw, if you think the changes now unrelated to routing are still relevant, it's worth still pursuing them 👍, but probably in a follow up PR

src/distributed_planner/task_estimator.rs
gabotechs: With a struct like this, it'll be easy add new fields as we find convenient while maintaining backwards compatibility.

src/execution_plans/distributed.rs
```diff
@@ -185,39 +175,18 @@ impl DistributedExec {
...
-            })?))
+            distribute_tasks(
+                ctx,
+                plan,
+                &codec,
+                &urls,
+                &metrics,
+                &mut join_set,
+                Arc::clone(&self.task_metrics),
+            )
```
gabotechs: I'd avoid this refactor, the new function introduced is at just 1 argument of violating a Clippy rule, and it's resulting in a more verbose implementation with no clear benefit (10 more lines of code for doing the same thing).

Typically, if you find yourself creating functions with big signatures that are only used once, that's a strong smell that the function should not be there, and it's preferable to just inline your changes.

Something like this does the same thing without paying the price of the big function signature, it's less code, and it's further from violating our clippy rules:

<details><summary> Same code just inlined in prepare_plan()</summary>

```rust
    /// Prepares the distributed plan for execution, which implies:
    /// 1. Perform some worker assignation, choosing randomly from the given URLs and assigning one
    ///    URL per task.
    /// 2. Sending the sliced subplans to the assigned URLs. For each URL assigned to a task, a
    ///    network call feeding the subplan is necessary.
    /// 3. In each network boundary, set the input plan to `None`. That way, network boundaries
    ///    become nodes without children and traversing them will not go further down in.
    /// 4. Spawn a background task per worker that waits for the worker to finish and collects
    ///    its metrics into [DistributedExec::task_metrics] via the coordinator channel.
    fn prepare_plan(&self, ctx: &Arc<TaskContext>) -> Result<PreparedPlan> {
        let worker_resolver = get_distributed_worker_resolver(ctx.session_config())?;
        let codec = DistributedCodec::new_combined_with_user(ctx.session_config());

        let urls = worker_resolver.get_urls()?;

        let metrics = CoordinatorToWorkerMetrics {
            // Metric that measures to total sum of bytes worth of subplans sent.
            plan_bytes_sent: MetricBuilder::new(&self.metrics)
                .with_label(Label::new(DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, "0"))
                .global_counter("plan_bytes_sent"),
            // Latency statistics about the network calls issued to the workers for feeding subplans.
            plan_send_latency: Arc::new(LatencyMetric::new(
                "plan_send_latency",
                |b| b.with_label(Label::new(DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, "0")),
                &self.metrics,
            )),
        };

        let mut join_set = JoinSet::new();
        let prepared = Arc::clone(&self.plan).transform_up(|plan| {
            // The following logic is only relevant to network boundaries.
            let Some(plan) = plan.as_network_boundary() else {
                return Ok(Transformed::no(plan));
            };

            let task_estimator = get_distributed_task_estimator(ctx.session_config())?;

            let stage = plan.input_stage();

            let mut spawner =
                CoordinatorToWorkerTaskSpawner::new(stage, &metrics, &codec, &mut join_set)?;

            // Default routing assigns tasks to URLs round-robin from a random starting point.
            let start_idx = rand::rng().random_range(0..urls.len());

            // If the user has not defined custom routing through the route_task API, we
            // default to round-robin task assignation with a randomized starting point.
            //
            // If the user has defined custom routing through the route_task interface,
            // we attempt to route the task to the specified URL.
            let routed_urls = match task_estimator.route_tasks(stage.tasks.clone(), &urls) {
                Ok(Some(routed_urls)) => routed_urls,
                Ok(None) => (0..stage.tasks.len())
                    .map(|i| urls[(start_idx + i) % urls.len()].clone())
                    .collect(),
                Err(e) => return Err(exec_datafusion_err!("error routing tasks to workers: {e}")),
            };

            if routed_urls.len() != stage.tasks.len() {
                return Err(exec_datafusion_err!(
                    "number of tasks ({}) was not equal to number of urls ({}) at execution time",
                    routed_urls.len(),
                    stage.tasks.len()
                ));
            }

            let mut tasks = Vec::with_capacity(stage.tasks.len());
            for (i, _task) in stage.tasks.iter().enumerate() {
                let url = routed_urls[i].clone();
                tasks.push(ExecutionTask {
                    url: Some(url.clone()),
                });
                // Spawn a task that sends the subplan to the chosen URL.
                // There will be as many as spawned tasks as workers.
                let (tx, worker_rx) = spawner.send_plan_task(Arc::clone(ctx), i, url)?;
                spawner.metrics_collection_task(i, worker_rx, Arc::clone(&self.task_metrics));
                spawner.work_unit_feed_task(Arc::clone(ctx), i, tx)?;
            }

            Ok(Transformed::yes(plan.with_input_stage(Stage {
                query_id: stage.query_id,
                num: stage.num,
                plan: None,
                tasks,
            })?))
        })?;
        Ok(PreparedPlan {
            plan: prepared.data,
            join_set,
        })
    }
```

</summary>

src/stage.rs
```diff
@@ -92,16 +92,27 @@ pub struct ExecutionTask {
 pub struct DistributedTaskContext {
     pub task_index: usize,
     pub task_count: usize,
+    pub task_url: Option<Url>,
 }
```
gabotechs, line 95: This struct is not really a good fit for propagating the `task_url`. Think for example how `ChildrenIsolatorUnion` work: they are capable of mutating the `DistributedTaskContext` that is propagated to their children, proxy-ing a different version depending on the specific children that are getting recursed.

See for example this implementation here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/execution_plans/distributed.rs#L402-L442

gabotechs: I tried myself a similar approach for propagating worker URLs like this in https://github.com/datafusion-contrib/datafusion-distributed/pull/427, and it does not work well when there are middle nodes that can mutate the `DistributedTaskContext` depending on its outer `DistributedTaskContext`.

Why do we need this here on the first place? I don't think we need this at all, do we?

src/distributed_planner/task_estimator.rs
JSOD11: Yes this looks like the right idea to me, just added it.

src/execution_plans/distributed.rs
JSOD11: Thank you, looking at it again you're definitely right. I've undone that change.

src/stage.rs
JSOD11: Moving forward with https://github.com/datafusion-contrib/datafusion-distributed/pull/435 as we discussed offline, thanks! 👍

examples/custom_execution_plan.rs
JSOD11: Yeah this sounds right to me, this is good feedback, thank you. I've updated the PR to be more tightly scoped.

src/distributed_planner/task_estimator.rs
```diff
@@ -121,6 +126,17 @@ pub trait TaskEstimator {
         task_count: usize,
         cfg: &ConfigOptions,
     ) -> Option<Arc<dyn ExecutionPlan>>;
+
+    /// Defines a custom protocol for routing tasks to specific URLs / physical machines.
+    fn route_tasks(&self, routing_ctx: &TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>>;
```
gabotechs: If we want to make this change fully backwards compatible, we'll need to add a default implementation for this method in the trait. We can do something like this here:

```rust
    /// Defines a custom protocol for routing tasks to specific URLs / physical machines.
    fn route_tasks(&self, routing_ctx: &TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>> {
        Ok(None)
    }
```

With this, people are not forced to implement the `fn route_tasks()` function.

src/distributed_planner/task_estimator.rs
```diff
@@ -121,6 +126,17 @@ pub trait TaskEstimator {
...
+}
+
+/// Context usable for routing tasks to worker URLs.
+pub struct TaskRoutingContext<'a> {
+    pub task_ctx: Arc<TaskContext>,
+    pub plan: &'a Arc<dyn ExecutionPlan>,
+    pub tasks: &'a [ExecutionTask],
```
gabotechs: 🤔 at this point `ExecutionTask` cannot be something other than `ExecutionTask { url: None }`, so people will always receive an array of always empty objects. From an API standpoint, this will be weird, as people are never going to be able to do anything other than calling `tasks.len()` with it.

It might be better to change this to `task_count: usize`.

src/distributed_planner/task_estimator.rs
```diff
@@ -121,6 +126,17 @@ pub trait TaskEstimator {
...
+    pub urls: &'a [Url],
```
gabotechs: These are the available worker URLs at the moment of planning, which for us it's clear, but for people might not be that clear, two things come to mind:
- rename this to `available_urls` and have a doc comment so that people now what these URLs are here for.
- Add a `DistributedConfig::get_urls()` method so that people can just take the `DistributedConfig` out of the `task_ctx` and retrieve the URLs themselves, allowing us to remove this field from here.

Don't have a strong opinion, whatever you prefer.

src/distributed_planner/task_estimator.rs
```diff
@@ -189,6 +211,18 @@ pub(crate) fn set_distributed_task_estimator(
     }
 }
 
+pub fn get_distributed_task_estimator(
```
gabotechs: As this function is not exposed to the public, it's fine to make it private to the crate:

```suggestion
pub(crate) fn get_distributed_task_estimator(
```

tests/task_estimator_test.rs
```diff
@@ -0,0 +1,371 @@
...
+          └──────────────────────────────────────────────────
+        +------------+------------+------+--------------+
+        | task_count | task_index | tag  | worker_url   |
+        +------------+------------+------+--------------+
+        | 5          | 0          | logs | http://url-4 |
+        | 5          | 1          | logs | http://url-3 |
+        | 5          | 2          | logs | http://url-2 |
+        | 5          | 3          | logs | http://url-1 |
+        | 5          | 4          | logs | http://url-0 |
+        +------------+------------+------+--------------+
```
gabotechs, line 45: 🎉 amazing tests!

gabotechs (commented):
This is almost there! I think we just need a tiny tweak and some docs for `TaskRoutingContext` and we can ship this.

src/distributed_planner/task_estimator.rs
JSOD11: Yeah good idea, this allowed me to delete a few unnecessary implementations.

src/distributed_planner/task_estimator.rs
JSOD11: Good call, fixed.

src/distributed_planner/task_estimator.rs
JSOD11: I went with the `available_urls` approach and left a comment there, let me know what you think.

tests/task_estimator_test.rs
JSOD11: Thanks! 🚀

gabotechs (commented):
GitHub seems broken, I cannot comment on specific chunks of code, adding the comment here:

about:

```rust
/// Context usable for routing tasks to worker URLs.
pub struct TaskRoutingContext<'a> {
    pub task_ctx: Arc<TaskContext>,
    pub plan: &'a Arc<dyn ExecutionPlan>,
    pub task_count: usize,
    // Contains a list of URLs representing machines available to receive a task. These URLs are
    // sourced at execution time and thus should closely reflect the real state of the cluster.
    pub available_urls: &'a [Url],
}
```

The comment is:

```
Nice!, we just miss a couple of things here:
- Doc comments to the other public attributes. 🤔 I thought we had a linting rule for this actually, it does not seems to be kicking in. Ideally, all public entities, whether they are fields, structs, traits, functions, etc... should have a valid doc comment. LSPs and IDEs will show this documentation when hovering over the fields, which provides a really nice experience.
- If we want the doc comments to render in LSP actions, docs.rs, etc... we need them to be actual doc comments. Doc comments start with three `///`, instead of just two `//`. So we'll need to change this in the `available_urls` field.
```

gabotechs (approved):
💯 awesome work @JSOD11! this is ready to go

jayshrivastava:
🚀

PR #411, opened 2026-04-22
Introduce WorkUnit concept for streaming units of work to leaf nodes at runtime

### LINKED ISSUE #374: [draft] stream metadata from DistributedExec to leaf nodes

gabotechs:
Second iteration of https://github.com/datafusion-contrib/datafusion-distributed/pull/399, with some further learnings reflected in the code.

Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/374

---

This PR introduces the concept of a `WorkUnit`, which is pretty much anything the user wants to that is susceptible too be fed into leaf nodes at runtime for performing execution.

A `WorkUnit` is just anything representable in protobuf format:
- A Parquet file address in S3 that must be read.
- An HTTP query that should be issued to an external API.
- An external database query that should be executed externally.
- Etc...

## Disclaimer

This PR is only relevant for cases when all the work units cannot be known at planning time. If a project can afford to resolve all work units at planning time and bake them into the execution plans themselves (e.g., the classical `FileScanConfig`), it's better to move on without this feature, as it's just opt-in.

## Why is this new concept introduced?

In order to understand what this new concept brings to the table, it's important to understand what we have today in plain DataFusion:

Let's look for example at a Parquet file source. When reading Parquet from a remote storage, all the Parquet files are resolved at planning time, and they are backed into the data source node as `FileGroup`s, which looks something like this:

```rust
┌─────────────────────────────────────────────────────┐
│                                                     │
│                                                     │
│              ... rest of the plan ...               │
│                                                     │
│                                                     │
└───────▲────────────▲────────────▲────────────▲──────┘
        │            │            │            │
        │            │            │            │
        │            │            │            │
┌───────┼────────────┼────────────┼────────────┼──────┐
│       │            │MyDataSource│            │      │
│ ┌─────┴────┐ ┌─────┴────┐ ┌─────┴────┐ ┌─────┴────┐ │
│ │    P0    │ │    P1    │ │    P2    │ │    P3    │ │ ( ) Parquet file
│ │( )( )    │ │( )       │ │( )       │ │( )( )    │ │     
│ └──────────┘ └──────────┘ └──────────┘ └──────────┘ │     
└─────────────────────────────────────────────────────┘
```

This works fine when the number of files to read is in the order of magnitude suitable for a single machine, but on a distributed context, there can be millions of files, and discovering all of them might take time, time in which already discovered files can start executing.

This gets worst with other patterns different from files, for example, when there is a two step API where:
- The first API resolves addresses or query fragments that should be issued to a second API
- The second API returns for each query the actual data that should be computed within DataFusion

## How does this new concept solve the issue

What a WorkUnit brings to the table is that it's designed to be discovered and streamed at execution time to nodes, rather than fully discovered and stored at planning time.

Feeding `WorkUnit`s into nodes is handled by `WorkUnitFeed`, and it works like this:

```rust
┌──────────────────────────────────────────────────────┐
│                    ExecutionPlan                     │
│                                                      │
│                                                      │
│┌────────────────────────────────────────────────────┐│
││                    WorkUnitFeed                    ││
││ ┌───────────┐     ┌───────────┐     ┌───────────┐  ││
││ │ .feed(0)  │     │ .feed(1)  │     │ .feed(2)  │  ││
││ └────┬──────┘     └──┬────────┘     └──┬────────┘  ││
│└──────┼───────────────┼─────────────────┼───────────┘│  .─.
│┌──────┼─────────┐┌────┼───────────┐┌───.▼.──────────┐│ (   ) WorkUnit
││      │P0       ││   .▼. P1       ││  (   )P2       ││  `─'  (e.g., a file address)
││     .▼.        ││  (   )         ││   `┬'          ││
││    (   )       ││   `┬'          ││    │           ││
││     `─'        ││    │           ││   .▼.          ││
││      │         ││   .▼.          ││  (   )         ││
││      │         ││  (   )         ││   `┬'          ││
││      │         ││   `─'          ││    │           ││
││      ▼         ││    ▼           ││    ▼           ││
││  processing... ││  processing... ││  processing... ││
││      │         ││    │           ││    │           ││
││      │         ││    │           ││    │           ││
│└──────┼─────────┘└────┼───────────┘└────┼───────────┘│
└───────┼───────────────┼─────────────────┼────────────┘
  ┌─────▼─────┐     ┌───▼───────┐      ┌──▼────────┐
  │RecordBatch│     │RecordBatch│      │RecordBatch│
  └───────────┘     └───────────┘      └───────────┘
```

But the challenge does not end here. As we are executing in a distributed context, WorkUnits need be suitable to be streamed over the wire to the worker nodes that are actually going to handle them.

This works out of the box with no extra action item required from the user's side, the DistributedExec head node (which at this point we should probably start calling it CoordinatorExec), is responsible for capturing all the streams locally, and fanning them out over the wire to the appropriate workers. Something like this:

```rust
                                                                                                    ┌──────────────────┐
                                                                                                    │Coordinating Stage│
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┴──────────────────┘
                                                                                                                       │
│
 ┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐│
││                                                    WorkUnitFeed                                                    │
 │  ┌───────────┐     ┌───────────┐     ┌───────────┐            ┌───────────┐      ┌───────────┐    ┌───────────┐    ││
││  │ .feed(0)  │     │ .feed(1)  │     │ .feed(2)  │            │ .feed(3)  │      │ .feed(4)  │    │ .feed(5)  │    │
 │  └────┬──────┘     └──┬────────┘     └──┬────────┘            └─────┬─────┘      └──┬────────┘    └───┬───────┘    ││
│└───────┼───────────────┼─────────────────┼───────────────────────────┼───────────────┼─────────────────┼────────────┘
         │               │                 │                           │               │                .┴.            │
└ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ .┴. ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─(   )─ ─ ─ ─ ─ ─
         │              .┴.                │                         (   )             │                `┬'
         │             (   )               │                          `┬'              │                .┴.
        .┴.             `┬'               .┴.                          │               │               (   )
       (   )             │               (   )                         │              .┴.               `┬'
        `┬'             .┴.               `┬'────────────┐             │             (   )               │┌────────────┐
         │             (   )               ││  Worker 1  │             │              `┬'                ││  Worker 2  │
┌ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ `┬' ─ ─ ─ ─ ─ ─ ─ ─│┴────────────┘    ┌ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─ ─│┴────────────┘
 ┌───────┼───────────────┼─────────────────┼────────────┐│     ┌───────┼───────────────┼─────────────────┼────────────┐│
││       │            Exe│utionPlan        │            │     ││       │            Exe│utionPlan        │            │
 │       │               │                 │            ││     │       │               │                 │            ││
││┌──────┼───────────────┼─────────────────┼───────────┐│     ││┌──────┼───────────────┼─────────────────┼───────────┐│
 ││      │          Remot│WorkUnitFeed     │           │││     ││      │          Remot│WorkUnitFeed     │           │││
│││ ┌────▼──────┐     ┌──▼────────┐     ┌──▼────────┐  ││     │││ ┌────▼──────┐     ┌──▼────────┐     ┌──▼────────┐  ││
 ││ │ .feed(0)  │     │ .feed(1)  │     │ .feed(2)  │  │││     ││ │ .feed(0)  │     │ .feed(1)  │     │ .feed(2)  │  │││
│││ └────┬──────┘     └──┬────────┘     └──┬────────┘  ││     │││ └────┬──────┘     └──┬────────┘     └──┬────────┘  ││
 │└──────┼───────────────┼─────────────────┼───────────┘││     │└──────┼───────────────┼─────────────────┼───────────┘││
││       │               │                 │            │     ││       │               │                 │            │
 │┌──────┼─────────┐┌────┼───────────┐┌───.▼.──────────┐││     │┌──────┼─────────┐┌────┼───────────┐┌───.▼.──────────┐││
│││      │P0       ││   .▼. P1       ││  (   )P2       ││     │││      │P0       ││   .▼. P1       ││  (   )P2       ││
 ││     .▼.        ││  (   )         ││   `┬'          │││     ││     .▼.        ││  (   )         ││   `┬'          │││
│││    (   )       ││   `┬'          ││    │           ││     │││    (   )       ││   `┬'          ││    │           ││
 ││     `─'        ││    ┼           ││   .▼.          │││     ││     `─'        ││    ┼           ││   .▼.          │││
│││      │         ││   .▼.          ││  (   )         ││     │││      │         ││   .▼.          ││  (   )         ││
 ││      │         ││  (   )         ││   `┬'          │││     ││      │         ││  (   )         ││   `┬'          │││
│││      │         ││   `─'          ││    │           ││     │││      │         ││   `─'          ││    │           ││
 ││      ▼         ││                ││    ▼           │││     ││      ▼         ││                ││    ▼           │││
│││  processing... ││  processing... ││  processing... ││     │││  processing... ││  processing... ││  processing... ││
 ││      │         ││    │           ││    │           │││     ││      │         ││    │           ││    │           │││
│││      │         ││    │           ││    │           ││     │││      │         ││    │           ││    │           ││
 │└──────┼─────────┘└────┼───────────┘└────┼───────────┘││     │└──────┼─────────┘└────┼───────────┘└────┼───────────┘││
│└───────┼───────────────┼─────────────────┼────────────┘     │└───────┼───────────────┼─────────────────┼────────────┘
   ┌─────▼─────┐     ┌───▼───────┐      ┌──▼────────┐    │       ┌─────▼─────┐     ┌───▼───────┐      ┌──▼────────┐    │
│  │RecordBatch│     │RecordBatch│      │RecordBatch│         │  │RecordBatch│     │RecordBatch│      │RecordBatch│
   └───────────┘     └───────────┘      └───────────┘    │       └───────────┘     └───────────┘      └───────────┘    │
└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
```

This allows the users to just have this code snippet in their custom leaf node implementations:

```rust
    fn execute(
        &self,
        partition: usize,
        context: Arc<TaskContext>,
    ) -> Result<SendableRecordBatchStream> {
        let work_unit_stream = self.my_work_unit_feed.feed(partition, context)?;
    ...
    }
```

And it will work out of the box no matter if single-node or distributed.


src/work_unit_feed/remote_work_unit_feed.rs
```diff
@@ -0,0 +1,82 @@
...
+        for partition in 0..partitions {
+            let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+            self.receivers.insert((id, partition), Mutex::new(Some(rx)));
+            self.senders.insert((id, partition), tx);
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct RemoteFeedProvider {
```
jayshrivastava, line 55: I feel like a diagram / comment would be nice here

src/work_unit_feed/work_unit_feed_provider.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// If no [DistributedWorkUnitFeedContext] is present, returns one valid for single-node
+    /// execution.
+    pub fn from_ctx(ctx: &Arc<TaskContext>) -> Arc<Self> {
+        ctx.session_config()
+            .get_extension::<Self>()
+            .unwrap_or(Arc::new(DistributedWorkUnitFeedContext {
+                fan_out_tasks: 1,
+            }))
+    }
+}
```
jayshrivastava, line 106: Yeah... I see why you did this. It's cleaner than having a param in `feed()`.

So as a user, if I want to use this, I must
1. `TaskEstimator::task_estimation`. Say I calculate X tasks
2. Implement `scale_up_leaf_node`, which might say there's Y tasks. I don't really know what a user would do - what leaf node changes would they possibly need to make?
3. Implement `WorkUnitFeedProvider::feed` which get's Y from the DistributedWorkUnitFeedContext and update my feed to split up the work into P partitions.

src/execution_plans/distributed.rs
```diff
@@ -319,6 +389,113 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
...
+                                    partition: partition as u64,
+                                    body: data_or_err?.encode_to_bytes(),
+                                })),
+                            })
+                            .is_err()
+                        {
+                            break; // channel closed.
+                        };
+                    }
+                    Ok::<_, DataFusionError>(())
```
jayshrivastava, line 477: We're swallowing the error right?

src/execution_plans/distributed.rs
```diff
@@ -319,6 +389,113 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
...
+    ) -> Result<()> {
+        let d_cfg = DistributedConfig::from_config_options(ctx.session_config().options())?;
+        /// Recurses into the plan looking for [WorkUnitFeedExec] nodes that should be handled by
+        /// the provided [task_i]. Because of [ChildrenIsolatorUnionExec]s being present in the
+        /// plan, there might be some present [WorkUnitFeedExec] that will not necessarily get
+        /// executed, so we don't want to stream any [WorkUnit] to those.
+        ///
+        /// It places in `out` the list of futures that should be polled for driving the [WorkUnit]
+        /// network streams forward.
+        fn gather_work_unit_feed_tasks(
```
jayshrivastava, line 414: 😵‍💫

Okay I suppose I see why this is necessary. Similarly, we may have some problems if we have different flavors of leaf nodes and want to execute queries across them. Ie. different types of leafs that get joined. That's a general problem we have though not specific to work units.

src/execution_plans/metrics.rs
```diff
@@ -68,8 +68,17 @@ impl ExecutionPlan for MetricsWrapperExec {
         unimplemented!("MetricsWrapperExec does not implement execute")
     }
 
-    // metrics returns the wrapped metrics.
+    /// returns the wrapped metrics merged with any other present in
```
jayshrivastava, line 71: Ok interesting. I thought we only wrapped remote nodes? Ie. nodes which have no metrics bc they were not executed on the coordinator.

jayshrivastava: How did you find this one?

src/work_unit_feed/remote_work_unit_feed.rs
gabotechs: Added a comment and a link to the `WorkUnitFeed` diagram, which should be good for explaining this one.

src/work_unit_feed/work_unit_feed_provider.rs
gabotechs: > Yeah... I see why you did this. It's cleaner than having a param in feed().

That's right, as users are also responsible for calling `.feed()`, that means that having extra params there introduces unnecessary friction

> So as a user, if I want to use this, I must

1. yeap.
2. No strict requirement to do anything on `scale_up_leaf_node` here. The fact that there will be `Y` tasks is already given by `DistributedWorkUnitFeedContext` while calling `WorkUnitFeedProvider::feed`.
3. yeap.

src/execution_plans/distributed.rs
gabotechs: There are two errors that can happen here:
- tx.send() -> channel closed error. If this happens we are fine, we just need to stop the loop
- data_or_err?.encode_to_bytes() -> this is getting surfaced, it will fail in `futures::future::try_join_all(futures).await?;`

src/execution_plans/distributed.rs
gabotechs: 🤔 That should not be a problem, it should be fine for us to have different flavors of leaf nodes, as this function will recurse independently to them.

src/execution_plans/metrics.rs
gabotechs: Just with the integration test I added to `metrics_collection.rs`. For leaf nodes that contain a `WorkUnitFeed`, part of that leaf node will execute on the "coordinator" and part of the leaf node will execute in each worker. This means that if the `WorkUnitFeed` contains metrics, they will be overwritten by the ones coming remotely.

What we want is to keep both, so just letting both co-live works out of the box.

jayshrivastava (approved):

gabotechs:
Thanks @jayshrivastava for the review!

Issue #412, opened 2026-04-23
Metrics lost when flight stream is dropped before completion


LiaCastaneda:
When a flight stream is dropped before completion, the metrics on the last FlightData message never reach the coordinator. We've seen this happen in queries with `limit` that trigger  early stream termination, and aborted queries as well.

More recently, after https://github.com/apache/datafusion/pull/21068 was merged, I've found this in queries with HashJoins where the probe side is short circuited (I still need to add a repro for this).

PR #375 introduced a worker protocol side channel that should enable sending metrics in a separate channel, which would be the proper fix. Related: #187, #341

gabotechs:
The test introduced by @gferrate in https://github.com/datafusion-contrib/datafusion-distributed/pull/341 should be a pretty good starting point indeed. Things that I imagine will be needed:
- Add a new `WorkerToCoordinator` message type that holds metrics information for a task
- Stop sending the metrics information as a last message in the worker-to-worker Arrow stream
- Start sending the metrics information in this new `WorkerToCoordinator` message, which `DistributedExec` will collect

This means that `NetworkBoundary` implementations are probably no longer going to be the ones responsible for sending these metric collection messages recursively upstream. It's likely going to happen somewhere near to https://github.com/datafusion-contrib/datafusion-distributed/blob/abcdd96b440a36fa2ee56cb6fda4720886626b10/src/worker/worker_service.rs#L219-L219

Which today is an empty stream of messages as there's no worker-to-coordinator info sent anywhere.

LiaCastaneda:
I managed to do a repro in our Fork [here](https://github.com/DataDog/datafusion-distributed/commit/a44ceb4d85f8961219ddc5e4d5f93721d24a23c5) for when metrics are lost because of `HashJoinExec` dropping the probe side (left and right sides have to be in different stages), we can't commit that to this repo yet because the PR introduced in DataFusion is not in any release yet.

gabotechs:
Closing as completed by https://github.com/datafusion-contrib/datafusion-distributed/pull/415

PR #415, opened 2026-04-24
Fix metrics loss on early stream termination 

LiaCastaneda:
issue: https://github.com/datafusion-contrib/datafusion-distributed/issues/412
Previously, worker metrics were sent on the last `FlightData` message of the execution stream. If the stream was dropped before completion ( due to a `LIMIT` or `HashJoin` dropping the probe side), the last message was never received and metrics were silently lost.

This PR fixes the issue by sending metrics via a dedicated coordinator channel (`WorkerToCoordinatorMsg`) instead of the data stream. Each worker sends its metrics back to the coordinator after all partitions finish or are dropped early.

```
Coordinator                                    Worker
──────────────────────────────────────────────────────

DistributedExec.execute()
  │
  └─ for each worker:
       ├─ SetPlanRequest ───────────────────────► stores plan in cache
       │
       └─ detached tokio::spawn
            (waiting on response_stream...)

Network boundary node
  │
  └─ ExecuteTask ─────────────────────────────► plan.execute()
       ◄── FlightData                                  │
       ◄── FlightData                                  │
       ◄── FlightData (last, or stream dropped)        │
                                                       │
       ◄── MetricsCollection ──────────────────────────┘
            │
            ▼
       detached tokio::spawn inserts
       into DistributedExec.task_metrics
            │
            ▼
       TaskMetricsCollector.collect(
           ..., Some(&task_metrics))

```


LiaCastaneda:
The test `test_metrics_collection_in_union` (and probably other ones) are flaky because metrics are now delivered asynchronously via the coordinator channel, so they may not be there yet by the time the test asserts on them.

gabotechs:
> The test test_metrics_collection_in_union (and probably other ones) are flaky because metrics are now delivered asynchronously via the coordinator channel, so they may not be there yet by the time the test asserts on them.

I wonder if it makes sense to add a method on `DistributedExec` for awaiting the metrics to come. Something like `DistributedExec::wait_metrics_collection` that the users can decide to call.

I can imagine how people would want to call that post-request as part of their observability.

src/test_utils/metrics.rs
```diff
@@ -1,4 +1,34 @@
...
+use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Waits until all worker tasks have reported their metrics back via the coordinator channel.
+pub async fn wait_for_all_metrics(plan: &Arc<dyn ExecutionPlan>) {
+    let Some(dist_exec) = plan.as_any().downcast_ref::<DistributedExec>() else {
+        return;
+    };
```
LiaCastaneda: makes sense i added it as a helper for tests here, but it would be cleaner to have this as an api yes

LiaCastaneda: https://github.com/datafusion-contrib/datafusion-distributed/pull/415/changes/ef172d827b82cc5fc30e0dabe77328de65ecb902

src/execution_plans/distributed.rs
```diff
@@ -377,18 +445,58 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
         );
 
         let metrics = self.metrics.clone();
-        self.join_set.spawn(async move {
+
+        // Sending the plan and waiting for metrics both run in a detached tokio::spawn so they
+        // are not cancelled when the output stream is dropped early.
+        let (plan_sent_tx, plan_sent_rx) = tokio::sync::oneshot::channel::<Result<()>>();
+
+        tokio::spawn(async move {
```
gabotechs: There should be a clearner way of coding this. Note the pattern here:

`CoordinatorToWorkerTaskSpawner` is a struct responsible for spawning the different tasks related to coordinator-to-worker communication. Currently it spawns two tasks:
1) A task that sends the plan to the workers
2) A task that streams `WorkUnit`s to the workers

I can imagine how we can have:
3) A task that receives metrics from the workers

I think that would very well fit the purpose of the `CoordinatorToWorkerTaskSpawner` abstraction, producer simpler and more decoupled code.

gabotechs: Here's one idea that comes to mind:

1. Change the function signature to:
```rust
    fn send_plan_task(
        &mut self,
        ctx: Arc<TaskContext>,
        task_i: usize,
        url: Url,
    ) -> Result<(UnboundedSender<CoordinatorToWorkerMsg>, UnboundedReceiver<WorkerToCoordinatorMsg>)> {
```

2. Create two channels instead of one. One for inbound worker messages and other for outbound coordinator messages:

```rust
        let (coordinator_to_worker_tx, coordinator_to_worker_rx) =
            tokio::sync::mpsc::unbounded_channel();
        let (worker_to_coordinator_tx, worker_to_coordinator_rx) =
            tokio::sync::mpsc::unbounded_channel();
```

3. Inject the collected worker-to-coordinator messages in the channel in the spawned task:
```rust
        self.join_set.spawn(async move {
            let start = Instant::now();
            let mut client = channel_resolver.get_worker_client_for_url(&url).await?;
            let response = client.coordinator_channel(request).await.map_err(|e| {
                tonic_status_to_datafusion_error(&e).unwrap_or_else(|| {
                    exec_datafusion_err!("Error sending plan to worker {url}: {e}")
                })
            })?;
            metrics.plan_send_latency.record(&start);
            metrics.plan_bytes_sent.add(plan_size);
            let mut stream = response.into_inner();
            while let Some(msg) = stream.next().await {
                if worker_to_coordinator_tx.send(msg).is_err() {
                    break; // closed
                };
            }
            Ok::<_, DataFusionError>(())
        });
        Ok((coordinator_to_worker_tx, worker_to_coordinator_rx))
```

4. Create a separate `CoordinatorToWorkerTaskSpawner::metric_collection_task()` that receives the `worker_to_coordinator_rx` and encapsulates all the logic regarding collecting metrics

gabotechs (commented):
Still reviewing. Flushing comments as I go.

src/metrics/task_metrics_collector.rs
```diff
@@ -92,26 +55,37 @@ impl TaskMetricsCollector {
...
-            input_task_metrics: self.input_task_metrics,
+            input_task_metrics,
         })
     }
 }
 
+fn dashmap_to_hashmap(
+    map: &DashMap<TaskKey, Vec<pb::MetricsSet>>,
+) -> HashMap<TaskKey, Vec<pb::MetricsSet>> {
+    map.iter()
```
gabotechs: Calling `.iter()` in a DashMap is a bit dangerous. From the official docs:

> Locking behaviour: May deadlock if called when holding a mutable reference into the map.

The fact that we need to convert a DashMap to a HashMap doesn't smell good. I'll try to propose a better alternative in another comment.

src/worker/impl_execute_task.rs
```diff
@@ -211,44 +215,51 @@ fn missing(field: &'static str) -> impl FnOnce() -> Status {
...
+fn collect_metrics_collection(
     task_key: TaskKey,
     plan: Arc<dyn ExecutionPlan>,
-) -> Result<flight_app_metadata::Content, FlightError> {
-    // Get the metrics for the task executed on this worker + child tasks.
-    let mut result = TaskMetricsCollector::new()
-        .collect(plan)
-        .map_err(|err| FlightError::ProtocolError(err.to_string()))?;
+) -> MetricsCollection {
+    let result = match TaskMetricsCollector::new().collect(plan, None) {
```
gabotechs: I think re-using the `TaskMetricsCollector` is forcing you to do some weird stuff for adapting to how `TaskMetricsCollector` was shaped some time ago.

I'd try to not use `TaskMetricsCollector` at all here, and instead, just recurse the plan by hand and collect the metrics that way. I bet that the outcome can actually be simpler.

gabotechs (commented):
Still reviewing, flushing a couple more comments.

src/worker/worker.proto
```diff
@@ -25,7 +25,12 @@ message CoordinatorToWorkerMsg {
 }
 
 message WorkerToCoordinatorMsg {
-  // For now, there are no messages that can flow back from worker to coordinator.
+  oneof inner {
+    // Sends the metrics collected during task execution back to the coordinator.
+    // This is sent after all partitions of a task have finished (or been dropped),
+    // ensuring metrics are never lost due to early stream termination.
+    MetricsCollection metrics_collection = 1;
```
gabotechs: The old `MetricsCollection` message contains a list of `TaskMetrics`, which are a list of metrics belong to a specific `TaskKey`.

The old style of propagating metrics required `TaskKey` to be present, because the same `MetricsCollection` might contain information for multiple `TaskKey`s. Now, this can never happen, because a `WorkerToCoordinatorMsg` message is already scoped by definition to the same `TaskKey` (The one sent by `SetPlanRequest` on the first place).

I suggest to not reuse the `MetricsCollection` message, and to create one alternative message instead. For example:

```protobuf
message PreOrderTaskMetrics {
  repeated MetricsSet metrics = 1;
}
```

src/worker/impl_execute_task.rs
gabotechs: For example, these two `send_metrics_collection_via_channel` and `collect_metrics_collection` functions could be implemented in a single much simpler function that looks like this:

```rust
fn send_metrics_collection_via_channel(
    metrics_tx: &Arc<Mutex<Option<Sender<MetricsCollection>>>>,
    task_key: TaskKey,
) {
    let mut proto_task_metrics = vec![];
    let _ = plan.apply(|plan| {
        proto_task_metrics.push(
            plan.metrics()
                .and_then(|v| df_metrics_set_to_proto(&v).ok())
                .unwrap_or_default(),
        );
        Ok(TreeNodeRecursion::Continue)
    });

    let task = TaskMetrics {
        task_key: Some(task_key),
        metrics: proto_task_metrics,
    };

    let tx = {
        let mut guard = match metrics_tx.lock() {
            Ok(g) => g,
            Err(_) => return,
        };
        guard.take()
    };
    let Some(tx) = tx else { return };
    // Ignore send errors — the coordinator channel may have been dropped (e.g. query cancelled).
    let _ = tx.send(MetricsCollection { tasks: vec![task] });
}
```

gabotechs: Also, citing https://github.com/datafusion-contrib/datafusion-distributed/pull/415#discussion_r3152961867, it does not make sense to reuse the `MetricsCollection` here, as there's no way now that multiple `TaskKey`s will be sent in the same `MetricsCollection`

src/execution_plans/distributed.rs
LiaCastaneda: I struggled a bit to understand this change, but I think I understand now, we want two channels: one for `worker-> coordinator` and another one for `coordinator -> worker`. then whatever we receive from the worker (either it is metrics or work units?) we will forward it to the `worker_to_coordinator_tx` so `metrics_collection_task` can pick up the metrics.

gabotechs: Yeap, exactly. Being consistent in how we treat the the `worker -> coordinator` channel vs how we treat the `coordinator -> worker` channel is likely to produce a pretty clean implementation.

LiaCastaneda: https://github.com/datafusion-contrib/datafusion-distributed/pull/415/changes/5a0411ae909c38973edc3a567ba8a63b5ea8f404

src/worker/worker.proto
LiaCastaneda: makes sense, and yes, there is abunch of code that gets removed bc of this

LiaCastaneda: https://github.com/datafusion-contrib/datafusion-distributed/pull/415/changes/a5d606faeaa14202ae2d1528e02744a0a2435725

src/metrics/task_metrics_collector.rs
LiaCastaneda: `dashmap_to_hashmap` is called from `collect()` which is called when we call `rewrite_distributed_plan_with_metrics` which in theory should be called after all metrics are collected, but I agree this doesn't look pretty... I will try to think on how to remove this

LiaCastaneda:
@gabotechs do you think it would be worth calling `wait_for_metrics` inside `rewrite_distributed_plan_with_metrics` or should we leave that up to the api users? 🤔

src/metrics/task_metrics_collector.rs
LiaCastaneda: left a related comment here https://github.com/datafusion-contrib/datafusion-distributed/pull/415#issuecomment-4336897758

gabotechs:
> @gabotechs do you think it would be worth calling wait_for_metrics inside rewrite_distributed_plan_with_metrics or should we leave that up to the api users? 🤔

🤔 It might actually be nice to call it ourselves in `rewrite_distributed_plan_with_metrics` actually. I don't see a reason why users would want to call that function, but not the `wait_for_metrics` one.

src/metrics/task_metrics_collector.rs
LiaCastaneda: I added this https://github.com/datafusion-contrib/datafusion-distributed/pull/415/changes/5c76a811755a5338a0d5699eb33d4b5722634b39 , it passes  the dash map as an Arc instead of copying it into a new HashMap (and converting it)

src/metrics/task_metrics_rewriter.rs
```diff
@@ -41,19 +38,28 @@ impl DistributedMetricsFormat {
 
 /// Rewrites a distributed plan with metrics. Does nothing if the root node is not a [DistributedExec].
 /// Returns an error if the distributed plan was not executed.
-pub fn rewrite_distributed_plan_with_metrics(
+///
+/// Waits for all worker task metrics to arrive before rewriting, so the result is always complete.
+pub async fn rewrite_distributed_plan_with_metrics(
```
LiaCastaneda, line 42: now we can await on this function, one concern I have is that this might hang on forever if a task worker dies unexpectedly and not gracefully, wondering what a reasonable timeout for this might be, technically it has to be the amount of the query duration (+ a little bit more) but if the pod dies while reading data it might be less than the query duration limit...
Maybe the user should set the timout for this

src/metrics/task_metrics_collector.rs
```diff
@@ -1,37 +1,29 @@
...
 /// MetricsCollectorResult is the result of collecting metrics from a task.
 pub struct MetricsCollectorResult {
-    // metrics is a collection of metrics for a task ordered using a pre-order traversal of the task's plan.
+    /// Metrics for each node in the task plan, in pre-order traversal order.
     pub task_metrics: Vec<MetricsSet>,
-    // input_task_metrics contains metrics for child tasks if they were collected.
-    pub input_task_metrics: HashMap<TaskKey, Vec<pb::MetricsSet>>,
+    /// Metrics collected from all workers across all stages, keyed by [TaskKey].
+    /// Sourced directly from [DistributedExec::task_metrics].
+    pub input_task_metrics: Option<Arc<MetricsStore>>,
```
gabotechs: :thinking: this is pretty weird, now this field is here just so that it gets passed through `TaskMetricsCollector::collect`. At that point, you might as well not have it here, so that it does not need to get passed around in `collect` just to be immediately accessed again from the outside.

src/metrics/task_metrics_collector.rs
```diff
@@ -1,37 +1,29 @@
...
 pub struct TaskMetricsCollector {
     /// metrics contains the metrics for the current task.
     task_metrics: Vec<MetricsSet>,
-    /// input_task_metrics contains metrics for tasks from child [StageExec]s if they were
-    /// collected.
-    input_task_metrics: HashMap<TaskKey, Vec<pb::MetricsSet>>,
 }
 
 /// MetricsCollectorResult is the result of collecting metrics from a task.
 pub struct MetricsCollectorResult {
```
gabotechs: This structure made sense when we needed to do a recursive metrics collection stage by stage, but now it's doing close to nothing, and I see it's being used in a very weird way.

I think we should just remove it, and collect metrics simply calling `plan.appy`

src/metrics/task_metrics_rewriter.rs
```diff
@@ -204,7 +210,7 @@ pub fn rewrite_local_plan_with_metrics(
 /// Note: Metrics may be aggregated by name (ex. output_rows) automatically by various datafusion utils.
 pub fn stage_metrics_rewriter(
     stage: &Stage,
-    metrics_collection: Arc<HashMap<TaskKey, Vec<pb::MetricsSet>>>,
+    metrics_collection: Option<Arc<MetricsStore>>,
```
gabotechs: I think this function can never accept `None`, all the places I see it used are passing `Some()`, so we might as well just have `metrics_collection: Arc<MetricsStore>`

src/execution_plans/distributed.rs
```diff
@@ -41,12 +44,50 @@ use std::sync::Mutex;
...
+/// Stores the metrics collected from all worker tasks, and notifies waiters when new entries arrive.
+#[derive(Debug)]
+pub struct MetricsStore {
+    pub map: DashMap<TaskKey, Vec<pb::MetricsSet>>,
+    count_tx: watch::Sender<usize>,
+}
+
+impl MetricsStore {
+    fn new() -> (Self, watch::Receiver<usize>) {
+        let (count_tx, count_rx) = watch::channel(0);
```
gabotechs: Typically, `Self::new()` in Rust return `Self` by convention, I see this one is returning a receiver for notifying of changes in the inner map, but why is it that the case, how it is that we cannot just store the `count_rx` also inside the `MetricsStore` map?

From the only place I see it used (`DistributedExec`), the return `count_tx` is also immediately stored in the `DistributedExec`, as well as the `MetricsStore`. If both things are going to be stored in the same place, you might as well store them together in `MetricsStore`.

src/execution_plans/distributed.rs
```diff
@@ -41,12 +44,50 @@ use std::sync::Mutex;
...
+            },
+            count_rx,
+        )
+    }
+
+    pub fn insert(&self, key: TaskKey, metrics: Vec<pb::MetricsSet>) {
+        self.map.insert(key, metrics);
+        // Notify waiters that a new entry was inserted.
+        self.count_tx.send_modify(|n| *n += 1);
+    }
```
gabotechs: There is a race condition here:
1. The last `self.map.insert` is called, with the last arriving `TaskKey`
2. `count_rx.wait_for` is called inside the `wait_for_metrics` method
3. Now here, `self.count_tx.send_modify` is called, nanoseconds after step 2
4. The `wait_for_metrics` method deadlocks forever

Overall, I see the `MetricsStore` and their methods are a bit weirdly structured (see other comments), we might need to find a better way of representing it.

gabotechs (commented):
It's getting there! I think we need some further restructure, but I'd bet the result will look pretty clean and efficient.

src/metrics/task_metrics_collector.rs
gabotechs: From https://github.com/datafusion-contrib/datafusion-distributed/pull/415#discussion_r3158865661, I think we should just remove the `TaskMetricsCollector`.

src/execution_plans/distributed.rs
gabotechs: I think part of the weirdness and the race condition are here because not the right synchronization primitives where used, I think it's worth trying doing something like this:

```rust
/// Stores the metrics collected from all worker tasks, and notifies waiters when new entries arrive.
#[derive(Debug)]
pub struct MetricsStore {
    rx: watch::Receiver<HashMap<TaskKey, Vec<pb::MetricsSet>>>,
    tx: watch::Sender<HashMap<TaskKey, Vec<pb::MetricsSet>>>,
}

impl MetricsStore {
    fn new() -> Self {
        let (tx, rx) = watch::channel(HashMap::new());
        Self { tx, rx }
    }

    pub fn insert(&self, key: TaskKey, metrics: Vec<pb::MetricsSet>) {
        self.tx.send_modify(|map| {
            map.insert(key, metrics);
        });
    }

    pub fn get(&self, key: &TaskKey) -> Option<Vec<pb::MetricsSet>> {
        self.rx.borrow().get(key).cloned()
    }

    #[cfg(test)]
    pub(crate) fn from_entries(
        entries: impl IntoIterator<Item = (TaskKey, Vec<pb::MetricsSet>)>,
    ) -> Self {
        let (tx, rx) = watch::channel(HashMap::from_iter(entries));
        Self { tx, rx }
    }
}
```

src/execution_plans/distributed.rs
gabotechs: The change proposed here should solve it https://github.com/datafusion-contrib/datafusion-distributed/pull/415#discussion_r3158990811

src/metrics/task_metrics_rewriter.rs
gabotechs: If a worker dies, the gRPC coordinator->worker channel will die, so probably the wait condition should not be "wait for all the task keys to be present", it should be more "wait for all the coordinator-> worker" gRPC connections to be finished.

src/execution_plans/distributed.rs
LiaCastaneda: I'm not sure I understand the deadlock scenario 🤔  why would step 4 wait forrever if `wait_for` already has the map insertion? I imagine `wait_for` would see the insertion from before no?

gabotechs: 🤔 I think you are right yes

LiaCastaneda: It's true separating the sender and the receiver makes things easier to understand. I included your suggestion here https://github.com/datafusion-contrib/datafusion-distributed/pull/415/commits/5c4b79ee75ddd43ac8dae2da09a454012aeeb7df

gabotechs: I think it's worth trying anyways https://github.com/datafusion-contrib/datafusion-distributed/pull/415#discussion_r3158990811, it will allow you to use a normal HashMap, which is nice.

LiaCastaneda: yep `DashMap` was removed

src/execution_plans/distributed.rs
```diff
@@ -290,14 +371,19 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
...
     fn send_plan_task(
         &mut self,
         ctx: Arc<TaskContext>,
         task_i: usize,
         url: Url,
-    ) -> Result<UnboundedSender<CoordinatorToWorkerMsg>> {
+    ) -> Result<(
+        UnboundedSender<CoordinatorToWorkerMsg>,
+        WorkerResponseRx,
+        TaskKey,
```
gabotechs: It's a bit weird that this returns a `TaskKey`, as this is something that folloing methods of this struct can compute on its own.

I'd aim towards having the signature of the `CoordinatorToWorkerTaskSpawner` as consistent as possible: either all of them receive `task_i` as an input, or all of them receive `TaskKey` as in input (instead of `task_i`).

gabotechs (approved):
💯 a negative LOC PR for fixing this sounds amazing, thanks @LiaCastaneda! nice work.

LiaCastaneda:
thanks for the review @gabotechs !

Issue #377, opened 2026-03-18
[Epic] Adaptative Query Execution


gabotechs:
The `TaskEstimator` API currently performs a very rough estimation of an appropriate number of tasks for the different stages of a query. It's mechanism relies on some details that are typically flawed:
- How well can a user of this library estimate the number of tasks for a leaf node at planning time.
- How accurate is to increase/decrease the task count just because a node is "reducing" or not the cardinality of the data.

Because of this, an attempt to ship a cost-based task estimator was attempted in https://github.com/datafusion-contrib/datafusion-distributed/pull/311, which aimed to rely on the standard DataFusion stats propagation machinery to estimate how many rows are going to be flowing through the nodes. There are several issues with that:
- It still relies on users being able to provide accurate stats at planning time, which might be true if reading parquet files, but might not be true if the data source is an API call over the network or similar.
- It assumes a correct stats propagation in upstream DataFusion, which unfortunately is far from being the case. There are several efforts for improving this situation in DataFusion (https://github.com/apache/datafusion/pull/20292, https://github.com/apache/datafusion/issues/8227, https://github.com/apache/datafusion/issues/20766, etc...)

As a consequence, [Cost based planning](https://github.com/datafusion-contrib/datafusion-distributed/pull/311) is not capable of demonstrating any kind of improvement.

I don't think there's any technical limitation preventing this project from implementing adaptative query execution, it has been mentioned before https://github.com/datafusion-contrib/datafusion-distributed/issues/171, and with the amount of uncertainty there is for all attempted stats-based approaches, I think it might be worth trying to ship some form of adaptative query execution to this project.

---

This effort can be split in several deliverables, which intend to add value on their own even if we decide that adaptative query execution is not a good idea:
- [x] https://github.com/datafusion-contrib/datafusion-distributed/issues/373
- [x] https://github.com/datafusion-contrib/datafusion-distributed/issues/382
- [x] https://github.com/datafusion-contrib/datafusion-distributed/issues/374
- [ ] https://github.com/datafusion-contrib/datafusion-distributed/issues/376

Also, special mention to this one, that is tangentially related with future API evolutions:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/378

gabotechs:
Closing this one as completed per issue description. However, the story continues in https://github.com/datafusion-contrib/datafusion-distributed/issues/522.

PR #416, opened 2026-04-27
Remove 2-pass planner (annotator + distribution) in favor of a 1-pass planner

gabotechs:
This is a preparatory step towards:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/377

- https://github.com/datafusion-contrib/datafusion-distributed/pull/422
- https://github.com/datafusion-contrib/datafusion-distributed/pull/424
- https://github.com/datafusion-contrib/datafusion-distributed/pull/416 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/425
- https://github.com/datafusion-contrib/datafusion-distributed/pull/426

The main purpose of this PR is to make distributed planning in a single pass, rather than the current two that communicate each other via an intermediate struct (`AnnotatedPlan`). This change cascades into several other changes that produce a nicer public API for building custom distributed plans, but also produce a big diff.

## Dropping the two-step annotation + NB injection

On a dynamic task assignation context, choosing the task count for a stage based on the previous one can no longer
be done statically. 

After "annotating" a stage, and before "annotating" the
one above, we need to be able to send it for execution, collect runtime
metrics, and based on that decide the task count for the stage above.

This means that the stage below should be good to be sent for execution
before the full annotation process has finished, meaning that we need to 
do everything there is to be done in the "annotation" process, we can no
longer divide the distribution process in several steps that recurse the
whole plan.

## Network boundaries no longer mutate their children

In order for network boundaries to know what mutations to apply to their
children, they need to now how many consumer tasks are they going to be 
running, but this might not be know until execution time, so if we want to
dynamically assign tasks to stages, there's no way at planning time that 
we can know how to mutate the children.

For example, we do not now how to scale up a `RepartitionExec` if we don't
know how many `NetworkShuffleExec`s are going to be consuming it.

The responsibility of preparing network boundaries inputs (e.g., scaling RepartitionExec)
is now factor out into a separate `network_boundary_scale_input()` function
that can be called either at planning time or at execution time.

Right now, it's still just called at planning time.


tests/tpcds_plans_test.rs
```diff
@@ -502,8 +502,8 @@ mod tests {
           │         AggregateExec: mode=FinalPartitioned, gby=[web_site_id@0 as web_site_id], aggr=[sum(salesreturns.sales_price), sum(salesreturns.profit), sum(salesreturns.return_amt), sum(salesreturns.net_loss)]
           │           [Stage 11] => NetworkShuffleExec: output_partitions=3, input_tasks=4
           └──────────────────────────────────────────────────
-            ┌───── Stage 3 ── Tasks: t0:[p0..p2] t1:[p0..p2] t2:[p0..p2] t3:[p0..p2] 
-            │ RepartitionExec: partitioning=Hash([s_store_id@0], 3), input_partitions=3
+            ┌───── Stage 3 ── Tasks: t0:[p0..p11] t1:[p0..p11] t2:[p0..p11] t3:[p0..p11]
+            │ RepartitionExec: partitioning=Hash([s_store_id@0], 12), input_partitions=3
```
gabotechs: 🤔 I noticed this while local testing. This was wrong before, as the stage above has 4 tasks, so each `RepartitionExec` should be outputing `3 * 4 = 12` partitions, which is what we see now.

I believe this was only a visualization bug, otherwise correctness would have been affected.

tests/tpcds_plans_test.rs
```diff
@@ -772,39 +772,43 @@ mod tests {
...
-        │                       [Stage 1] => NetworkBroadcastExec: partitions_per_consumer=2, stage_partitions=2, input_tasks=2
-        │                     HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(CAST(date_dim.d_date_sk AS Float64)@1, ss_sold_date_sk@0)], projection=[ss_store_sk@3, ss_net_profit@4]
-        │                       CoalescePartitionsExec
-        │                         [Stage 2] => NetworkBroadcastExec: partitions_per_consumer=3, stage_partitions=3, input_tasks=2
-        │                       DataSourceExec: file_groups={3 groups: [[/testdata/tpcds/plans_sf1_partitions4/store_sales/part-0.parquet:<int>..<int>, /testdata/tpcds/plans_sf1_partitions4/store_sales/part-1.parquet:<int>..<int>], [/testdata/tpcds/plans_sf1_partitions4/store_sales/part-1.parquet:<int>..<int>, /testdata/tpcds/plans_sf1_partitions4/store_sales/part-2.parquet:<int>..<int>], [/testdata/tpcds/plans_sf1_partitions4/store_sales/part-2.parquet:<int>..<int>, /testdata/tpcds/plans_sf1_partitions4/store_sales/part-3.parquet:<int>..<int>]]}, projection=[ss_sold_date_sk, ss_store_sk, ss_net_profit], file_type=parquet, predicate=DynamicFilter [ empty ] AND DynamicFilter [ empty ]
```
gabotechs: These changes in broadcasts happen because we no longer force-throttle the build side to have less producer tasks than that consumer side.

tests/tpcds_plans_test.rs
gabotechs: This was actually an issue with ChildrenIsolatorExec, fixed now

tests/tpcds_plans_test.rs
gabotechs: I factored this out to https://github.com/datafusion-contrib/datafusion-distributed/pull/422

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1157 @@
...
+    let sf = sf.unwrap_or(1.0);
+    match plan.cardinality_effect() {
+        CardinalityEffect::LowerEqual => sf / ctx.d_cfg.cardinality_task_count_factor,
+        CardinalityEffect::GreaterEqual => sf * ctx.d_cfg.cardinality_task_count_factor,
+        _ => sf,
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
gabotechs, line 525: Unfortunately, Github is not smart enough to show a proper diff. These tests have barely changed.

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,82 @@
...
-    if stage_id == 1 {
+    // Run some preparations on the network boundaries, like optimizing them out if there is one
+    // task at both sides of the boundary or assigning it proper unique identifiers.
+    plan = prepare_network_boundaries(plan)?;
+    if !plan.exists(|plan| Ok(plan.is_network_boundary()))? {
         return Ok(None);
     }
 
-    // Insert PartialReduce aggregation nodes above hash repartitions to reduce shuffle data volume.
+    // Push partial aggregation below hash shuffles to shrink the volume of data sent over
```
asolimando: This is a great idea, and it's generally going to help as data gets indeed reduced.

Today we don't have enough support for statistics propagation for quality NDV, but if this changes tomorrow we might use that to avoid doing that in case the NDV gets closer to the number of rows.

I am aware that there is a runtime escape hatch for this, after reading the first 100k rows, but if we had many it might still add up.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1162 @@
...
+
+    _inject_network_boundaries(plan, ctx).await
+}
+
+#[derive(Clone)]
+struct Context<'a> {
+    parent: Option<&'a Arc<dyn ExecutionPlan>>,
+    cfg: &'a ConfigOptions,
+    d_cfg: &'a DistributedConfig,
+    task_counts: &'a Mutex<HashMap<usize, TaskCountAnnotation>>,
```
asolimando, line 155: I got into a similar need to have a hashmap keyed by node and ended-up doing it similarly, but I think we should start considering a stable node id (longer discussion [here](https://github.com/apache/datafusion/pull/21815#issuecomment-4351618242) if interested!), as plan mutations might dispose (and later even re-use!) those pointers.

I am wondering if the following comment has any relationship with the pointer or it's unrelated:
> /// WARNING: every return statement in this function must funnel through
/// [Context::with_task_count] (or [Context::set_task_count] on the way through) so the returned
/// node has a recorded task count. Callers downstream depend on this invariant.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1162 @@
...
+fn calculate_scale_factor(plan: &Arc<dyn ExecutionPlan>, ctx: &Context) -> f64 {
+    if plan.is_network_boundary() {
+        return 1.0;
+    };
+
+    let mut sf = None;
+    for plan in plan.children() {
+        sf = match sf {
+            None => Some(calculate_scale_factor(plan, ctx)),
+            Some(sf) => Some(sf.max(calculate_scale_factor(plan, ctx))),
```
asolimando, line 512: Q: the scale factor seems reasonable but I am sure we will have to fine-tune it at some point, I am wondering if there would be a way to make this customizable somehow?

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,82 @@
...
     original: Arc<dyn ExecutionPlan>,
     cfg: &ConfigOptions,
 ) -> datafusion::common::Result<Option<Arc<dyn ExecutionPlan>>> {
-    // Keep this function idempotent.
+    // The plan already contains network boundaries set by the user. Just ensure they have nice
+    // unique identifiers for each stage, and move forward with it.
     if original.exists(|plan| Ok(plan.is_network_boundary()))? {
-        return Ok(None);
+        // Ensure the stages in the plan have nice unique identifiers.
+        let plan = prepare_network_boundaries(original)?;
```
asolimando: Nit: maybe it's marginal cost to re-walk the plan, but I guess we could tell already from the `prepare_network_boundaries` and skip walking the plan again?

asolimando (approved):
LGTM, I focused more on the conceptual part than on the details of the code itself, and I think this looks very good!

One pass planning and the clear split for stages over network boundaries is the perfect place to tune the "dimensionality" of the plan (partitions and scaling).

This is a very sensible trade-off between purely static planning which can suffer from bad estimations for the number of partitions, without paying the price (for now?) for full replanning: you fix the structure early on based on the plan structure single-node DF computes, which anyway something that is way harder to adjust dynamically, but you tune the level of parallelism which in a good class of query shapes is the dominating cost, and this tuning stays tractable.

The clear separation in stages you are defining here also makes per-stage observability easier, and keeps the door open for having fault-tolerance/retries at the stage level tomorrow, if needed.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: That's spot on. Right now the thing that prevents this whole system from falling apart are the unit tests, but from a development standpoint is very easy to dispose a pointer address in favor of a new one and not track it properly here.

I have found no better alternative though... definitely some stable IDs for nodes will greatly help here.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: The factor can be configured with:

https://github.com/datafusion-contrib/datafusion-distributed/blob/d31fc71530d8138fdcda4bbb2754eca1db40a8ab/src/distributed_planner/distributed_config.rs#L19-L24

But this is anyways a very very coarse way of estimating an appropriate task count. A better approach would be to:
- be stats based: tried that without success due to the extreme inaccuracy of stats currently in DataFusion
- by adaptive: this is what this stack of PRs aim to do. So far it looks like a promising approach.

src/distributed_planner/distribute_plan.rs
gabotechs: 🤔 good idea, I'll try this

gabotechs: I've tried this, but it results in significantly messier code in `prepare_network_boundaries`, as we need to recursively propagate whether the plan gets mutated or not.

Typically, traversing the plan it's dirt cheap as long as you don't mutate. It's plan mutations what recursively trigger `Arc` operations for all nodes. Here, the first `original.exists()` call just partially recurses the plan until the first network boundary, without transforming it, so it's price should be << than `prepare_network_boundaries()`.

src/distributed_planner/distribute_plan.rs
gabotechs: This is unfortunately disabled by default as it implies some regressions and there was not a net positive impact in latency of the current benchmarks.

Maybe there's something to be followed up there.

There's some more detailed analysis here https://github.com/datafusion-contrib/datafusion-distributed/pull/396

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -1,169 +1,82 @@
...
+///    for each node, and inserts `NetworkShuffleExec` / `NetworkBroadcastExec` /
+///    `NetworkCoalesceExec` above the nodes that delimit a stage (hash `RepartitionExec`s,
+///    build-side `BroadcastExec`s, and any node sitting under a `CoalescePartitionsExec` /
+///    `SortPreservingMergeExec`).
 ///
-/// 3. Based on the [AnnotatedPlan] returned by [annotate_plan], place all the appropriate
-///    network boundaries ([NetworkShuffleExec] and [NetworkCoalesceExec]) with the task count
-///    assignation that the annotations required. After this, the plan is already a distributed
-///    executable plan.
+/// 3. **Boundary preparation.** [prepare_network_boundaries] readies each injected boundary
```
gene-bordegaray, line 33: Nit: Prefer to say "materializes" each.

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,82 @@
...
-/// 3. Based on the [AnnotatedPlan] returned by [annotate_plan], place all the appropriate
-///    network boundaries ([NetworkShuffleExec] and [NetworkCoalesceExec]) with the task count
-///    assignation that the annotations required. After this, the plan is already a distributed
-///    executable plan.
+/// 3. **Boundary preparation.** [prepare_network_boundaries] readies each injected boundary
+///    for execution: elides ones that aren't actually needed and finalises the survivors. If
+///    no boundary survives, this function returns `None`.
 ///
-/// This function returns None if the plan was left undistributed.
+/// 4. **Shuffle-volume optimisation.** [partial_reduce_below_network_shuffles] inserts partial
```
gene-bordegaray: spelling: "optimization"

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,82 @@
...
 pub(super) async fn distribute_plan(
     original: Arc<dyn ExecutionPlan>,
     cfg: &ConfigOptions,
 ) -> datafusion::common::Result<Option<Arc<dyn ExecutionPlan>>> {
-    // Keep this function idempotent.
+    // The plan already contains network boundaries set by the user. Just ensure they have nice
+    // unique identifiers for each stage, and move forward with it.
     if original.exists(|plan| Ok(plan.is_network_boundary()))? {
-        return Ok(None);
+        // Ensure the stages in the plan have nice unique identifiers.
```
gene-bordegaray: I think we can reduce comments here

src/distributed_planner/prepare_network_boundaries.rs
```diff
@@ -0,0 +1,81 @@
...
+    plan: Arc<dyn ExecutionPlan>,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    prepare(plan, 1, Uuid::new_v4(), &mut 1)
+}
+
+fn prepare(
+    plan: Arc<dyn ExecutionPlan>,
+    consumer_task_count: usize,
+    query_id: Uuid,
+    num: &mut usize,
```
gene-bordegaray: I don't love the parameter being named num, maybe something a bit more descriptive to show the relationship it plays with identifying the stage

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1162 @@
...
+///
+/// Starting from the leaves, we climb the plan, asking the [TaskEstimator] for a task count
+/// at each leaf and merging children's task counts at each inner node (`Desired` takes the
+/// max; a `Maximum(N)` anywhere in the subtree caps and dominates). We keep going until the
+/// current node is one that requires a network boundary **above it** — currently a hash
+/// `RepartitionExec` (→ shuffle), a `BroadcastExec` whose parent is collecting partitions
+/// (→ broadcast), or any other node whose parent is `CoalescePartitionsExec` /
+/// `SortPreservingMergeExec` (→ coalesce). At that moment, that node and the subtree
+/// underneath it form the stage we've just delimited; the boundary will be injected above it
+/// in Phase 3.
```
gene-bordegaray: For the documentation on passes like this I typically find bullet points with "rules" or "steps" much easier to follow. If we could make short lists of the invariants and conditions we check for I think it would be easier to digest

src/distributed_planner/task_estimator.rs
```diff
@@ -35,15 +36,24 @@ impl From<TaskCountAnnotation> for usize {
...
+            Maximum(maximum) => Maximum(maximum.min(limit)),
+        }
+    }
+
+    pub(crate) fn merge(self, other: TaskCountAnnotation) -> Self {
+        match (self, other) {
+            (Desired(a), Desired(b)) => Desired(std::cmp::max(a, b)),
+            (Desired(_), Maximum(b)) => Maximum(b),
+            (Maximum(a), Desired(_)) => Maximum(a),
+            (Maximum(a), Maximum(b)) => Maximum(std::cmp::max(a, b)),
```
gene-bordegaray: The `Max(a), Max(b)` case should be:

```rust
(Maximum(a), Maximum(b)) => Maximum(std::cmp::min(a, b)),
```

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1162 @@
...
+        // in a non-distributed context. The ChildrenIsolatorUnionExec itself will make sure to
+        // determine which children to run and which to exclude depending on the task index in
+        // which it's running.
+        let children = plan.children();
+        let c_i_union = ChildrenIsolatorUnionExec::from_children_and_task_counts(
+            children.iter().map(|v| Arc::clone(*v)),
+            children
+                .iter()
+                // TODO: rather than defaulting to 1 when no task count is recorded, it's better to error out.
+                .map(|v| ctx.task_count(v).map_or(1, |v| v.as_usize())),
```
gene-bordegaray: Just comment to remind addressing this

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1162 @@
...
+            let plan = propagate_task_count(&plan, task_count, &ctx)?;
+
+            let f = calculate_scale_factor(&plan, &ctx);
+            let input_stage = LocalStage {
+                query_id: ctx.query_id,
+                num: ctx.fetch_add_stage_id(),
+                plan,
+                tasks: task_count.as_usize(),
+            };
+            let plan = Arc::new(NetworkShuffleExec::from_stage(input_stage));
```
gene-bordegaray: I dont love that we are creating this middle state of plan. We know the producer info but not the consumer so we are creating these placeholder states.

I think this is find but it would be nice to have this be explicit like having something like:

```rust
NetworkShuffleExec::pending_from_stage()
```

with helpers like:

```rust
LocalStage::with_plan()
...
```

src/distributed_planner/network_boundary.rs
```diff
@@ -42,3 +40,39 @@ impl NetworkBoundaryExt for dyn ExecutionPlan {
...
+        Arc::clone(&input),
+        consumer_partitions,
+        consumer_task_count,
+    )?;
+    if transformed.transformed {
+        return Ok(transformed.data);
+    }
+
+    Ok(input)
+}
```
gene-bordegaray, line 78: I think this would be nice being on the `NetworkBoundary` trait itself. Itwould be a bit cleanr and less susceptible to unforseen plan bugs

src/execution_plans/network_shuffle.rs
```diff
@@ -108,52 +105,50 @@ pub struct NetworkShuffleExec {
...
-            worker_connections: WorkerConnectionPool::new(input_task_count),
-            properties: input.properties().clone(),
-        })
+    /// Creates a new [NetworkShuffleExec] fed by the provided [RepartitionExec]. The input plan
+    /// will be executed in a remote worker in `producer_tasks` number of tasks.
+    pub fn try_new(input: Arc<dyn ExecutionPlan>, producer_tasks: usize) -> Result<Self> {
+        if !input.as_any().is::<RepartitionExec>() {
+            return plan_err!("The input of a NetworkShuffleExec can only be a RepartitionExec");
+        }
+
```
gene-bordegaray, line 146: I think we shoudl also chekc that partitioning is Hash

src/execution_plans/network_coalesce.rs
```diff
@@ -80,37 +81,62 @@ pub struct NetworkCoalesceExec {
...
-        let max_input_task_count = input_task_count.div_ceil(task_count).max(1);
-        Ok(Self {
-            properties: scale_partitioning_props(input.properties(), |p| p * max_input_task_count),
-            input_stage: Stage::Local(LocalStage {
-                query_id,
-                num,
+        producer_tasks: usize,
+        consumer_tasks: usize,
+    ) -> Self {
+        Self::from_stage(
```
gene-bordegaray: Should we check that consumer_tasks isn't 0 do dive ceil doesnt panic

gene-bordegaray (commented):
I think this is a good separation of passes 👍 . The only thing I am a bit wary of in the "pending" or hald-built network stages. I think this could be more explicit to be clearer.

src/distributed_planner/distributed_query_planner.rs
gabotechs: This is not like before, where we had "annotations" that materialize into "network boundaries", here, the network boundaries present in the plan are already fully materialized, and the only thing pending is to assigned them a query identifier and a number, but besides that they are fully material from the beginning.

src/distributed_planner/distribute_plan.rs
gabotechs: 👍 yeap, I don't think we need most comments here actually, the fact that everything is scoped in functions with nice names makes it clear enough

src/distributed_planner/prepare_network_boundaries.rs
gabotechs: It's refers to `LocalStage.num`, which agree is not a nice name, it was chosen a while ago. Let's change it in the function to `stage_id` instead.

src/distributed_planner/task_estimator.rs
gabotechs: Nice catch!

src/distributed_planner/inject_network_boundaries.rs
gabotechs: Addressed now

src/distributed_planner/inject_network_boundaries.rs
gabotechs: This will become clearer in follow up PRs in the stack, here's a key detail:

We cannot assume we know the consumer tasks at planning time, in fact, the consumer task count might be known even after the query has started executing.

Because of this, not knowing the consumer tasks is not a "pending" state of a network boundary, but the only viable state at any moment. This means that they are not just placeholders, they are the final thing, and consumer task count is never never assumed until after execution.

In a follow up PR, this happens:

https://github.com/datafusion-contrib/datafusion-distributed/blob/96dfd617ad3744cfb231388f735ef59ea92e8f29/src/worker/impl_execute_task.rs#L56-L60

Which means that "scaling up" a RepartitionExec or a BroadcastExec based on the consumer tasks happens lazily right at the moment of starting execution.

src/distributed_planner/network_boundary.rs
gabotechs: Yeah, completely agree, however, this is a static method, meaning that it does not accept `&self` as first argument, and baking a static method in the `NetworkBoundary` trait would make it non dyn-compatible.

This means that we would not be able to do `Arc<dyn NetworkBoundary>`. It's a bit annoying, I'm not sure if there's a Rust compiler trick that workaround this, the option I chose is to just be consistent in all `NetworkBoundary` implementations, and if there's something to catch, catch it at runtime.

src/execution_plans/network_shuffle.rs
gabotechs: Yeap, let's do that

src/execution_plans/network_coalesce.rs
gabotechs: Yeap, let's do that

src/distributed_planner/inject_network_boundaries.rs
gabotechs: I have reframed this explanations to follow this feedback, let me know what you think

src/distributed_planner/network_boundary.rs
gene-bordegaray: I don't see why this can't be an instance method?

```rust
      fn prepare_input_stage(
          &self,
          input: Arc<dyn ExecutionPlan>,
          consumer_partitions: usize,
          consumer_task_count: usize,
      ) -> Result<Arc<dyn ExecutionPlan>>;
```

then just impl for each boundary right?

gabotechs: Because then, that means that we need a `NetworkBoundary` instantiated in order to call `network_boundary_scale_input`, but at the moment of calling the function, we don't have one.

gene-bordegaray: we do have the one here: https://github.com/datafusion-contrib/datafusion-distributed/pull/416/changes#diff-e656b8889efe1812c6ba8c46d4709a11fc2938114e7f6e441abdf79d6c40cfb0R40 -- could we link to that even if it isn't the rebuilt one?

gabotechs: We'll need to call this method from other places where there's no NetworkBoundary built. For example, in `impl_execute_task.rs` at execution time:

https://github.com/datafusion-contrib/datafusion-distributed/blob/96dfd617ad3744cfb231388f735ef59ea92e8f29/src/worker/impl_execute_task.rs#L56-L60

src/distributed_planner/inject_network_boundaries.rs
gene-bordegaray: Yes these are great, thank you 😄

src/distributed_planner/inject_network_boundaries.rs
gene-bordegaray: 👍 sorry missing some stuff with the follow ups after. But with this, makes sense

src/distributed_planner/network_boundary.rs
gene-bordegaray: ditto to comment above about follow ups 👍

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,72 @@
...
+/// Transforms a single-node physical plan into a distributed plan by injecting network
+/// boundaries between stages.
 ///
-/// It performs the following operations:
+/// The pipeline runs four passes in order:
 ///
-/// 1. It prepares the plan for distribution, adding some extra single-node nodes like
-///    [BroadcastExec] or [CoalescePartitionsExec] that will signal the following steps to
-///    introduce network boundaries in the appropriate places.
+/// 1. **Pre-distribution shaping.** A `CoalescePartitionsExec` is wrapped on top of the plan
```
jayshrivastava:
```suggestion
/// 1. **Pre-distribution shaping.** A [`CoalescePartitionsExec`] is wrapped on top of the plan
```

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -1,169 +1,72 @@
...
+///    when it has more than one output partition (so [inject_network_boundaries] later sees a
+///    partition-collecting parent and injects a `NetworkCoalesceExec` above its child). Then
+///    [insert_broadcast_execs] adds `BroadcastExec` nodes on the build side of `CollectLeft`
+///    hash joins so those build sides can later be wrapped in `NetworkBroadcastExec`.
 ///
-/// 2. Annotate the plan with [annotate_plan]: adds some annotations to each node about how
-///    many distributed tasks should be used in the stage containing them, and whether they
-///    need a network boundary below or not.
-///    For more information about this step, read [annotate_plan] docs.
+/// 2. **Boundary injection.** [inject_network_boundaries] walks the plan, computes a task count
```
jayshrivastava, line 27: nit: bullet points might be better for each case

src/distributed_planner/distribute_plan.rs
```diff
@@ -1,169 +1,72 @@
...
+/// 4. **Shuffle-volume optimization.** [partial_reduce_below_network_shuffles] inserts partial
+///    aggregation nodes underneath hash shuffles where it can, so less data crosses the network.
 pub(super) async fn distribute_plan(
     original: Arc<dyn ExecutionPlan>,
     cfg: &ConfigOptions,
 ) -> datafusion::common::Result<Option<Arc<dyn ExecutionPlan>>> {
-    // Keep this function idempotent.
+    // The plan already contains network boundaries set by the user. Just ensure they have nice
+    // unique identifiers for each stage, and move forward with it.
     if original.exists(|plan| Ok(plan.is_network_boundary()))? {
```
jayshrivastava: This is confusing. I have many questions.

"The plan already contains network boundaries set by the user". How would they do this? 

- `prepare_network_boundaries` already walks the plan. Why do we call `exists` and walk the plan again?
- Why would you get a malformed plan where there's network boundaries but they aren't prepared properly (ex. missing unique IDs)? Can't we safely assume that `distribute_plan` was called and both `inject_network_boundaries` and `prepare_network_boundaries` ran.
- I don't think `create_physical_plan` needs to be idempotent in datafusion... we call `distribute_plan` once when planning a query
- Should we return some enum to say "this plan was already distributed"?

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+
+    fn fetch_add_stage_id(&self) -> usize {
+        self.stage_id.fetch_add(1, Ordering::Acquire)
+    }
+}
+
+/// Identity key for a plan node. The pointer is only used as a hash-map key, never dereferenced,
+/// so casting it to `usize` is safe and makes the key `Send + Sync`.
+fn plan_ptr_key(plan: &Arc<dyn ExecutionPlan>) -> usize {
+    Arc::as_ptr(plan) as *const () as usize
```
jayshrivastava, line 214: 🙈

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+}
+
+/// WARNING: every return statement in this function must funnel through
+/// [Context::with_task_count] (or [Context::set_task_count] on the way through) so the returned
+/// node has a recorded task count. Callers downstream depend on this invariant.
+async fn _inject_network_boundaries(
+    plan: Arc<dyn ExecutionPlan>,
+    ctx: Context<'_>,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    let broadcast_joins = ctx.d_cfg.broadcast_joins;
```
jayshrivastava: nit: `broadcast_joins_enabled`

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+    }
+
+    fn set_task_count(&self, plan: &Arc<dyn ExecutionPlan>, task_count: TaskCountAnnotation) {
+        self.task_counts
+            .lock()
+            .expect("task counts mutex poisoned")
+            .insert(plan_ptr_key(plan), task_count);
+    }
+
+    fn with_task_count(
```
jayshrivastava: bit of a weird name, i'd expect it to return a `Context` but it doesn't

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+        task_counts: &Mutex::new(HashMap::new()),
+        query_id: Uuid::new_v4(),
+        stage_id: &AtomicUsize::new(1),
+    };
+
+    _inject_network_boundaries(plan, ctx).await
+}
+
+#[derive(Clone)]
+struct Context<'a> {
```
jayshrivastava, line 152: This is also confusing. This is a per-plan node Context which basically annotates it with the stage id. But it's also global context which contains the config and task_counts map. Why not use 2 contexts and make `parent` non optional?

The "global context" can just be params if you want. Or, you just pass the parent as an argument and let this be the global context.

src/distributed_planner/inject_network_boundaries.rs
jayshrivastava: It doesn't even use any of the fields in the context other than the map. This is hard to follow.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+                query_id: ctx.query_id,
+                num: ctx.fetch_add_stage_id(),
+                plan,
+                tasks: task_count.as_usize(),
+            };
+            let plan = Arc::new(NetworkShuffleExec::from_stage(input_stage));
+            let task_count = Desired((f * task_count.as_usize() as f64).ceil() as usize);
+            return Ok(ctx.with_task_count(plan, task_count));
+        }
+    } else if let Some(parent) = ctx.parent
```
jayshrivastava: Add a comment saying why this happens? Only happens if this node has siblings, right?

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+                plan,
+                tasks: task_count.as_usize(),
+            };
+            let plan = Arc::new(NetworkBroadcastExec::from_stage(input_stage));
+            let task_count = Desired((f * task_count.as_usize() as f64).ceil() as usize);
+            Ok(ctx.with_task_count(plan, task_count))
+        } else {
+            // The subtree below this point belongs to one stage. Propagate the chosen task
+            // count down so every node in that stage has it recorded.
+            let plan = propagate_task_count(&plan, task_count, &ctx)?;
```
jayshrivastava:
```suggestion
            let plan = propagate_task_count_until_stage_or_leaf(&plan, task_count, &ctx)?;
```

src/distributed_planner/prepare_network_boundaries.rs
```diff
@@ -0,0 +1,81 @@
...
+/// producer and consumer sides both run on a single task, scales the producer-stage head of
+/// the survivors to feed all consumer tasks, and stamps each surviving stage with a unique
+/// `(query_id, num)` identifier.
+pub(crate) fn prepare_network_boundaries(
+    plan: Arc<dyn ExecutionPlan>,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    prepare(plan, 1, Uuid::new_v4(), &mut 1)
+}
+
+fn prepare(
```
jayshrivastava, line 20: doc comment would be nice.

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -1,169 +1,72 @@
...
+///    build-side `BroadcastExec`s, and any node sitting under a `CoalescePartitionsExec` /
+///    `SortPreservingMergeExec`).
 ///
-/// 3. Based on the [AnnotatedPlan] returned by [annotate_plan], place all the appropriate
-///    network boundaries ([NetworkShuffleExec] and [NetworkCoalesceExec]) with the task count
-///    assignation that the annotations required. After this, the plan is already a distributed
-///    executable plan.
+/// 3. **Boundary preparation.** [prepare_network_boundaries] readies each injected boundary
+///    for execution: elides ones that aren't actually needed and finalises the survivors. If
+///    no boundary survives, this function returns `None`.
```
jayshrivastava, line 35: note that we also scale up other plan nodes (ex. repartition exec for shuffles) at this point

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1161 @@
...
+            Ok(ctx.with_task_count(plan, estimate.task_count))
+        } else {
+            // We could not determine how many tasks this leaf node should run on, so
+            // assume it cannot be distributed and use just 1 task.
+            Ok(ctx.with_task_count(plan, Maximum(1)))
+        };
+    }
+
+    let mut futures = Vec::with_capacity(plan.children().len());
+    for child in plan.children() {
```
jayshrivastava, line 242: Do we really want to spawn tasks? I would think that this is all synchronous. I saw the mutex around the hashmap

src/distributed_planner/distributed_query_planner.rs
gabotechs: :thinking: but aren't they already bullet points? step 1., step 2., step 3., ...

src/distributed_planner/distributed_query_planner.rs
gabotechs: Yeah, this is stated in `prepare_network_boundaries`, I preferred to not repeat it here.

src/distributed_planner/distribute_plan.rs
gabotechs: > "The plan already contains network boundaries set by the user". How would they do this?

They just literally call `NetworkShuffleExec::try_new()` themselves inside their own `PhysicalOptimizerRule` for example. After this PR, they can just do it.

gabotechs: > prepare_network_boundaries already walks the plan. Why do we call exists and walk the plan again?

Because it results in cleaner code, and it does not have a performance impact (no allocations and atomic operations in the `.exists()` path). This question has been raised before, so I might try to do it in one pass again.

gabotechs: > Why would you get a malformed plan where there's network boundaries but they aren't prepared properly (ex. missing unique IDs)? Can't we safely assume that distribute_plan was called and both inject_network_boundaries and prepare_network_boundaries ran.

If users injected their own network boundaries, then `inject_network_boundaries` should not (and will not) run. At that point, a plan with network boundaries already placed in by the users might be place, and rather than we requiring users to assign a `query_id` and a `stage_num`, we do it here in the preparation step.

gabotechs: > I don't think create_physical_plan needs to be idempotent in datafusion... we call distribute_plan once when planning a query

And this function in fact does not try to be idempotent, doing several passes will produce different `query_id`s actually.

gabotechs: > Should we return some enum to say "this plan was already distributed"?

This has not really changed vs `main`, the plan either:
- got distributed
- or not

and for that just uses the same `Option` as before.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: This is pretty good feedback. I'll separate it.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: Maybe `propagate_task_count_until_network_boundaries`, which looking at the body of the function is exactly what it does.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: Done in https://github.com/datafusion-contrib/datafusion-distributed/pull/416/changes/661490547f3aef48cafdcfc781e7c9ee7782ba11. I should look better now.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: True, in fact the rusty convenion is that methods beginning with `with_` should take `mut self` and return `Self`.

Changed the name to something else.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: I definitely do not love this... but I really have not found a better way of doing this...

If you have better ideas they are more than welcome.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: We are not really spawning tasks here, this is just joining them into a single future without further spawns, so it's just 1 task.

The `Mutex` thing is indeed annoying:

We need mutable access to the task count map, so in theory, we would be fine with just a simply `RefCell`, as we know it's impossible that more than 1 tokio task polls this at any moment. However, because this is getting called from within an `#[async_trait]` here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/gabrielmusat%2Frefactor-distributed-planning/src/distributed_planner/session_state_builder_ext.rs#L35-L37

Which requires this method to be `Send` and `Sync`. Using a `Mutex` is fine from a performance standpoint, because we are sure that there's never going to be more than 1 concurrent task accessing it, but unfortunately, the compiler does not enforce it, and therefore, we are bound to comply with `Send` and `Sync`.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: Added

src/distributed_planner/prepare_network_boundaries.rs
gabotechs: The comment of the function above is the one applicable here. This is just a separate function for technical reasons (we need it for dfs recursion), but besides that, it follows what's stated in publicly exposed function.

I tweaked the comment a bit in that function to make it clearer

src/distributed_planner/inject_network_boundaries.rs
jayshrivastava: Ope yes. Sorry this one was a stale comment before I read the code. I also have a bad habit of saying task instead of future.

src/distributed_planner/distribute_plan.rs
gene-bordegaray: I think that we can wrap a `DistributedExec` in a `DistributedExec` here in a weird edge cases where someone calls this on an already distributed plan. Idk why they would want to do that but could just return None on a `DistributedExec`

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -0,0 +1,1157 @@
...
+) -> Result<Arc<dyn ExecutionPlan>> {
+    let broadcast_joins_enabled = ctx.d_cfg.broadcast_joins;
+    let estimator = &ctx.d_cfg.__private_task_estimator;
+
+    if plan.children().is_empty() {
+        // This is a leaf node, maybe a DataSourceExec, or maybe something else custom from the
+        // user. We need to estimate how many tasks are needed for this leaf node, and we'll take
+        // this decision into account when deciding how many tasks will be actually used.
+        return if let Some(estimate) = estimator.task_estimation(&plan, ctx.cfg) {
+            Ok(ctx.plan_with_task_count(plan, estimate.task_count))
```
gene-bordegaray: Should we cap this leaf estimate before storing?

I saw that Unions read the child task counts so they should probably already be capped?

src/execution_plans/network_coalesce.rs
```diff
@@ -80,37 +81,65 @@ pub struct NetworkCoalesceExec {
 }
 
 impl NetworkCoalesceExec {
-    /// Builds a new [NetworkCoalesceExec] in "Pending" state.
+    /// Does nothing, but it's here for explicitly stating that this network boundary does not
+    /// need mutate the input plan in other to account for more consumer tasks.
```
gene-bordegaray: nit: "does not need to mutate"

src/distributed_planner/distribute_plan.rs
gene-bordegaray: I am also thinking here that I don't love the `Option` as it isn't clear what being distributed means. Like is it if it contains a `DistributedExec` or if users just injected any type of network boundary?

gene-bordegaray (commented):
mostly extending existing conversations

jayshrivastava (approved):

jayshrivastava:
Looks good from my end.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: :thinking: the task count cap is applied some lines below to the node above, but yeah, I think it's more correct to also apply it here. Doing that now.

src/distributed_planner/distribute_plan.rs
gabotechs: :thinking: I'm thinking that it might be clearer if this function does not even exist. It's introducing some hidden control flows based on whether `None` or `Some` are returned.

gabotechs: Pushed something, it should look better now

gene-bordegaray (approved):
woooop 💯

gabotechs:
🙏 Thanks for the reviews guys! I know this is pretty tricky code, and getting you to also be aware of it is great.

PR #419, opened 2026-04-30
Buffering improvements to inter-worker communication

gabotechs:
## Summary
- Replaces the bounded per-partition channels in `WorkerConnection` with unbounded ones to eliminate head-of-line blocking on the demux task. A single bounded queue could block the demux and starve all sibling partitions, deadlocking queries with cross-partition dependencies (e.g. TPC-H Q18).
- Backpressure is now enforced globally per worker connection: the demux loop pauses the gRPC pull while `memory_reservation.size()` is at or above a configurable byte budget, and consumers wake it via `Notify` after each `shrink`. A single oversized message is always admitted to avoid livelock, so the peak per connection is `budget + max_message_size`.
- Adds a new `DistributedConfig::worker_connection_buffer_budget_bytes` knob (default 64 MiB) and the matching `with_distributed_worker_connection_buffer_budget_bytes` / `set_…` builder methods on all `DistributedExt` impls.
- Refactored `spawn_select_all` to use one channel per partition instead of a global channel for all partitions. This makes sure that slow partitions do not delay fast ones.

### DIFF
--- a/src/worker/spawn_select_all.rs
+++ b/src/worker/spawn_select_all.rs
@@ -21,13 +21,14 @@ where
     El: MemoryFootPrint + Send + 'static,
     Err: Send + 'static,
 {
-    let (tx, rx) = tokio::sync::mpsc::channel(queue_size);
+    let reservation = Arc::new(MemoryConsumer::new("NetworkBoundary").register(&pool));
 
-    let mut tasks = vec![];
+    let mut tasks = Vec::with_capacity(inner.len());
+    let mut in_rxs = Vec::with_capacity(inner.len());
     for mut t in inner {
-        let tx = tx.clone();
-        let pool = Arc::clone(&pool);
-        let consumer = MemoryConsumer::new("NetworkBoundary");
+        let (in_tx, in_rx) = tokio::sync::mpsc::channel(queue_size);
+        in_rxs.push(ReceiverStream::new(in_rx));
+        let reservation = Arc::clone(&reservation);
 
         tasks.push(SpawnedTask::spawn(async move {
             loop {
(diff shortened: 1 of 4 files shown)


src/worker/worker_connection_pool.rs
```diff
@@ -192,9 +202,8 @@ impl WorkerConnection {
             per_partition_rx.insert(partition, rx);
         }
```
LiaCastaneda, line 203: Looks like this was already unbounded before no?

src/worker/spawn_select_all.rs
```diff
@@ -21,38 +21,41 @@ where
...
+    let reservation = Arc::new(MemoryConsumer::new("NetworkBoundary").register(&pool));
 
-    let mut tasks = vec![];
+    let mut tasks = Vec::with_capacity(inner.len());
+    let mut in_rxs = Vec::with_capacity(inner.len());
     for mut t in inner {
-        let tx = tx.clone();
-        let pool = Arc::clone(&pool);
-        let consumer = MemoryConsumer::new("NetworkBoundary");
+        let (in_tx, in_rx) = tokio::sync::mpsc::channel(queue_size);
```
LiaCastaneda, line 29: iuc this is the core improvement right? Before we had 1 channel for all partitions, now we have 1 per partition. Is `queue_size` still needed for backpressure? 🤔 Technically messages will flow through more quickly and there's less possibility of having many messages stranded here no?

LiaCastaneda (approved):
This makes sense to me, I just have a few questions for my understanding

src/worker/worker_connection_pool.rs
gabotechs: Yes, before it was YOLO-style unbounded, meaning that there was no back-pressuring mechanism. You can see this translated to a very high memory usage in our (Datadog's) metrics.

Now the channels are still unbounded (we cannot bound them because that results in deadlocks), but we can back-pressure based on our memory readings.

src/worker/spawn_select_all.rs
gabotechs: yes, this is right, although I think there are still other ways a message can show high network latency, as right now I don't think the measurement is perform in the correct place.

One thing that comes to mind is that the "created_timestamp" is set here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/d12d12bc75fd874171b4708b49a79dd2dd7fa0f1/src/worker/impl_execute_task.rs#L105-L112

However, that code is executed within a stream that is one of the inputs of `spawn_select_all`. If we wanted to measure real network latency, that "created_timestamp" should be set at output of `spawn_select_all`, not at the input.

gabotechs: I think this improvement is worth doing, either in this one, or in a follow up. Whatever you prefer.

LiaCastaneda: I'm not sure, if we fix the network latency metric will we still see that delay (from the queue block) or not?  I think maybe we should do it as a follow-up and see how the network latency metric (as it is now) changes after this change.

gabotechs: Sounds good. Merging this one now.

PR #420, opened 2026-04-30
Remove trailing whitespace from plan task display

gene-bordegaray:
eliminate whitespaces at end of plan lines. they keep causing churn in snapshot reviews with `cargo insta review` because these strings dont have the trailing space

 don't know if this is happening to others as well.


gabotechs (approved):
🙏 yes please!

PR #423, opened 2026-05-01
fix union recomputing child properties

gene-bordegaray:
The `ChildrenIsolatorUnionExec` when replacing children would keep the old cached properties (things like `output_partitioning()`) rather than recomputing them via the replaced children. This would cause stale properties to remain.

This is evident in a scenario like this:

```text
ChildrenIsolatorUnionExec -> advertises output partitions = 2
    child 0: 2 partitions
    child 1: 1 partition
```

This is ok because:
```
children say: max child/task output = 2
ChildrenIsolatorUnionExec properties: output_partitioning = 2
```

But say an optimizer rule replaces child 0 with a new plan that has more partitions:

```text
ChildrenIsolatorUnionExec -> advertises output partitions = 2  (now stale)
    child 0: 8 partitions (updated)
    child 1: 1 partition
```

Now `ChildrenIsolatorUntionExec is incorrect about its output.

**If `ChildrenIsolatorUntionExec` advertises too few partitions:**

```text
actual child partitions: 0 1 2 3
advertised ChildrenIsolatorUntionExec partitions:  0 1

scheduled downstream work:  0 1
missing work: 2 3
```

This can cause correctness issues if those extra partitions might never be executed.

**If `ChildrenIsolatorUntionExec` advertises too many partitions:**

```text
actual child partitions: 0 1
advertised CIU partitions:  0 1 2 3
```

Then downstream operators might schedule extra empty work.

### DIFF
--- a/tests/work_unit_feed.rs
+++ b/tests/work_unit_feed.rs
@@ -376,9 +376,9 @@ mod tests {
         assert_snapshot!(plan + &results, @r"
         ┌───── DistributedExec ── Tasks: t0:[p0]
         │ SortPreservingMergeExec: [tag@0 ASC NULLS LAST, task@1 ASC NULLS LAST, partition@2 ASC NULLS LAST, letter@3 ASC NULLS LAST]
-        │   [Stage 1] => NetworkCoalesceExec: output_partitions=12, input_tasks=3
+        │   [Stage 1] => NetworkCoalesceExec: output_partitions=6, input_tasks=3
         └──────────────────────────────────────────────────
-          ┌───── Stage 1 ── Tasks: t0:[p0..p3] t1:[p4..p7] t2:[p8..p11]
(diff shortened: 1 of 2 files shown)


tests/work_unit_feed.rs
```diff
@@ -376,9 +376,9 @@ mod tests {
         assert_snapshot!(plan + &results, @r"
```
gene-bordegaray, line 376: this happened because the `RowGenerator` initially mkaes 4 partitions -> `ChildrenIsolatorUnionExec` advertises this 4 partitions -> children rewritten to 2 output partitions but the `ChildrenUsolatorUnionExec` is using stale properties

gabotechs (approved):
👍 good catch!

tests/work_unit_feed.rs
gabotechs: This sounds right 👍

PR #422, opened 2026-05-01
Remove artificial broadcast task limit in the build side

gabotechs:
PR factored out from https://github.com/datafusion-contrib/datafusion-distributed/pull/416.

This is one PR from the following stack of PRs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/422 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/424
- https://github.com/datafusion-contrib/datafusion-distributed/pull/416
- https://github.com/datafusion-contrib/datafusion-distributed/pull/425
- https://github.com/datafusion-contrib/datafusion-distributed/pull/426
- https://github.com/datafusion-contrib/datafusion-distributed/pull/427
- https://github.com/datafusion-contrib/datafusion-distributed/pull/432

Previously, we where force-propagating a max task count assignation below
the NetworkBroadcast so that the remote build side has never more tasks
than the stage above.

In a dynamic task count assignation context, we can no longer do this, as by the time 
you realize a remote build side is going to have more tasks than the stage above, the build side might have
already started executing, and by that time its task count is set in stone.

This is fine, build side in broadcast should have an arbitrarily more expensive
build side. What matters there is not that the build side is cheap to
execute, but that it returns little amount of data. A build side can return
very little data (just a couple of rows) and still be very expensive to execute

This is actually the reason why there is a small speedup in benchmarks.

---

<details><summary>tpch_sf10 1.09 faster ✔</summary>

```text
=== Comparing tpch_sf10 results from engine 'datafusion-distributed-main' [prev] with 'datafusion-distributed-dynamic-task-allocation' [new] ===
      q1: prev=1269 ms, new=1286 ms, diff=1.01 slower ✖
      q2: prev= 390 ms, new= 395 ms, diff=1.01 slower ✖
      q3: prev= 784 ms, new= 826 ms, diff=1.05 slower ✖
      q4: prev= 413 ms, new= 392 ms, diff=1.05 faster ✔
      q5: prev=1306 ms, new=1242 ms, diff=1.05 faster ✔
      q6: prev= 534 ms, new= 528 ms, diff=1.01 faster ✔
      q7: prev=1483 ms, new=1420 ms, diff=1.04 faster ✔
      q8: prev=3001 ms, new=1585 ms, diff=1.89 faster ✅
      q9: prev=2054 ms, new=2009 ms, diff=1.02 faster ✔
     q10: prev= 951 ms, new= 921 ms, diff=1.03 faster ✔
[...]

PR #424, opened 2026-05-01
Refactor Stage struct to be more explicit about states

gabotechs:
An independent refactor factored out from https://github.com/datafusion-contrib/datafusion-distributed/pull/416

- https://github.com/datafusion-contrib/datafusion-distributed/pull/422
- https://github.com/datafusion-contrib/datafusion-distributed/pull/424 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/416
- https://github.com/datafusion-contrib/datafusion-distributed/pull/425
- https://github.com/datafusion-contrib/datafusion-distributed/pull/426

Previously the stage struct was a "hidden" state machine that could have two states:

1. A state where the Stage contains the input plan and is locally accessible and traversible.

```rust
pub struct Stage {
    query_id: ...
    num: ...
    plan: Some(plan),
    tasks: vec![None, None, None]
}
```

2. A state where the input plan is serialized, and the worker URLs are assigned. This happens in `DistributedExec` right before execution on `prepare_plan()`

```rust
pub struct Stage {
    query_id: ...
    num: ...
    plan: None,
    tasks: vec![Some("http://1"), Some("http://2"), Some("http://3")]
}
```

This PR makes this behavior explicit, and represented with an `enum`:

```rust
pub enum Stage {
    Local(LocalStage),
    Remote(RemoteStage),
}

pub struct LocalStage {
    pub query_id: Uuid,
    pub num: usize,
    pub plan: Arc<dyn ExecutionPlan>,
    pub tasks: usize,
}

pub struct RemoteStage {
    pub query_id: Uuid,
    pub num: usize,
    pub workers: Vec<Url>,
}
```

### DIFF
--- a/src/metrics/task_metrics_rewriter.rs
+++ b/src/metrics/task_metrics_rewriter.rs
@@ -4,8 +4,9 @@ use crate::execution_plans::{DistributedExec, MetricsStore, MetricsWrapperExec};
 use crate::metrics::DISTRIBUTED_DATAFUSION_TASK_ID_LABEL;
 use crate::metrics::collect_plan_metrics;
 use crate::metrics::proto::metrics_set_proto_to_df;
-use crate::stage::Stage;
+use crate::stage::{LocalStage, Stage};
 use crate::worker::generated::worker::TaskKey;
+use datafusion::common::plan_err;
 use datafusion::common::tree_node::Transformed;
 use datafusion::common::tree_node::TreeNode;
 use datafusion::common::tree_node::TreeNodeRecursion;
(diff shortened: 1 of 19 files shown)


src/protobuf/distributed_codec.rs
```diff
@@ -66,13 +66,30 @@ impl PhysicalExtensionCodec for DistributedCodec {
...
                 return Err(proto_error("Empty StageProto"));
             };
-
-            Ok(Stage {
-                query_id: deserialize_uuid(proto.query_id.as_ref())?,
-                num: proto.num as usize,
-                plan: inputs.first().cloned(),
-                tasks: decode_tasks(proto.tasks)?,
-            })
+            if let Some(input) = inputs.first().cloned() {
```
gene-bordegaray, line 69: We are determining `Local` vs `Remote` based on if the network boundary has a child, I think this is fine but if wea re trying to make things more explicit would it be better to have this in the protobuf?

gene-bordegaray: but of course the tradeoffs of the proto churn so honestly your call on this guy

src/metrics/task_metrics_rewriter.rs
```diff
@@ -68,17 +69,19 @@ pub async fn rewrite_distributed_plan_with_metrics(
...
             // For now, we are okay with trading off performance for simplicity.
             let plan_with_metrics =
                 stage_metrics_rewriter(stage, Arc::clone(&metrics_collection), format)?;
-            let network_boundary = network_boundary.with_input_stage(Stage::new(
-                stage.query_id,
-                stage.num,
-                plan_with_metrics,
-                stage.tasks.len(),
-            ))?;
+            let network_boundary = network_boundary.with_input_stage(Stage::Local(LocalStage {
```
gene-bordegaray, line 79: nit: might be nice to have constructors

gene-bordegaray (approved):
This is a great change 💯 just some nits

src/protobuf/distributed_codec.rs
gabotechs: yeah, I don't love it, but doing anything else is a breaking change in the protocol. 

As cleaning that up would be breaking change, it might be better to do it in an isolated PR so that it can be more easily reverted if something goes wrong.

src/metrics/task_metrics_rewriter.rs
gabotechs: There's one thing with constructors that I don't love when some there's 2 or more arguments with the same type: the compiler does not catch mixups with which argument gets passed in each position.

For example, `LocalStage` struct has `num: usize` and `tasks: usize` in the field. With a constructor, This will look like `Stage::new_local(Uuid::new_v4(), task_count, Arc::clone(&input), stage_num)`, and it's hard to miss that `task_count` and `stage_num` are swapped, whether with something like this is pretty obvious.
```rust
Stage::Local(LocalStage {
  query_id: Uuid::new_v4(),
  num: task_count,
  plan: Arc::clone(&input),
  tasks: stage_num
})
```

gene-bordegaray: hm that makes sense, I never thought of that. Thank you 😄

PR #427, opened 2026-05-04
Use in-memory comms if both workers involved in an exchange happen to be the same 

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/427 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/469
- https://github.com/datafusion-contrib/datafusion-distributed/pull/461

This is a preparatory step towards:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/377

This is an optimization that allows workers to communicate in-memory avoiding network calls and serialization in case it needs to communicate to itself.

This optimization shows good improvements if using a small number of workers, and fades away as more workers are used.

Even if this shows improvements today, it will become very meaningful in adaptative query execution: there will be times when two consecutive stages are assigned 1 task each, and as that was done dynamically, we cannot eagerly do the optimization that collapses those two stages into one, so instead, we assigned both stages to the same worker, and the optimization in this PR kicks in, executing the plan fully locally.

### DIFF
--- a/src/worker/impl_execute_task.rs
+++ b/src/worker/impl_execute_task.rs
@@ -3,12 +3,13 @@ use crate::common::{map_last_stream, on_drop_stream};
 use crate::metrics::proto::df_metrics_set_to_proto;
 use crate::protobuf::datafusion_error_to_tonic_status;
 use crate::worker::generated::worker::{FlightAppMetadata, PreOrderTaskMetrics};
-use crate::worker::worker_service::Worker;
+use crate::worker::worker_service::{TaskDataEntries, Worker};
 use arrow_flight::encode::{DictionaryHandling, FlightDataEncoder, FlightDataEncoderBuilder};
 use arrow_flight::error::FlightError;
 use arrow_select::dictionary::garbage_collect_any_dictionary;
 use datafusion::arrow::array::{Array, AsArray, RecordBatch, RecordBatchOptions};
 use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
+use datafusion::common::{Result, exec_err, internal_err};
 
 use crate::worker::generated::worker::ExecuteTaskRequest;
 use crate::worker::generated::worker::worker_service_server::WorkerService;
@@ -17,8 +18,9 @@ use datafusion::arrow::ipc::CompressionType;
 use datafusion::arrow::ipc::writer::IpcWriteOptions;
 use datafusion::common::exec_datafusion_err;
 use datafusion::error::DataFusionError;
-use datafusion::execution::SendableRecordBatchStream;
+use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use futures::TryStreamExt;
 use prost::Message;
 use std::sync::Arc;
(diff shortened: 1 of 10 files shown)


src/worker/impl_execute_task.rs
```diff
@@ -3,171 +3,199 @@ use crate::common::{map_last_stream, on_drop_stream};
...
 const RECORD_BATCH_BUFFER_SIZE: usize = 2;
 const WAIT_PLAN_TIMEOUT_SECS: u64 = 10;
 
-impl Worker {
-    pub(crate) async fn impl_execute_task(
-        &self,
-        request: Request<ExecuteTaskRequest>,
-    ) -> Result<Response<<Worker as WorkerService>::ExecuteTaskStream>, Status> {
-        let body = request.into_inner();
+pub(crate) async fn execute_local_task(
```
jayshrivastava, line 43: doc comment?

src/worker/impl_execute_task.rs
```diff
@@ -3,171 +3,199 @@ use crate::common::{map_last_stream, on_drop_stream};
...
+    entry: Arc<SingleWriteMultiRead<ResultTaskData>>,
```
jayshrivastava: kind of sucks to pass this whole thing just to evict the stale entry

src/worker/impl_execute_task.rs
jayshrivastava, line 43: this is a nice idea

src/worker/worker_connection_pool.rs
```diff
@@ -384,7 +410,75 @@ impl WorkerConnection {
...
+            },
+        }
+    }
+}
+
+impl WorkerConnection for LocalWorkerConnection {
+    fn execute(&self, partition: usize) -> Result<BoxStream<'static, Result<RecordBatch>>> {
+        let mut request = self.request_template.clone();
+        request.target_partition_start = partition as u64;
+        request.target_partition_end = (partition + 1) as u64;
```
jayshrivastava: Why does this request have a range if we only execute 1 partition in the remote and local case?

src/worker/impl_execute_task.rs
gabotechs: Yeap, it does... I'll try to think of something

src/worker/worker_connection_pool.rs
gabotechs: In the remote case we execute multiple partitions:

The range is passed here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/b0cb0825d70a1909cbb1d1969740c465f1055e04/src/worker/worker_connection_pool.rs#L175-L175

And one of the places where the range is given is here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/b0cb0825d70a1909cbb1d1969740c465f1055e04/src/execution_plans/network_shuffle.rs#L231-L231

src/worker/impl_execute_task.rs
gabotechs: Ok, managed to get something better! https://github.com/datafusion-contrib/datafusion-distributed/pull/427/changes/aea9cc49216cf3c447b71e5f9ae023fc6dcce402.

Thanks for pointing this out.

src/worker/worker_connection_pool.rs
gene-bordegaray: Yeah, the range matters for remote because one request owns multiple target partitions, but I think that the local request to just the partition being executed makes it so unrequested local partitions never get the drop wrapper that decrements `num_partitions_remaining`.

Could we either make the local connection own the full range or mark the unstarted local partitions done on connection drop?

gene-bordegaray (commented):
just one follow up on a thread but I am using this in a brnach haha 😆

src/worker/worker_connection_pool.rs
gabotechs: 🤔 Good suggestion, trying this now...

gabotechs: Implemented this here https://github.com/datafusion-contrib/datafusion-distributed/pull/427/changes/bf7cadcea378f1429bdc19d6064540354a4af33a, let me know what you think

src/worker/worker_connection_pool.rs
```diff
@@ -384,7 +409,93 @@ impl WorkerConnection {
...
+        Self {
+            partition_start,
+            local_streams,
+        }
+    }
+}
+
+impl WorkerConnection for LocalWorkerConnection {
+    fn execute(&self, partition: usize) -> Result<BoxStream<'static, Result<RecordBatch>>> {
+        let relative_i = partition - self.partition_start;
```
gene-bordegaray: nit: might as well `checked_sub` since partition is passed in

gene-bordegaray (approved):
looks good

PR #428, opened 2026-05-04
speed up test exeuction time by transitioning to in_memory worker

### LINKED ISSUE #370: Deflake and speed up TPCDS tests
These tests are
- huge, so they take really long to run (~10m)
- flake sometimes (see https://github.com/datafusion-contrib/datafusion-distributed/pull/369#issuecomment-4068396208)

Let's improve this.

Rich-T-kid:
Closes #370 
replaced the SessionContext for tpcds & tpch. This lead to a noticeable improvement in test execution time.

Pre
Image 5-3-26 at 11 32 PM

Post 
Image 5-3-26 at 11 35 PM

(this was for tpcds_correctness_test)

these changes dont make the test non-flaky, each time that I ran them they worked fine so im not sure how to reproduce the errors @jayshrivastava  mentioned [ here ](https://github.com/datafusion-contrib/datafusion-distributed/pull/369#issuecomment-4068396208)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,22 @@ jobs:
       - uses: ./.github/actions/setup
       - run: cargo test --features tpch --test 'tpch_*'
 
-  tpcds-test:
+  tpcds-correctness-test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: ["01", "02", "03", "04", "05", "06", "07", "08", "09", "10"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+      - uses: actions/cache@v4
+        with:
+          path: testdata/tpcds/main.zip
+          key: "main.zip"
+      - run: cargo test --features tpcds --test tpcds_correctness_test shard${{ matrix.shard }}
+
+  tpcds-plans-test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -57,7 +72,7 @@ jobs:
         with:
           path: testdata/tpcds/main.zip
           key: "main.zip"
-      - run: cargo test --features tpcds --test 'tpcds_*'
+      - run: cargo test --features tpcds --test tpcds_plans_test
 
   clickbench-test:
     runs-on: ubuntu-latest
--- a/src/test_utils/in_memory_channel_resolver.rs
+++ b/src/test_utils/in_memory_channel_resolver.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use datafusion::common::DataFusionError;
+use datafusion::prelude::SessionContext;
 use hyper_util::rt::TokioIo;
 use tonic::transport::{Endpoint, Server};
 
@@ -56,6 +57,23 @@ impl InMemoryChannelResolver {
 
         this_clone
     }
+}
+pub fn start_in_memory_context<B>(num_workers: usize, session_builder: B) -> SessionContext
+where
+    B: WorkerSessionBuilder + Send + Sync + Clone + 'static,
+{
+    use crate::SessionStateBuilderExt;
+    use datafusion::execution::SessionStateBuilder;
+    let mut state = SessionStateBuilder::new()
+        .with_default_features()
+        .with_distributed_planner()
+        .with_distributed_worker_resolver(InMemoryWorkerResolver::new(num_workers))
+        .with_distributed_channel_resolver(InMemoryChannelResolver::from_session_builder(
+            session_builder,
+        ))
+        .build();
+    state.config_mut().options_mut().execution.target_partitions = 3;
+    SessionContext::from(state)
 }
 
 impl Default for InMemoryChannelResolver {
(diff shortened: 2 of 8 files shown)


tests/tpcds_correctness_test.rs
```diff
@@ -567,7 +567,7 @@ mod tests {
         let s_ctx = SessionContext::new();
 
         // Make distributed localhost context to run queries
-        let (d_ctx, _guard, _) = start_localhost_context(NUM_WORKERS, DefaultSessionBuilder).await;
+        let d_ctx = start_in_memory_context(NUM_WORKERS, DefaultSessionBuilder);
```
gabotechs: Nice! Now that we are here, how about also using this in `clickbench_correctness_test.rs`?

gabotechs:
😞 it seems like they still run in 22 mins https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/25314967550/job/74224224116...

Rich-T-kid:
> 😞 it seems like they still run in 22 mins https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/25314967550/job/74224224116...

@gabotechs Yea I wonder whats making it run so much slower on the CI. each query/test should be running on its own thread, maybe since theres almost 100 and each query is computationally expensive theres a lot of context switching? Im going to run it locally and try to change up the number of threads uses.

gabotechs:
@Rich-T-kid we are using the default Github runners here, which are pretty small. If you find a better config, it's probably worth adding that to the `cargo test` command in the CI definition inside the `.github/` folder

Rich-T-kid:
tpcds ran in 22 mins even after it was split up. I also ran some benchmarks & as expected the more threads the faster it was up until 24 threads. the default git runner is a 4cpu, im going to look into profiling this with **cargo-flamegraph**

gabotechs:
I'm also noticing something weird in https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/25314967550/job/74224224116.

The pipeline took 22 mins, Cargo reports 190s for running the tests here:

```
running 99 tests
test tests::test_tpcds_1 has been running for over 60 seconds
test tests::test_tpcds_10 has been running for over 60 seconds
...
test tests::test_tpcds_96 ... ok
test tests::test_tpcds_97 ... ok
test tests::test_tpcds_98 ... ok
test tests::test_tpcds_99 ... ok

test result: ok. 96 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 190.63s
```

🤔 Where are the other 19mins going?

gabotechs:
Maybe what is slow is just the tpcds data generation? I'd expect the tpcds data to be cached in the GitHub runners, but maybe there's an issue in the caching implementation.

Rich-T-kid:
I've benchmarked this locally and used samply to profile CPU usage. On smaller machines, it seems like there's a hard ceiling on how fast it can run.
[tpcds_correctness_test-8c4eae474fd09fc9 2026-05-04 14.15 profile.json.gz](https://github.com/user-attachments/files/27372114/tpcds_correctness_test-8c4eae474fd09fc9.2026-05-04.14.15.profile.json.gz)
The profile looks normal, the majority of execution time is in DataFusion itself, so this doesn't appear to be an I/O issue. Looking at the [test action](https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/25325835466/job/74246466248?pr=428), the dataset is being cached as expected. The total test time is split roughly between building the binary (~7 minutes) and executing the tests.
I have a few possible directions:

1. Upgrade the runner hardware, beefier machines with more cores.
2. Split the tests across more jobs. GitHub Actions matrix jobs would let us shard the test suite across parallel runners. This has a limit, though: since building the binary takes ~7 minutes per runner, there's a point where the build cost outweighs the parallelism gains.
3. Add a caching layer. I need to do more research on what would be effective here.

src/test_utils/in_memory_channel_resolver.rs
```diff
@@ -57,6 +58,23 @@ impl InMemoryChannelResolver {
         this_clone
     }
 }
+pub fn start_in_memory_context<B>(num_workers: usize, session_builder: B) -> SessionContext
```
gabotechs: This test function might actually be useful to @JSOD11 for testing some of his newer changes. I think we should ship it even if it does not speed up the tests.

.github/workflows/ci.yml
```diff
@@ -41,14 +41,18 @@ jobs:
...
       - uses: ./.github/actions/setup
-      - run: cargo test --features tpch --test 'tpch_*'
+      - uses: actions/cache@v4
+        with:
+          path: testdata/tpcds/main.zip
+          key: "main.zip"
+      - run: cargo test --features tpcds --test tpcds_correctness_test
 
-  tpcds-test:
+  tpcds-plans-test:
```
gabotechs, line 66: The `tpcds-plans-test` are pretty trivial, I'd not expect them to take a lot of time. One thing we can do though, is cache the full `testdata/tpcds/`, and not just `testdata/tpcds/main.zip`, that might be something to look at.

Rich-T-kid: okay ill update the CI to cache the entire directory.

Rich-T-kid:
tpcds-test are no longer passing after caching the entire directory 😕

gabotechs:
🎉 The longest test now takes 7 mins instead of 22! nice job @Rich-T-kid! this is ready to go

gabotechs (approved):

PR #432, opened 2026-05-04
Adaptive task count assignation

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/432 <- you are here

---

This PR implements support for dynamic task count assignation, allowing the distributed planner to determine the optimal number of tasks for each stage based on runtime characteristics and available resources. This enables adaptive query execution where task counts are optimized for the specific execution context rather than being determined statically.

There are two fundamental pieces delivered in this PR that make adaptive task count assignation possible:

1. [SamplerExec](https://github.com/datafusion-contrib/datafusion-distributed/pull/432/changes#diff-17bd8d290d703ba259143955ade0265fd22463f45cfe752c6071ab136331fe7f)
2. [prepare_dynamic_plan](https://github.com/datafusion-contrib/datafusion-distributed/pull/432/changes#diff-01ffef8aa9ef0e8104dfe1c133320f1c1edd157fd968798b9206a3e04dd95228)

## SamplerExec

This is a new `ExecutionPlan` implementation that peeks a few initial record batches before execution, gathers statistics over them, and then reports a `LoadInfo` message to the `prepare_dynamic_plan` running on the coordinator

```rust
                    ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐                       
                        ┌─────────────────────┐                           
                    │   │                     │   │                       
   ┌───────────────────▶│   DistributedExec   │◀──────────────────────┐   
   │                │   │                     │   │                   │   
   │                    └─────────────────────┘                       │   
┌──┴─────┐          │   ┌─────────────────────┐   │             ┌─────┴──┐
│LoadInfo│              │                     │                 │LoadInfo│
└──┬─────┘          │   │         ...         │   │             └─────┬──┘
   │                    │                     │                       │   
   │                │   └─────────────────────┘   │                   │   
   │                 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─                    │   
   │                                                                  │   
   │ ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐   ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐│   
   │     ┌─────────────────────┐           ┌─────────────────────┐    │   
   │ │   │    ProducerHead     │   │   │   │    ProducerHead     │   ││   
   │     │ (RepartitionExec or │           │ (RepartitionExec or │    │   
   │ │   │   BroadcastExec)    │   │   │   │   BroadcastExec)    │   ││   
   │     └─────────────────────┘           └─────────────────────┘    │   
   │ │   ┌─────────────────────┐   │   │   ┌─────────────────────┐   ││   
   │     │                     │           │                     │    │   
   └─┼───│     SamplerExec     │   │   │   │     SamplerExec     │───┼┘   
         │                     │           │                     │        
     │   └─────────────────────┘   │   │   └─────────────────────┘   │    
         ┌─────────────────────┐           ┌─────────────────────┐        
     │   │                     │   │   │   │                     │   │    
         │         ...         │           │         ...         │        
     │   │                     │   │   │   │                     │   │    
         └─────────────────────┘           └─────────────────────┘        
     │   ┌─────────────────────┐   │   │   ┌─────────────────────┐   │    
         │                     │           │                     │        
     │   │   DataSourceExec    │   │   │   │   DataSourceExec    │   │    
         │                     │           │                     │        
     │   └─────────────────────┘   │   │   └─────────────────────┘   │    
      ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     
```

Each `SamplerExec` node contains one `PartitionSampler` per partition, and each partition sampling is treated independently:

```rust
┌────────────────────────────────────────────────────────────────────────────┐
│                                SamplerExec                                 │
│ ┌────────────────┐┌────────────────┐                    ┌────────────────┐ │
│ │                ││                │                    │                │ │
│ │PartitionSampler││PartitionSampler│         ...        │PartitionSampler│ │
│ │                ││                │                    │                │ │
│ └────────────────┘└────────────────┘                    └────────────────┘ │
└────────────────────────────────────────────────────────────────────────────┘
```

The way a `PartitionSampler` works is that, before actual execution, it will peek a representative number of `RecordBatch`es and gather stats over them (per column byte count, per column NDV, per column null count, total row count, etc...). For that the following things happen:

1. Pulls all the `RecordBatch`es that are available synchronously without `async` gaps. Some operators like `Aggregate(Partial)` retain big internal representations in memory that can yield more than 1 `RecordBatch` synchronously, because they were already computed and stored in memory.

```rust
┌───────────────────────────────────────────────────────────────┐
│                       PartitionSampler                        │
│                                                               │
│┌───────────┬───────────┬───────────┐                          │
││           │           │           │                          │
││RecordBatch│RecordBatch│RecordBatch│                          │
││           │           │           │                          │
│└───────────┴───────────┴───────────┘                          │
└───────────────────────────────────────────────────────────────┘
```

2. Once an `async` gap is reached, it pulls just 1 more record batch, and it measures the time it takes in this `async` gap:

```rust
┌───────────────────────────────────────────────────────────────┐
│                       PartitionSampler                        │
│                                                               │
│┌───────────┬───────────┬───────────┐             ┌───────────┐│
││           │           │           │             │           ││
││RecordBatch│RecordBatch│RecordBatch│  async wait │RecordBatch││
││           │           │           │             │           ││
│└───────────┴───────────┴───────────┘             └───────────┘│
└───────────────────────────────────────────────────────────────┘
```

3. With this information, two sources of useful information are present:
  - Bytes ready to be yielded immediately
  - The velocity at with new `RecordBatch`es are arriving

```rust
┌───────────────────────────────────────────────────────────────┐
│                       PartitionSampler                        │
│                                                               │
│┌───────────┬───────────┬───────────┐             ┌───────────┐│
││           │           │           │             │           ││
││RecordBatch│RecordBatch│RecordBatch│  async wait │RecordBatch││
││           │           │           │             │           ││
│└─────┬─────┴─────┬─────┴───┬─────┬─┘             └────▲────┬─┘│
└──────┼───────────┼─────────┼─────┼────────────────────┼────┼──┘
       │           │         │     ─────────────────────│    │   
       │           │         │                               │   
       │           │         │      RecordBatch velocity     │   
       │           │         │                               │   
       │           │         │                               │   
       │           │         │                               │   
       │           │         │     ┌─────────────────────────┘   
       ▼           ▼         ▼     ▼                             
   Batches to be yielded immediately                                                                                         
```

This velocity information is what really matters towards deciding how many task count should be yielded. 

- Scanning a lot of data is not problematic if you can do it in a streaming fashion and the DataSources are bottleneck on IO, in this situations, 1 node might be enough.
- If DataSources yield data very quickly, or if the compute cost over that data is very expensive, even if it's not a lot of data, it might still bottleneck the query en CPU power, so we'll benefit from more CPUs (more machines).

# prepare_dynamic_plan

This is analogous to the existing `prepare_static_plan`, but with a runtime based dynamic task count assignation.

In order to understand how it works, let's use the classical aggregation example:

```rust
┌───────────────────┐
│CoalescePartitions │
└───────────────────┘
┌───────────────────┐
│  ProjectionExec   │
└───────────────────┘
┌───────────────────┐
│ Aggregate(final)  │
└───────────────────┘
┌───────────────────┐
│  RepartitionExec  │
└───────────────────┘
┌───────────────────┐
│Aggregate(partial) │
└───────────────────┘
┌───────────────────┐
│    FilterExec     │
└───────────────────┘
┌───────────────────┐
│  DataSourceExec   │
└───────────────────┘
```

The plan walks from bottom to top, until we delimit a stage, same as before:

```rust
                  Stage 1 
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐ 
  ┌───────────────────┐   
│ │  RepartitionExec  │ │ 
  └───────────────────┘   
│ ┌───────────────────┐ │ 
  │Aggregate(partial) │   
│ └───────────────────┘ │ 
  ┌───────────────────┐   
│ │    FilterExec     │ │ 
  └───────────────────┘   
│ ┌───────────────────┐ │ 
  │  DataSourceExec   │   
│ └───────────────────┘ │ 
 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─  
```

We'll use runtime information of "Stage 1" in order to determine the task count for the future "Stage 2". For that, the plan is modified and a `SamplerExec` is inserted just below the producer head (`RepartitionExec`):

```rust
                  Stage 1           
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐           
  ┌───────────────────┐             
│ │  RepartitionExec  │ │           
  └───────────────────┘             
│ ┌───────────────────┐ │           
  │    SamplerExec    │◀────────────
│ └───────────────────┘ │           
  ┌───────────────────┐             
│ │Aggregate(partial) │ │           
  └───────────────────┘             
│ ┌───────────────────┐ │           
  │    FilterExec     │             
│ └───────────────────┘ │           
  ┌───────────────────┐             
│ │  DataSourceExec   │ │           
  └───────────────────┘             
└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘           
```

Let's say that the `TaskEstimator` for the leaf node `DataSourceExec` decided a `Desired(3)` task estimation. This subplan is then immediately sent to three workers, and upon setting the plan on the workers, all the `SamplerExec`s are kicked off and they start sampling even before any call to `.execute()` in any node:

```rust
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─Stage 1 
   ┌ ─ ─ ─ ─ ─ ─ ─Worker 1  ┌ ─ ─ ─ ─ ─ ─ ─Worker 2  ┌ ─ ─ ─ ─ ─ ─ ─Worker 3       
│   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
   ││  RepartitionExec  ││  ││  RepartitionExec  ││  ││  RepartitionExec  ││       
│   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
   │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
│   │    SamplerExec    │    │    SamplerExec    │    │    SamplerExec    │   │    
   │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
│             │                        │                        │             │    
   │   Eager kick off    │  │   Eager kick off    │  │   Eager kick off    │       
│             ▼                        ▼                        ▼             │    
   │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
│   │Aggregate(partial) │    │Aggregate(partial) │    │Aggregate(partial) │   │    
   │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
│   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
   ││    FilterExec     ││  ││    FilterExec     ││  ││    FilterExec     ││       
│   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
   │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
│   │  DataSourceExec   │    │  DataSourceExec   │    │  DataSourceExec   │   │    
   │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │    
 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     
```

All the `SamplerExec`s start sampling, and they start reporting `LoadInfo` messages to the `prepare_dynamic_plan` function, which builds the next `NetworkShuffleExec` boundary with some pre-defined `Statistics`, the ones that where collected at runtime by the `SamplerExec`s below:

```rust
                               ┌───────────────────┐  *                              
                               │  ProjectionExec   │  *                              
                               └───────────────────┘  *                              
                               ┌───────────────────┐  *                              
                               │ Aggregate(final)  │  *                              
                               └───────────────────┘  *   Still not planned          
                               ┌───────────────────┐  *                              
                               │NetworkShuffleExec │  *                              
                               │ (+ runtime stats) │  *                              
                               └───────▲──▲──▲─────┘  *                              
                                       │  │  │                                       
┌──────────────────────────────────────┘  │  └────────┐                              
│                            ┌────────────┘           │                              
│                            │                        │                              
│ ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─Stage 1 
│    ┌ ─ ─ ─ ─ ─ ─ ─Worker 1 │┌ ─ ─ ─ ─ ─ ─ ─Worker 2 │┌ ─ ─ ─ ─ ─ ─ ─Worker 3       
│ │   ┌───────────────────┐  │ ┌───────────────────┐  │ ┌───────────────────┐   │    
│    ││  RepartitionExec  ││ │││  RepartitionExec  ││ │││  RepartitionExec  ││       
│ │   └───────────────────┘  │ └───────────────────┘  │ └───────────────────┘   │    
│    │┌───────────────────┐│ ││┌───────────────────┐│ ││┌───────────────────┐│       
└─┼───┤    SamplerExec    │  └─┤    SamplerExec    │  └─┤    SamplerExec    │   │    
     │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
  │             │                        │                        │             │    
     │   Eager kick off    │  │   Eager kick off    │  │   Eager kick off    │       
  │             ▼                        ▼                        ▼             │    
     │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
  │   │Aggregate(partial) │    │Aggregate(partial) │    │Aggregate(partial) │   │    
     │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
  │   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
     ││    FilterExec     ││  ││    FilterExec     ││  ││    FilterExec     ││       
  │   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
     │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
  │   │  DataSourceExec   │    │  DataSourceExec   │    │  DataSourceExec   │   │    
     │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
  │   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─    ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │    
   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     
```

While delimiting the next stage, the one that contains the `ProjectionExec` and the `Aggregate(final)`, the task count for that stage is decided purely over its compute cost, based on stats. At this point, the stats we have in that slice of the plan are very accurate, as we manage to gather them and condense them in `NetworkShuffleExec`, which at that point acts as a leaf node (as it's stage was already set to `Stage::Remote` because it already was sent and started sampling).

Based on the compute cost inferred from the runtime stats, we suddenly realize that there's very very little data flowing through the `SamplerExec`, and as a consequence, the compute cost of "Stage 2" is estimated to be super low, so we collapse early to a single node:

```rust
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─Stage 2   
                            ┌ ─ ─ ─ ─ ─ ─ ─Worker 1                                
│                            ┌───────────────────┐                            │    
                            ││  ProjectionExec   ││                                
│                            └───────────────────┘                            │    
                            │┌───────────────────┐│                                
│                            │ Aggregate(final)  │  Very little data flowing, │    
                            │└───────────────────┘│     1 task is enough           
│                            ┌───────────────────┐                            │    
                            ││NetworkShuffleExec ││                                
│                            │ (+ runtime stats) │                            │    
                            │└───────────────────┘│                                
│                            ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─                            │    
 ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─     
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─Stage 1 
   ┌ ─ ─ ─ ─ ─ ─ ─Worker 1  ┌ ─ ─ ─ ─ ─ ─ ─Worker 2  ┌ ─ ─ ─ ─ ─ ─ ─Worker 3       
│   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
   ││  RepartitionExec  ││  ││  RepartitionExec  ││  ││  RepartitionExec  ││       
│   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
   │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
│   │    SamplerExec    │    │    SamplerExec    │    │    SamplerExec    │   │    
   │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
│   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
   ││Aggregate(partial) ││  ││Aggregate(partial) ││  ││Aggregate(partial) ││       
│   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
   │┌───────────────────┐│  │┌───────────────────┐│  │┌───────────────────┐│       
│   │    FilterExec     │    │    FilterExec     │    │    FilterExec     │   │    
   │└───────────────────┘│  │└───────────────────┘│  │└───────────────────┘│       
│   ┌───────────────────┐    ┌───────────────────┐    ┌───────────────────┐   │    
   ││  DataSourceExec   ││  ││  DataSourceExec   ││  ││  DataSourceExec   ││       
│   └───────────────────┘    └───────────────────┘    └───────────────────┘   │    
   └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘  └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘  └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘       
└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘    
```

# Benchmarks

Extracted comparing runs of this same branch with `--dynamic false` VS `--dynamic true`:

<details><summary>TPCH SF1: Tasks 279 -> 318, prev=14631 ms, new=6727 ms, diff=2.17 faster ✅</summary>

```
      q1: prev= 319 ms, new= 307 ms, diff=1.04 faster ✔
      q2: prev=1054 ms, new= 393 ms, diff=2.68 faster ✅
      q3: prev= 989 ms, new= 504 ms, diff=1.96 faster ✅
      q4: prev= 436 ms, new= 449 ms, diff=1.03 slower ✖
      q5: prev=1019 ms, new= 314 ms, diff=3.25 faster ✅
      q6: prev= 255 ms, new= 206 ms, diff=1.24 faster ✅
      q7: prev= 934 ms, new= 310 ms, diff=3.01 faster ✅
      q8: prev=1307 ms, new= 467 ms, diff=2.80 faster ✅
      q9: prev=1208 ms, new= 331 ms, diff=3.65 faster ✅
     q10: prev= 585 ms, new= 381 ms, diff=1.54 faster ✅
     q11: prev= 380 ms, new= 183 ms, diff=2.08 faster ✅
     q12: prev= 328 ms, new= 170 ms, diff=1.93 faster ✅
     q13: prev= 477 ms, new= 176 ms, diff=2.71 faster ✅
     q14: prev= 644 ms, new= 222 ms, diff=2.90 faster ✅
     q15: prev= 946 ms, new= 175 ms, diff=5.41 faster ✅
     q16: prev= 268 ms, new= 297 ms, diff=1.11 slower ✖
     q17: prev= 828 ms, new= 281 ms, diff=2.95 faster ✅
     q18: prev= 838 ms, new= 261 ms, diff=3.21 faster ✅
     q19: prev= 424 ms, new= 313 ms, diff=1.35 faster ✅
     q20: prev= 374 ms, new= 200 ms, diff=1.87 faster ✅
     q21: prev= 815 ms, new= 608 ms, diff=1.34 faster ✅
     q22: prev= 203 ms, new= 179 ms, diff=1.13 faster ✔
   TOTAL: prev=14631 ms, new=6727 ms, diff=2.17 faster ✅
```

</details>

<details><summary>TPCH SF10: Tasks 363 -> 734, prev=35577 ms, new=13014 ms, diff=2.73 faster ✅</summary>

```
      q1: prev=1498 ms, new= 604 ms, diff=2.48 faster ✅
      q2: prev=2258 ms, new= 529 ms, diff=4.27 faster ✅
      q3: prev=2005 ms, new= 567 ms, diff=3.54 faster ✅
      q4: prev= 841 ms, new= 468 ms, diff=1.80 faster ✅
      q5: prev=2051 ms, new= 701 ms, diff=2.93 faster ✅
      q6: prev= 679 ms, new= 575 ms, diff=1.18 faster ✔
      q7: prev=1810 ms, new= 723 ms, diff=2.50 faster ✅
      q8: prev=3028 ms, new= 769 ms, diff=3.94 faster ✅
      q9: prev=3011 ms, new= 885 ms, diff=3.40 faster ✅
     q10: prev=1699 ms, new= 764 ms, diff=2.22 faster ✅
     q11: prev=1142 ms, new= 304 ms, diff=3.76 faster ✅
     q12: prev= 913 ms, new= 427 ms, diff=2.14 faster ✅
     q13: prev= 789 ms, new= 925 ms, diff=1.17 slower ✖
     q14: prev=1261 ms, new= 489 ms, diff=2.58 faster ✅
     q15: prev=2132 ms, new= 541 ms, diff=3.94 faster ✅
     q16: prev= 610 ms, new= 242 ms, diff=2.52 faster ✅
     q17: prev=1771 ms, new= 589 ms, diff=3.01 faster ✅
     q18: prev=2162 ms, new= 795 ms, diff=2.72 faster ✅
     q19: prev=1157 ms, new= 480 ms, diff=2.41 faster ✅
     q20: prev=1350 ms, new= 669 ms, diff=2.02 faster ✅
     q21: prev=2830 ms, new= 747 ms, diff=3.79 faster ✅
     q22: prev= 580 ms, new= 221 ms, diff=2.62 faster ✅
   TOTAL: prev=35577 ms, new=13014 ms, diff=2.73 faster ✅
```

</details>

<details><summary>TPCH SF100: Tasks 387 -> 1175, prev=307187 ms, new=62943 ms, diff=4.88 faster ✅</summary>

```
      q1: prev=16724 ms, new=3086 ms, diff=5.42 faster ✅
      q2: prev=3282 ms, new=1038 ms, diff=3.16 faster ✅
      q3: prev=14599 ms, new=2683 ms, diff=5.44 faster ✅
      q4: prev=9351 ms, new=1378 ms, diff=6.79 faster ✅
      q5: prev=16400 ms, new=3628 ms, diff=4.52 faster ✅
      q6: prev=9168 ms, new=1552 ms, diff=5.91 faster ✅
      q7: prev=17370 ms, new=4049 ms, diff=4.29 faster ✅
      q8: prev=22325 ms, new=4329 ms, diff=5.16 faster ✅
      q9: prev=24036 ms, new=5788 ms, diff=4.15 faster ✅
     q10: prev=14957 ms, new=5010 ms, diff=2.99 faster ✅
     q11: prev=2642 ms, new= 852 ms, diff=3.10 faster ✅
     q12: prev=10561 ms, new=1726 ms, diff=6.12 faster ✅
     q13: prev=4839 ms, new=1503 ms, diff=3.22 faster ✅
     q14: prev=9423 ms, new=1625 ms, diff=5.80 faster ✅
     q15: prev=26224 ms, new=3120 ms, diff=8.41 faster ✅
     q16: prev=1426 ms, new= 566 ms, diff=2.52 faster ✅
     q17: prev=21121 ms, new=4623 ms, diff=4.57 faster ✅
     q18: prev=25218 ms, new=5256 ms, diff=4.80 faster ✅
     q19: prev=11387 ms, new=1751 ms, diff=6.50 faster ✅
     q20: prev=13517 ms, new=2141 ms, diff=6.31 faster ✅
     q21: prev=29777 ms, new=6310 ms, diff=4.72 faster ✅
     q22: prev=2840 ms, new= 929 ms, diff=3.06 faster ✅
   TOTAL: prev=307187 ms, new=62943 ms, diff=4.88 faster ✅
```

</details>

<details><summary>TPCDS SF1: Tasks 3079->2770, prev=62815 ms, new=44737 ms, diff=1.40 faster ✅</summary>

```
     q1: prev=1039 ms, new= 531 ms, diff=1.96 faster ✅
      q2: prev= 677 ms, new= 396 ms, diff=1.71 faster ✅
      q3: prev= 618 ms, new= 288 ms, diff=2.15 faster ✅
      q4: prev=3420 ms, new=2582 ms, diff=1.32 faster ✅
      q5: prev= 767 ms, new= 554 ms, diff=1.38 faster ✅
      q6: prev=1002 ms, new= 967 ms, diff=1.04 faster ✔
      q7: prev= 614 ms, new= 421 ms, diff=1.46 faster ✅
      q8: prev= 484 ms, new= 422 ms, diff=1.15 faster ✔
      q9: prev= 267 ms, new= 341 ms, diff=1.28 slower ❌
     q10: prev= 648 ms, new= 826 ms, diff=1.27 slower ❌
     q11: prev=2281 ms, new=1663 ms, diff=1.37 faster ✅
     q12: prev= 287 ms, new= 229 ms, diff=1.25 faster ✅
     q13: prev= 962 ms, new= 720 ms, diff=1.34 faster ✅
     q14: prev=1095 ms, new= 737 ms, diff=1.49 faster ✅
     q15: prev= 314 ms, new= 212 ms, diff=1.48 faster ✅
     q16: prev= 528 ms, new= 597 ms, diff=1.13 slower ✖
     q17: prev= 768 ms, new= 257 ms, diff=2.99 faster ✅
     q18: prev= 780 ms, new= 339 ms, diff=2.30 faster ✅
     q19: prev= 280 ms, new= 269 ms, diff=1.04 faster ✔
     q20: prev= 180 ms, new= 169 ms, diff=1.07 faster ✔
     q21: prev= 708 ms, new= 382 ms, diff=1.85 faster ✅
     q22: prev= 490 ms, new= 435 ms, diff=1.13 faster ✔
     q23: prev=1149 ms, new= 567 ms, diff=2.03 faster ✅
     q24: prev=1148 ms, new= 586 ms, diff=1.96 faster ✅
     q25: prev= 558 ms, new= 275 ms, diff=2.03 faster ✅
     q26: prev= 346 ms, new= 265 ms, diff=1.31 faster ✅
     q27: prev= 480 ms, new= 328 ms, diff=1.46 faster ✅
     q28: prev= 329 ms, new= 221 ms, diff=1.49 faster ✅
     q29: prev= 663 ms, new= 251 ms, diff=2.64 faster ✅
     q31: prev=1268 ms, new= 218 ms, diff=5.82 faster ✅
     q32: prev= 426 ms, new= 190 ms, diff=2.24 faster ✅
     q33: prev= 429 ms, new= 321 ms, diff=1.34 faster ✅
     q34: prev= 424 ms, new= 373 ms, diff=1.14 faster ✔
     q35: prev= 655 ms, new= 546 ms, diff=1.20 faster ✔
     q36: prev= 388 ms, new= 290 ms, diff=1.34 faster ✅
     q37: prev= 389 ms, new= 384 ms, diff=1.01 faster ✔
     q38: prev= 568 ms, new= 252 ms, diff=2.25 faster ✅
     q39: prev= 547 ms, new= 479 ms, diff=1.14 faster ✔
     q40: prev= 612 ms, new= 417 ms, diff=1.47 faster ✅
     q41: prev= 128 ms, new= 143 ms, diff=1.12 slower ✖
     q42: prev= 331 ms, new= 139 ms, diff=2.38 faster ✅
     q43: prev= 158 ms, new= 204 ms, diff=1.29 slower ❌
     q44: prev= 514 ms, new= 353 ms, diff=1.46 faster ✅
     q45: prev= 398 ms, new= 237 ms, diff=1.68 faster ✅
     q46: prev= 641 ms, new= 495 ms, diff=1.29 faster ✅
     q47: prev= 964 ms, new= 488 ms, diff=1.98 faster ✅
     q48: prev= 514 ms, new= 418 ms, diff=1.23 faster ✅
     q49: prev= 550 ms, new= 314 ms, diff=1.75 faster ✅
     q50: prev= 549 ms, new= 445 ms, diff=1.23 faster ✅
     q51: prev= 425 ms, new= 226 ms, diff=1.88 faster ✅
     q52: prev= 252 ms, new= 117 ms, diff=2.15 faster ✅
     q53: prev= 202 ms, new= 223 ms, diff=1.10 slower ✖
     q54: prev= 679 ms, new= 383 ms, diff=1.77 faster ✅
     q55: prev= 191 ms, new= 155 ms, diff=1.23 faster ✅
     q56: prev= 448 ms, new= 332 ms, diff=1.35 faster ✅
     q57: prev= 673 ms, new= 288 ms, diff=2.34 faster ✅
     q58: prev= 644 ms, new= 289 ms, diff=2.23 faster ✅
     q59: prev= 447 ms, new= 294 ms, diff=1.52 faster ✅
     q60: prev= 301 ms, new= 291 ms, diff=1.03 faster ✔
     q61: prev=1017 ms, new=1095 ms, diff=1.08 slower ✖
     q62: prev= 776 ms, new= 761 ms, diff=1.02 faster ✔
     q63: prev= 243 ms, new= 223 ms, diff=1.09 faster ✔
     q64: prev=1907 ms, new=1370 ms, diff=1.39 faster ✅
     q65: prev= 432 ms, new= 259 ms, diff=1.67 faster ✅
     q66: prev= 839 ms, new= 781 ms, diff=1.07 faster ✔
     q67: prev= 486 ms, new= 389 ms, diff=1.25 faster ✅
     q68: prev= 432 ms, new= 373 ms, diff=1.16 faster ✔
     q69: prev= 596 ms, new= 618 ms, diff=1.04 slower ✖
     q70: prev= 440 ms, new= 541 ms, diff=1.23 slower ❌
     q71: prev= 406 ms, new= 298 ms, diff=1.36 faster ✅
     q72: prev=5276 ms, new=3156 ms, diff=1.67 faster ✅
     q73: prev= 259 ms, new= 294 ms, diff=1.14 slower ✖
     q74: prev= 733 ms, new= 656 ms, diff=1.12 faster ✔
     q75: prev=1004 ms, new= 542 ms, diff=1.85 faster ✅
     q76: prev= 362 ms, new= 232 ms, diff=1.56 faster ✅
     q77: prev= 549 ms, new= 305 ms, diff=1.80 faster ✅
     q78: prev= 954 ms, new= 299 ms, diff=3.19 faster ✅
     q79: prev= 278 ms, new= 266 ms, diff=1.05 faster ✔
     q80: prev= 689 ms, new= 398 ms, diff=1.73 faster ✅
     q81: prev= 441 ms, new= 344 ms, diff=1.28 faster ✅
     q82: prev= 315 ms, new= 310 ms, diff=1.02 faster ✔
     q83: prev= 551 ms, new= 299 ms, diff=1.84 faster ✅
     q84: prev= 321 ms, new= 359 ms, diff=1.12 slower ✖
     q85: prev= 594 ms, new= 545 ms, diff=1.09 faster ✔
     q86: prev= 168 ms, new= 186 ms, diff=1.11 slower ✖
     q87: prev= 394 ms, new= 283 ms, diff=1.39 faster ✅
     q88: prev= 479 ms, new= 484 ms, diff=1.01 slower ✖
     q89: prev= 262 ms, new= 207 ms, diff=1.27 faster ✅
     q90: prev= 352 ms, new= 274 ms, diff=1.28 faster ✅
     q91: prev= 541 ms, new= 404 ms, diff=1.34 faster ✅
     q92: prev= 326 ms, new= 207 ms, diff=1.57 faster ✅
     q93: prev= 287 ms, new= 226 ms, diff=1.27 faster ✅
     q94: prev= 401 ms, new= 451 ms, diff=1.12 slower ✖
     q95: prev= 444 ms, new= 354 ms, diff=1.25 faster ✅
     q96: prev= 193 ms, new= 235 ms, diff=1.22 slower ❌
     q97: prev= 268 ms, new= 174 ms, diff=1.54 faster ✅
     q98: prev= 165 ms, new= 176 ms, diff=1.07 slower ✖
     q99: prev=1038 ms, new=1229 ms, diff=1.18 slower ✖
   TOTAL: prev=62815 ms, new=44737 ms, diff=1.40 faster ✅
```

</details>

<details><summary>ClickBench 0-100: Tasks 912->609, prev=35116 ms, new=34185 ms, diff=1.03 faster ✔</summary>

```
      q0: prev=   2 ms, new=   2 ms, diff=1.00 slower ✖
      q1: prev= 506 ms, new= 396 ms, diff=1.28 faster ✅
      q2: prev= 477 ms, new= 337 ms, diff=1.42 faster ✅
      q3: prev= 323 ms, new= 452 ms, diff=1.40 slower ❌
      q4: prev= 405 ms, new= 439 ms, diff=1.08 slower ✖
      q5: prev= 585 ms, new= 555 ms, diff=1.05 faster ✔
      q6: prev=   2 ms, new=   2 ms, diff=1.00 slower ✖
      q7: prev= 253 ms, new= 363 ms, diff=1.43 slower ❌
      q8: prev= 492 ms, new= 477 ms, diff=1.03 faster ✔
      q9: prev= 620 ms, new= 635 ms, diff=1.02 slower ✖
     q10: prev= 478 ms, new= 461 ms, diff=1.04 faster ✔
     q11: prev= 449 ms, new= 392 ms, diff=1.15 faster ✔
     q12: prev= 596 ms, new= 550 ms, diff=1.08 faster ✔
     q13: prev= 771 ms, new= 792 ms, diff=1.03 slower ✖
     q14: prev= 536 ms, new= 549 ms, diff=1.02 slower ✖
     q15: prev= 477 ms, new= 414 ms, diff=1.15 faster ✔
     q16: prev= 775 ms, new= 753 ms, diff=1.03 faster ✔
     q17: prev= 749 ms, new= 726 ms, diff=1.03 faster ✔
     q18: prev=1114 ms, new=1038 ms, diff=1.07 faster ✔
     q19: prev= 380 ms, new= 391 ms, diff=1.03 slower ✖
     q20: prev=1886 ms, new=1786 ms, diff=1.06 faster ✔
     q21: prev=1807 ms, new=1775 ms, diff=1.02 faster ✔
     q22: prev=2144 ms, new=2053 ms, diff=1.04 faster ✔
     q23: prev=5267 ms, new=5354 ms, diff=1.02 slower ✖
     q24: prev= 461 ms, new= 554 ms, diff=1.20 slower ❌
     q25: prev= 409 ms, new= 482 ms, diff=1.18 slower ✖
     q26: prev= 582 ms, new= 463 ms, diff=1.26 faster ✅
     q27: prev=1958 ms, new=1955 ms, diff=1.00 faster ✔
     q28: prev=2504 ms, new=2367 ms, diff=1.06 faster ✔
     q29: prev= 297 ms, new= 261 ms, diff=1.14 faster ✔
     q30: prev= 614 ms, new= 589 ms, diff=1.04 faster ✔
     q31: prev= 714 ms, new= 656 ms, diff=1.09 faster ✔
     q32: prev= 946 ms, new= 891 ms, diff=1.06 faster ✔
     q33: prev=2557 ms, new=2397 ms, diff=1.07 faster ✔
     q34: prev=2521 ms, new=2440 ms, diff=1.03 faster ✔
     q35: prev= 459 ms, new= 438 ms, diff=1.05 faster ✔
   TOTAL: prev=35116 ms, new=34185 ms, diff=1.03 faster ✔
```

</details>

### DIFF
--- a/src/common/vec.rs
+++ b/src/common/vec.rs
@@ -1,0 +1,80 @@
+use datafusion::common::internal_err;
+use datafusion::error::Result;
+use num_traits::AsPrimitive;
+use std::ops::{AddAssign, DivAssign, MulAssign};
+
+/// Converts a slice of type `I` to a `Vec<O>` using `as`-style primitive casting.
+pub(crate) fn vec_cast<I, O>(input: &[I]) -> Vec<O>
+where
+    I: AsPrimitive<O>,
+    O: Copy + 'static,
+{
+    input.iter().map(|v| v.as_()).collect()
+}
+
+/// Adds each element of `other` into the corresponding element of `one`, converting types via `AsPrimitive`.
+pub(crate) fn element_wise_sum<I, O>(mut one: Vec<I>, other: &[O]) -> Result<Vec<I>>
+where
+    I: AddAssign + Copy + 'static,
+    O: AsPrimitive<I> + 'static,
+{
+    if one.len() != other.len() {
+        return internal_err!("Cannot do an element wise sum of two vectors of different lengths");
+    }
+    for i in 0..one.len() {
+        one[i] += other[i].as_();
+    }
+    Ok(one)
+}
+
+/// Multiplies every element of `one` by the scalar `other`, converting types via `AsPrimitive`.
+pub(crate) fn vec_mul<I, O>(mut one: Vec<I>, other: O) -> Vec<I>
+where
+    I: MulAssign + Copy + 'static,
+    O: AsPrimitive<I> + 'static,
+{
+    for el in one.iter_mut() {
+        *el *= other.as_();
+    }
+    one
+}
+
+/// Divides every element of `one` by the scalar `other`, converting types via `AsPrimitive`.
+pub(crate) fn vec_div<I, O>(mut one: Vec<I>, other: O) -> Vec<I>
+where
+    I: DivAssign + Copy + 'static,
+    O: AsPrimitive<I> + 'static,
+{
+    for el in one.iter_mut() {
+        *el /= other.as_();
+    }
+    one
+}
+
+/// Reduces a collection of same-length `f32` vectors into a single vector by averaging element-wise.
+/// Empty inner vecs are skipped; returns an empty vec if all inputs are empty.
+pub(crate) fn vec_avg_reduce(vecs: Vec<Vec<f32>>) -> Result<Vec<f32>> {
+    let sample_count = vecs.len();
+    let mut iter = vecs.into_iter();
+    let mut acc = loop {
+        let Some(v) = iter.next() else {
+            return Ok(vec![]);
+        };
+        if !v.is_empty() {
+            break v;
+        }
+    };
+    for v in iter {
+        if v.is_empty() {
+            continue;
+        } else if acc.len() != v.len() {
+            return internal_err!(
+                "vec_avg_reduce: length mismatch — first vec has {} elements, got {}",
+                acc.len(),
+                v.len()
+            );
+        }
+        acc = element_wise_sum(acc, &v)?;
+    }
+    Ok(vec_div(acc, sample_count as f32))
+}
--- a/src/coordinator/prepare_dynamic_plan.rs
+++ b/src/coordinator/prepare_dynamic_plan.rs
@@ -1,0 +1,361 @@
+use crate::TaskCountAnnotation::{Desired, Maximum};
+use crate::common::{TreeNodeExt, element_wise_sum, vec_avg_reduce, vec_div, vec_mul};
+use crate::coordinator::distributed::PreparedPlan;
+use crate::coordinator::query_coordinator::QueryCoordinator;
+use crate::distributed_planner::{
+    InjectNetworkBoundaryContext, NetworkBoundaryBuilderResult, ProducerHead, calculate_cost,
+    inject_network_boundaries,
+};
+use crate::execution_plans::SamplerExec;
+use crate::stage::{LocalStage, RemoteStage};
+use crate::worker::generated::worker as pb;
+use crate::{BytesCounterMetric, NetworkBoundaryExt, NetworkCoalesceExec, Stage};
+use dashmap::DashMap;
+use datafusion::common::stats::Precision;
+use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeRecursion};
+use datafusion::common::{Result, exec_err, plan_err};
+use datafusion::physical_plan::metrics::MetricsSet;
+use datafusion::physical_plan::{
+    ColumnStatistics, ExecutionPlan, ExecutionPlanProperties, Statistics,
+};
+use futures::{Stream, StreamExt};
+use std::any::TypeId;
+use std::sync::Arc;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+
+pub(super) async fn prepare_dynamic_plan(
+    query_coordinator: &QueryCoordinator,
+    base_plan: &Arc<dyn ExecutionPlan>,
+) -> Result<PreparedPlan> {
+    let plans_for_viz = Arc::new(PlanReconstructor::default());
+
+    let head_stage = inject_network_boundaries(
+        Arc::clone(base_plan),
+        |mut input_stage: LocalStage, nb_type: TypeId, nb_ctx: &InjectNetworkBoundaryContext| {
+            let mut metrics = MetricsSet::new();
+
+            // At this point, input_stage.plan has two kind of leaf nodes:
+            // - The ones that naturally do not read from any children, like DataSourceExec
+            // - Network boundaries whose Stage was set to Stage::Remote by a previous iteration
+            //   of this same function.
+            // Both types of leaf nodes contain very valuable and accurate statistics that are used
+            // here for computing an estimation of the compute cost (measured in bytes):
+            // - DataSourceExec (or natural leaf nodes) contain stats pulled directly from their
+            //   data source, like parquet files.
+            // - Network boundaries contain statistics collected from runtime information, gathered
+            //   by the SamplerExec injected by this same function.
+            let cost = calculate_cost(&input_stage.plan)?;
+            metrics.push(BytesCounterMetric::new_metric(
+                "cpu_cost",
+                *cost.cpu.get_value().unwrap_or(&0),
+            ));
+            metrics.push(BytesCounterMetric::new_metric(
+                "memory_cost",
+                *cost.memory.get_value().unwrap_or(&0),
+            ));
+            metrics.push(BytesCounterMetric::new_metric(
+                "network_cost",
+                *cost.network.get_value().unwrap_or(&0),
+            ));
+            let compute_based_task_count = cost
+                .cpu
+                .get_value()
+                .unwrap_or(&0)
+                .div_ceil(nb_ctx.d_cfg.bytes_per_partition_per_second.max(1))
+                .div_ceil(input_stage.plan.output_partitioning().partition_count())
+                .clamp(1, nb_ctx.max_tasks()?);
+            let task_count = nb_ctx
+                .task_count(&input_stage.plan)?
+                .merge(Desired(compute_based_task_count));
+
+            // Propagate the final task_count inferred based on runtime statistics and compute cost.
+            // Here is where leaf nodes are scaled up by TaskEstimator::scale_up_leaf_node, and the
+            // plan is finally left ready for distribution.
+            input_stage.plan = nb_ctx
+                .propagate_task_count_until_network_boundaries(&input_stage.plan, task_count)?;
+            input_stage.tasks = task_count.as_usize();
+            // In order to infer the compute the cost of the stage above this one, here a sampler
+            // is injected to gather runtime statistics.
+            input_stage.plan = ProducerHead::insert_sampler(input_stage.plan)?;
+
+            let mut stage_coordinator = query_coordinator.stage_coordinator(&input_stage);
+
+            let mut workers = Vec::with_capacity(input_stage.tasks);
+            let mut load_info_rxs = Vec::with_capacity(input_stage.tasks);
+
+            let routed_urls = if input_stage.tasks == 1 {
+                // If there's an input stage with a single worker, and the current stage is also
+                // going to run in a single worker, we want to co-locate them so that unnecessary
+                // network transfers are avoided.
+                match stage_coordinator.find_input_stage_with_single_url() {
+                    Some(single_url) => vec![single_url],
+                    None => stage_coordinator.routed_urls()?,
+                }
+            } else {
+                stage_coordinator.routed_urls()?
+            };
+
+            for (i, routed_url) in routed_urls.into_iter().enumerate() {
+                workers.push(routed_url.clone());
+                // Spawns the task that feeds this subplan to this worker. There will be as
+                // many as this spawned tasks as workers.
+                let (worker_tx, worker_rx) = stage_coordinator.send_plan_task(i, routed_url)?;
+                load_info_rxs.push({
+                    let rx = stage_coordinator.worker_to_coordinator_task(i, worker_rx);
+                    UnboundedReceiverStream::new(rx)
+                });
+                stage_coordinator.coordinator_to_worker_task(i, worker_tx)?;
+            }
+
+            let plans_for_viz = Arc::clone(&plans_for_viz);
+            Ok(async move {
+                let (stats, consumer_tc) = if nb_type == TypeId::of::<NetworkCoalesceExec>() {
+                    (None, Maximum(1))
+                } else {
+                    let stats = gather_runtime_statistics(load_info_rxs, &input_stage.plan).await?;
+                    let sampled_bytes = *stats.total_byte_size.get_value().unwrap_or(&0);
+                    metrics.push(BytesCounterMetric::new_metric(
+                        "sampled_bytes",
+                        sampled_bytes,
+                    ));
+                    // returning Desired(1) here is our way to tell the planner that we don't care
+                    // about the task count assigned to the network boundary in the consumer stage,
+                    // and we don't want it to affect other task count decisions.
+                    (Some(Arc::new(stats)), Desired(1))
+                };
+
+                // Capture the output partitioning of the (rescaled, sampler-wrapped) input plan
+                // before it's moved: the returned stage is remote and carries no plan to read it
+                // back from.
+                let input_properties = Arc::clone(input_stage.plan.properties());
+                plans_for_viz.insert(input_stage.num, input_stage.plan, metrics);
+                Ok(NetworkBoundaryBuilderResult {
+                    consumer_task_count: consumer_tc,
+                    input_stage: Stage::Remote(RemoteStage {
+                        query_id: input_stage.query_id,
+                        num: input_stage.num,
+                        workers,
+                        runtime_stats: stats,
+                    }),
+                    input_properties,
+                })
+            })
+        },
+        query_coordinator.session_config().options(),
+    )
+    .await?;
+
+    Ok(PreparedPlan {
+        plan_for_viz: plans_for_viz.reconstruct(&head_stage)?,
+        head_stage,
+    })
+}
+
+/// Reconstructs the plan dynamically as stages get transitioned to Remote and get sent to the
+/// respective workers.
+///
+/// As the [prepare_dynamic_plan] function recurses and progressively sends the plan to workers, the
+/// original plan gets modified, and subplans belong to the different [Stage]s get lost as they get
+/// transitioned to [Stage::Remote].
+///
+/// This struct is in charge of tracking the [prepare_dynamic_plan] process and storing the final
+/// version of all the subplans so that it can be reconstructed into a fully blown plan for
+/// visualization purposes.
+#[derive(Default)]
+struct PlanReconstructor {
+    stage_map: DashMap<usize, (Arc<dyn ExecutionPlan>, MetricsSet)>,
+}
+
+impl PlanReconstructor {
+    fn insert(&self, stage: usize, plan: Arc<dyn ExecutionPlan>, metrics_set: MetricsSet) {
+        self.stage_map.insert(stage, (plan, metrics_set));
+    }
+
+    fn reconstruct(&self, head_stage: &Arc<dyn ExecutionPlan>) -> Result<Arc<dyn ExecutionPlan>> {
+        let head_stage = Arc::clone(head_stage);
+        let reconstructed = head_stage.transform_down_with_task_count(1, |plan, tc| {
+            let Some(nb) = plan.as_network_boundary() else {
+                return Ok(Transformed::no(plan));
+            };
+            let input_stage = nb.input_stage();
+            let Some((_, entry)) = self.stage_map.remove(&input_stage.num()) else {
+                return exec_err!(
+                    "Failed to retrieve plan for stage {} for visualization purposes",
+                    input_stage.num()
+                );
+            };
+            let (plan_for_viz, metrics_set) = entry;
+
+            let plan_for_viz = nb.producer_head(tc).insert(plan_for_viz)?;
+
+            let nb = nb.with_input_stage(Stage::Local(LocalStage {
+                query_id: input_stage.query_id(),
+                num: input_stage.num(),
+                plan: plan_for_viz,
+                tasks: input_stage.task_count(),
+                metrics_set,
+            }))?;
+
+            Ok(Transformed::yes(nb))
+        })?;
+        Ok(reconstructed.data)
+    }
+}
+
+/// Estimates the bytes per second flowing through a stage by reading sample information.
+async fn gather_runtime_statistics(
+    per_task_load_info_stream: Vec<impl Stream<Item = pb::LoadInfo> + Unpin>,
+    plan: &Arc<dyn ExecutionPlan>,
+) -> Result<Statistics> {
+    const ESTIMATED_QUERY_TIME_S: usize = 10;
+    const BYTES_READY_SAMPLE_PERCENTAGE: f32 = 0.2;
+    const BYTES_PER_SECOND_SAMPLE_PERCENTAGE: f32 = 0.2;
+
+    let Some(sampler) = find_sampler(plan) else {
+        return plan_err!("Mising SamplerExec while gathering load report");
+    };
+    let n_cols = sampler.schema().fields.len();
+
+    fn apply_pct(value: usize, pct: f32) -> usize {
+        (value as f32 * pct).round() as usize
+    }
+
+    let partitions_per_task = sampler.partition_samplers.len();
+    let task_count = per_task_load_info_stream.len();
+    let total_partitions = partitions_per_task * task_count;
+
+    let mut partitions_with_bytes_per_second_done = 0;
+    let mut partitions_with_bytes_ready_done = 0;
+    let mut partitions_done = 0;
+    let mut rows_ready = 0;
+    let mut rows_per_second = 0;
+    let mut per_col_bytes_ready = vec![0usize; n_cols];
+    let mut per_col_bytes_per_second = vec![0usize; n_cols];
+
+    let mut ndv_pct = vec![];
+    let mut null_pct = vec![];
+
+    let mut load_info_stream = futures::stream::select_all(per_task_load_info_stream);
+    while let Some(load_info) = load_info_stream.next().await {
+        rows_per_second += load_info.rows_per_second as usize;
+        rows_ready += load_info.rows_ready as usize;
+        per_col_bytes_per_second = element_wise_sum(
+            per_col_bytes_per_second,
+            &load_info.per_column_bytes_per_second,
+        )?;
+        per_col_bytes_ready =
+            element_wise_sum(per_col_bytes_ready, &load_info.per_column_bytes_ready)?;
+        ndv_pct.push(load_info.per_column_ndv_percentage);
+        null_pct.push(load_info.per_column_null_percentage);
+
+        partitions_with_bytes_per_second_done +=
+            load_info.per_column_bytes_per_second.iter().any(|v| *v > 0) as usize;
+        partitions_with_bytes_ready_done +=
+            load_info.per_column_bytes_ready.iter().any(|v| *v > 0) as usize;
+        partitions_done += 1;
+
+        // Short circuit if we collected enough bytes_ready measurements.
+        if partitions_with_bytes_ready_done
+            >= apply_pct(total_partitions, BYTES_READY_SAMPLE_PERCENTAGE).max(1)
+        {
+            break;
+        }
+
+        // Short circuit if we collected enough bytes_per_second measurements.
+        if partitions_with_bytes_per_second_done
+            >= apply_pct(total_partitions, BYTES_PER_SECOND_SAMPLE_PERCENTAGE).max(1)
+        {
+            break;
+        }
+
+        // Short circuit if there are no further partitions remaining to sample from.
+        if partitions_done == total_partitions {
+            break;
+        }
+    }
+
+    if partitions_done == 0 {
+        return Ok(zero_stats(plan.schema().fields.len()));
+    }
+
+    let per_col_bytes_ready = vec_div(
+        vec_mul(per_col_bytes_ready, total_partitions),
+        partitions_done,
+    );
+    let per_col_bytes_per_second = vec_div(
+        vec_mul(per_col_bytes_per_second, total_partitions),
+        partitions_done,
+    );
+
+    let rows_ready = rows_ready * total_partitions / partitions_done;
+    let rows_per_second = rows_per_second * total_partitions / partitions_done;
+
+    let total_num_rows = rows_ready + rows_per_second * ESTIMATED_QUERY_TIME_S;
+
+    if total_num_rows == 0 {
+        return Ok(zero_stats(n_cols));
+    }
+
+    let per_col_byte_size = element_wise_sum(
+        per_col_bytes_ready,
+        &vec_mul(per_col_bytes_per_second, ESTIMATED_QUERY_TIME_S),
+    )?;
+    let total_byte_size: usize = per_col_byte_size.iter().sum();
+
+    let ndv_pct = vec_avg_reduce(ndv_pct)?;
+    if ndv_pct.len() != n_cols {
+        return plan_err!("Expected {n_cols} ndv values, but got {}", ndv_pct.len());
+    }
+    let null_pct = vec_avg_reduce(null_pct)?;
+    if null_pct.len() != n_cols {
+        return plan_err!("Expected {n_cols} null values, but got {}", null_pct.len());
+    }
+
+    Ok(Statistics {
+        num_rows: Precision::Inexact(total_num_rows),
+        total_byte_size: Precision::Inexact(total_byte_size),
+        column_statistics: ndv_pct
+            .into_iter()
+            .zip(null_pct)
+            .zip(per_col_byte_size)
+            .map(|((ndv, null), col_bytes)| ColumnStatistics {
+                null_count: Precision::Inexact((null * total_num_rows as f32) as usize),
+                distinct_count: Precision::Inexact((ndv * total_num_rows as f32) as usize),
+                byte_size: Precision::Inexact(col_bytes),
+                max_value: Precision::Absent,
+                min_value: Precision::Absent,
+                sum_value: Precision::Absent,
+            })
+            .collect(),
+    })
+}
+
+fn find_sampler(plan: &Arc<dyn ExecutionPlan>) -> Option<&SamplerExec> {
+    let mut sampler = None;
+    plan.apply(|plan| {
+        if let Some(node) = plan.downcast_ref::<SamplerExec>() {
+            sampler = Some(node);
+            return Ok(TreeNodeRecursion::Stop);
+        };
+        Ok(TreeNodeRecursion::Continue)
+    })
+    .expect("Cannot fail");
+    sampler
+}
+
+fn zero_stats(n_cols: usize) -> Statistics {
+    Statistics {
+        num_rows: Precision::Exact(0),
+        total_byte_size: Precision::Exact(0),
+        column_statistics: (0..n_cols)
+            .map(|_| ColumnStatistics {
+                null_count: Precision::Exact(0),
+                max_value: Precision::Absent,
+                min_value: Precision::Absent,
+                sum_value: Precision::Absent,
+                distinct_count: Precision::Exact(0),
+                byte_size: Precision::Exact(0),
+            })
+            .collect(),
+    }
+}
--- a/src/coordinator/query_coordinator.rs
+++ b/src/coordinator/query_coordinator.rs
@@ -12,17 +12,19 @@ use crate::worker::generated::worker::coordinator_to_worker_msg::Inner;
 use crate::worker::generated::worker::set_plan_request::WorkUnitFeedDeclaration;
 use crate::{
     BytesCounterMetric, BytesMetricExt, DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, DistributedCodec,
-    DistributedConfig, DistributedTaskContext, DistributedWorkUnitFeedContext, TaskEstimator,
-    TaskKey, TaskRoutingContext, get_distributed_channel_resolver, get_distributed_worker_resolver,
+    DistributedConfig, DistributedTaskContext, DistributedWorkUnitFeedContext, NetworkBoundaryExt,
+    Stage, TaskEstimator, TaskKey, TaskRoutingContext, get_distributed_channel_resolver,
+    get_distributed_worker_resolver,
 };
 use datafusion::common::instant::Instant;
 use datafusion::common::runtime::JoinSet;
-use datafusion::common::tree_node::{Transformed, TreeNodeRecursion};
+use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeRecursion};
 use datafusion::common::{DataFusionError, exec_datafusion_err};
 use datafusion::common::{Result, exec_err};
 use datafusion::execution::TaskContext;
 use datafusion::physical_expr_common::metrics::{ExecutionPlanMetricsSet, Label, MetricBuilder};
 use datafusion::physical_plan::ExecutionPlan;
+use datafusion::prelude::SessionConfig;
 use datafusion_proto::physical_plan::AsExecutionPlan;
 use datafusion_proto::protobuf::PhysicalPlanNode;
 use futures::{Stream, StreamExt};
@@ -86,6 +88,11 @@ impl QueryCoordinator {
             end_stream_notifier: &self.end_stream_notifier,
             join_set: &self.join_set,
         }
+    }
+
+    /// Returns the [SessionConfig] for the current query.
+    pub(super) fn session_config(&self) -> &SessionConfig {
+        self.task_ctx.session_config()
     }
 
     /// returns a guard that, when dropped, it signals all the coordinator->worker connections that
@@ -200,8 +207,8 @@ impl<'a> StageCoordinator<'a> {
             let mut worker_to_coordinator_stream = response.into_inner();
             while let Some(msg_or_err) = worker_to_coordinator_stream.next().await {
                 let msg = msg_or_err.map_err(|err| {
-                    tonic_status_to_datafusion_error(err).unwrap_or_else(|| {
-                        exec_datafusion_err!("Unknown error on worker to coordinator stream")
+                    tonic_status_to_datafusion_error(&err).unwrap_or_else(|| {
+                        exec_datafusion_err!("Unknown error on worker to coordinator stream: {err}")
                     })
                 })?;
                 if worker_to_coordinator_tx.send(msg).is_err() {
@@ -221,13 +228,15 @@ impl<'a> StageCoordinator<'a> {
         &mut self,
         task_i: usize,
         mut worker_to_coordinator_rx: UnboundedReceiver<pb::WorkerToCoordinatorMsg>,
-    ) {
+    ) -> UnboundedReceiver<pb::LoadInfo> {
         let task_key = TaskKey {
             query_id: serialize_uuid(&self.query_id),
             stage_id: self.stage_id as u64,
             task_number: task_i as u64,
         };
         let task_metrics = self.metrics_store.clone();
+        let (load_info_tx, load_info_rx) = tokio::sync::mpsc::unbounded_channel();
+        let mut load_info_tx_opt = Some(load_info_tx);
 
         // Cannot use self.join_set because that's tied to the lifetime of the query, and the
         // metrics collection process might outlive the query's lifetime.
@@ -242,9 +251,18 @@ impl<'a> StageCoordinator<'a> {
                             task_metrics.insert(task_key.clone(), pre_order_metrics);
                         }
                     }
+                    pb::worker_to_coordinator_msg::Inner::LoadInfo(load_info) => {
+                        if let Some(tx) = &load_info_tx_opt {
+                            let _ = tx.send(load_info);
+                        }
+                    }
+                    pb::worker_to_coordinator_msg::Inner::LoadInfoEos(_) => {
+                        let _ = load_info_tx_opt.take();
+                    }
                 }
             }
         });
+        load_info_rx
     }
 
     /// Spawns a background task in charge of sending messages to workers. Some things that are sent
@@ -400,6 +418,28 @@ impl<'a> StageCoordinator<'a> {
             );
         }
         Ok(routed_urls)
+    }
+
+    pub(super) fn find_input_stage_with_single_url(&self) -> Option<Url> {
+        let mut single_stage_url = None;
+        self.plan
+            .apply(|plan| {
+                let Some(nb) = plan.as_network_boundary() else {
+                    return Ok(TreeNodeRecursion::Continue);
+                };
+
+                if let Stage::Remote(remote) = nb.input_stage()
+                    && remote.workers.len() == 1
+                {
+                    single_stage_url = Some(remote.workers[0].clone());
+                    return Ok(TreeNodeRecursion::Stop);
+                }
+
+                Ok(TreeNodeRecursion::Jump)
+            })
+            .expect("Cannot fail");
+
+        single_stage_url
     }
 }
 
--- a/src/distributed_planner/distributed_query_planner.rs
+++ b/src/distributed_planner/distributed_query_planner.rs
@@ -108,6 +108,14 @@ impl QueryPlanner for DistributedQueryPlanner {
 
         plan = insert_broadcast_execs(plan, cfg)?;
 
+        if d_cfg.dynamic_task_count {
+            // The task count will be decided dynamically at execution time.
+            return Ok(Arc::new(
+                DistributedExec::new(plan).with_metrics_collection(d_cfg.collect_metrics),
+            ));
+        }
+
+        // Compute per-node task counts and inject `Network*Exec` nodes at the stage boundaries.
         plan = inject_network_boundaries(plan, CardinalityBasedNetworkBoundaryBuilder, cfg).await?;
 
         plan = prepare_network_boundaries(plan)?;
--- a/src/distributed_planner/network_boundary.rs
+++ b/src/distributed_planner/network_boundary.rs
@@ -1,3 +1,4 @@
+use crate::execution_plans::SamplerExec;
 use crate::{BroadcastExec, NetworkBroadcastExec, NetworkCoalesceExec, NetworkShuffleExec, Stage};
 use datafusion::common::Result;
 use datafusion::physical_expr::Partitioning;
@@ -13,7 +14,7 @@ pub trait NetworkBoundary: ExecutionPlan {
     /// information to perform any internal transformations necessary for distributed execution.
     ///
     /// Typically, [NetworkBoundary]s will use this call for transitioning from "Pending" to "ready".
-    fn with_input_stage(&self, input_stage: Stage) -> Result<Arc<dyn ExecutionPlan>>;
+    fn with_input_stage(&self, input_stage: Stage) -> Result<Arc<dyn NetworkBoundary>>;
 
     /// Returns the assigned input [Stage], if any.
     fn input_stage(&self) -> &Stage;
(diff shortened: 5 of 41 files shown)


src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+    /// Time since [SamplerExec::kick_off_first_sampler] was called until the first batch from
+    /// the input arrived
+    kick_off_to_fist_batch_p50: P50LatencyMetric,
+    kick_off_to_fist_batch_max: MaxLatencyMetric,
+    /// Time since [SamplerExec::kick_off_first_sampler] was called until the [pb::LoadInfo] message
+    /// was sent.
+    kick_off_to_load_info_sent_p50: P50LatencyMetric,
+    kick_off_to_load_info_sent_max: MaxLatencyMetric,
+    /// Time since [SamplerExec::kick_off_first_sampler] was called until the first batch from
+    /// the input arrived
```
jayshrivastava: Wrong comment.

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+    kick_off_to_execution_p50: P50LatencyMetric,
+    kick_off_to_execution_max: MaxLatencyMetric,
+    /// Maximum number of record batches buffered by a sampler.
+    max_batches_buffered: MaxGaugeMetric,
+    /// Peak memory buffered by any partition sampler during the sampling phase.
+    max_mem_used: Gauge,
```
jayshrivastava, line 72: I dont think this measures what you want. Right now we do this for every batch in all partitions.
```
        self.max_mem_used.add(batch_size);
```
So this measures the size of all the batches pushed to the buffer.

You probably want to use an int to store the total size of bytes in the buffer on every push/pop. Then make this a  `MaxGaugeMetric` and update it on every push/pop.

jayshrivastava: Also none of these metrics are tagged by partition. We should probably do that. SamplerExecMetrics::new() can take a partition id.

jayshrivastava: And the comment on `MaxGaugeMetric` should say that it only stores a max.
```
/// Similar to DataFusion's Gauge metric, but aggregates between instances using `max` instead of
/// `sum`.
```
I honestly thought it was a regular Guage with a different aggregator.

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+impl SamplerExec {
+    pub(crate) fn new(input: Arc<dyn ExecutionPlan>) -> Self {
+        let metric_set = ExecutionPlanMetricsSet::new();
+        let metric_set_clone = metric_set.clone();
+        // Metrics need to be lazily initialized, otherwise the coordinator side will register
+        // them when they are never relevant there, they are just relevant in workers.
+        let metrics: Arc<LazyLock<_, Box<dyn FnOnce() -> SamplerExecMetrics + Send>>> =
+            Arc::new(LazyLock::new(Box::new(move || {
+                SamplerExecMetrics::new(&metric_set_clone)
+            })));
```
jayshrivastava, line 117: I don't understand this, or the comment.

> otherwise the coordinator side will register them when they are never relevant there

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+
+        let input = Arc::clone(&self.input);
+        let partition_idx = self.partition_idx;
+        let schema = input.schema();
+        let elapsed_compute = self.metrics.elapsed_compute.clone();
+        let first_batch_at = Arc::clone(&self.first_batch_at);
+        let n_cols = self.input.schema().fields.len();
+
+        let mut reporter = LoadInfoDropHandler {
+            load_info: pb::LoadInfo {
```
jayshrivastava: Nit: can we implement `new` on this?

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+            });
+        }
+        Self {
+            input,
+            metric_set,
+            partition_samplers: samplers,
+        }
+    }
+
+    pub(crate) fn kick_off_first_sampler(
```
jayshrivastava, line 141: Why is it called "first"?

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+        self.stream
+            .lock()
+            .expect("poisoned lock")
+            .replace(Box::pin(RecordBatchStreamAdapter::new(schema, stream)));
+
+        Ok(sampling_rx)
+    }
+}
+
+struct LoadInfoDropHandler {
```
jayshrivastava, line 332: nit: commentary

src/common/vec.rs
```diff
@@ -0,0 +1,80 @@
+use datafusion::common::internal_err;
```
jayshrivastava, line 1: Would an extension trait make sense here to make all of these methods?

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -0,0 +1,337 @@
...
+            //   by the SamplerExec injected by this same function.
+            let compute_cost = calculate_cost(&input_stage.plan, nb_ctx.d_cfg)?;
+            metrics.push(BytesCounterMetric::new_metric("compute_cost", compute_cost));
+            let compute_based_task_count = compute_cost
+                .div_ceil(nb_ctx.d_cfg.bytes_per_partition_per_second)
+                .div_ceil(input_stage.plan.output_partitioning().partition_count())
+                .clamp(1, nb_ctx.max_tasks()?);
+            let task_count = nb_ctx
+                .task_count(&input_stage.plan)?
+                .merge(Desired(compute_based_task_count));
```
jayshrivastava, line 69: `merge` basically only ever increases the task count, which is interesting. Do you imagine any cases where we might want to decrease it?

src/execution_plans/sampler.rs
```diff
@@ -0,0 +1,561 @@
...
+        let Some(stream) = self.partition_samplers[partition].start_stream() else {
+            return exec_err!("SamplerExec[{partition}] was not kicked off");
+        };
+        Ok(stream)
+    }
+
+    fn metrics(&self) -> Option<MetricsSet> {
+        Some(self.metric_set.clone_inner())
+    }
+}
```
jayshrivastava, line 594: This is all pretty clean. Some unit tests would be nice.

src/distributed_planner/network_boundary.rs
```diff
@@ -84,3 +85,16 @@ pub(crate) fn insert_producer_head(
     };
     Ok(plan)
 }
+
+/// Injects a [SamplerExec] right below a [RepartitionExec] or [BroadcastExec].
+pub(crate) fn insert_sampler(input: Arc<dyn ExecutionPlan>) -> Result<Arc<dyn ExecutionPlan>> {
```
jayshrivastava: I feel like this one-off function should be tied to the `NetworkBoundary` trait or `ProducerHead`.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -0,0 +1,337 @@
...
+            };
+
+            for (i, routed_url) in routed_urls.into_iter().enumerate() {
+                workers.push(routed_url.clone());
+                // Spawns the task that feeds this subplan to this worker. There will be as
+                // many as this spawned tasks as workers.
+                let (worker_tx, worker_rx) = stage_coordinator.send_plan_task(i, routed_url)?;
+                load_info_rxs.push({
+                    let rx = stage_coordinator.worker_to_coordinator_task(i, worker_rx);
+                    UnboundedReceiverStream::new(rx)
```
jayshrivastava, line 105: How many messages do we get? Maybe we can set the size.

1 load info per partition, 1 load info EoS, and 1 metrics message?

src/execution_plans/network_shuffle.rs
```diff
@@ -238,4 +246,9 @@ impl ExecutionPlan for NetworkShuffleExec {
     fn metrics(&self) -> Option<MetricsSet> {
         Some(self.worker_connections.metrics.clone_inner())
     }
+
+    fn partition_statistics(&self, partition: Option<usize>) -> Result<Arc<Statistics>> {
```
jayshrivastava, line 250: I wonder if this could use a little test coverage. Maybe it's covered elsewhere, I'm still reviewing.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -0,0 +1,337 @@
...
+        head_stage,
+    })
+}
+
+#[derive(Default)]
+struct PlanReconstructor {
+    stage_map: DashMap<usize, (Arc<dyn ExecutionPlan>, MetricsSet)>,
+}
+
+impl PlanReconstructor {
```
jayshrivastava, line 169: this feels super random. it would be nice if it was shared between here and `prepare_network_boundaries` which also does things like `insert_producer_head`

src/coordinator/query_coordinator.rs
```diff
@@ -404,6 +422,28 @@ impl<'a> StageCoordinator<'a> {
...
+
+    pub(super) fn find_input_stage_with_single_url(&self) -> Option<Url> {
+        let mut single_stage_url = None;
+        self.plan
+            .apply(|plan| {
+                let Some(nb) = plan.as_network_boundary() else {
+                    return Ok(TreeNodeRecursion::Continue);
+                };
+
+                if let Stage::Remote(remote) = nb.input_stage()
```
jayshrivastava, line 431: Would it be ok to add the remote stage to the `NetworkBoundaryBuilder`? So you don't have to recurse here. You could store an `Option<RemoteStage>`

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -108,6 +108,14 @@ impl QueryPlanner for DistributedQueryPlanner {
 
         plan = insert_broadcast_execs(plan, cfg)?;
 
+        if d_cfg.dynamic_task_count {
+            // The task count will be decided dynamically at execution time.
+            return Ok(Arc::new(
+                DistributedExec::new(plan).with_metrics_collection(d_cfg.collect_metrics),
```
jayshrivastava, line 114: Just a little bit of a smell.

The dynamic prepare plan does a lot of stuff that `inject_network_boundaries`, `prepare_network_boundaries` is supposed to do. Should we move those function calls to the static prepare plan? So the `DistributedExec` is always responsible for that stuff?

The downside is that the prepare static plan which happens below during planning would happen during execution in `DistributedExec`

src/execution_plans/sampler.rs
jayshrivastava: Oh I get it now. Bc it's the first one seen in the plan.

src/worker/impl_coordinator_channel.rs
```diff
@@ -172,19 +177,35 @@ impl Worker {
...
+                Err(err) => Err(Status::internal(err.to_string())),
+            })
+            .chain(futures::stream::once(async move {
+                Ok(WorkerToCoordinatorMsg {
+                    inner: Some(worker_to_coordinator_msg::Inner::LoadInfoEos(true)),
+                })
+            }));
+
         // Stream back the metrics once the task finishes executing.
         // The oneshot receiver resolves when impl_execute_task sends the collected
```
jayshrivastava, line 195: Also I noticed that we still have the old code in the `impl_execute_task.rs`
```
            if num_partitions_remaining.fetch_sub(1, Ordering::SeqCst) == 1 {
```

jayshrivastava (commented):
Pretty promising results. Left a few comments. I also had codex do a sweep:

# Last Commit Review Findings

Review target: last commit, `Support dynamic task count assignation`.

## 1. Blocking: dynamic planning can deadlock on set-but-never-executed stages

`DistributedExec::execute` keeps the end-query guard alive until after `drain_pending_tasks()`:

- `src/coordinator/distributed.rs:199`
- `src/coordinator/distributed.rs:218`

At the same time, `send_plan_task` keeps the coordinator-to-worker request stream open with `keep_stream_alive`:

- `src/coordinator/query_coordinator.rs:184`

The worker response stream also waits on the metrics stream:

- `src/worker/impl_coordinator_channel.rs:180`

If the dynamic planner sets and samples a stage but later never executes that stage, metrics never completes. Then `drain_pending_tasks()` waits for the worker response stream to end, but the response stream is kept alive by the coordinator channel, and that channel only closes when the guard drops. The guard cannot drop because execution is waiting inside `drain_pending_tasks()`.

This is a circular wait.

## 2. Blocking: task-data invalidation is still coupled to execution finish

The commit rationale says task entry invalidation moves from task execution finish to coordinator-channel lifetime, but the old invalidation path is still present:

- `src/worker/impl_execute_task.rs:97`

The new coordinator-channel invalidation is also present:

- `src/worker/impl_coordinator_channel.rs:167`

That means task data can still be removed when partition execution drains, even while the coordinator channel is intentionally held open. This keeps the old early-cleanup behavior and can break dynamic cases with late consumers, retries, or future execution requests that rely on the coordinator channel still owning task-data lifetime.

## 3. High: per-column throughput stats are almost always zero

`set_per_col_bytes_per_second` computes:

```rust
ready / total_ready * total_bytes_per_second
```

at:

- `src/execution_plans/sampler.rs:357`

Because this is integer division, any column with `ready < total_ready` becomes `0` before multiplication. In ordinary multi-column cases, most or all columns will report zero bytes per second.

The fix is to multiply before dividing, with overflow-safe arithmetic if needed:

```rust
ready.saturating_mul(total_bytes_per_second as u64) / total_ready
```

## 4. High: `bytes_per_partition_per_second = 0` can panic

The config setter accepts any `usize`:

- `src/distributed_ext.rs:761`

But dynamic planning uses the value as a divisor:

- `src/coordinator/prepare_dynamic_plan.rs:49`

If a user sets `bytes_per_partition_per_second` to `0`, planning can panic in `div_ceil`. The setter should reject zero, or planning should clamp or error before dividing.

## 5. High: remote partition statistics ignore the requested partition

`Stage::partition_statistics` returns the whole remote stage statistics regardless of the requested partition:

- `src/stage.rs:149`

If DataFusion calls `partition_statistics(Some(partition))`, each partition can appear to contain the full stage statistics. That can inflate rows and bytes by partition count and distort AQE decisions.

The safer behavior is either:

- return unknown stats for per-partition requests when only global stats are available, or
- track real per-partition stats and return the matching partition.

## 6. Medium: `SamplerExec::execute` can panic on stale partition ids

`SamplerExec::execute` indexes into `partition_samplers` before validating the partition:

- `src/execution_plans/sampler.rs:547`

If a stale or mismatched dynamic partition id reaches this plan, this panics instead of returning a DataFusion execution error.

This should be changed to `get(partition)` and return `exec_err!` when the partition is invalid.

## 7. Medium: dynamic coalesce boundaries discard runtime stats

Dynamic planning explicitly skips runtime-stat collection for `NetworkCoalesceExec`:

- `src/coordinator/prepare_dynamic_plan.rs:98`

This may be fine for the final gather stage, but if a coalesce boundary feeds another planned stage, downstream task sizing gets unknown stats. That can make later dynamic decisions less accurate or effectively static.

If `NetworkCoalesceExec` can appear below additional distributed operators, it should either gather stats too or document why it is always terminal for dynamic planning.

## 8. Medium: plan reconstruction is destructive

`PlanReconstructor` removes entries from `stage_map` while reconstructing:

- `src/coordinator/prepare_dynamic_plan.rs:157`

That makes reconstruction one-shot. If a stage is referenced by more than one network boundary, or reconstruction is retried or reused for diagnostics, the second lookup fails.

Prefer borrowing and cloning from the map instead of removing entries.

## 9. Medium: sampler/load-info errors can surface too late

Worker load-info errors are emitted through the same response stream as metrics:

- `src/worker/impl_coordinator_channel.rs:180`

The dynamic planner waits on the separate load-info receiver, while `send_plan_task` sees response-stream errors later:

- `src/coordinator/query_coordinator.rs:202`

This can let planning continue with partial or zero stats, while the actual sampler error surfaces later during pending-task draining. Runtime-stat collection should receive explicit errors, not just missing load-info messages.

## 10. Low: the new lifecycle behavior needs a targeted regression test

The highest-risk scenario is:

1. Dynamic planner sends `SetPlanRequest` for a stage.
2. Worker starts sampling and sends `LoadInfo`.
3. AQE chooses a different shape and that stage never reaches `execute_task`.
4. The query still must finish and `drain_pending_tasks()` must not hang.

Existing correctness tests likely exercise stages that are eventually consumed, so they may not catch this lifecycle bug. Add a targeted test for a planned-but-abandoned dynamic stage and assert that pending-task draining completes.

src/execution_plans/sampler.rs
gabotechs: It do is a regular Gauge with a max aggregator.

> So this measures the size of all the batches pushed to the buffer.

This is equal to the max memory used. Note that entries in the buffer are never popped. Instead, they are converted into a chained stream of the buffered entries + the ones that are yet to come, but this is an only-push buffer.

This is actually not really a buffer, it's more like a peek, where some entries are peeked and then streamed back as-is without any extra buffering in the middle.

I'll update the name.

src/execution_plans/sampler.rs
gabotechs: I extended a bit the comment to explain it a bit more

src/execution_plans/sampler.rs
gabotechs: This is a protobuf message autogenerated, and typically further methods are not implemented for protobuf messages.

I can add a separate function though, I'll do that.

src/common/vec.rs
gabotechs: I thought about that, but the conclusion was that it was not worth it as that does not save much code, and these are not typically used during method chaining that is where extension traits really help.

It's also nice to make it explicit for readers that these are not `std` methods.

src/coordinator/prepare_dynamic_plan.rs
gabotechs: Nah, I don't think so. If there is one branch of the plan tree that benefits from greater parallelism, rather than constrain it so that it adecuates to lower parallelism branches, it's more performant to adecuate the lower parallelism branches to match the greater parallelism branch.

The only reason I can think of for decreasing parallelism is when there's a hard requirement given by one node, which is what typically `Maximum()` task counts are used for.

src/execution_plans/sampler.rs
gabotechs: Yeah, I have rewritten from scratch so many times this piece of code that I've always deferred having actual unit tests until the end.

This is starting to settle now (unless you have other concerns/ideas), so it might be a good time to start adding unit tests

src/distributed_planner/network_boundary.rs
gabotechs: 🤔 We cannot add it to the `NetworkBoundary` trait because it will make the trait non `dyn` compatible, but to the `ProducerHead`... maybe that's a good idea actually, I'll do that

src/coordinator/prepare_dynamic_plan.rs
gabotechs: It's not very worth it setting the size, it's not like a vector that you can avoid allocations when you know the size before hand. 

Making this a bounded channel will force us to use async methods for producing elements in it, which is not really necessary as we don't care about backpressuring or anything like that for these messages, and it can more easily result in deadlocks.

src/execution_plans/network_shuffle.rs
gabotechs: You might notice that there's very little new unit tests overall. This is mainly because this new feature is covered by all tpch, tpcds and clickbench integration tests, by actually running the full queries with AQE enabled

gabotechs: I have explicitly not added almost any new unit test because of this, and because of the amount of from scratch iterations I've done over this, but as this stabilizes, if you see opportunities for covering things that are not already covered by the integration tests, those are more than welcome

src/coordinator/prepare_dynamic_plan.rs
gabotechs: Both `prepare_network_boundaries` and this code are doing very different things:
- `prepare_network_boundaries` is actually laying out the plan that will eventually get executed and sent to workers
- This is just reconstructing the plan dynamically as different stages get sent to the workers during AQE so that we can then have something to visualize. The `insert_producer_head` for example is really just for the sake of a future visualization

gabotechs: I'm going to add some comments about this

src/coordinator/query_coordinator.rs
gabotechs: It would be a bit messy, because there can be an arbitrary number of `RemoteStage`s present in the subplan here.

I think it's not too bad, recursing is negligible as long as no Arc operations or plan restructures happen during the recursion. `.apply` takes the plan by reference, and at no point it needs an owned reference to an `Arc<dyn ExecutionPlan>`, so the cost of having such recursion is ~0, unlike if you were down a `transform_down()` returning `Transformed::yes` at some point in the query

src/distributed_planner/distributed_query_planner.rs
gabotechs: Yeah, I think that's a big drawback. For static planning, we really like to see the full plan before it's send for execution.

The split of responsibilities seems correct here:
- static planner: we plan statically inside `distributed_query_planner.rs`, so all further logic is here (e.g., `prepare_network_boundaries`)
- dynamic planner: we plan during execution, so the logic is in `DistributedExec`

src/worker/impl_coordinator_channel.rs
gabotechs: Yeah, we still need that old code for a couple of things:
- Marking execution as finished for stage metrics
- Sending the runtime metrics over the wire back to the coordinator

Additionally, the cache entry is still promptly invalidated here rather than waiting for the coordinator->worker channel to drop.

gabotechs:
> 1. Blocking: dynamic planning can deadlock on set-but-never-executed stages

This should not be a problem, if the query finishes streaming arrow data, it does not matter what is waiting for what, everything is abruptly cancelled (but the metrics collection).

This is actually not even related to this PR, although I think there might be better ways of doing this. As it's not related to this PR, I think it might be worth evaluating in a different one

gabotechs:
> 2. Blocking: task-data invalidation is still coupled to execution finish

This is fine, the only consequence of this is that task data will get invalidated a bit earlier, which is acceptable and even desirable.

Again, this did not change in this PR

gabotechs:
> 3. High: per-column throughput stats are almost always zero

👍 This is a good call, doing this

gabotechs:
> 4. High: bytes_per_partition_per_second = 0 can panic

👍 Doing that

gabotechs:
> 5. High: remote partition statistics ignore the requested partition

Because of there being repartitions in between samplers and network boundaries, it's not possible to attribute some statistics to specific partitions.

One thing that comes to mind is, if a specific partition is requested, maybe it's better to just divide the stats by the amount of partitions available, assuming that the partitions are going to be perfectly evenly distributed

gabotechs:
> 6. Medium: SamplerExec::execute can panic on stale partition ids

👍 Sounds reasonable

gabotechs:
> 7. Medium: dynamic coalesce boundaries discard runtime stats

This is fine an expected. We don't care about sampling this boundaries because we always know that the stage above should have `Maximum(1)` tasks

gabotechs:
> 8. Medium: plan reconstruction is destructive

This is fine, we are good with taking owned references for this, we don't need to hold them in memory.

gabotechs:
> 9. Medium: sampler/load-info errors can surface too late

I think this is fine, the timing for all of this should be negligible, and I'm afraid that doing it differently can bring complexity to the code

gabotechs:
> 10. Low: the new lifecycle behavior needs a targeted regression test

I actually have fought this as this was surfaced by the existing tests. This situation happens in many TPCH and TPC-DS queries, so I think we are covered on this front

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -0,0 +1,361 @@
...
+        Ok(reconstructed.data)
+    }
+}
+
+/// Estimates the bytes per second flowing through a stage by reading sample information.
+async fn gather_runtime_statistics(
+    per_task_load_info_stream: Vec<impl Stream<Item = pb::LoadInfo> + Unpin>,
+    plan: &Arc<dyn ExecutionPlan>,
+) -> Result<Statistics> {
+    const ESTIMATED_QUERY_TIME_S: usize = 10;
```
gabotechs, line 210: I don't love having this constant here. It seems like something users should be able to provide if they wanted.

src/execution_plans/sampler.rs
jayshrivastava: Feel free to file an issue. I can take it

jayshrivastava (approved):

gabotechs:
Opened the following issues to track follow-ups from this PR:

- #518 — Add unit tests for `SamplerExec`
- #519 — Make `ESTIMATED_QUERY_TIME_S` user-configurable
- #520 — Leverage memory and network cost for dynamic task count decisions
- #521 — Dynamic planning can deadlock on set-but-never-executed stages

gabotechs:
Thanks @jayshrivastava for the quality review! Also, @gene-bordegaray, please feel free to leave any comments even if this is merged, I'll very gladly address them as if it wasn't.

PR #421, opened 2026-05-01
Gene.bordegaray/2026/04/add network exchange assignment

gene-bordegaray:
# Network Exchange Assignment Approach

**Quick Summary:** removes `N * M` shuffle partition inflation, assigns network exchange ownership
explicitly, and recovers selected post-shuffle local parallelism with `LocalExchangeSplitExec`.

## Findings Addressed

In the shuffle report we identified two major problem families.

1. Partition inflation
2. Per-RPC pipeline execution

This PR directly addresses the first issue and reduces the second issue as a side-effect.

## Old Shape

Before this PR, the planner created consumer parallelism by multiplying the shuffle partition count
by the consumer task count.

```text
N = single-node hash partitions
M = consumer tasks

          ┌─┐ ... ┌─┐                         ┌─┐ ... ┌─┐                         ┌─┐ ... ┌─┐
          │1│     │N│                         │1│     │N│                         │1│     │N│
┌─────────┴─┴─────┴─┴─────────┐     ┌─────────┴─┴─────┴─┴─────────┐     ┌─────────┴─┴─────┴─┴─────────┐
│                             │     │                             │     │                             │
│     NetworkShuffleExec      │     │     NetworkShuffleExec      │ ... │     NetworkShuffleExec      │
│          (Task 1)           │     │          (Task 2)           │     │          (Task M)           │
│                             │     │                             │     │                             │
└─────────┬─┬─────┬─┬─────────┘     └─────────┬─┬─────┬─┬─────────┘     └─────────┬─┬─────┬─┬─────────┘
          │1│     │N│                         │1│     │N│                         │1│     │N│
          └▲┘ ... └▲┘                         └▲┘ ... └▲┘                         └▲┘ ... └▲┘
           │       │                           │       │                           │       │
           │ slots                               slots                               slots
           │ [1, N]                             (N, 2N]                          (M-1(N), MN]
           │
           │                                                                               │
           │                                                                               │
           └────────────────────────────┐                      ┌───────────────────────────┘
                                        │                      │
                                       ┌─┐        ...        ┌──┐
                                       │1│                   │MN│
                                    ┌──┴─┴───────────────────┴──┴─┐
                                    │                             │
                                    │       RepartitionExec       │
                                    │          (Task 1)           │
                                    │                             │
                                    └─────────┬─┬─────┬─┬─────────┘
                                              │1│     │N│
                                              └─┘ ... └─┘
```

This create two costs that we are trying to eliminate (`N` = `RepartitionExec` output partitions,
[...]

PR #440, opened 2026-05-10
Refactor network exchange layouts

gene-bordegaray:
## Which issue does this PR close?

- This is a first digestible slice out of #421.

## Rationale for this change

#421 contains the larger local exchange / network exchange assignment implementation. This PR extracts the network exchange layout refactor as a smaller, behavior-preserving first step so the assignment model can be reviewed independently before introducing fanout reduction or local exchange insertion changes.

## What changes are included in this PR?

- Adds `ExchangeLayout` and `SlotReadPlan` to centralize network exchange slot assignment.
- Moves shuffle, coalesce, and broadcast producer/consumer mapping into explicit layouts.
- Derives exchange layouts in the distributed planner before constructing network boundaries.
- Preserves existing fanout behavior and partitioning assignment for shuffle, coalesce and broadcast.
- Serializes layout dimensions through `DistributedCodec` and adds direct codec roundtrip coverage.

## Are these changes tested?

Yes.

Also, no snapshot change 💯

### DIFF
--- a/src/execution_plans/network_shuffle.rs
+++ b/src/execution_plans/network_shuffle.rs
@@ -2,9 +2,9 @@ use crate::common::require_one_child;
 use crate::execution_plans::common::scale_partitioning;
 use crate::stage::{LocalStage, Stage};
 use crate::worker::WorkerConnectionPool;
-use crate::{DistributedTaskContext, NetworkBoundary};
+use crate::{DistributedTaskContext, ExchangeAssignment, NetworkBoundary, SlotReadPlan};
 use datafusion::common::tree_node::{Transformed, TreeNodeRecursion};
-use datafusion::common::{Result, not_impl_err, plan_err};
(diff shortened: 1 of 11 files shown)


src/execution_plans/network_shuffle.rs
```diff
@@ -105,24 +107,47 @@ pub struct NetworkShuffleExec {
...
     pub(crate) worker_connections: WorkerConnectionPool,
+    pub(crate) layout: Arc<ExchangeLayout>,
 }
 
 impl NetworkShuffleExec {
     /// Builds a new [NetworkShuffleExec] in "Pending" state.
     ///
     /// Typically, the `input` to this
     /// node is a [RepartitionExec] with a [Partitioning::Hash] partition scheme.
+    /// This count-based constructor keeps the public API stable; the distributed planner uses
```
gene-bordegaray: dont know how much we care about keeping this stable but I thin kwe should only derive from layouts going forward for clarity

I would not be opposed to deprecating 👍

PR #442, opened 2026-05-10
Fix TPCDS runner explain analyze await

gene-bordegaray:
## Summary

Fixes the `console/examples/tpcds_runner.rs` example after `explain_analyze` became async.
## Validation

- `cargo test --workspace`


gabotechs (approved):
👍 Thanks!

PR #443, opened 2026-05-10
feat: use tpchgen-rs for benchmarks

### LINKED ISSUE #128: Move out of Docker in favor of tpchgen-rs for benchmarks
We have two different ways of generating TPCH data:
- A docker image in the benchmarks, similar to https://github.com/apache/datafusion/tree/main/benchmarks
- https://github.com/clflushopt/tpchgen-rs for the integration tests

The argument of keeping the docker-based generation in the benchmarks was to keep the approach close to upstream DataFusion, but as the benchmarks crate evolves, this is becoming less and less relevant, and the approaches are naturally diverging.

It would be nice to clean it up and use the same TPCH generation approach for both benchmarks and integration tests

clflushopt:
I'am trying to assist with using `tpchgen-rs` in data projects, this PR closes #128 and introduces the following changes :

* `dfbench generate-tpch` subcommand (`--output`, `--scale-factor` and `--partitions`).
* Remove `dfbench prepare-tpch` (no longer needed).
* Update `gen-tpch.sh` to call `generate-tpch` instead of Docker.

I didn't run any extensive benchmarking but when testing this is on average 2.8x faster than going through Docker at scale factor 1. The improvements are more dramatic at higher scale factors.


benchmarks/src/main.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) const RESULTS_DIR: &str = ".results";
 enum Options {
     Run(run::RunOpt),
     Compare(compare::CompareOpt),
-    PrepareTpch(prepare_tpch::PrepareTpchOpt),
+    GenerateTpch(generate_tpch::GenerateTpchOpt),
     PrepareTpcds(prepare_tpcds::PrepareTpcdsOpt),
     PrepareClickbench(prepare_clickbench::PrepareClickBenchOpt),
```
gabotechs: As the other methods for the other datasets are called `Prepare*`, how about either also calling this one `PrepareTpch`, or rename the other two to `GenerateTpcds` and `GenerateClickbench`?

Not a strong opinion on the naming, but worth being consistent with it.

gabotechs (approved):
:100: nice! this was pretty easy in fact. Thanks @clflushopt!

benchmarks/src/main.rs
clflushopt: Agree on that I aligned with the other functions by renaming to `PrepareTpch`

gabotechs:
Thanks @clflushopt 🙏, much appreciated.

Issue #167, opened 2025-09-29
Use `datafusion::common::{HashMap, HashSet}` always


jayshrivastava:
`datafusion::common::{HashMap, HashSet}` use `hashbrown::{HashMap, HashSet}` which are favorable compared to the std implementations

gabotechs:
Done in https://github.com/datafusion-contrib/datafusion-distributed/pull/444

PR #444, opened 2026-05-10
swap std for hashbrown

Rich-T-kid:
see #167
I didn't change the calls to `std::collections::Hashmap/set` for benchmarks.


gabotechs (approved):
👍 Looks good. Thanks @Rich-T-kid!

If you want to go one step beyond, we can configure Clippy to not accept usages of the standard library's `HashMap` and `HashSet`.

I'll merge this now, feel free to open a new PR if you also want to add the Clippy config.

PR #446, opened 2026-05-11
set up git hooks

### LINKED ISSUE #434: Add pre-commit hooks
Right now, before any commit I manually run

```
cargo fmt --all
```

and then 

```
cargo clippy --all-targets --all-features -- -D warnings
```

These are run in our CI checks. We could just add them as pre-commit hooks so that they are automatically run whenever a contributor tries to make a commit. This way, it is impossible to forget and then have the issue be caught later in CI.

Rich-T-kid:
### this resolves #434 
I avoided added a step that also runs the test because it would be time consuming. I think it would be worth it to add a git hook for `git push` that runs the tpch/tpchs/clickhouse test.


docs/source/contributor-guide/setup.md
```diff
@@ -16,6 +16,17 @@ git lfs install
...
 ```
 
+## Pre-commit Hook Setup
+
+Install the pre-commit hook to catch issues before committing:
+
+```bash
+cp hook-scripts/pre-commit .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
```
gabotechs: I'd rather have the full copy-pastable command here for setting up the hooks rather than it be in an another top-level folder in the workspace. Folder real state at the root of the project is a bit expensive.

You can just put something like this here:
```md

```bash
cat > .git/hooks/pre-commit << 'EOF'
  #!/bin/sh

  # this pre-commit hook avoids leaving the repo in a broken state.

  echo "Running cargo fmt..."
  cargo fmt --all -- --check || exit 1

  echo "Running cargo clippy..."
  cargo clippy --all-targets --all-features -- -D warnings || exit 1

  echo "Running cargo check..."
  cargo check || exit 1
  
  echo "All pre-commit checks passed!"
EOF
chmod +x .git/hooks/pre-commit
```

And people can just copy-paste this directly.

Also, running both `cargo clippy` and `cargo check` should be redundant. I think just running `cargo clippy` should be enough.

Rich-T-kid: makes sense, I updated the PR

gabotechs (approved):
👍 Thanks @Rich-T-kid! this looks good

gabotechs:
💯 Just 7 mins of CI, I wonder who made this!

Rich-T-kid:
😆

Issue #445, opened 2026-05-11
configure Clippy to not accept usages of the standard library's HashMap and HashSet


Rich-T-kid:
Follow up of #444, [see comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/444#pullrequestreview-4261454101) that started this issue.

Rich-T-kid:
take

PR #447, opened 2026-05-11
Create Clippy rule dis-allowing std::collection::hashmap

Rich-T-kid:
see #445


clippy.toml
```diff
@@ -0,0 +1,4 @@
+disallowed-types = [
+    { path = "std::collections::HashMap", reason = "Use hashbrown::HashMap (or datafusion_common::HashMap) for better performance" },
+    { path = "std::collections::HashSet", reason = "Use hashbrown::HashSet (or datafusion_common::HashSet) for better performance" },
```
gabotechs: We should also prohibit `tokio::spawn`. Now that we are here, could you also add that? the preferred alternative is `datafusion::common::runtime::SpawnedTask`, as that makes sure that several things are correctly propagated across the task context (like tracing info).

We have some instances where we use `tokio::spawn` currently, but those are fine, we can:
- Add `tokio::spawn` to disallowed-methods here
- Add #[allow(clippy::disallowed_methods)] to the existing instances in the project.

Rich-T-kid: Updated the PR 😄

src/execution_plans/distributed.rs
```diff
@@ -527,7 +527,7 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
             stage_id: self.stage_id as u64,
             task_number: task_i as u64,
         };
-        let task_metrics_collection = Arc::clone(self.task_metrics);
+        #[allow(clippy::disallowed_methods)]
```
gabotechs, line 531: Seems like this will result in a compilation error

Rich-T-kid: im running into a weird merge issue.should be resolved now

gabotechs: 🤔 I think `#[allow(clippy::disallowed_methods)]` should be above `tokio::spawn` and not here, right?

Rich-T-kid: yes exactly, when I push it up its showing a merge conflict. I clicked accept both changes. Currently trying to fix this locally before pushing it up

Rich-T-kid: @gabotechs  finally fixed it ✅

gabotechs (approved):
👍 Thanks @Rich-T-kid!

PR #418, opened 2026-04-28
Harden work unit tests

gabotechs:
This is one PR from a stack of PRs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/418 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/448

---

Just some extra test to the WorkUnitFeed machinery.

Extends the work unit testing tooling to support statements like this:

```sql
SELECT * FROM test_work_unit('a', 1, 'rows(1), err(boom_single)')
```

or

```sql
SELECT * FROM test_work_unit('a', 1, 'rows(1), wait(800), rows(1)')
```

Replicating weird behaviors in `WorkUnitFeedProvider` implementations.

---

Additionally, ships a testing `WorkUnitFileScanConfig` implementation that replicates a `FileScanConfig` but streaming the `FileGroup`s as work units instead of baking them in the plan. Suitable for benchmarking and ensuring that the `WorkUnitFeed` machinery meets performance expectations

PR #448, opened 2026-05-11
Add WorkUnitFileScanConfig for testing WorkUnits with real data

gabotechs:
This is one PR from a stack of PRs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/418
- https://github.com/datafusion-contrib/datafusion-distributed/pull/448 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/453

---

Ships a testing WorkUnitFileScanConfig implementation that replicates a FileScanConfig but streaming the FileGroups as work units instead of baking them in the plan. Suitable for benchmarking and ensuring that the WorkUnitFeed machinery meets performance expectations.

Benchmarks comparing the normal `FileScanConfig` vs the new `WorkUnitFileScan` show no performance difference, which is expected as transmitting work unit feeds over the wire is efficient.

### DIFF
--- a/src/test_utils/work_unit_file_scan.rs
+++ b/src/test_utils/work_unit_file_scan.rs
@@ -1,0 +1,433 @@
+//! WorkUnit-based [`FileScanConfig`] alternative for benchmarking purposes.
+//!
+//! Streams the per-partition `PartitionedFile`s through a [`WorkUnitFeed`]
+//! instead of embedding them directly in the serialized plan. Used to measure
+//! the latency impact of routing file scan inputs through the work unit
+//! pipeline as compared to the regular [`FileScanConfig`] path.
+
+use crate::{DistributedConfig, TaskCountAnnotation, TaskEstimation, TaskEstimator};
+use crate::{WorkUnitFeed, WorkUnitFeedProto, WorkUnitFeedProvider};
+use datafusion::catalog::memory::DataSourceExec;
+use datafusion::common::tree_node::{Transformed, TreeNode};
+use datafusion::common::{Result, internal_datafusion_err};
+use datafusion::common::{Statistics, internal_err};
+use datafusion::config::ConfigOptions;
+use datafusion::datasource::physical_plan::{FileGroup, FileScanConfig, FileScanConfigBuilder};
+use datafusion::datasource::source::DataSource;
+use datafusion::execution::{SendableRecordBatchStream, TaskContext};
+use datafusion::physical_expr::{EquivalenceProperties, LexOrdering};
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+use datafusion::physical_plan::execution_plan::SchedulingType;
+use datafusion::physical_plan::metrics::ExecutionPlanMetricsSet;
+use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion::physical_plan::{DisplayFormatType, ExecutionPlan, Partitioning};
+use datafusion_proto::physical_plan::{
+    AsExecutionPlan, DefaultPhysicalExtensionCodec, PhysicalExtensionCodec,
+};
+use datafusion_proto::protobuf as df_proto;
+use datafusion_proto::protobuf::proto_error;
+use futures::stream::BoxStream;
+use futures::{StreamExt, TryStreamExt};
+use prost::Message;
+use std::any::Any;
+use std::fmt::{Debug, Formatter};
+use std::sync::Arc;
+
+/// Per-partition work unit emitted by [`FileScanWorkUnitProvider`]: the (encoded)
+/// `PartitionedFile` that the receiving worker partition should scan. `None`
+/// means "no file for this slot" — used as padding when the total file count
+/// isn't divisible by `task_count` so the global feed indexing layout
+/// (`task_index * partitions_per_task + p`) stays consistent across tasks.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct FileScanWorkUnit {
+    #[prost(message, tag = "1")]
+    pub file: Option<df_proto::PartitionedFile>,
+}
+
+/// Local provider that holds the per-partition `PartitionedFile` assignment
+/// for a [`WorkUnitFileScanConfig`]. It is only invoked on the coordinator
+/// (the worker side gets a `RemoteFeedProvider` automatically when the plan is
+/// decoded).
+#[derive(Debug, Clone)]
+pub struct FileScanWorkUnitProvider {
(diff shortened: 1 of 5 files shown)


src/test_utils/work_unit_file_scan.rs
```diff
@@ -0,0 +1,433 @@
...
+    type WorkUnit = FileScanWorkUnit;
+
+    fn feed(
+        &self,
+        partition: usize,
+        _ctx: Arc<TaskContext>,
+    ) -> Result<BoxStream<'static, Result<Self::WorkUnit>>> {
+        let Some(file_group) = self.file_groups.get(partition) else {
+            return Ok(futures::stream::empty().boxed());
+        };
```
LiaCastaneda, line 79: Is this possible? the number of partitions == the number of `FileGroups` no? (in this case)

src/test_utils/work_unit_file_scan.rs
```diff
@@ -0,0 +1,433 @@
...
+        });
+        Ok(stream.boxed())
+    }
+}
+
+/// [`DataSource`] that defers obtaining its [`PartitionedFile`](datafusion_datasource::PartitionedFile)
+/// until execution time, pulling it off a [`WorkUnitFeed`] before delegating
+/// the actual file scan to the wrapped [`FileScanConfig`]. The wrapped config
+/// carries no file groups while it is being serialized — the per-partition
+/// file travels through the feed instead.
```
LiaCastaneda, line 96: for my understanding, the main advantage of not passing the file locations in the plan is in making the plan smaller (and I guess less expensive to serialize)?

src/test_utils/work_unit_file_scan.rs
```diff
@@ -0,0 +1,433 @@
...
+        &self,
+        partition: usize,
+        context: Arc<TaskContext>,
+    ) -> Result<SendableRecordBatchStream> {
+        let inner = self.fsc.clone();
+        let schema = inner.projected_schema()?;
+
+        let stream = self
+            .feed
+            .feed(partition, Arc::clone(&context))?
```
LiaCastaneda, line 132: I'm not sure I understand where the workers wait for the coordinator to send the Work units 🤔 is it on `feed`?

src/test_utils/work_unit_file_scan.rs
```diff
@@ -0,0 +1,433 @@
...
+            file_groups,
+            metrics: ExecutionPlanMetricsSet::new(),
+        }
+    }
+}
+
+impl WorkUnitFeedProvider for FileScanWorkUnitProvider {
+    type WorkUnit = FileScanWorkUnit;
+
+    fn feed(
```
LiaCastaneda, line 72: This `feed` is run on the coordinator always no?

src/test_utils/work_unit_file_scan.rs
```diff
@@ -0,0 +1,433 @@
...
+        let Some(inner) = dse.data_source().as_any().downcast_ref::<FileScanConfig>() else {
+            return Err(proto_error(
+                "Expected the WorkUnitFileScan template DataSource to be a FileScanConfig",
+            ));
+        };
+        let Some(feed_proto) = proto.feed else {
+            return Err(proto_error("WorkUnitFileScanProto missing feed"));
+        };
+        Ok(DataSourceExec::from_data_source(WorkUnitFileScanConfig {
+            feed: WorkUnitFeed::from_proto(feed_proto)?,
```
LiaCastaneda, line 280: Ok this answers my second question I think, this creates a `RemoteFeedProvider` which is where we wait for the `FileGroups`

LiaCastaneda (approved):
I had to ramp up with the new Work Unit APIs since I was unfamiliar with them. This makes sense to me, just left a few questions for my understanding

src/test_utils/work_unit_file_scan.rs
gabotechs: Yes, that's right. There's an option for these feeds to be run inside workers instead, for that, two things need to happen: 
1) The `WorkUnitFeed` must **not** be registered with [DistributedExt::with_distributed_work_unit_Feed](https://github.com/datafusion-contrib/datafusion-distributed/blob/c5e62fad814238f03fd443bc419af8330ce585d0/src/distributed_ext.rs#L571-L571)
2) The protobuf serialization/deserialization must be done manually instead of using the integrated `.to_proto()` and `.from_proto()` methods.

However, this `FileScanWorkUnitProvider` concretely does not do that, and therefore, it runs on the coordinator exclusively.

src/test_utils/work_unit_file_scan.rs
gabotechs: I did this a long time ago, and I'm not 100% sure, but I think this should be possible, for example, if `.file_groups` could not be split in P (partitions) * T (tasks) entries because there's not enough underlaying `PartitionedFile` entries.

This is expected though.

src/test_utils/work_unit_file_scan.rs
gabotechs: In practice, this is negligible, and there's no real advantage in doing this for the specific `WorkUnitFileScanConfig`. The sole purpose of its existence is for benchmarking and ensuring that there's nothing funny happening that could make passing `WorkUnit`s introduce unexpected latency.

This could be actually beneficial if, rather than resolving all Parquet files at planning time, these files could be resolved dynamically at execution time, and streamed to workers progressively as soon as they become available.

As the DataFusion default planner does not do this, and it waits for all Parquet files to be resolved at planning time, then the real benefit here is just pure benchmarking that we are not doing anything funny with introducing `WorkUnit` feeds into the equation.

src/test_utils/work_unit_file_scan.rs
gabotechs: That's right.

If this `WorkUnitFileScanConfig` is executed inside a worker, `self.feed` will **not** be a `WorkUnitFeed<FileScanWorkUnitProvider>`, it will instead be a `WorkUnitFeed<RemoteWorkUnitProvider>`, pulling work units over the network instead of executing `FileScanWorkUnitProvider::feed` locally.

This is the charm of `WorkUnitFeed`s, that you as a developer just call `.feed()` and do not need to care whether this is executed on the coordinator or inside a worker, that's handled "magically" inside Distributed DataFusion.

src/test_utils/work_unit_file_scan.rs
gabotechs: Yeap, that's it.

gabotechs:
Thanks for the review @LiaCastaneda!

Issue #451, opened 2026-05-12
Remove ignore flag from test


Rich-T-kid:
Ignored tests: categorized triage

  **Query took too long to execute**

  - tests/tpcds_correctness_test.rs — q72, q99

  **Query did not get distributed**

  - tests/tpcds_correctness_test.rs — q13, q48, q62, q91
  - tests/clickbench_correctness_test.rs — q0, q6

  **Result sets were not equal**

  - tests/tpcds_correctness_test.rs — q14, q59
  - tests/clickbench_correctness_test.rs — q3, q17, q21, q22, q24, q31, q32

  **Ordering mismatch**

  - tests/clickbench_correctness_test.rs — q42

  **Column 'c_last_review_date_sk' not found**

  - tests/tpcds_correctness_test.rs — q30
  - tests/tpcds_plans_test.rs — q30

  **Expected no error but got (Arrow empty-batch error)**

  - tests/tpcds_correctness_test.rs — q9

  **Non-deterministic column ordering in plan snapshots**

  - tests/tpcds_plans_test.rs — q70, q86

  ---
###   Possible solutions
TODO: I plan on updating this before implementing the PR.
_Query took too long_ : these queries are broken into 15 stages and take 746's. 
_did not get distributed_ : 
_ordering mismatch_  : 
_results set not equal_ : 
_non-deterministic plan column order_ : 
_c_last_review_date_sk_ : I think this has to do with schema mismatch in the dataset

Rich-T-kid:
easiest solution for the slow queries is to run them on there own partition for the CI pipeline but this would still make them slow for local development

gabotechs:
Yeah... I think we can address the other ones first, the slow ones might be tricky to address. Maybe hiding them behind a `--feature slow-tests`?

PR #452, opened 2026-05-13
Un-ignore test

Rich-T-kid:
see #451
color codes original ignored test
Image 5-13-26 at 2 28 PM
Image 5-13-26 at 2 28 PM (1)

### change log
- Added PerTestConfig in src/test_utils/property_based.rs with two booleans:
    - uses_undeterministic_limit_operator — LIMIT without ORDER BY (q17).
    - non_deterministic_sort — ORDER BY with unbroken ties at the LIMIT boundary (q21, q22, q24, q31, q32).
  - compare_result_set takes &PerTestConfig and short-circuits to Ok(()) when either flag is set, with a comment explaining that a strict set comparison would false-positive on SQL-legal answer variation.
  - Each call site is now explicit; only flagged queries opt out.
  - Threaded the same param through test_tpcds_query for symmetry; all tpcds correctness call sites pass PerTestConfig::default().

 Clickbench correctness q3 : shifted precision back by one digit. [see comment for details](https://github.com/datafusion-contrib/datafusion-distributed/pull/452/changes#diff-ff8aabeb67bceb6ab0a320cf9c644092c802c3c24b0246d5a6d549a17115f7d6R221)
 q0 & q6 are still ignored because datafusions planner is smart enough to resolve the answer through statistics.
 
### Still ignored (3)

- Clickbench q42 : compare_ordering reports LexOrdering PartialEq inequality even though the Debug output is byte-identical on both sides. fields are missing that are needed to debug. the current diff shows identical structs even though the comparison fails
- TPCDs q70 & q86 : GROUP BY rollup(s_state, s_county) with two grouping() calls in the projection. The trailing two projection columns swap positions between runs because the planner iterates a HashMap to figure out which columns the rank window function needs. That swap propagates upward through SortExec / SortPreservingMergeExec as @5/@6 index flips. these are very flaky and it flips between passing and failing
- TPCDs q30 - this works locally & fails on the CI, it might make sense to change the query to match [what was done in upstream datafusion](https://github.com/apache/datafusion/pull/22086) I.E changing the query. Will wait for @gabotechs thoughts on this.

### DIFF
--- a/src/test_utils/property_based.rs
+++ b/src/test_utils/property_based.rs
@@ -11,10 +11,24 @@ use datafusion::{
 };
 use std::sync::Arc;
 
+/// Per-test configuration knobs for the property-based comparison helpers.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct PerTestConfig {
+    /// If true, skip the set-equality comparison in [`compare_result_set`]. Use for queries
+    /// whose final operator is a non-deterministic LIMIT (no ORDER BY at all), where two
+    /// engines may correctly return different valid 10-row slices.
+    pub uses_undeterministic_limit_operator: bool,
+    /// If true, skip the set-equality comparison in [`compare_result_set`]. Use for queries
+    /// whose ORDER BY does not produce a total ordering (e.g. `ORDER BY c DESC` with ties
+    /// at the LIMIT boundary), where two engines may correctly pick different tied rows.
+    pub non_deterministic_sort: bool,
+}
+
 /// compares the set of record batches for equality
 pub fn compare_result_set(
     actual_result: &Result<Vec<RecordBatch>>,
     expected_result: &Result<Vec<RecordBatch>>,
+    config: &PerTestConfig,
 ) -> Result<()> {
     let test_batches = match actual_result.as_ref() {
         Ok(batches) => batches,
@@ -36,7 +50,7 @@ pub fn compare_result_set(
         }
     };
 
-    records_equal_as_sets(test_batches, compare_batches)
+    records_equal_as_sets(test_batches, compare_batches, config)
         .map_err(|e| internal_datafusion_err!("result sets were not equal: {}", e))
 }
 
@@ -55,7 +69,6 @@ pub fn compare_ordering(
 
     let actual_ordering = actual_physical_plan.properties().output_ordering();
     let expected_ordering = expected_physical_plan.properties().output_ordering();
-
     if actual_ordering != expected_ordering {
         return internal_err!(
             "ordering mismatch: expected ordering: {:?}, actual ordering: {:?}",
@@ -93,6 +106,7 @@ pub fn compare_ordering(
 fn records_equal_as_sets(
     left: &[RecordBatch],
     right: &[RecordBatch],
+    config: &PerTestConfig,
 ) -> Result<(), DataFusionError> {
     // First check if total row counts match
     let left_rows: usize = left.iter().map(|b| b.num_rows()).sum();
@@ -113,6 +127,13 @@ fn records_equal_as_sets(
             left[0].schema(),
             right[0].schema()
         );
+    }
+    // Skip the row-by-row set comparison for queries where SQL semantics allow two engines
+    // to correctly return different result sets (e.g. LIMIT without ORDER BY, or ORDER BY
+    // with unbroken ties at the LIMIT boundary). Running the full comparison on these
+    // queries would surface false positives even though both engines are right.
+    if config.uses_undeterministic_limit_operator || config.non_deterministic_sort {
+        return Ok(());
     }
 
     detailed_batch_comparison(left, right)
@@ -197,11 +218,17 @@ fn batch_rows_to_strings(batches: &[RecordBatch]) -> Vec<String> {
                 if array.is_null(row_idx) {
                     row_values.push("NULL".to_string());
                 } else if let Some(arr) = array.as_any().downcast_ref::<Float16Array>() {
-                    row_values.push(format!("{:.1$}", arr.value(row_idx), 2));
+                    // 14 significant digits — leaves ~1-2 digits of headroom below f64 precision
+                    // to tolerate summation-order drift between single-node and distributed
+                    // aggregation. [15 digits still diverged at the last digit]
+                    // Example values that 15 digits could not reconcile:
+                    //  - Single-node: 2.533767602294735e18  → 2,533,767,602,294,735,000
+                    //  - Distributed: 2.5337676022947354e18 → 2,533,767,602,294,735,400
+                    row_values.push(format!("{:.13e}", arr.value(row_idx).to_f64()));
                 } else if let Some(arr) = array.as_any().downcast_ref::<Float32Array>() {
-                    row_values.push(format!("{:.1$}", arr.value(row_idx), 2));
+                    row_values.push(format!("{:.13e}", arr.value(row_idx) as f64));
                 } else if let Some(arr) = array.as_any().downcast_ref::<Float64Array>() {
-                    row_values.push(format!("{:.1$}", arr.value(row_idx), 2));
+                    row_values.push(format!("{:.13e}", arr.value(row_idx)));
                 } else {
                     // Use Arrow's deterministic string representation
                     let value_str = array_value_to_string(array, row_idx)
@@ -290,7 +317,7 @@ mod tests {
     async fn test_records_equal_as_sets_empty() {
         let left: Vec<RecordBatch> = vec![];
         let right: Vec<RecordBatch> = vec![];
-        assert!(records_equal_as_sets(&left, &right).is_ok());
+        assert!(records_equal_as_sets(&left, &right, &PerTestConfig::default()).is_ok());
     }
 
     #[tokio::test]
@@ -323,7 +350,7 @@ mod tests {
         let right = vec![batch2];
 
         // Should be equal as sets (order independent)
-        assert!(records_equal_as_sets(&left, &right).is_ok());
+        assert!(records_equal_as_sets(&left, &right, &PerTestConfig::default()).is_ok());
     }
 
     #[tokio::test]
@@ -345,7 +372,7 @@ mod tests {
         let left = vec![batch1];
         let right = vec![batch2];
 
-        assert!(records_equal_as_sets(&left, &right).is_err());
+        assert!(records_equal_as_sets(&left, &right, &PerTestConfig::default()).is_err());
     }
 
     #[tokio::test]
--- a/src/test_utils/property_based.rs
+++ b/src/test_utils/property_based.rs
@@ -11,24 +11,10 @@ use datafusion::{
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


Rich-T-kid:
seems like `test_tpcds_shard03_q30` has a schema mismatch since the datasource that we are using uses **c_last_review_date** but the query expects **c_last_review_date_sk**. It seems datafusion renamed the column

Rich-T-kid:
seems like the #[ignore = "reason"] is a bit outdated. Updating this. seems to be the exact same issue across multiple test & it has something to do with how snapshots are handled.

Rich-T-kid:
>  datasource that we are using uses c_last_review_date but the query expects c_last_review_date_sk. 
taking a look at the q30.sql file this seems to be true 
```
-- SQLBench-DS query 30 derived from TPC-DS query 30 under the terms of the TPC Fair Use Policy.
-- TPC-DS queries are Copyright 2021 Transaction Processing Performance Council.
-- This query was generated at scale factor 1.
with customer_total_return as
 (select wr_returning_customer_sk as ctr_customer_sk
        ,ca_state as ctr_state, 
 	sum(wr_return_amt) as ctr_total_return
 from web_returns
     ,date_dim
     ,customer_address
 where wr_returned_date_sk = d_date_sk 
   and d_year =2000
   and wr_returning_addr_sk = ca_address_sk 
 group by wr_returning_customer_sk
         ,ca_state)
  select  c_customer_id,c_salutation,c_first_name,c_last_name,c_preferred_cust_flag
       ,c_birth_day,c_birth_month,c_birth_year,c_birth_country,c_login,c_email_address
       ,c_last_review_date_sk,ctr_total_return
 from customer_total_return ctr1
     ,customer_address
     ,customer
 where ctr1.ctr_total_return > (select avg(ctr_total_return)*1.2
 			  from customer_total_return ctr2 
                  	  where ctr1.ctr_state = ctr2.ctr_state)
       and ca_address_sk = c_current_addr_sk
       and ca_state = 'KS'
       and ctr1.ctr_customer_sk = c_customer_sk
 order by c_customer_id,c_salutation,c_first_name,c_last_name,c_preferred_cust_flag
                  ,c_birth_day,c_birth_month,c_birth_year,c_birth_country,c_login,c_email_address
                  ,c_last_review_date_sk,ctr_total_return
 LIMIT 100;
```
spoke to claude about this and it might be a github caching issue. ill try invalidating the cache to see if it re-downloads the data. But besides q30 the rest of the issues are snapshot issues which I need to do more research on.

gabotechs:
I think q30 is failing for something else, see this PR that just got merged upstream:

- https://github.com/apache/datafusion/pull/22086

testdata/clickbench/queries/q17.sql
```diff
@@ -1,4 +1,11 @@
 -- Must set for ClickBench hits_partitioned dataset. See https://github.com/apache/datafusion/issues/16591
 -- set datafusion.execution.parquet.binary_as_string = true
 
-SELECT "UserID", "SearchPhrase", COUNT(*) FROM hits GROUP BY "UserID", "SearchPhrase" LIMIT 10;
+--SELECT "UserID", "SearchPhrase", COUNT(*) FROM hits GROUP BY "UserID", "SearchPhrase" LIMIT 10; 
+-- previous query did not have an ORDER BY, so the results were non-deterministic. Adding ORDER BY to make the test deterministic.
```
Rich-T-kid: @gabotechs  what do you think of this approach? certain queries can return correct results while not being an exact match of vanilla data fusion. The most straightforward approach to me is to update queries to only have one correct answer. This avoids updating the testing logic everywhere.
I commented out the old query as a refrence, if this looks good ill update the remaining breaking test.

Rich-T-kid:
chaning the queries is not a valid solution since there are standardized. The main issue that comes to mind is that record batches aren't reaching the final coordinator node. Assuming the comparison logic itself is correct, if the rows aren't there to be re-ordered into a deterministic order, the comparison will always fail.
For example: take a query that returns 100 rows, run it twice, once on vanilla DF and once on DDF. Since there's no ORDER BY, the engine can decide how to apply the LIMIT and either result is correct. The problem when we go to order the final results is that the data may have already been filtered before the final record batch was produced, which is likely what's causing the mismatches we're seeing.
### fix
1. the initial approach I want to try is looking at the raw output of both queries & then looking at the raw output after applying a limit 10. 
2. dig into the comparison logic.  this is assuming that the produced record batch from the queries contain the same rows from **step 1** if not comparing batches that include different rows will always be wrong.
3. double check how vanilla df handles Limit's. does it have free range to select any rows? since DDF is backed by vanilla DF it should do the exact same processing, but if vanila DF allows for non-deterministic limits we may need to change the approach

testdata/clickbench/queries/q17.sql
gabotechs: This queries are standard industry benchmarks, it might be better to stick to them, otherwise we would be testing a different thing.

Rich-T-kid:
did steps 1 &2 and it seems like this comes down to the engines picking two different 10-row slices. Going to add a flag to comparison functions that check for the limit operator. if it exist we cannot guarantee that the results are the exact same.this boils down to skipping the `detailed_batch_comparison()` call while keeping the row count check & schema validation.

src/test_utils/property_based.rs
```diff
@@ -35,6 +49,13 @@ pub fn compare_result_set(
             return Ok(()); // Both errored, so the query is valid
         }
     };
+    // Skip the row-by-row set comparison for queries where SQL semantics allow two engines
+    // to correctly return different result sets (e.g. LIMIT without ORDER BY, or ORDER BY
+    // with unbroken ties at the LIMIT boundary). Running the full comparison on these
+    // queries would surface false positives even though both engines are right.
+    if config.uses_undeterministic_limit_operator || config.non_deterministic_sort {
```
Rich-T-kid: this has the draw back of not actually checking if the results are correct/identical, but I don't see another way of handling this. the record batches returned by run(ctx,query) has already filter/truncated the results.

src/test_utils/property_based.rs
```diff
@@ -114,6 +128,13 @@ fn records_equal_as_sets(
             right[0].schema()
         );
     }
+    // Skip the row-by-row set comparison for queries where SQL semantics allow two engines
+    // to correctly return different result sets (e.g. LIMIT without ORDER BY, or ORDER BY
+    // with unbroken ties at the LIMIT boundary). Running the full comparison on these
+    // queries would surface false positives even though both engines are right.
+    if config.uses_undeterministic_limit_operator || config.non_deterministic_sort {
```
Rich-T-kid: moved this function a little lower so that it runs after a schema check and a row count check

Rich-T-kid:
@gabotechs this is good to review now 😃

tests/clickbench_correctness_test.rs
```diff
@@ -24,228 +24,263 @@ mod tests {
     const FILE_RANGE: Range<usize> = 0..3;
 
     #[tokio::test]
-    #[ignore = "Query 0 did not get distributed"]
+    #[ignore = "Query q0 (SELECT COUNT(*) FROM hits) still does not get distributed. The planner correctly chooses a single-task plan because of parquet statistics."]
```
gabotechs: 🤔 "still does not get distributed"? this sounds like an LLM generated this, pasted into the code, and was not revisited by the developer.

If the reason why this test does not get distributed has not changed, I'd suggest just leaving the `ingore` comment as before.

tests/clickbench_correctness_test.rs
```diff
@@ -24,228 +24,263 @@ mod tests {
...
 
     #[tokio::test]
     async fn test_clickbench_5() -> Result<()> {
-        test_clickbench_query("q5").await
+        test_clickbench_query("q5", PerTestConfig::default()).await
     }
 
     #[tokio::test]
-    #[ignore = "Query 6 did not get distributed"]
+    #[ignore = "Query q6 (SELECT MIN/MAX FROM hits) still does not get distributed. The planner correctly chooses a single-task plan because of parquet statistics."]
```
gabotechs: Same as above.

src/test_utils/property_based.rs
gabotechs: 🤔 what this is doing essentially is `if onfig.uses_undeterministic_limit_operator || config.non_deterministic_sort`, then just don't test anything.

I don't think this is a satisfying solution, we should still find a way to somehow compare the two result sets rather than just omitting it.

Rich-T-kid: I agree. I think a better would be to run the DataFusion query, and then run a slightly altered query for distributed DataFusion.
This would not mean altering the original query. for example query 17 is 
```SELECT "UserID", "SearchPhrase", COUNT(*) FROM hits GROUP BY "UserID", "SearchPhrase" LIMIT 10;```
we could take this query and run it on vanilla datafusion and we'll get back 10 rows

we could then take the results of this query and run the query again on distributed datafusion by adding a `in` clause?
for example
```
WITH agg AS (
    SELECT "UserID", "SearchPhrase", COUNT(*) AS cnt
    FROM hits
    GROUP BY "UserID", "SearchPhrase"
)
SELECT *
FROM agg
WHERE ("UserID", "SearchPhrase") IN (
    (123, 'phrase_a'),
    (456, 'phrase_b'),
    -- ...  from vanilla
)
```
This way no matter what 10 rows vanilla DataFusion returns we can always validate the aggregations produced line up with what vanilla DataFusion produces.
At least thats what comes to mind at the moment, ill have to think this over.

Comment (gabotechs) [reply src/test_utils/property_based.rs]: 🤔 That might get a bit convoluted though... for this PR, I think it's fine to leave it as is and keep the tests ignored, and we can think of a more complex solution in its own isolated PR.

Comment (Rich-T-kid) [reply src/test_utils/property_based.rs]: make sense! I pushed up a revised PR. I'll create a separate issue for the non-deterministic queries and put all the context/approaches in there.

File: tests/clickbench_correctness_test.rs [hunk 6]
```diff
@@ -289,7 +289,6 @@ mod tests {
 
         let (s_plan, s_results) = run(&s_ctx, &query_sql).await;
         let (d_plan, d_results) = run(&d_ctx, &query_sql).await;
-
         if !d_plan.as_any().is::<DistributedExec>() {
```
gabotechs, line 292: Not a big deal on this one, but it's typically a good practice to avoid including unrelated formatting changes, as they are not necessary, and they have a chance of introducing `git` conflicts with other colleagues PRs.

tests/tpcds_correctness_test.rs
```diff
@@ -173,7 +170,7 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore = "Fails with column 'c_last_review_date_sk' not found"]
+    #[ignore = "fails on CI but works locally, see https://github.com/datafusion-contrib/datafusion-distributed/pull/452#issuecomment-4439115012"]
```
gabotechs, line 173: 🤔 I remember seeing an issue upstream about this specific test

gabotechs (approved):
Thanks @Rich-T-kid! this looks good 👍

PR #455, opened 2026-05-18
Add reproducer tests for ChildrenIsolatorUnionExec bug

gabotechs:
Stack of some PRs for fixing a bug in `ChildrenIsolatorUnionExec`:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/455 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/456

---

Adds a regression test (`nested_union_budget_exceeds_children_sum` in `tests/work_unit_feed.rs`) that reproduces a production planning error:

  > Internal error: ChildrenIsolatorUnionExec had a task count N, which is greater than the sum of child task counts M

The bug is triggered when a `UnionExec` with cheap children (low combined task count) sits on the probe side of a broadcast `HashJoinExec` whose build side carries a higher task count into the same stage. Bottom-up reconciliation at the join takes `max(probe, build)`, and the top-down propagation pass then hands that inflated value down to the union as its CIU budget — exceeding the sum of the union's children's task counts.

### DIFF
--- a/tests/distributed_unions.rs
+++ b/tests/distributed_unions.rs
@@ -177,6 +177,102 @@ mod tests {
         exact_same_data(ctx.task_ctx(), physical, physical_distributed).await
     }
 
+    #[tokio::test]
+    async fn nested_unions() -> Result<(), Box<dyn Error>> {
+        let (ctx_distributed, _guard, _) = start_localhost_context(3, DefaultSessionBuilder).await;
+
+        // The LIMIT on the inner subqueries prevents the logical optimizer from
+        // flattening the nested `UNION ALL`s into a single `Union`, so the resulting
+        // physical plan contains a `UnionExec` whose child is another `UnionExec`.
+        let query = r#"
+        SELECT * FROM (
+            SELECT "MinTemp", "RainToday" FROM weather WHERE "MinTemp" > 10.0
+            UNION ALL
+            SELECT "MaxTemp", "RainToday" FROM weather WHERE "MaxTemp" < 30.0
+            LIMIT 1000000
+        )
+        UNION ALL
+        SELECT * FROM (
+            SELECT "Temp9am", "RainToday" FROM weather WHERE "Temp9am" > 15.0
+            UNION ALL
+            SELECT "Temp3pm", "RainToday" FROM weather WHERE "Temp3pm" < 25.0
+            LIMIT 1000000
+        )
+        ORDER BY "MinTemp", "RainToday"
+        "#;
+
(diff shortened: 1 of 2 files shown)


tests/distributed_unions.rs
```diff
@@ -177,6 +177,102 @@ mod tests {
...
+            UNION ALL
+            SELECT "MaxTemp", "RainToday" FROM weather WHERE "MaxTemp" < 30.0
+            LIMIT 1000000
+        )
+        UNION ALL
+        SELECT * FROM (
+            SELECT "Temp9am", "RainToday" FROM weather WHERE "Temp9am" > 15.0
+            UNION ALL
+            SELECT "Temp3pm", "RainToday" FROM weather WHERE "Temp3pm" < 25.0
+            LIMIT 1000000
```
gabotechs, line 199: This is essentially just me trying to reproduce the bug unsuccessfully. It's anyways a pretty interesting test, so I thought that it would be nice to commit it anyways.

tests/work_unit_feed.rs
```diff
@@ -360,6 +360,108 @@ mod tests {
...
+            UNION ALL
+            SELECT * FROM (
+                SELECT * FROM test_work_unit('c', 2, 'rows(3)', 'rows(1)', 'rows(1)', 'rows(2)')
+                UNION ALL
+                SELECT * FROM test_work_unit('d', 2, 'rows(1)', 'rows(1)', 'rows(2)', 'rows(1)')
+                LIMIT 1000000
+            )
+            ORDER BY tag, task, partition, letter
+        "#,
+        )
```
gabotechs, line 386: same as above

tests/work_unit_feed.rs
```diff
@@ -699,6 +801,33 @@ mod tests {
...
+            "expected the planner to reject the inconsistent CIU input (budget > children sum)",
+        );
+        let msg = err.to_string();
+        assert!(
+            msg.contains("ChildrenIsolatorUnionExec had a task count")
+                && msg.contains("greater than the sum of child task counts"),
+            "unexpected error message: {msg}"
+        );
+    }
+
```
gene-bordegaray, line 862: Would it be better to just ignore this with reasoning until the fix is merged

gene-bordegaray (approved):
You can address comment or not, up to you. But what an edge case haha

tests/work_unit_feed.rs
gabotechs: I instead added the actual snapshot expectations and `#[ignore]`ed the test

PR #456, opened 2026-05-18
Fix ChildrenIsolatorUnionExec budget mismatch via weight-based proportional allocation

gabotechs:
Stack of some PRs for fixing a bug in `ChildrenIsolatorUnionExec`:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/455
- https://github.com/datafusion-contrib/datafusion-distributed/pull/456 <- you are here

---

## Summary

Fixes the planning error introduced when a `UnionExec`'s stage budget exceeds the sum of its children's task counts.

### Root cause

`ChildrenIsolatorUnionExec` was built with **hard task counts** per child. The planner passed each child's bottom-up task count as a fixed budget, but when a sibling subtree in the same stage drove the stage budget higher, `from_children_and_task_counts` would receive a `task_count_budget` larger than the sum of those hard counts and reject it with:

> ChildrenIsolatorUnionExec had a task count N, which is greater than the sum of child task counts M

### Fix

Replace the hard-task-count API with a **weight-based proportional allocation**:

- `from_children_and_task_counts` → `from_children_and_weights`, taking `Vec<ChildWeight>` where each `ChildWeight` carries a relative `f64` weight and an optional hard `max` cap (used to surface `Maximum(N)` annotations).
- The stage budget is distributed across children using the **Hare largest-remainder method**: `floor(budget × wᵢ / Σw)` for each child, with leftover slots going to the children with the largest fractional remainders.
- A child whose proportional share rounds to zero is **packed into the last occupied task slot** rather than stealing a slot from a heavier child — this also unifies the `budget < n` and `budget >= n` cases into a single code path.
- `Maximum(N)` caps are honoured: a capped child never receives more than `N` slots; excess budget flows to uncapped siblings, or leaves empty trailing slots if every child is capped.
- Weights are `f64`; `0.0` is valid (no share, packed); negative and non-finite values are rejected upfront.

### DIFF
--- a/src/execution_plans/children_isolator_union.rs
+++ b/src/execution_plans/children_isolator_union.rs
@@ -99,25 +99,15 @@ pub struct ChildrenIsolatorUnionExec {
     >,
 }
 
-/// Per-child allocation hint passed to
-/// [`ChildrenIsolatorUnionExec::from_children_and_weights`].
-///
-/// `weight` sets the child's *relative* share of the stage's task budget — only the ratios
-/// among the non-zero entries matter. Valid values are any finite, non-negative `f64`:
-/// `0.0` means "no proportional share" (the child still runs but gets packed into an
-/// existing slot rather than earning its own). Negative and non-finite (`NaN`, `±∞`)
-/// values are rejected with an error. If every weight is `0.0`, the distribution falls
-/// back to uniform so the planner never has to special-case "no signal".
-///
-/// `max` is an optional *absolute* cap: the allocator will never assign the child more than
-/// `max` task slots, even if its proportional share would be larger. Use it to surface
-/// annotations like `Maximum(N)` that hard-limit how much parallelism a child can take.
-/// Leaving `max` unset means "no cap" — the child can absorb surplus budget from other
-/// capped siblings.
+/// Per-child allocation hint passed to [`ChildrenIsolatorUnionExec::from_children_and_weights`].
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct ChildWeight {
-    pub weight: f64,
-    pub max: Option<usize>,
+    /// Weight relative to other children. The higher the weight vs other children, the more tasks
+    /// will be allocated to it.
+    pub(crate) weight: f64,
+    /// Maximum task count cap for this child. While allocating tasks for this child, it cannot
+    /// exceed the specified `max` no matter its `weight`
+    pub(crate) max: Option<usize>,
 }
 
 impl ChildWeight {
--- a/src/execution_plans/children_isolator_union.rs
+++ b/src/execution_plans/children_isolator_union.rs
@@ -86,6 +86,10 @@ pub struct ChildrenIsolatorUnionExec {
     pub(crate) properties: Arc<PlanProperties>,
     pub(crate) metrics: ExecutionPlanMetricsSet,
     pub(crate) children: Vec<Arc<dyn ExecutionPlan>>,
+    /// The original per-child weights (and their optional hard caps) used to build the
+    /// `task_idx_map`. Stored so `with_new_children` can re-run the allocator with the same
+    /// inputs and preserve `Maximum(N)` caps across plan rewrites.
+    pub(crate) child_weights: Vec<ChildWeight>,
     pub(crate) task_idx_map: Vec<
         /* outer distributed task idx */
         Vec<(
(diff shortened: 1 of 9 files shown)


src/execution_plans/children_isolator_union.rs
```diff
@@ -95,24 +99,63 @@ pub struct ChildrenIsolatorUnionExec {
...
+/// `0.0` means "no proportional share" (the child still runs but gets packed into an
+/// existing slot rather than earning its own). Negative and non-finite (`NaN`, `±∞`)
+/// values are rejected with an error. If every weight is `0.0`, the distribution falls
+/// back to uniform so the planner never has to special-case "no signal".
+///
+/// `max` is an optional *absolute* cap: the allocator will never assign the child more than
+/// `max` task slots, even if its proportional share would be larger. Use it to surface
+/// annotations like `Maximum(N)` that hard-limit how much parallelism a child can take.
+/// Leaving `max` unset means "no cap" — the child can absorb surplus budget from other
+/// capped siblings.
```
gene-bordegaray: nit: just always find it easier to digest info in short bullets

src/execution_plans/children_isolator_union.rs
```diff
@@ -336,54 +380,49 @@ impl Stream for ObservedStream {
...
-///     `task_count_per_children`: [1, 1, 1]
-///     `task_count`: 3
-/// It returns the following output:
-///     `[[(0, 0/1)], [(1, 0/1)], [(2, 0/1)]]`
-/// That means that:
-/// - Task 0 will execute task 0 from child 0
-/// - Task 1 will execute task 0 from child 1
-/// - Task 2 will execute task 0 from child 2
+/// Behavior (a single proportional allocation handles every regime — `budget < n`, `budget
+/// == n`, and `budget > n`):
```
gene-bordegaray: It is not exactly clear what this means when first reading it.

src/execution_plans/children_isolator_union.rs
```diff
@@ -336,54 +380,49 @@ impl Stream for ObservedStream {
...
+///  - The budget is allocated proportionally with the Hare largest-remainder method (ties
+///    broken by lower child index first, so the output is deterministic).
```
gene-bordegaray: Would prefer to just have it stated what this method is briefly

src/execution_plans/children_isolator_union.rs
```diff
@@ -336,54 +380,49 @@ impl Stream for ObservedStream {
...
+///   weights: [w(1), w(1)], budget: 3  →  more budget than the total weight — the heavier-
+///       share child (after tiebreak) covers two slots:
+///       [[(0, 0/2)], [(0, 1/2)], [(1, 0/1)]]
 ///
-/// The function looks pretty much like the solution of a LeetCode problem, so it's not super
-/// easy to understand just be looking at the code. The best way to get a grasp of what its doing
-/// is by looking at the tests.
+///   weights: [Maximum(1), Maximum(1)], budget: 3  →  both children are capped at 1; the
+///       third slot stays empty:
+///       [[(0, 0/1)], [(1, 0/1)], []]
```
gene-bordegaray, line 391: Coudl empty be something like `[(2, N/A)]`

src/execution_plans/children_isolator_union.rs
```diff
@@ -396,98 +435,167 @@ fn split_children(
...
+            result[last_occupied].push((
+                child_idx,
+                DistributedTaskContext {
+                    task_index: 0,
+                    task_count: 1,
+                },
+            ));
+        }
+    }
     Ok(result)
```
gene-bordegaray: I think we could spread our the remaining children a bit better. If we had a situatiion like:

```text
weights = [1, 1, 1, 1, 1]
budget = 3

floor gives slots to c0, c1, c2
c3 and c4 get zero -> both are packed into the last occupied slot

result:
  t0: c0
  t1: c1
  t2: c2, c3, c4
```

So now t2 has 3 children when we could have 2 dispersed more evenly and I think this can get skewed much worse which would cause a bottleneck.

Could we change the packing to redistirbute based on load not always just last slot to get something like:

```text
result:
  t0: c0
  t1: c1, c4
  t2: c2, c3
```

tests/distributed_unions.rs
```diff
@@ -147,10 +147,10 @@ mod tests {
             @r"
         ┌───── DistributedExec ── Tasks: t0:[p0]
         │ SortPreservingMergeExec: [MinTemp@0 ASC NULLS LAST, RainToday@1 ASC NULLS LAST]
-        │   [Stage 1] => NetworkCoalesceExec: output_partitions=18, input_tasks=3
+        │   [Stage 1] => NetworkCoalesceExec: output_partitions=27, input_tasks=3
         └──────────────────────────────────────────────────
-          ┌───── Stage 1 ── Tasks: t0:[p0..p5] t1:[p6..p11] t2:[p12..p17]
-          │ DistributedUnionExec: t0:[c0, c1] t1:[c2, c3] t2:[c4]
+          ┌───── Stage 1 ── Tasks: t0:[p0..p8] t1:[p9..p17] t2:[p18..p26]
+          │ DistributedUnionExec: t0:[c0] t1:[c1] t2:[c2, c3, c4]
```
gene-bordegaray: Ya, this is kinda what I was mentioning about packing only the last guy, could we have this be more even

gene-bordegaray (commented):
I like this change, I think the model makes more sense around doing distributed unions. I am just wary about packing that last task

src/execution_plans/children_isolator_union.rs
gabotechs: I should have removed this, this is a leftover from a previous iteration, no longer relevant.

src/execution_plans/children_isolator_union.rs
gabotechs: That implies there's a third children, but there isn't in this case, therefore empty is a length 0 array

src/execution_plans/children_isolator_union.rs
gabotechs: I think with the comments I left in each step of the function, just the first phrase should be enough. If someone wants to take a look at how it works, probably just stepping through the function body is enough.

src/execution_plans/children_isolator_union.rs
gabotechs: :thinking: I can imagine how that has the risk of allocating more children to the first slots, which are at risk of being the most loaded ones.

I'll try to come up with a general satisfying rule for this.

gabotechs: I changed this to round-robin the remainders across slots, rather than putting them all at the end

gabotechs:
I'm double checking that this does not introduce any performance regression

gabotechs:
TPCH SF10: 2% faster
TPCDS SF1: 3% faster

So within noise, no performance impact.

tests/distributed_unions.rs
```diff
@@ -150,7 +150,7 @@ mod tests {
         │   [Stage 1] => NetworkCoalesceExec: output_partitions=18, input_tasks=3
         └──────────────────────────────────────────────────
           ┌───── Stage 1 ── Tasks: t0:[p0..p5] t1:[p6..p11] t2:[p12..p17]
-          │ DistributedUnionExec: t0:[c0, c1] t1:[c2, c3] t2:[c4]
```
gene-bordegaray, line 153: nice

gene-bordegaray (approved):
👍

PR #461, opened 2026-05-23
Collect stage-level metrics

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/469
- https://github.com/datafusion-contrib/datafusion-distributed/pull/461 <- you are here

This PR introduces stage-level metrics collection to provide better observability into the execution of distributed query plans. By collecting metrics at the stage level, we can understand performance characteristics of individual stages and optimize task distribution and scheduling decisions based on actual runtime data.


src/common/time.rs
```diff
@@ -0,0 +1,9 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// Nanoseconds elapsed since UNIX epoch.
+pub(crate) fn now_ns() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|duration| duration.as_nanos() as u64)
+        .unwrap_or(0)
+}
```
gene-bordegaray, line 9: Something about this is bothering me.

I dont think we should unwrap or 0 as that would be confusing.

Maybe like:

```rust
 pub(crate) fn unix_epoch_nanos() -> Option<u64> {
      SystemTime::now()
          .duration_since(UNIX_EPOCH)
          .ok()
          .and_then(|d| u64::try_from(d.as_nanos()).ok())
  }
```

and just handle the `Option`?

src/worker/impl_execute_task.rs
```diff
@@ -88,10 +90,11 @@ pub(crate) async fn execute_local_task(
...
         let key_clone = key.clone();
         let plan = Arc::clone(&plan);
-        let plan_for_drop = Arc::clone(&plan);
+        let plan_cloned = Arc::clone(&plan);
         let fully_finished = Arc::new(AtomicBool::new(false));
         let fully_finished_cloned = Arc::clone(&fully_finished);
-        let d_ctx_for_last = d_ctx.clone();
+        let d_ctx_cloned = d_ctx.clone();
+        let task_data_metrics = Arc::clone(&task_data.task_data_metrics);
         let stream = map_last_stream(stream, move |msg, last_msg_in_stream| {
```
gene-bordegaray: Could we introduce a helper to eliminate some of these clones. I think if we make a helper and call it in the colsures that shoudl clean things up nicely.

src/stage.rs
```diff
@@ -316,6 +328,33 @@ fn display_inner_ascii(
...
+        query_id: serialize_uuid(&stage.query_id()),
+        stage_id: stage.num() as u64,
+        task_number: 0,
+    };
+    let mut all_metrics = MetricsSet::new();
+    while let Some(task_metrics) = metrics_store.get(&task_key) {
+        let Some(metrics_set) = task_metrics.task_metrics else {
+            continue;
+        };
+        for mut metric in metrics_set.metrics {
```
gene-bordegaray: possible infinite loop?

src/worker/task_data.rs
```diff
@@ -22,7 +26,80 @@ pub struct TaskData {
...
+    /// When the plan execution was triggered by the parent worker.
+    pub(super) plan_executed_at: MaxLatencyMetric,
+    /// When the execution stream finished.
+    pub(super) plan_finished_at: MaxLatencyMetric,
+}
+
+impl TaskDataMetrics {
+    pub(super) fn new(query_start_time_ns: u64) -> Self {
+        let plan_added_at = MaxLatencyMetric::default();
+        plan_added_at.add_duration(Duration::from_nanos(now_ns() - query_start_time_ns));
```
gene-bordegaray: We use `now_ns()` here with `query_start_time_ns` but these are coming from different sources: `now_ns()` from the worker we are on and `query_start_time_ns` from the coordinator.

This means if there is clock skew we can have some weird stuff happen. I don't know if there is a good fix for this but it should at least be documented that this is possible behavior.

Could we extract the same info in a less brittle way?

gene-bordegaray (commented):
Mostly small cleanups, this will be very useful 👍 

Some notes:
- could we get a test for the stage metrics display?
- would it be useful to have a task runtime duration that uses local `Instant`, I think it would be a good baseline to compare with the other added metrics and a source of truth

src/common/time.rs
gabotechs: Based on the docs:

> Returns an Err if earlier is later than self, and the error contains how far from self the time is.

So in practice, while comparing `SystemTime::now()` against `UNIX_EPOCH`, this can never fail. It would be a shame to take a hit in usability of this function for something that can never fail.

Although, in these situations where there's something that can never fail for 100% sure, the Rusty way of handling it is by panicking. It just means that we cannot send this code back in time to the 60s.

src/worker/task_data.rs
gabotechs: Indeed this is a known challenge, we've discussed this before for the metrics we fire inside `worker_connection_pool.rs`, which suffer from this same problem.

Unfortunately, besides ensuring no clock skew across machines with a centralized system, I don't think there's a way addressing this problem without defying the fabric of space-time itself.

In practice, what we are measuring here is in the order of milliseconds/seconds, and at that scale clock skew across a typical homogeneous k8s deployment is not noticeable, and for now it has proved useful, so our stance has been to just stick with this approach.

src/stage.rs
gabotechs: :thinking: yeah, let's improve this

src/worker/impl_execute_task.rs
gabotechs: I think we should be able to clean this up more heavily, and remove this `map_last_stream` function.

It's something inherited from a previous approach, but I think we no longer need it.

gabotechs: This is a great cleanup indeed, did this in its own isolated commit: https://github.com/datafusion-contrib/datafusion-distributed/pull/461/changes/78aae87d974e88c67920fcdb10ab6c69dc328d75

gabotechs:
> would it be useful to have a task runtime duration that uses local Instant, I think it would be a good baseline to compare with the other added metrics and a source of truth

This should be redundant with the current metrics, as it can be derived from `plan_finished_at - plan_executed_at`, the runtime duration is exactly that with 100% accuracy.

src/worker/impl_execute_task.rs
gene-bordegaray: that's a good looking diff 😆

src/worker/task_data.rs
gene-bordegaray: ah ya we cant bend space and time, but this is feasible

can we add safety with a `saturating_sub()` along with the other arithmetic in the file 👍

src/worker/task_data.rs
```diff
@@ -22,7 +26,80 @@ pub struct TaskData {
...
+pub(crate) const PLAN_EXECUTED_AT_METRIC: &str = "plan_executed_at";
+pub(crate) const PLAN_FINISHED_AT_METRIC: &str = "plan_finished_at";
+
+#[derive(Debug)]
+pub(super) struct TaskDataMetrics {
+    pub(super) query_start_time_ns: u64,
+    /// When the plan was set by the coordinator.
+    pub(super) plan_added_at: MaxLatencyMetric,
+    /// When the plan execution was triggered by the parent worker.
+    pub(super) plan_executed_at: MaxLatencyMetric,
```
gene-bordegaray, line 46: should this be a `FirstLatencyMetric` instead of Max so we dont keep a later time if two callers execute almost at the same time?

gene-bordegaray (approved):
just two small comments 👍

src/worker/task_data.rs
gabotechs: Good suggestion, done in https://github.com/datafusion-contrib/datafusion-distributed/pull/461/commits/3923c296e88158f0ecad4ee21b2209f45fcb6b4a

src/worker/task_data.rs
gabotechs: While consuming this in:

I found it a bit more useful to report the maximum one. This is because typically what drives overall latency up in a query is the slowest partition, not necessarily the first one, so reporting the max there provides a really good hint of which task contained the partition that was the biggest offender towards slowness.

If we reported first latency, it'd likely be always the minimum, and that has the risk of not exposing offending partitions.

gene-bordegaray: I see that makes sense, thanks for clarifying

PR #311, opened 2026-01-26
Cost based planning

gabotechs:
Reworks the distributed planner so that it integrates with DataFusion upstream statistics system in favor of having users provide a custom TaskEstimator.

---

There are some key changes in this PR that rework how this project assign tasks and stages to a plan:

### Remove TaskEstimator

Users are no longer free to specific how many tasks should be used for a distributed query:
- previously: users provide for their own nodes the number of tasks to be used in the stage containing their nodes, and upper stages are estimated based on an arbitrary cardinality reduction factor.
- now: users provide runtime statistics for their nodes, using DataFusion upstream `ExecutionPlan::partition_statistics()` method. Based on this statistics, Distributed DataFusion calculates how many tasks are appropriate

### Rely on upstream statistics system

This PR heavily consumes the `Statistics` provided by the different nodes in order to estimate how much data is going to flow through them. 

There are still some gaps in upstream statistics that are bridge in this project with some sane defaults.

### Compute cost assignation

One of the biggest additions in this PR is the compute complexity estimation for the different nodes. Each node has cost attached that is estimated based on how compute heavy they are.

The cost is measured in "bytes processed", which estimates how many bytes are expected to be processed by the node given the node itself, and the estimated rows and bytes that are going to flow through it given by upstream's `ExecutionPlan::partition_statistics`.

The computational complexity is taking into account for the different operators with the following enum:

```rust
pub enum Complexity {
    /// Constant complexity
    Constant,
    /// Linear with a specific column from a specific child.
    Linear(LinearComplexity),
    /// NLogM
    Log(Box<Complexity>, Box<Complexity>),
    /// N+M
    Plus(Box<Complexity>, Box<Complexity>),
    /// N*M
    Multiply(Box<Complexity>, Box<Complexity>),
}
```

---

# Results

TODO: choose better defaults in order to improve performance

```
=== Comparing tpch_sf1 results from engine 'datafusion-distributed-main' [prev] with 'datafusion-distributed-cost-based-planning' [new] ===
      q1: prev= 471 ms, new= 586 ms, diff=1.24 slower ❌
      q2: prev= 636 ms, new= 725 ms, diff=1.14 slower ✖
      q3: prev= 969 ms, new=1191 ms, diff=1.23 slower ❌
      q4: prev= 474 ms, new= 470 ms, diff=1.01 faster ✔
      q5: prev= 477 ms, new= 790 ms, diff=1.66 slower ❌
[...]

PR #463, opened 2026-05-23
Add NetworkBoundaryBuilder argument to inject_network_boundaries

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/463 <- you are here

This PR introduces a NetworkBoundaryBuilder argument to the network boundary injection logic, allowing more flexible and configurable strategies for determining which exchanges require network communication. This enables better optimization of data movement across distributed tasks.

### DIFF
--- a/src/distributed_planner/inject_network_boundaries.rs
+++ b/src/distributed_planner/inject_network_boundaries.rs
@@ -3,19 +3,21 @@ use crate::execution_plans::{ChildWeight, ChildrenIsolatorUnionExec};
 use crate::stage::LocalStage;
 use crate::{
     BroadcastExec, DistributedConfig, NetworkBoundaryExt, NetworkBroadcastExec,
-    NetworkCoalesceExec, NetworkShuffleExec, TaskCountAnnotation, TaskEstimator,
+    NetworkCoalesceExec, NetworkShuffleExec, Stage, TaskCountAnnotation, TaskEstimator,
 };
+use async_trait::async_trait;
 use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
 use datafusion::common::{HashMap, Result, plan_err};
 use datafusion::config::ConfigOptions;
 use datafusion::physical_expr::Partitioning;
-use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
 use datafusion::physical_plan::execution_plan::CardinalityEffect;
 use datafusion::physical_plan::joins::{HashJoinExec, PartitionMode};
 use datafusion::physical_plan::repartition::RepartitionExec;
 use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
 use datafusion::physical_plan::union::UnionExec;
+use datafusion::physical_plan::{ExecutionPlan, PlanProperties};
+use std::any::TypeId;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -25,9 +27,9 @@ use uuid::Uuid;
 /// [NetworkCoalesceExec] nodes wherever a stage boundary is needed. The returned plan has the
 /// same shape as the input except for these inserted boundary nodes.
 ///
-/// Per-node task counts are recorded in a side map on the [Context] (keyed by plan-pointer
-/// identity) rather than mutated into the plan itself. Later passes look them up via
-/// [Context::task_count].
+/// Per-node task counts are recorded in a side map on the [InjectNetworkBoundaryContext] (keyed by
+/// plan-pointer identity) rather than mutated into the plan itself. Later passes look them up via
+/// [InjectNetworkBoundaryContext::task_count].
 ///
 /// # The three-phase loop
 ///
@@ -131,13 +133,15 @@ use uuid::Uuid;
 /// boundary injection, so the head stage is closed by running one final Phase 2 pass over
 /// the whole plan. This guarantees every node (including head-stage nodes that never sat
 /// directly above a boundary) has a task count recorded.
-pub(super) async fn inject_network_boundaries(
+pub(crate) async fn inject_network_boundaries(
     plan: Arc<dyn ExecutionPlan>,
+    nb_builder: impl NetworkBoundaryBuilder + Send + Sync,
     cfg: &ConfigOptions,
 ) -> Result<Arc<dyn ExecutionPlan>> {
-    let ctx = Context {
+    let ctx = InjectNetworkBoundaryContext {
         cfg,
         d_cfg: DistributedConfig::from_config_options(cfg)?,
+        nb_builder: &nb_builder,
         task_counts: &Mutex::new(HashMap::new()),
         query_id: Uuid::new_v4(),
         stage_id: &AtomicUsize::new(1),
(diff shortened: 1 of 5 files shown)


src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -473,6 +488,37 @@ fn propagate_task_count_until_network_boundaries(
     }
 }
 
+pub(crate) struct NetworkBoundaryBuilderResult {
+    pub(crate) task_count_above: TaskCountAnnotation,
```
gene-bordegaray: should we validate the invariant that the task count must be greater than 0. The cardinality struct won't do this but I could see this happening in AQE?

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -315,31 +324,37 @@ async fn _inject_network_boundaries(
...
                 num: ctx.fetch_add_stage_id(),
                 plan,
                 tasks: task_count.as_usize(),
-            };
-            let plan = Arc::new(NetworkBroadcastExec::from_stage(input_stage));
-            let task_count = Desired((f * task_count.as_usize() as f64).ceil() as usize);
-            Ok(ctx.plan_with_task_count(plan, task_count))
+            });
+            let result = ctx.apply_nb_builder(Arc::new(plan)).await?;
+            return Ok(ctx.plan_with_task_count(result.network_boundary, result.task_count_above));
```
gene-bordegaray: The control bock here is a bit more confusing than the old one. Could we just hvae this return Ok() like the older shape?

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -506,24 +552,57 @@ fn propagate_task_count_until_network_boundaries(
...
+        cfg: &ConfigOptions,
+    ) -> Result<NetworkBoundaryBuilderResult> {
+        if nb.as_any().is::<NetworkCoalesceExec>() {
+            return Ok(NetworkBoundaryBuilderResult {
+                task_count_above: Maximum(1),
+                network_boundary: nb,
+            });
         }
-    }
+        let d_cfg = DistributedConfig::from_config_options(cfg)?;
```
gene-bordegaray: nit: could we just pas the distributed config to avoid this per boundary

gene-bordegaray (approved):
Overall some small nits but nothing blocking

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -473,6 +488,37 @@ fn propagate_task_count_until_network_boundaries(
...
+    F: Send,
+{
+    async fn build(
+        &self,
+        nb: Arc<dyn NetworkBoundary>,
+        cfg: &ConfigOptions,
+    ) -> Result<NetworkBoundaryBuilderResult> {
+        self(nb, cfg)?.await
+    }
+}
```
gene-bordegaray, line 535: also just wanted to say this is quite superb

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -390,90 +408,134 @@ async fn _inject_network_boundaries(
...
-                scaled_up.apply(|plan| {
-                    ctx.set_task_count(plan, task_count);
-                    Ok(TreeNodeRecursion::Continue)
-                })?;
-                Ok(ctx.plan_with_task_count(scaled_up, task_count))
+impl InjectNetworkBoundaryContext<'_> {
+    fn propagate_task_count_until_network_boundaries(
+        &self,
+        plan: &Arc<dyn ExecutionPlan>,
+        task_count: TaskCountAnnotation,
```
gabotechs, line 415: This function has not changed, it's just now a method of the `InjectNetworkBoundaryContext`. Because the fact that it's not under 1 extra indentation level, the diff gets pretty funny here, but I actually left it untouched.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -390,90 +408,134 @@ async fn _inject_network_boundaries(
...
+
+#[async_trait]
+pub(crate) trait NetworkBoundaryBuilder {
+    async fn build<'a>(
+        &'a self,
+        input_stage: LocalStage,
+        nb_type: TypeId,
+        nb_ctx: &'a InjectNetworkBoundaryContext<'a>,
+    ) -> Result<NetworkBoundaryBuilderResult>;
+}
```
gabotechs, line 517: I've needed to change significantly the signature here VS last time you took a look @gene-bordegaray. This comes from the fact that I integrated a cost based planner + runtime collected statistics for choosing an appropriate task count size for the different stages, which implies several things:
- Now, upon delimiting a stage, we might want to scaled up/down the `TaskCountAnnotation` based on cost+runtime statistics, which means that the `propagate_task_count_until_network_boundaries` needs to be called inside the `NetworkBoundaryBuilder` implementation.
- The `NetworkBoundaryBuilder`, as it's now capable of propagating any arbitrary desired task count, it needs access to the `InjectNetworkBoundaryContext` (previously named just `Context`). As this contextual struct now can no longer be private to this file, I chose a better name upon making it `pub(crate)`
- The type of network boundary needs to be propagated somehow, and the way to do is by it's `TypeId` (from Rust std library), so that we can do things like `nb_type == TypeId::of::<NetworkCoalesceExec>()`. This is the equivalent to downcasting, but when no specific struct implementation is passed.

gabotechs: For getting a sense of why this is necessary, a good place to look at is:

https://github.com/datafusion-contrib/datafusion-distributed/pull/432/changes#diff-01ffef8aa9ef0e8104dfe1c133320f1c1edd157fd968798b9206a3e04dd95228

Which has the dynamic implementation of this trait

gene-bordegaray: awesome sounds like some cool stuff. I will read over this guy and the cost model 😄

gabotechs: I rescued the cost model from https://github.com/datafusion-contrib/datafusion-distributed/pull/311, which I dropped as the conclusion was that for a cost model to work, we'd need good stats, and we don't have them.

However, AQE collects stats at runtime, so they are pretty good stats, which gives another opportunity for the cost model to shine.

I'm also getting some pretty wild benchmark results, it's way more code than what I anticipated (because of the cost model mainly), but I'm seeing that it's worth it. Just posted some benchmark results from that branch on discord here:

https://discord.com/channels/885562378132000778/885562378132000781/1513488324461858956

And I'm about to update [Adaptive task count assignation](https://github.com/datafusion-contrib/datafusion-distributed/pull/432) with the new benchmark results.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: Changed

src/distributed_planner/inject_network_boundaries.rs
gabotechs: I think it's not a big deal. Validating helps when an external user is the one that sends this `TaskCountAnnotation`, but as this is still a private trait, and we have full control over the implementations, I'd lean towards just doing the validation in they dynamic implementation.

gabotechs:
I'm going to pull this one in for unblocking some follow up PRs in the stack, but @gene-bordegaray please feel free to re-review the new changes anyways and I'll very gladly address the comments in a follow up PR.

PR #466, opened 2026-05-25
update benchmark docs to match --path -> --dataset rename

kentkr:
Simple one liner to update outdated benchmark docs

The arg `--path` was updated to `--dataset` for `dfbench` in #300


gabotechs (approved):
Looks good, thanks @kentkr!

PR #464, opened 2026-05-24
Add MaxGauge metric

gabotechs:
This is one PR from the following stack of PRs:

- https://github.com/datafusion-contrib/datafusion-distributed/pull/477
- https://github.com/datafusion-contrib/datafusion-distributed/pull/463
- https://github.com/datafusion-contrib/datafusion-distributed/pull/464 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/478
- https://github.com/datafusion-contrib/datafusion-distributed/pull/479
- https://github.com/datafusion-contrib/datafusion-distributed/pull/486
- https://github.com/datafusion-contrib/datafusion-distributed/pull/432

This PR introduces a MaxGauge metric to provide better tracking of peak values in distributed metrics collection. This enables more accurate monitoring of resource utilization and helps identify bottlenecks in the execution pipeline.

PR #469, opened 2026-05-26
Factor out distributed recursion

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/469 <- you are here
- https://github.com/datafusion-contrib/datafusion-distributed/pull/461

This PR factors out common recursion patterns used in the distributed planner into reusable helpers, reducing complexity in critical parts of the codebase. By centralizing recursion logic, we improve code maintainability and reduce the risk of subtle bugs in recursive tree traversals.

### DIFF
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -14,8 +14,8 @@ pub(crate) trait TreeNodeExt {
     ///
     /// For example, the presence of [ChildrenIsolatorUnionExec] will make this function
     /// not recurse into nodes that would be ignored because of the contextual
-    /// [DistributedTaskContext], and while recursing into its children, the propagated
-    /// [DistributedTaskContext] will be mutated.
+    /// [DistributedTaskContext], and while recursing into its children, a different
+    /// [DistributedTaskContext] will be passed.
     ///
     /// The return [`TreeNodeRecursion`] controls the recursion and can cause an early return.
     ///
@@ -43,8 +43,9 @@ pub(crate) trait TreeNodeExt {
         Self: Sized;
 
     /// Recursively rewrite the tree using `f` in a bottom-up (post-order) fashion, propagating
-    /// the appropriate task count based on the presence of nodes that can isolate tasks, like
-    /// [ChildrenIsolatorUnionExec].
+    /// the appropriate task count based on the presence of nodes that can isolate tasks (e.g.,
+    /// [ChildrenIsolatorUnionExec]) and the presence of network boundaries that change the task
+    /// count.
     ///
     /// `f` is applied to the node's children first, and then to the node itself.
     fn transform_up_with_task_count<F: FnMut(Self, usize) -> Result<Transformed<Self>>>(
@@ -56,8 +57,9 @@ pub(crate) trait TreeNodeExt {
         Self: Sized;
 
     /// Recursively rewrite the tree using `f` in a top-down (pre-order) fashion, propagating
-    /// the appropriate task count based on the presence of nodes that can isolate tasks, like
-    /// [ChildrenIsolatorUnionExec].
+    /// the appropriate task count based on the presence of nodes that can isolate tasks (e.g.,
+    /// [ChildrenIsolatorUnionExec]) and the presence of network boundaries that change the task
+    /// count.
     ///
     /// `f` is applied to the node first, and then its children.
     #[allow(dead_code)] // Used in follow up work.
@@ -115,41 +117,39 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
         Self: Sized,
     {
         // None = skip this subtree (irrelevant CIU child for our task index).
-        let stack = RefCell::new(vec![Some(dt_ctx)]);
-        self.transform_down_up(
-            |node| {
-                let Some(dt_ctx) = stack.borrow_mut().pop().unwrap() else {
-                    return Ok(Transformed {
-                        data: node,
-                        transformed: false,
-                        tnr: TreeNodeRecursion::Jump,
-                    });
-                };
-                let transformed = f(node, dt_ctx.clone())?;
-                if transformed.tnr != TreeNodeRecursion::Continue
-                    || transformed.data.is_network_boundary()
-                {
-                    return Ok(Transformed {
-                        tnr: TreeNodeRecursion::Jump,
-                        ..transformed
-                    });
+        let mut stack = vec![Some(dt_ctx)];
+        self.transform_down(|node| {
+            let Some(dt_ctx) = stack.pop().unwrap() else {
+                return Ok(Transformed {
+                    data: node,
+                    transformed: false,
+                    tnr: TreeNodeRecursion::Jump,
+                });
+            };
+            let transformed = f(node, dt_ctx.clone())?;
+            if transformed.tnr == TreeNodeRecursion::Stop {
+                return Ok(transformed);
+            }
+            if transformed.tnr != TreeNodeRecursion::Continue
+                || transformed.data.is_network_boundary()
+            {
+                return Ok(Transformed {
+                    tnr: TreeNodeRecursion::Jump,
+                    ..transformed
+                });
+            }
+            let node = &transformed.data;
+            if let Some(ciu) = node.as_any().downcast_ref::<ChildrenIsolatorUnionExec>() {
+                let mut child_ctxs = vec![None; ciu.children.len()];
+                for (child_idx, child_ctx) in &ciu.task_idx_map[dt_ctx.task_index] {
+                    child_ctxs[*child_idx] = Some(child_ctx.clone());
                 }
-                let node = &transformed.data;
-                if let Some(ciu) = node.as_any().downcast_ref::<ChildrenIsolatorUnionExec>() {
-                    let mut child_ctxs = vec![None; ciu.children.len()];
-                    for (child_idx, child_ctx) in &ciu.task_idx_map[dt_ctx.task_index] {
-                        child_ctxs[*child_idx] = Some(child_ctx.clone());
-                    }
-                    stack.borrow_mut().extend(child_ctxs.into_iter().rev());
-                } else {
-                    stack
-                        .borrow_mut()
-                        .extend(node.children().iter().map(|_| Some(dt_ctx.clone())).rev());
-                }
-                Ok(transformed)
-            },
-            |node| Ok(Transformed::no(node)),
-        )
+                stack.extend(child_ctxs.into_iter().rev());
+            } else {
+                stack.extend(node.children().iter().map(|_| Some(dt_ctx.clone())).rev());
+            }
+            Ok(transformed)
+        })
     }
 
     fn transform_up_with_task_count<F: FnMut(Self, usize) -> Result<Transformed<Self>>>(
@@ -214,14 +214,14 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::NetworkCoalesceExec;
     use crate::execution_plans::ChildWeight;
+    use crate::stage::RemoteStage;
+    use crate::{NetworkCoalesceExec, Stage};
     use datafusion::arrow::datatypes::Schema;
     use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
     use datafusion::physical_plan::empty::EmptyExec;
     use datafusion::physical_plan::union::UnionExec;
     use insta::assert_snapshot;
-
     // ── apply_with_dt_ctx ────────────────────────────────────────────────────────
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+pub(crate) trait TreeNodeExt {
+    /// Applies `f` to the node then each of its children, recursively (a
+    /// top-down, pre-order traversal), propagating the [DistributedTaskContext] correctly
+    /// across nodes that mutate this context, and ignoring nodes that do not belong to
+    /// the passed [DistributedTaskContext].
+    ///
+    /// For example, the presence of [ChildrenIsolatorUnionExec] will make this function
+    /// not recurse into nodes that would be ignored because of the contextual
+    /// [DistributedTaskContext], and while recursing into its children, the propagated
+    /// [DistributedTaskContext] will be mutated.
```
jayshrivastava: mega nit: this confused me. I was wondering how/why the parameter might get mutated.
```suggestion
    /// [DistributedTaskContext], and the [DistributedTaskContext] may be modified before
    /// recursing into its children.
```

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+    fn apply_with_dt_ctx<F: FnMut(&Self, DistributedTaskContext) -> Result<TreeNodeRecursion>>(
+        &self,
+        ctx: DistributedTaskContext,
+        mut f: F,
+    ) -> Result<TreeNodeRecursion> {
+        fn recurse<
+            F: FnMut(&Arc<dyn ExecutionPlan>, DistributedTaskContext) -> Result<TreeNodeRecursion>,
+        >(
+            plan: &Arc<dyn ExecutionPlan>,
+            ctx: DistributedTaskContext,
```
jayshrivastava, line 85: Would it make sense to make `DistributedTaskContext` implement `Copy`?

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+                if let Some(ciu) = plan.as_any().downcast_ref::<ChildrenIsolatorUnionExec>() {
+                    // Just recurse to children that will actually get executed by this
+                    // ChildrenIsolatorUnionExec.
+                    ciu.task_idx_map[ctx.task_index].iter().apply_until_stop(
+                        |(child_i, child_ctx)| {
+                            recurse(&ciu.children[*child_i], child_ctx.clone(), f)
+                        },
+                    )
+                } else if plan.is_network_boundary() {
+                    Ok(TreeNodeRecursion::Continue)
```
jayshrivastava, line 98: I'd expect this to be `TreeNodeRecursion::Jump` because the trait method says "This function does not recurse into the input of network boundaries.". Network Boundaries have no children, so this doesn't really matter I guess.... Unless you call this during planning, in which case the NB has a `LocalStage`, meaning this will recurse into the children.

jayshrivastava: We should have some better comments on `LocalStage` and `RemoteStage`. They are basically `PlanningStage` and `ExecutionStage` at this point.

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+        self,
+        dt_ctx: DistributedTaskContext,
+        mut f: F,
+    ) -> Result<Transformed<Self>>
+    where
+        Self: Sized,
+    {
+        // None = skip this subtree (irrelevant CIU child for our task index).
+        let stack = RefCell::new(vec![Some(dt_ctx)]);
+        self.transform_down_up(
```
jayshrivastava: Why use `transform_down_up` and pass `|node| Ok(Transformed::no(node)),` to the up function? Isn't `transform_down` simpler?

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+                let Some(dt_ctx) = stack.borrow_mut().pop().unwrap() else {
+                    return Ok(Transformed {
+                        data: node,
+                        transformed: false,
+                        tnr: TreeNodeRecursion::Jump,
+                    });
+                };
+                let transformed = f(node, dt_ctx.clone())?;
+                if transformed.tnr != TreeNodeRecursion::Continue
+                    || transformed.data.is_network_boundary()
```
jayshrivastava: Ok so we don't traverse into network boundaries. Should all the trait methods say "    /// This function does not recurse into the input of network boundaries."?

When I started reading this PR, it seemed like you only wanted one method to skip traversing into them because only one had that comment.

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+            |node| {
+                let Some(dt_ctx) = stack.borrow_mut().pop().unwrap() else {
+                    return Ok(Transformed {
+                        data: node,
+                        transformed: false,
+                        tnr: TreeNodeRecursion::Jump,
+                    });
+                };
+                let transformed = f(node, dt_ctx.clone())?;
+                if transformed.tnr != TreeNodeRecursion::Continue
```
jayshrivastava: Do we really need the input closure to declare the `TreeNodeRecursion`? It's a weird abstraction. The user may intend to `TreeNodeRecursion::stop()` but we override that with `TreeNodeRecursion::Jump`.

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+    >(
+        self,
+        dt_ctx: DistributedTaskContext,
+        mut f: F,
+    ) -> Result<Transformed<Self>>
+    where
+        Self: Sized,
+    {
+        // None = skip this subtree (irrelevant CIU child for our task index).
+        let stack = RefCell::new(vec![Some(dt_ctx)]);
```
jayshrivastava: Don't need `RefCell` bc you only pass it to one closure. It can just be `mut Vec`

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+                        .borrow_mut()
+                        .extend(node.children().iter().map(|_| Some(dt_ctx.clone())).rev());
+                }
+                Ok(transformed)
+            },
+            |node| Ok(Transformed::no(node)),
+        )
+    }
+
+    fn transform_up_with_task_count<F: FnMut(Self, usize) -> Result<Transformed<Self>>>(
```
jayshrivastava, line 155: For this and `transform_down_with_task_count`, whether or not we recurse into the children will depend if the NB has a LocalStage or RemoteStage. In other words, behavior will depend on when you use this recursion helper (execution or planning).

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+
+    fn union(children: Vec<Arc<dyn ExecutionPlan>>) -> Arc<dyn ExecutionPlan> {
+        UnionExec::try_new(children).unwrap()
+    }
+
+    fn network_boundary(
+        input: Arc<dyn ExecutionPlan>,
+        producer_tasks: usize,
+    ) -> Arc<dyn ExecutionPlan> {
+        Arc::new(NetworkCoalesceExec::try_new(input, producer_tasks, 1).unwrap())
```
jayshrivastava, line 585: This means you're testing with LocalStage only

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    #[test]
+    fn apply_ciu_multi_children_per_task() {
+        // 4 children split across 2 tasks → each task sees 2 children
+        let plan = ciu(vec![leaf(), leaf(), leaf(), leaf()], vec![1, 1, 1, 1], 2).unwrap();
+        assert_snapshot!(trace_apply(&plan, ctx(0, 2)), @r"
+        CIU [ctx=0/2]
+        Leaf [ctx=0/1]
+        Leaf [ctx=0/1]
```
jayshrivastava, line 346: nit: not sure if it's worth the extra coverage but you don't assert the order of these leaf nodes

src/common/recursion.rs
```diff
@@ -0,0 +1,678 @@
...
+        ");
+    }
+
+    #[test]
+    fn apply_nested_ciu() {
+        let inner0 = ciu(vec![leaf(), leaf()], vec![1, 1], 2).unwrap();
+        let inner1 = ciu(vec![leaf(), leaf()], vec![1, 1], 2).unwrap();
+        let plan = ciu(vec![inner0, inner1], vec![2, 2], 4).unwrap();
+        assert_snapshot!(trace_apply(&plan, ctx(0, 4)), @r"
```
jayshrivastava, line 317: nit: you don't differentiate between the two children CIUs

```
        assert_snapshot!(trace_apply(&plan, ctx(0, 4)), @r"
        CIU [ctx=0/4]
        CIU [ctx=0/2]
        Leaf [ctx=0/1]
        ");
        assert_snapshot!(trace_apply(&plan, ctx(1, 4)), @r"
        CIU [ctx=1/4]
        CIU [ctx=1/2] -> this looks identical to the one below
        Leaf [ctx=0/1]
        ");
        assert_snapshot!(trace_apply(&plan, ctx(2, 4)), @r"
        CIU [ctx=2/4]
        CIU [ctx=0/2]
        Leaf [ctx=0/1]
        ");
        assert_snapshot!(trace_apply(&plan, ctx(3, 4)), @r"
        CIU [ctx=3/4]
        CIU [ctx=1/2] -> this one
        Leaf [ctx=0/1]
        ");
```

src/common/recursion.rs
gabotechs: :thinking: I think "mutated" or "modified" is still not the right word, we are not mutating it really, we are just building a new one. Let me improve this.

src/common/recursion.rs
gabotechs: Absolutely. I wanted to defer this to another PR as this will imply some changes all across the codebase. This will probably make a good "good first issue" for random people on the internet.

src/common/recursion.rs
gabotechs: :thinking: `PlanningStage` and `ExecutionStage` are actually good names...

About the `TreeNodeRecursion::Jump`: this function never recurses into network boundaries, regardless of whether they are `LocalStage` or `RemoteStage`. Now that I see it it's a bit confusing, but even if we are returning here `TreeNodeRecursion::Continue`, we are not calling `apply_until_stop`, so we don't recurse.

This is being called within `TreeNodeRecursion::visit_children()`, and `TreeNodeRecursion::Continue` here does not mean "keep recursing down", like in the `.transform_down()` API, it means instead "keep visiting children", and we do want to move to the next children without recursing into NBs.

It's a bit unfortunate that upstream API decided to reuse the same `TreeNodeRecursion` for two different things...

src/common/recursion.rs
gabotechs: True! let's avoid it.

src/common/recursion.rs
gabotechs: Yeap! the only reason why this is `transform_down_up` is because that's how I implemented every method on the first place, but I'm realizing the `RefCell` + `transform_down_up` trick is only needed for implementing `transform_up` variants.

src/common/recursion.rs
gabotechs: In the `*_with_dt_ctx` methods we cannot traverse into network boundaries because the `DistributedTaskContext` can no longer be propagated there.

However, in the `*_with_task_count` methods we do can traverse into network boundaries, and we do. I just expanded a bit the doc comments to reflect this.

src/common/recursion.rs
gabotechs: The idea is to keep the API identical to upstream `.transform_down` and `.transform_up`, and those return `Transformed<Arc<dyn ExecutionPlan>>`, which uses `TreeNodeRecursion` under the hood.

That being said, it's true that here, if the user said `TreeNodeRecursion::Stop` we should be stopping instead of jumping, I'll fix that.

src/common/recursion.rs
gabotechs: Yeap, same as with upstream's `.transform_down` and `.transform_up`. This is expected though: if there is a `RemoteStage`, the input plan was already sent over the wire to another worker, so there's nothing to keep recursing on.

src/common/recursion.rs
gabotechs: :thinking: but I think it's not a big deal. Good coverage for CIU's is already provided in children_isolator_union.rs, and here I think the snapshots show the necessary info (the `[ctx=X/Y]`).

On one iteration I tried having a HashMap from `Arc<dyn ExecutionPlan>` pointer address to pre-order index so that it also gets rendered here, but I removed it because it seemed like an overkill that did not add a lot of value to the tests.

src/common/recursion.rs
gabotechs: Same as below, I think we should be fine. The extra mess of annotating each node with a pre-order index or something like that gets a bit messy and does not add that much extra value.

If you think it's worth bringing back the pre-order index annotation we can do it though.

src/common/recursion.rs
gabotechs: Yeah, testing with RemoteStage is not very interesting as there's nothing to recurse on. But it does not hurt to add a test for this.

gabotechs: Added some. They are not very interesting though, as remote network boundaries are always leafs, but it does not harm to have them.

jayshrivastava (approved):

PR #475, opened 2026-05-29
Make DistributedTaskContext Copy

### LINKED ISSUE #470: Implement `Copy` for `DistributedTaskContext`
DistributedTaskContext is a pretty simple struct that we keep cloning over an over:

```rust
struct DistributedTaskContext {
  task_index: usize,
  task_count: usize
}
```

It would be nice to implement `Copy` for it and remove manual `.clone()` calls from the codebase.

homa31:
## Summary
- `DistributedTaskContext` is two `usize` fields with no heap data, so deriving `Copy` is a strict refinement over `Clone` — every existing call site keeps working, and the explicit `.clone()` ceremony can go.
- Dropped the now-redundant `.clone()` calls at the internal sites that pass it by value:
  - `src/common/recursion.rs` — `apply_with_dt_ctx` / `transform_down_with_dt_ctx` no longer clone the context when recursing.
  - `src/worker/impl_execute_task.rs` — the `d_ctx_for_last` / `d_ctx_for_drop` rebindings (which only existed because `d_ctx` would otherwise be moved into the first `move` closure) are gone; both closures now capture by copy.
- `from_ctx() -> Arc<Self>` is unchanged — the `Arc` is there to satisfy `with_extension`, not because the inner struct is expensive.

Resolves https://github.com/datafusion-contrib/datafusion-distributed/issues/470

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -85,21 +85,19 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             ctx: DistributedTaskContext,
             f: &mut F,
         ) -> Result<TreeNodeRecursion> {
-            f(plan, ctx.clone())?.visit_children(|| {
+            f(plan, ctx)?.visit_children(|| {
                 if let Some(ciu) = plan.as_any().downcast_ref::<ChildrenIsolatorUnionExec>() {
                     // Just recurse to children that will actually get executed by this
                     // ChildrenIsolatorUnionExec.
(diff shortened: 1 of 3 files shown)


gabotechs (approved):
Looks good, thanks @homa31! There seems to be a formatting issue, but otherwise, it looks good to go.

gabotechs:
Thanks @homa31

PR #477, opened 2026-06-01
No eager buffering in network connections

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/477 <- you are here

---

Network boundaries in this project are currently breaking one assumption from upstream DataFusion:

`SendableRecordBatchStream`s yield record batches in two situations:
- If explicitly polled
- Eagerly on an spawned task triggered by the first poll

https://github.com/apache/datafusion/blob/d9ea38b95123159161c017840d3e6256e41988dd/datafusion/physical-plan/src/execution_plan.rs#L973-L988

Today, network boundaries pulling from remote sources are breaking this rule, because they start yielding `RecordBatches` over the network even if no poll has ever happened to the `SendableRecordBatchStream` returned by the network boundary.

This has two consequences:
1. Greater memory consumption, as data will get accumulated in the network boundaries while nobody is polling for it.
2. Greater speed on JOINs, as an artifact of eagerly buffering right sides even before they are ever polled

Consequence 2 is nice, but it should be delivered using standard upstream mechanisms:

https://github.com/apache/datafusion/blob/d9ea38b95123159161c017840d3e6256e41988dd/datafusion/common/src/config.rs#L695-L709

Not accidentally by how remote network boundaries work.

---

This PR makes it so that remote network boundaries only start the network stream on first poll, instead of on `.execute()` call, as stated by the `EvaluationType::Eager` docs.

### DIFF
--- a/src/worker/worker_connection_pool.rs
+++ b/src/worker/worker_connection_pool.rs
@@ -22,7 +22,7 @@ use datafusion::execution::memory_pool::{MemoryConsumer, MemoryReservation};
 use datafusion::physical_expr_common::metrics::{ExecutionPlanMetricsSet, MetricValue};
 use datafusion::physical_plan::metrics::{MetricBuilder, Time};
 use futures::stream::BoxStream;
-use futures::{Stream, StreamExt, TryFutureExt, TryStreamExt};
+use futures::{FutureExt, Stream, StreamExt, TryFutureExt, TryStreamExt};
 use http::Extensions;
 use pin_project::{pin_project, pinned_drop};
 use prost::Message;
@@ -160,6 +160,7 @@ struct RemoteWorkerConnection {
     cancel_token: CancellationToken,
     per_partition_rx: DashMap<usize, UnboundedReceiver<WorkerMsg>>,
 
+    first_poll_notify: Arc<Notify>,
     // Signals the demux task that buffered memory has been freed by a consumer.
     mem_available_notify: Arc<Notify>,
 
@@ -246,6 +247,9 @@ impl RemoteWorkerConnection {
         let mem_available_notify = Arc::new(Notify::new());
         let mem_available_notify_for_task = Arc::clone(&mem_available_notify);
 
+        let first_poll_notify = Arc::new(Notify::new());
+        let first_poll_notify_for_task = Arc::clone(&first_poll_notify);
+
         // Cancellation token allows us to stop the background task promptly when all partition
         // streams are dropped (e.g., when the query is cancelled).
         let cancel_token = CancellationToken::new();
@@ -255,6 +259,12 @@ impl RemoteWorkerConnection {
         // fan them out to the appropriate `per_partition_rx` based on the "partition" declared
         // in each individual record batch flight metadata.
         let task = SpawnedTask::spawn(async move {
+            tokio::select! {
+                biased;
+                _ = cancel.cancelled() => return,
+                _ = first_poll_notify_for_task.notified() => {}
+            }
+
             let mut client = match channel_resolver.get_worker_client_for_url(&url).await {
                 Ok(v) => v,
                 Err(err) => {
@@ -364,6 +374,7 @@ impl RemoteWorkerConnection {
             not_consumed_streams: Arc::new(AtomicUsize::new(per_partition_rx.len())),
             per_partition_rx,
             mem_available_notify,
+            first_poll_notify,
 
             // metrics stuff
             memory_reservation: memory_reservation_clone,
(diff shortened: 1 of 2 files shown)


src/worker/worker_service.rs
```diff
@@ -48,7 +48,7 @@ pub struct Worker {
 impl Default for Worker {
     fn default() -> Self {
         let cache = Cache::builder()
-            .time_to_idle(Duration::from_secs(60))
+            .time_to_idle(Duration::from_mins(10))
```
gabotechs: As we are no longer eagerly calling the remote streams, there's now a legitimate use case where the time gap between when the plan was sent in `CoordinatorChannel` until it's executed with `ExecuteTask` is significant.

This typically happens in JOINs, where the stage on the build side is called immediately, but the stage on the probe side will be delayed until the left side has been fully gathered. Having 10 mins here gives enough head room for the join to fully collect the build side until it starts executing the probe side.

src/worker/worker_service.rs
gene-bordegaray: nit: comment still 60 seconds

src/worker/worker_connection_pool.rs
gene-bordegaray: I think this comment is now stale about happening before init()

gene-bordegaray (approved):
The idea makes sense here and I understand the issue with buffering memory before on late consumers.

I am curious though do you see a case where this was actually beneficial. In the previous behavior we could be like prefetching / buffering data before it is actually request so it is ready to go right away when the consumer asks for it. I was thinking there may be cases where the time to first batch now for the consumer might be slower without this?

Could we bench this? But trust your judgement on the call 👍 

EDIT:

I am seeing your point about the JOINs and ya I think I agree here and this is more like a happy mistake from uninteded behavior

src/worker/worker_service.rs
gene-bordegaray: on second thought, could we actually make this a const or a config?

gabotechs: I tried introducing something like:
```rust

pub struct WorkerConfig {
    /// Time to idle for storing tasks in-memory.
    task_tti: Duration,
}

impl Default for WorkerConfig {
    fn default() -> Self {
        Self {
            task_tti: Duration::from_mins(10),
        }
    }
}
```

But then we enter in a rabbit hole of API design for allowing configuration on the `Worker` struct:
- Some things would be passed on `WorkerConfig`
- Some other things are currently configured with `Worker` methods (e.g., `Worker::with_max_message_size`)

This yields some inconsistency, and it would require some more thoughtful API design, so I'm inclined towards just making this a constant in this PR, and leave the configuration API design for a different one.

gabotechs:
> Could we bench this?

Doing that now. I do expect some performance impact (which is expected).

gene-bordegaray:
> > Could we bench this?
> 
> Doing that now. I do expect some performance impact (which is expected).

@gabotechs nice feel free to ping when some results are ready 😄

gabotechs:
This is the benchmark comparison against `main`, as expected, slower:

```
  ┌────────────┬────────┬────────┬───────────────────────────────┐
  │  Dataset   │  main  │ branch │            verdict            │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-H SF1  │ 17.1s  │ 24.3s  │ 1.42x slower ❌               │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-H SF10 │ 59.3s  │ 83.3s  │ 1.41x slower ❌               │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-DS SF1 │ 129.9s │ 171.5s │ 1.32x slower ❌               │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ ClickBench │ 151.2s │ 154.8s │ 1.02x slower ✖ (within noise) │
  └────────────┴────────┴────────┴───────────────────────────────┘
```

I'll try now running with `hash_join_buffering_capacity: 1_000_000`. With this, I'd expect it to be ~equally performant than `main`

gabotechs:
Now against `main` with `hash_join_buffering_capacity: 1_000_000`:

```
  ┌────────────┬────────┬────────┬───────────────────────────────┐
  │  Dataset   │  main  │ branch │            verdict            │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-H SF1  │ 17.1s  │ 16.3s  │ 1.05x faster ✔                │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-H SF10 │ 59.3s  │ 65.5s  │ 1.11x slower ✖                │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ TPC-DS SF1 │ 129.9s │ 117.4s │ 1.11x faster ✔                │
  ├────────────┼────────┼────────┼───────────────────────────────┤
  │ ClickBench │ 151.2s │ 171.0s │ 1.13x slower ✖                │
  └────────────┴────────┴────────┴───────────────────────────────┘
```

With hash join buffering enabled, things look as before

gene-bordegaray:
> Now against `main` with `hash_join_buffering_capacity: 1_000_000`:
> 
> ```
>   ┌────────────┬────────┬────────┬───────────────────────────────┐
>   │  Dataset   │  main  │ branch │            verdict            │
>   ├────────────┼────────┼────────┼───────────────────────────────┤
>   │ TPC-H SF1  │ 17.1s  │ 16.3s  │ 1.05x faster ✔                │
>   ├────────────┼────────┼────────┼───────────────────────────────┤
>   │ TPC-H SF10 │ 59.3s  │ 65.5s  │ 1.11x slower ✖                │
>   ├────────────┼────────┼────────┼───────────────────────────────┤
>   │ TPC-DS SF1 │ 129.9s │ 117.4s │ 1.11x faster ✔                │
>   ├────────────┼────────┼────────┼───────────────────────────────┤
>   │ ClickBench │ 151.2s │ 171.0s │ 1.13x slower ✖                │
>   └────────────┴────────┴────────┴───────────────────────────────┘
> ```
> 
> With hash join buffering enabled, things look as before

@gabotechs noice, thank you 🙇

gabotechs:
I think the CI failure is because of https://github.com/datafusion-contrib/datafusion-distributed/issues/480

PR #479, opened 2026-06-02
Refactor task spawner into QueryCoordinator and StageCoordinator

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/479 <- you are here

---

This is a pure refactor PR with a couple of implementation detail changes:

# Reducing the bloat of complex functions in already complicated parts of the codebase.

For this, two key structs are introduced:
- `QueryCoordinator`: scoped to a whole distributed query, it handles references to pieces of data global to a query's lifetime, like the `TaskContext`, the metrics, the `JoinSet` used for spawning tasks, etc... it's in charge also of building `StageCoordinator` instances, which are scoped per-stage instead of per-query.
- `StageCoordinator`: this is the old `CoordinatorToWorkerTaskSpawner`, but with some more methods that allow reusability and some better naming. It handles all the comms between workers and coordinator needed for driving a stage forward.

This allows reducing the bloat in `prepare_static_plan` and the future `prepare_dynamic_plan` functions.

# Ensuring a coordinator->worker channel is held active for as long as the `DistributedExec` node is executing the query on the coordinator.

For the static planner, this is a noop, as the previous model worked fine before, but this will become important in the future for the dynamic planner. In the dynamic planner, the plan can be set for some stages, but they might never reach execution, so instead of coupling the task entry cache invalidation to the task execution finish, it's coupled instead of the coordinator->channel lifetime.

This has one collateral effect: `WorkUnit` feeds can no longer rely on the global `coordinator->worker` EOS signal for ensuring that no further `WorkUnit` feed is going to be sent by the coordinator, so they need an explicit EOS message that signals that no further `WorkUnit`s will be received, even though the `coordinator->worker` channel will still be alive for a while.

# Add a `plan_for_viz` field in `DistributedExec`.

This is a new slot in `DistributedExec` that holds a reference of the plan that is supposed to be rewritten with metrics for visualization purposes.

Again, for the static planner this is a noop, because the plan meant for visualization is equal to the plan that arrived as child to `DistributedExec` on the first place. However, during dynamic planning, the plan that arrives to `DistributedExec` is not going to be the same as the final one after execution, so we need a slot for storing that final plan.


src/coordinator/latency_metric.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// DataFusion metrics system is pretty limited from an API standpoint. This intermediate struct
+/// bridges the gaps that are not satisfied by upstream API for measuring latency.
+pub(super) struct LatencyMetric {
+    max: Time,
+    avg: Time,
+    max_latency_micros: AtomicU64,
+    sum_latency_micros: AtomicU64,
+    count_latency_micros: AtomicU64,
+}
```
gabotechs, line 17: All the contents of this file are unchanged code that previously lived inside `task_spawner.rs`. I just extracted it to its own file as it's kind of isolated.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -51,27 +50,23 @@ pub async fn rewrite_distributed_plan_with_metrics(
...
 
-    let task_metrics = collect_plan_metrics(&prepared)?;
+    let head_stage = distributed_exec.head_stage()?;
+    let task_metrics = collect_plan_metrics(&head_stage)?;
 
     // Rewrite the DistributedExec's child plan with metrics.
     let dist_exec_plan_with_metrics = rewrite_local_plan_with_metrics(
         format.to_rewrite_ctx(0), // Task id is 0 for the DistributedExec plan
-        plan.children()[0].clone(),
+        distributed_exec.plan_for_viz()?,
```
gabotechs, line 65: During AQE, the plan meant for visualization is the one that gets constructed dynamically during execution, not necessarily the one that arrived on the first place to `DistributedExec`. This `plan_for_viz()` method ensures the appropriate plan is rewritten with metrics for visualization purposes.

src/coordinator/query_coordinator.rs
```diff
@@ -307,65 +298,141 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
...
+                routed_urls.len()
+            );
+        }
+        Ok(routed_urls)
+    }
 }
 
-/// DataFusion metrics system is pretty limited from an API standpoint. This intermediate struct
-/// bridges the gaps that are not satisfied by upstream API for measuring latency.
-pub(super) struct LatencyMetric {
```
gabotechs, line 325: Pretty unfortunate diff here:

- `LatencyMetric` was just moved to `latency_metric.rs` as-is
- New `keep_stream_alive` and `NotifyGuard` are introduced
- `CoordinatorToWorkerMetrics` was moved from the top of this file to the bottom as-is

src/coordinator/distributed.rs
```diff
@@ -26,22 +25,39 @@ use std::sync::Mutex;
...
-    plan: Arc<dyn ExecutionPlan>,
-    prepared_plan: Arc<Mutex<Option<Arc<dyn ExecutionPlan>>>>,
+    /// Initial [ExecutionPlan] present before execution.
+    /// - If the plan was distributed statically, this will be the final distributed plan with all
+    ///   the appropriate network boundaries in it.
+    /// - If the plan is going to be distributed dynamically during execution, this is the initial
+    ///   non-distributed plan.
+    base_plan: Arc<dyn ExecutionPlan>,
+    /// Resulting [ExecutionPlan] after execution ready for visualization purposes.
+    /// - If the plan was distributed statically, this is equal to the base plan.
```
jayshrivastava, line 35: nit: it's wrapped with the metrics wrappers technically, right?

src/coordinator/query_coordinator.rs
```diff
@@ -183,22 +170,26 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
...
-            futures::stream::once(async { msg }).chain(
-                UnboundedReceiverStream::new(coordinator_to_worker_rx).map(set_work_unit_send_time),
-            ),
+            futures::stream::once(async { msg })
+                .chain(UnboundedReceiverStream::new(coordinator_to_worker_rx))
+                .map(set_work_unit_send_time)
+                // Keep the request side of the channel open until the query ends: this tail emits
+                // no messages and only completes, once the `Notify` fires. Workers interpret this
+                // EOS of this stream as a query finished/aborted signal.
+                .chain(keep_stream_alive(Arc::clone(self.end_stream_notifier))),
```
jayshrivastava, line 187: Why do you want to keep this alive longer? It's ok for this stream to end when there's no more work units to send.

The `coordinator_to_worker_task` is owned by the `QueryCoordinator` (via the joinset which is passed to the `StageCoordinator`), so it's kept alive for the length of the query (and the `QueryCoordinator` isn't dropped until we've drained all the worker to coordinator messages).

src/worker/impl_coordinator_channel.rs
```diff
@@ -121,28 +121,51 @@ impl Worker {
...
+                                work_unit_senders.remove(&(id, partition));
+                                continue;
+                            }
+                        }
+                    }
+                    Inner::WorkUnitEos(_) => {
+                        // No further work unit message will be received here, so drop all the
+                        // sender sides so that receiver sides see an EOS upon draining the
+                        // remaining messages.
+                        let _ = work_unit_senders.take();
```
jayshrivastava, line 167: Why not return after this? We aren't going to get any more messages. Also why not just call `task_data_entries.invalidate(&key).await` here too.

src/worker/impl_coordinator_channel.rs
```diff
@@ -121,28 +121,51 @@ impl Worker {
...
+                    }
                 }
             }
+            #[allow(clippy::disallowed_methods)]
+            tokio::spawn(async move { task_data_entries.invalidate(&key).await });
```
jayshrivastava, line 172: We're already in a task, which has nothing to do. Might as well do this after dropping the senders?

src/coordinator/query_coordinator.rs
```diff
@@ -312,65 +305,141 @@ impl<'a> CoordinatorToWorkerTaskSpawner<'a> {
             }));
             Ok(TreeNodeRecursion::Continue)
         })?;
-        self.join_set.spawn(async move {
+
+        struct WorkUnitEosOnDrop(UnboundedSender<pb::CoordinatorToWorkerMsg>);
+        impl Drop for WorkUnitEosOnDrop {
+            fn drop(&mut self) {
+                let _ = self.0.send(pb::CoordinatorToWorkerMsg {
+                    inner: Some(Inner::WorkUnitEos(true)),
```
jayshrivastava, line 313: I still don't really know why we need the EoS signal on a remote worker. Previously, workers got this signal by seeing that the stream has ended.

src/coordinator/prepare_static_plan.rs
```diff
@@ -28,18 +18,9 @@ use std::sync::Arc;
 /// 4. Spawn a background task per worker that waits for the worker to finish and collects
 ///    its metrics into [DistributedExec::task_metrics] via the coordinator channel.
 pub(super) fn prepare_static_plan(
```
jayshrivastava, line 20: Could be a method on `&QueryCoordinator` now... but maybe leave that for a separate PR.

src/coordinator/distributed.rs
gabotechs: It will be wrapped during `rewrite_distributed_plan_with_metrics()`, but at this point no, this is just a normal plan.

src/coordinator/prepare_static_plan.rs
gabotechs: Yeap, in fact having it here as the first argument is almost identical to having it as a method.

The only thing that I don't love is the extra indentation leve, but besides that there's no difference.

src/coordinator/query_coordinator.rs
gabotechs: This is explained in the second point of the PR description:

We want to keep it alive for the length of the query, as with this PR we are coupling task data invalidation to the EOS signal of this stream. We want worker task data to get invalidated if this stream drops.

Before, task data invalidation was coupled to remote worker execution: once a worker calls `ExecuteTask` and finishes streaming Arrow data or drops the connection, the task data cache was invalidated.

This is fine for the static planner, where all the plans are set in workers at the same time and all remote `ExecuteTask` calls also happen ~ at the same time.

However, during dynamic planning, different parts of the plan will be set at random times during the query, and the remote `ExecuteTask` calls will also happen at random moments, so there might be a huge time difference between when a plan is set VS when it's executed.

The consequence of this is that, during dynamic planning, you'll have several plans that were set in workers, but that are never going to reach execution (e.g., a JOIN drops the right side at the middle of the query because it realizes the build side is empty), and therefore, you end up with a bunch of lingering task data entries that were never invalidated.

One of the scopes of this PR is to solve that issue by coupling task data invalidation to the EOS of `CoordinatorWorkerChannel` instead of `ExecuteTask`.

src/coordinator/query_coordinator.rs
gabotechs: This is explained in the second point in the description. I think that + https://github.com/datafusion-contrib/datafusion-distributed/pull/479#discussion_r3418798227 should clarify this, but let me know if you want me to dive a bit deeper.

src/worker/impl_coordinator_channel.rs
gabotechs: I think the explanation in https://github.com/datafusion-contrib/datafusion-distributed/pull/479#discussion_r3418798227 should answer this

src/worker/impl_coordinator_channel.rs
gabotechs: Same as above, https://github.com/datafusion-contrib/datafusion-distributed/pull/479#discussion_r3418798227 should answer this

src/worker/impl_coordinator_channel.rs
jayshrivastava: I get it now. I was confusing the `WorkUnitEos` with stream competition. It think this is pretty subtle and could use more comments.

Ex. "When the WorkUnitEos is seen, then drop the work unit senders. Note that this is different from stream completition.

Then on `task_data_entries.invalidate`, add a comment saying, when the stream is compelted, then we can drop the `task_data_entries`.

jayshrivastava (approved):

src/worker/impl_coordinator_channel.rs
gabotechs: Added a comment clarifying that

PR #485, opened 2026-06-08
Benchmarks housekeeping

gabotechs:
Just some housekeeping for the benchmarks:
- Update Ballista to 53
- Fix Spark TPCH query 15
- Use p50 instead of avg
- Bump default --iterations from 3 to 5
- Mirror the above behavior in local benchmarks
- Normalize aggregated bench time across runs with different --iterations value
- Use 12 instances by default, not 4
- Update Claude SKILL and README.md


gene-bordegaray (approved):
nice, saw the updated bench run too 👍

PR #486, opened 2026-06-08
Add cost model

gabotechs:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/486 <- you are here

---

Mostly rescued from:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/311

This PR is pretty big, but the whole purpose is to expose a very simple function:

```rust
pub(crate) fn calculate_cost(plan: &Arc<dyn ExecutionPlan>, cfg: &DistributedConfig) -> Result<Cost>;
```

While keeping everything else private to the rest of the codebase.

---

This PR introduces a cost model used in [Adaptive task count assignation](https://github.com/datafusion-contrib/datafusion-distributed/pull/432), that attributes a compute cost to a plan based on:
1. The `Statistics` available by normal DataFusion mechanisms (`ExecutionPlan::partition_statistics()`)
2. The compute time complexity of the different operators in a plan

The whole meat of this PR lives in the src/distributed_planner/statistics/compute_per_node.rs file: it's in charge of attributing a time complexity to the different `ExecutionPlan` implementations using big O notation. For example:

```rust
        let plan = plan_costs(r#"SELECT * FROM weather WHERE "MinTemp" > 5"#, 1).await;
        assert_snapshot!(plan, @r"
        O((out_Cols+Col0)) | FilterExec: MinTemp@0 > 5
         O(out_Cols) | DataSourceExec: file_groups={...}
        ");
```
In this example:
- `DataSourceExec` needs to allocate all the `RecordBatch`es for all the output columns, as it's the data source, so it's time complexity is `O(out_Cols)`, as its compute complexity linearly scales with the amount of columns it needs to allocate.
- `FilterExec` needs to:
   1. Perform a linear predicate evaluation over the `MinTemp` column (`O(Col0)`)
   2. Allocate full new `RecordBatch`es for all output columns based on rows filtered by the evaulated predicate (`O(out_Cols)`)

The tests in complexity_cpu.rs and complexity_memory.rs contains several examples of this for the different operators, which can go from pretty simple (`FilterExec`, `ProjectionExec`), to very complex (`HashJoinExec`).

Once the time complexity in big O notation is attributed to a node, doing a simple multiplication math of the columns appearing in the notation by the estimated `byte_size` of those columns (given by `Statistics`), gives the compute cost measured in bytes.


src/distributed_planner/distributed_config.rs
```diff
@@ -66,6 +66,10 @@ extensions_options! {
         /// budget will still be admitted (otherwise we would livelock), so the actual peak per
         /// connection is `worker_connection_buffer_budget_bytes + max_message_size`.
         pub worker_connection_buffer_budget_bytes: usize, default = 64 * 1024 * 1024
+        /// Distributed DataFusion relies on row count estimation in order to infer how many workers
+        /// should be used in serving the query. Some plans might not implement any kind of row count
+        /// estimation, and this parameter sets the default estimated row count for those plans.
+        pub default_estimated_row_count: Option<usize>, default = Some(0)
```
asolimando: Nit: I wonder if we shouldn't have `None` here and defer to the caller on what to do, so we can tell a very cheap plan from a case where we don't know

src/distributed_planner/statistics/compute_per_node.rs
```diff
@@ -0,0 +1,1029 @@
...
+                // `ilog2` panics on 0, which happens whenever the logged input has zero estimated
+                // bytes/rows (e.g. an empty or fully-pruned relation). Flooring at 1 makes log2
+                // contribute 0 there, i.e. sorting/merging nothing costs nothing.
+                n * m.checked_ilog2().unwrap_or(0) as usize
+            }
+            Self::Plus(n, m) => {
+                n.cost(output_stat, input_stats)? + m.cost(output_stat, input_stats)?
+            }
+            Self::Multiply(n, m) => {
+                n.cost(output_stat, input_stats)? * m.cost(output_stat, input_stats)?
```
asolimando: I think we should use `saturating_mul`/`saturating_add` instead of */+ for estimates as they can overflow

src/distributed_planner/statistics/plan_statistics.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Uses upstream DataFusion stats system with some small overrides.
+pub(super) fn plan_statistics(
+    node: &Arc<dyn ExecutionPlan>,
+    children_stats: &[Arc<Statistics>],
+    opts: &DistributedConfig,
+) -> Result<Arc<Statistics>> {
+    let mut stats = partition_statistics_with_children_override(node, None, children_stats)?;
+
+    // If rows are absent, be conservative and assume that all the rows from all the children
```
asolimando: The comment doesn't seem to match the code, as `Absent` children stats values are ignored and there is no rescaling of the result, right?

src/distributed_planner/statistics/plan_statistics.rs
```diff
@@ -0,0 +1,160 @@
...
+        };
+        stats.num_rows = Precision::Inexact(num_rows)
+    }
+
+    let schema = node.schema();
+
+    for (i, col_stats) in &mut stats.column_statistics.iter_mut().enumerate() {
+        let rows = stats.num_rows.get_value().unwrap_or(&0);
+
+        // If some of the NDVs are not present in one of the column-level stats, assume the
```
asolimando: It's hard to tell what is more reasonable without benchmarking/testing, but I am afraid defaulting to `nuw_rows` would be too much, another alternative would be to assume a normal distribution and interpolate missing values as proportionally closer to those of existing ones.

E.g., two stats: 100 rows, 2 NDV and 50 rows and Absent NDV, you could interpolate the missing value to 1.

We could leave it as-is for now and revisit when we have ways to see how much this moves and needle, and in which direction.

src/distributed_planner/statistics/complexity_cpu.rs
```diff
@@ -0,0 +1,1029 @@
...
+
+    // FilterExec: copies the output rows + evaluates the predicate over the input rows.
+    #[tokio::test]
+    async fn filter_exec() {
+        let plan = TestPlanBuilder::new()
+            .target_partitions(1)
+            .physical_plan(r#"SELECT * FROM weather WHERE "MinTemp" > 5"#)
+            .await;
+        assert_snapshot!(plan_costs(plan), @r"
+        O((out_Cols+Col0)) | FilterExec: MinTemp@0 > 5
```
asolimando, line 480: I think we should be printing the cost estimates instead of the cost model here. While this is interesting in a sense, it's basically written in the code, while the computation is the dynamic aspect we should be checking here.

src/distributed_planner/statistics/cost.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::sync::Arc;
+
+pub(crate) fn calculate_cost(
+    plan: &Arc<dyn ExecutionPlan>,
+    cfg: &DistributedConfig,
+) -> Result<usize> {
+    f(plan, cfg).map(|(cost, _stats)| cost)
+}
+
+fn f(plan: &Arc<dyn ExecutionPlan>, d_cfg: &DistributedConfig) -> Result<(usize, Arc<Statistics>)> {
```
asolimando: We briefly discussed offline the scalar vs vector cost, and I promised I would provide a little more details on this

We're collapsing everything into a single scalar (`usize` of bytes processed), but in a distributed engine network movement is often the deciding factor, and right now it's hard to have it move the needle: `BroadcastExec` is `Constant(1)` and I am under the impression that the data crossing a network boundary isn't charged as network at all.

I think we wouldn't later be able to write an effective CBO rule for broadcast-vs-partitioned or push-aggregation-below-exchange decision, as network doesn't influence the cost.

Trino for instance keeps cost as a multi-dimensional vector `<cpu, memory, network>` and collapses to a scalar only at comparison time with configurable weights (you have an example of how plans are displayed in [EXPLAIN doc](https://trino.io/docs/current/optimizer/cost-in-explain.html), and the definition of the vector in [CostComparator](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/cost/CostComparator.java)).

I think we could do the same with a small `<cpu, network, memory>` struct accumulated component-wise, although that would make it a little harder to use the `Complexity` abstraction.

src/distributed_planner/statistics/compute_per_node.rs
```diff
@@ -0,0 +1,1029 @@
...
+use datafusion::physical_plan::sorts::sort::SortExec;
+use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+use datafusion::physical_plan::union::{InterleaveExec, UnionExec};
+use datafusion::physical_plan::windows::{BoundedWindowAggExec, WindowAggExec};
+use datafusion::physical_plan::{ExecutionPlan, Partitioning};
+use std::fmt::{Debug, Formatter};
+use std::sync::Arc;
+
+#[derive(Clone, PartialEq, Eq)]
+pub(super) enum Complexity {
```
asolimando: Complexity feels like a nice abstraction as conceptually that's how we conceive the cost of operators, I traditionally see free-form formulas as they allow for better fine-tuning when things stabilize, and adding small deltas or tweaking coefficients might matter (it gets similar to fitting a function over data points, which is easier with a more traditional equation/formula than this notation).

asolimando (commented):
I did a first pass on the PR, it's surely heading in the right direction but there are a couple of choices around how to deal with missing values and how to represent cost that will probably benefit from more discussion.

src/distributed_planner/statistics/plan_statistics.rs
gabotechs: :thinking: I think the comment is incomplete, what it states is true just in case children stats are not absent.

I'm wondering if all these overrides with defaults and stuff are actually worth it. This is inherited from https://github.com/datafusion-contrib/datafusion-distributed/pull/311, and there, as soon as you hit an `Absent` row count, then this `Absent` is propagated through the rest of the plan. 

However, for AQE this is not really a problem, as at each stage we "reset" the stats with real runtime stats, so even if there's just a couple of nodes with `Absent` row count, it does not really propagate.

I'm going to try deleting all these defensive defaults, and see how it goes.

src/distributed_planner/statistics/compute_per_node.rs
gabotechs: This is actually intended to represent free-form formulas, just with operators narrowed down to representing time complexity.

Looking at other symbolic math libraries in the open-source, they definitely seem an overkill for what we actually need here.

I'll investigate a bit more if there's an alternative.

src/distributed_planner/statistics/compute_per_node.rs
gabotechs: :+1: good call

src/distributed_planner/statistics/complexity_cpu.rs
gabotechs: It turns out that these tests were super useful, not really as tests, but as an iteration loop tool for LLMs to get the symbolic expressions right for the different SQL statements.

When you tell an LLM to implement the `Complexity` symbolic function for a specific operator (e.g., a HashJoin), it will use these tests for ensuring that the appropriate formulas are being applied.

Printing the actual cost estimate here is not really going to provide much coverage, as one you have the cost expression, the actual cost value is just one multiplication away, so we would be in a similar situation regarding coverage.

src/distributed_planner/statistics/cost.rs
gabotechs: :+1: I think evaluating cpu, network and memory separately is worth doing. It'll enlarge this PR, but I think it can be worth it. I'll try that.

src/distributed_planner/statistics/plan_statistics.rs
gabotechs: :thinking: I think I'm gravitating more towards just removing all this, and rather than inventing defaults, let `Absent` stats be just absent.

src/distributed_planner/distributed_config.rs
gabotechs: I'm going to try to delete this at all

src/distributed_planner/statistics/plan_statistics.rs
gabotechs: Ok, I've tried, and unfortunately results are worst...

This is definitely kicking in and providing some sane defaults for when we have absent stats. I'll recover it, but just the bits that are strictly necessary.

src/distributed_planner/statistics/compute_per_node.rs
gabotechs: Saw this crate https://crates.io/crates/big_o, but I think it's a bit overkill, and I think it's worth having something more domain specific here.

My take is that it's worth keeping the `enum Complexy` expression tree for modeling BigO complexity, is there any formula that you think cannot be model by this? for example:
- A small delta can be modeled like this: `Plus(Linear::OutputColumns, Constant(100))`
- A coeficient can be modeled like this: `Multiply(Linear::OutputColumns, Constant(10))`

src/distributed_planner/statistics/cost.rs
gabotechs: Done! let me know what you think

src/distributed_planner/statistics/plan_statistics.rs
gabotechs: Ok, could not remove this, we'll actually need something here.

How about this: let's just have a constant that we can tweak and iterate on. With upstream improvements in stats propagation, we should more likely converge into a situation where this defensive code is never exercised.

src/distributed_planner/statistics/compute_per_node.rs
asolimando: I agree that those crates would be an overkill w.r.t. what we need here. I was mostly referring to the `Constant` being an `usize` and preventing the use of fractional values for better fine-tuning. If we can support that, I think we should be fine.

src/distributed_planner/statistics/complexity_cpu.rs
asolimando: I didn't think of LLMs as consumers, so it's OK to keep them. It's true that the final cost estimate is easy to compute, but it's not only the formula, it's also the statistics propagation that fuels the variables of the formula, I think it's informative to have those printed, but it's not a blocker at this stage.

src/distributed_planner/statistics/plan_statistics.rs
asolimando: Thanks for trying, already reducing the surface and keep only the significant part helps a lots!

src/distributed_planner/statistics/cost.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub(crate) cpu: usize,
+    pub(crate) memory: usize,
+    pub(crate) network: usize,
+}
+
+impl AddAssign for Cost {
+    fn add_assign(&mut self, rhs: Self) {
+        self.cpu += rhs.cpu;
+        self.memory += rhs.memory;
+        self.network += rhs.network;
```
asolimando: We should use `saturing_add` here too as at large scales we might overflow here too

src/distributed_planner/statistics/plan_statistics.rs
```diff
@@ -0,0 +1,153 @@
...
+        // If the per-column byte size stats are not present, estimate the byte size based on the
+        // data type and the row count.
+        let Some(dt) = schema.fields.get(i).map(|v| v.data_type()) else {
+            return plan_err!("Field with index {i} not present in schema: {schema:?}");
+        };
+
+        // If it turns out that we do not have `byte_size` stats, but we do have an estimated number
+        // of rows, do a best-effort in trying to infer the byte size for each column.
+        if matches!(col_stats.byte_size, Precision::Absent) {
+            col_stats.byte_size = Precision::Inexact(default_bytes_for_datatype(dt) * rows)
```
asolimando:
```suggestion
            col_stats.byte_size = Precision::Inexact(default_bytes_for_datatype(dt).saturating_mul(*rows))
```

src/distributed_planner/statistics/plan_statistics.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    // If bytes are absent, let's just infer them based on the schema and the
+    // number of rows.
+    if matches!(stats.total_byte_size, Precision::Absent) {
+        let mut total_byte_size = 0;
+        for col_stats in &stats.column_statistics {
+            total_byte_size += col_stats.byte_size.get_value().unwrap_or(&0);
```
asolimando:
```suggestion
            total_byte_size = total_byte_size.saturating_add(*col_stats.byte_size.get_value().unwrap_or(&0));
```

src/distributed_planner/statistics/complexity_network.rs
```diff
@@ -0,0 +1,13 @@
+use crate::NetworkBoundaryExt;
+use crate::distributed_planner::statistics::complexity::{Complexity, LinearComplexity};
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::Arc;
+
+/// Calculates the memory cost for the provided node, without recursing into children.
```
asolimando:
```suggestion
/// Calculates the network cost for the provided node, without recursing into children.
```

src/distributed_planner/statistics/plan_statistics.rs
asolimando: `FALLBACK_NDV_RATIO` is a pragmatic solution for now, we can always refine with interpolation later if needed, as the change will be local, but hopefully if we keep improving statistics propagation this will not needed at all.

The comment seems stale now though, we don't re-use the number of rows anymore, maybe something like "If a column's NDV is absent, fall back to a fraction of the row count". WDYT?

src/distributed_planner/statistics/cost.rs
```diff
@@ -0,0 +1,47 @@
...
+    let mut acc_cost = Cost::default();
+    for child in children {
+        let (cost, child_stat) = f(child)?;
+        acc_cost += cost;
+        child_stats.push(child_stat);
+    }
+
+    let stats = plan_statistics(plan, &child_stats)?;
+    let cpu = complexity_cpu(plan);
+    acc_cost.cpu += cpu.cost(&stats, &child_stats).unwrap_or(0);
```
asolimando: Here too we should use the `saturating_add` variant to be safe. I am also wondering if we should strive to preserve `Absent` at least for cases where there is no value to be used, to make it recognizable as different from 0 (with the split into three components it's fine to have 0 network for fully "local" subplans, so I think that this might now matter even more, before it was almost impossible to have 0 memory for real)

src/distributed_planner/statistics/complexity_network.rs
```diff
@@ -0,0 +1,13 @@
...
+pub(super) fn complexity_network(node: &Arc<dyn ExecutionPlan>) -> Complexity {
+    if node.is_network_boundary() {
+        return Complexity::Linear(LinearComplexity::AllOutputColumns);
```
asolimando:
```suggestion
        let bytes = Complexity::Linear(LinearComplexity::AllOutputColumns);
        // A broadcast boundary replicates its input to every consumer task, so the bytes that
        // cross the network scale with the consumer task count, unlike shuffle/coalesce which move
        // each row once. The broadcast boundary's only child is always a BroadcastExec (enforced
        // by NetworkBroadcastExec::try_new), which carries the replication factor.
        if let Some(bcast) = node
            .children()
            .first()
            .copied()
            .and_then(|c| c.downcast_ref::<BroadcastExec>())
        {
            return bytes.multiply(Complexity::Constant(bcast.consumer_task_count()));
        }
        return bytes;
```

(missing import of `BroadcastExec`)

Verified this locally on the PR branch with Claude: a broadcast join with `consumer_tasks=3` totals `network=22692` today vs `41724` once the boundary is charged `* consumer_task_count` (the broadcast term goes `9516` to `28548`, the rest unchanged).

So right now the dimension can't tell broadcast from a single-copy transfer of the same bytes, which is the main reason for having it, and I expect similar network-related decisions to be the dominating factor in this distributed setting.

This also makes the case for printing per-node cost as expressed in another comment: with it, "broadcast wins" and "shuffle wins" become two snapshot tests asserting the network column, I had to manually print totals to be able to verify.

All of the above is not a blocker, the impact is very local and we can refine later without jeopardizing the design, but I think it will quickly allow to have better readings of the benchmarks' results.

src/distributed_planner/statistics/complexity_cpu.rs
```diff
@@ -0,0 +1,874 @@
...
+use datafusion::physical_plan::repartition::RepartitionExec;
+use datafusion::physical_plan::sorts::sort::SortExec;
+use datafusion::physical_plan::sorts::sort_preserving_merge::SortPreservingMergeExec;
+use datafusion::physical_plan::union::{InterleaveExec, UnionExec};
+use datafusion::physical_plan::windows::{BoundedWindowAggExec, WindowAggExec};
+use datafusion::physical_plan::{ExecutionPlan, Partitioning};
+use std::sync::Arc;
+
+/// Calculates the CPU cost for the provided node, without recursing into children.
+pub(super) fn complexity_cpu(node: &Arc<dyn ExecutionPlan>) -> Complexity {
```
asolimando, line 27: This file was a great read to understand where the complexity of each operator really lies! Most of this useful knowledge applies identical to core DF, so when statistics will be mature enough there, we will be able to push CBO a little more, it will be possible to translate the tradeoffs you are detailing here into explicit choices between plans later.

asolimando (commented):
Thanks for this new iteration, Gabriel, the split between the three `complexity_*` files reinforces the need to keep the three dimensions separate, not only to write a better cost model, but also to tame complexity for maintainers.

I think the only must-have before merge is switching the remaining +/* on estimates to `saturating_add`/`saturating_mul`.

The broadcast fan-out and printing per-node costs I'd put as (very) nice to have. Both are additive and can land in a follow-up without touching the design, so no pressure to fit them here.

Everything else (the NDV ratio, the stale comments, naming) is minor.

Overall it's in great shape, thanks for the quick turnaround on the vector split.

src/distributed_planner/statistics/compute_per_node.rs
gabotechs: That's reasonable, I'll do that 👍

src/distributed_planner/statistics/cost.rs
gabotechs: :thinking: that makes sense, but I'd not know how to do that...

If everything is absent, that would be easy: `acc_cost.cpu` = Precision::Absent, and that's it. But I bet that the vast majority of the times a node has absent `acc_cost.cpu`, there's also going to be other nodes in the plan that do not contain absent stats, and how could we reconcile those?

src/distributed_planner/statistics/complexity_network.rs
gabotechs: Indeed, you are completely right! I'll pull in the code you suggested

src/distributed_planner/statistics/cost.rs
asolimando: One option is to wrap the measurements in `Precision` like DF stats, but with our own combine rule: an `Absent` operand contributes `0` to the magnitude and just downgrades the result to `Inexact`. 

We can't use `Precision::add`, which absorbs, but we can keep using the current `unwrap_or(0)`, by just detecting if any operand is `Absent`, and downgrading to `Inexact`.

A more complex option is interpolation, but we can lean on the runtime feedback loop to see whether that's even necessary.

asolimando (approved):
LGTM, one last point to address in https://github.com/datafusion-contrib/datafusion-distributed/pull/486#discussion_r3497443201 but I think the direction is clear now and all major comments have been addressed!

src/distributed_planner/statistics/cost.rs
gabotechs: Make sense. Now all the `Cost` fields are expressed in terms of `Precision`

gabotechs:
Thanks @asolimando for the thorough review! this looks much better with your suggestions

PR #488, opened 2026-06-08
Fix benchmarks build: remove nonexistent with_distributed_dynamic_task_count call

EdsonPetry:
## Summary

- Commit `0d4e239` (Benchmarks housekeeping) introduced a `dynamic: bool` field in `RunOpt` and called `.with_distributed_dynamic_task_count(self.dynamic)?` in `RunOpt::run_local`
- That method was never implemented in `distributed_ext.rs`, so the `dfbench` binary has been broken since that commit
- This PR removes the two dead lines (the struct field and the call site)

## Verification

Checked out to `0d4e239` and confirmed the compile error:
```
error[E0599]: no method named `with_distributed_dynamic_task_count` found for struct `SessionStateBuilder`
  --> benchmarks/src/run.rs:212:14
```

After this fix, `cargo build -p datafusion-distributed-benchmarks --bin dfbench` compiles cleanly.

### DIFF
--- a/benchmarks/src/run.rs
+++ b/benchmarks/src/run.rs
@@ -105,10 +105,6 @@ pub struct RunOpt {
     #[structopt(long, default_value = "0")]
     max_tasks_per_stage: usize,
 
-    /// Activate dynamic task count
-    #[structopt(long)]
-    dynamic: bool,
-
     /// Number of iterations of each test run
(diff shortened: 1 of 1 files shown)


gabotechs (approved):
Good catch! thanks @EdsonPetry

PR #490, opened 2026-06-08
fix: await async metric delivery in metrics collection tests

### LINKED ISSUE #487: Flaky test_metrics_collection_e2e_1 test
This test seems to constantly flake, for example:

- https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/27137166880/job/80101529756

sjhddh:
## Which issue does this PR close?

Closes #487.

## Rationale for this change

`test_metrics_collection_e2e_1` flakes intermittently in CI, panicking on `Option::unwrap()` of a `None` value at the task-key lookup in `run_metrics_collection_e2e_test` (e.g. [this run](https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/27137166880/job/80101529756)).

Per-task metrics are delivered to the coordinator asynchronously over the `WorkerToCoordinator` side channel by a background task, which may still be in flight after `execute_plan` returns. The test asserted on the collected metrics immediately, racing that delivery, so a task key was occasionally still absent from the `MetricsStore` when the assertion ran.

## What changes are included in this PR?

- Add a `wait_for_task_metrics` test helper that waits on the `MetricsStore` watch channel until every expected task key is present before asserting, bounded by a 10s timeout that reports any still-missing keys. This mirrors the existing `DistributedExec::wait_for_metrics` pattern but waits on exactly the task keys the assertions check.
- Use it in `run_metrics_collection_e2e_test` (covers `_e2e_1` and `_e2e_2`).
- `test_metrics_collection_with_limit_causing_early_stream_termination` had the same race papered over with a single `yield_now().await`; replace that with the same deterministic wait.

## Are there any user-facing changes?

No. Test-only change.

### DIFF
--- a/src/metrics/task_metrics_collector.rs
+++ b/src/metrics/task_metrics_collector.rs
@@ -28,7 +28,8 @@ mod tests {
     use datafusion::arrow::record_batch::RecordBatch;
     use futures::StreamExt;
 
-    use crate::coordinator::DistributedExec;
+    use crate::TaskKey;
+    use crate::coordinator::{DistributedExec, MetricsStore};
     use crate::test_utils::in_memory_channel_resolver::{
         InMemoryChannelResolver, InMemoryWorkerResolver,
     };
@@ -38,6 +39,7 @@ mod tests {
     };
     use crate::test_utils::session_context::register_temp_parquet_table;
     use crate::{DistributedExt, SessionStateBuilderExt};
+    use datafusion::common::HashSet;
     use datafusion::execution::{SessionStateBuilder, context::SessionContext};
     use datafusion::prelude::SessionConfig;
     use datafusion::{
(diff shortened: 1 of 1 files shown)


src/metrics/task_metrics_collector.rs
```diff
@@ -172,14 +196,12 @@ mod tests {
             DisplayableExecutionPlan::new(plan.as_ref()).indent(true)
         );
 
+        let metrics_store = dist_exec.metrics_store.as_ref().unwrap();
+        wait_for_task_metrics(metrics_store, &expected_task_keys).await;
```
gabotechs: There's no need to create an extra method coupled to implementation details of `MetricsStore` that ideally should be hidden here.

There's a method that does this already in `DistributedExec`, we should be able to just use it, something as simple as:

```rust
dist_exec.wait_for_metrics().await.unwrap();
```

sjhddh: Good call, done. Dropped the helper and switched both tests to `dist_exec.wait_for_metrics().await` — it derives the expected task keys from the plan itself, so the test no longer reaches into `MetricsStore`. (It returns `()`, so no `unwrap` needed.)

gabotechs (approved):
Thanks @sjhddh! this looks good. As soon as CI passes we are good to pull this in

Issue #313, opened 2026-01-26
Consolidate helper functions for testing plans


gabotechs:
Right now, we have several functions with overlapping intentions in our tests. They mainly generate plans in different ways.

It would be nice to do a small refactor and consolidate them in something more reusable and ergonomic. I'd like to propose a builder pattern for that, but I have not empirically verified if that's what yields the best results.

kentkr:
take!

kentkr:
@gabotechs you can close this

PR #491, opened 2026-06-08
#313: consolidate test plan helpers

kentkr:
This solves #313 by consolidating helpers for testing plans. It uses two new structs `TestPlanBuilder` and `TestPlan` to encapsulate the configuration of a context and creating a subsequent query plan. 

There are two things I want to mention, neither seem super problematic but they're odd:
1) [`register_parquet_tables`](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/test_utils/plans.rs#L111) is ran when creating the plan, which isn't what I originally planned (originally expected it to be in `TestPlanBuilder::build`). But the [SET query here](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/distributed_planner/distributed_query_planner.rs#L497) only takes effect if tables are registered after the SET query is called. This surprises me but I don't know enough to say if it's 1) expected, 2) a bug in datafusion, 3) or the plan is incorrect.
    - fyi the settings are confirmed to be updated by the query. And if `target_partitions` is set in the rust code the plan is correct. That cancels out 1 and 3.
2) many of the tests rely on [`display_plan_ascii`](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/stage.rs#L201). However some tests don't seem to need it ([eg](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/distributed_planner/distributed_query_planner.rs#L216)). Looks like the func has two if conditions, one returns ascii, the other doesnt. This is probably just a confusing name issue but wanted to raise it.


src/distributed_planner/distributed_query_planner.rs
```diff
@@ -150,23 +144,24 @@ mod tests {
...
-        })
-        .await;
-        assert_snapshot!(plan, @"DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[MinTemp, MaxTemp, Rainfall, Evaporation, Sunshine, WindGustDir, WindGustSpeed, WindDir9am, WindDir3pm, WindSpeed9am, WindSpeed3pm, Humidity9am, Humidity3pm, Pressure9am, Pressure3pm, Cloud9am, Cloud3pm, Temp9am, Temp3pm, RainToday, RISK_MM, RainTomorrow], file_type=parquet");
+        let physical_plan_ascii = TestPlanBuilder::default()
+            .build()
+            .physical_plan_as_ascii(query, false)
+            .await;
+        assert_snapshot!(physical_plan_ascii, @"DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[MinTemp, MaxTemp, Rainfall, Evaporation, Sunshine, WindGustDir, WindGustSpeed, WindDir9am, WindDir3pm, WindSpeed9am, WindSpeed3pm, Humidity9am, Humidity3pm, Pressure9am, Pressure3pm, Cloud9am, Cloud3pm, Temp9am, Temp3pm, RainToday, RISK_MM, RainTomorrow], file_type=parquet");
```
gabotechs, line 152: I see that in some tests the `.distributed_planner()` is not being enabled. We'd want to have it eneabled in every test. Even in the ones you see we are not actually distributing the plan, we still want it enable, as we still want to test those cases where the distributed planner decides that it's not worth distributing.

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -218,11 +215,12 @@ mod tests {
...
-            b.with_distributed_worker_resolver(InMemoryWorkerResolver::new(0))
-        })
-        .await;
-        assert_snapshot!(plan, @r"
+        let physical_plan_ascii = TestPlanBuilder::default()
+            .num_workers(0)
+            .build()
+            .physical_plan_as_ascii(query, false)
+            .await;
+        assert_snapshot!(physical_plan_ascii, @r"
```
gabotechs, line 219: Same as above, I'd even just remove the `distributed_planner()` option at all and have it always enabled.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -571,7 +566,7 @@ mod tests {
         let query = r#"
         SELECT * FROM weather
         "#;
-        let annotated = sql_to_annotated(query).await;
+        let annotated = sql_to_annotated_plan(query, 4, 4, false, None).await;
```
gabotechs: Looking at this line, it's a bit hard to know what are all those values (4, 4, false, None) without navigating back to the `sql_to_annotated_plan`.

I think what you did on `distributed_query_planner.rs` about having the `TestPlanBuilder` being used explicitly in each test is great, and it would be nice to have the same pattern here, removing the `sql_to_annotated_plan` function at all.

src/distributed_planner/partial_reduce_below_network_shuffles.rs
```diff
@@ -102,20 +102,15 @@ pub(crate) fn partial_reduce_below_network_shuffles(
...
-    use datafusion::prelude::{SessionConfig, SessionContext};
 
     #[tokio::test]
     async fn grouped_aggregation() {
-        let explain =
-            sql_to_explain(r#"SELECT "RainToday", COUNT(*) FROM weather GROUP BY "RainToday""#)
-                .await;
-        assert_snapshot!(explain, @r"
+        let query = r#"SELECT "RainToday", COUNT(*) FROM weather GROUP BY "RainToday""#;
+        let physical_plan_ascii = sql_to_physical_plan_ascii(query, true).await;
```
gabotechs: For consistency with the rest of the tests, it might even be good to just use the plain `TestPlanBuilder` directly here without the intermediate `sql_to_physical_plan_ascii`.

It'll be a bit more verbose, but the fact that it makes all the tests across the codebase read equally would be worth it IMO.

src/test_utils/plans.rs
```diff
@@ -87,89 +92,205 @@ fn find_input_stages(plan: &dyn ExecutionPlan) -> Vec<&Stage> {
...
+/// the order distributed settings must be applied in. [`TestPlanBuilder::build`]
+/// resolves that order: config, then default features, then the distributed planner
+/// and any [DistributedConfig] modifiers.
 #[cfg(test)]
-#[derive(Clone, Copy, Debug)]
-pub(crate) struct TestPlanOptions {
-    pub(crate) target_partitions: usize,
-    pub(crate) num_workers: usize,
-    pub(crate) broadcast_enabled: bool,
+pub(crate) struct TestPlanBuilder {
```
gabotechs, line 129: :rocket: this is very cool! what you did here is definitely working

src/test_utils/plans.rs
```diff
@@ -87,89 +92,205 @@ fn find_input_stages(plan: &dyn ExecutionPlan) -> Vec<&Stage> {
...
+    target_partitions: Option<usize>,
+    num_workers: Option<usize>,
+    distributed_planner: bool,
```
gabotechs, line 132: I think this is the bool flag that we would like to have always enabled.

I actually don't think we have any test where we would not want this set to `true`

gabotechs (commented):
This is looking very good! just left some minor comments, but it's overall a pretty good improvement. Thanks @kentkr!

> many of the tests rely on [display_plan_ascii](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/stage.rs#L201). However some tests don't seem to need it ([eg](https://github.com/kentkr/datafusion-distributed/blob/b284aec40f75239d28ffc963a0ecc4f4f84888f4/src/distributed_planner/distributed_query_planner.rs#L216)).

`display_plan_ascii` will only use ASCII graphics for distributed plans. If the plan does not get distributed (there arelegitimate cases for this to happen even when distribution is enabled), then the default DataFusion visualization for plans is used, which happens to not use any ASCII graphics.

Always relying on `display_plan_ascii` is the correct choice here 👍

src/distributed_planner/insert_broadcast.rs
```diff
@@ -215,8 +217,14 @@ mod tests {
         FROM weather a INNER JOIN weather b
         ON a."RainToday" = b."RainToday"
         "#;
-        let physical_plan = sql_to_physical_plan(query, 1, 4).await;
-        assert_snapshot!(physical_plan, @r"
+        let physical_plan_string = TestPlanBuilder::new()
+            .target_partitions(1)
```
gabotechs, line 221: :thinking: it looks like we were inconsistently setting some parameters with `SET` statements VS with Rust code.

It would be nice to unify this a bit:

There are some SET statement we could be using (e.g., `datafusion.execution.target_partitions`, `distributed.broadcast_joins`, `distributed.cardinality_effect_task_scale_factor`, `distributed.partial_reduce`), but while writing this message... I'm realizing that rather than using them, it might be better to rely on your `TestPlanBuilder` always, even if we could be using `SET` statements.

The issue is that there are things that we cannot set with `SET` statements (e.g., number of workers, task estimator, information schema enabled), and therefore, for consistency, it might be better to just always configure things in the `TestPlanBuilder`.

src/distributed_planner/inject_network_boundaries.rs
kentkr: glad you said something about this, i was thinking it but wasn't totally convinced. I'll mimic `distributed_query_planner.rs`

src/distributed_planner/insert_broadcast.rs
kentkr: Sounds good I'll drop the sets.

Two questions then:
1) from a datafusion perspective, should `register_parquet_tables` be moved to when the context gets created? (`TestPlanBuilder::build_context`)
2) i'm curious why certain things can't be set. Is it a datafusion thing?

kentkr: that also means we can drop the multiple query loop

https://github.com/kentkr/datafusion-distributed/blob/e72ec86622a9f3db05704eccaf43212e9897071e/src/test_utils/plans.rs#L107-L111

gabotechs: > from a datafusion perspective, should register_parquet_tables be moved to when the context gets created?

Good call. Yeah I think so. In practice it would work either way, but I think it's cleaner to do as you say here and call `register_parquet_tables` to when the context is created.

> i'm curious why certain things can't be set. Is it a datafusion thing?

SET statements in DataFusion allow to set configuration that can be expressed with simple primitive types, like an `int` or  a `string` (e.g., `target_partitions` or `enable_hash_joins`). But there are other things that cannot be expressed so easily, and instead we are required to provide actual `struct`s that implement some `trait`s. For those, we need to pass them with actual code.

kentkr:
All set. A small recap of the changes I made:
1) `distributed_planner` is now the default. but i kept the method to set it bc `inject_network_boundaries` [wants it to be false](https://github.com/kentkr/datafusion-distributed/blob/f0f08e64ade15aae1f36a3a1abad81bdf35b99b8/src/distributed_planner/inject_network_boundaries.rs#L572). I'm not sure exactly why but it's something related to [`annotate_test_plan`](https://github.com/kentkr/datafusion-distributed/blob/f0f08e64ade15aae1f36a3a1abad81bdf35b99b8/src/distributed_planner/inject_network_boundaries.rs#L1184) wanting a single node plan
2) the helper funcs with magic args are removed
3) `SET` queries were removed and new builder methods were added if needed

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -150,23 +144,25 @@ mod tests {
...
-        let plan = sql_to_explain(query, |b| {
-            b.with_distributed_worker_resolver(InMemoryWorkerResolver::new(3))
-        })
-        .await;
-        assert_snapshot!(plan, @"DataSourceExec: file_groups={3 groups: [[/testdata/weather/result-000000.parquet], [/testdata/weather/result-000001.parquet], [/testdata/weather/result-000002.parquet]]}, projection=[MinTemp, MaxTemp, Rainfall, Evaporation, Sunshine, WindGustDir, WindGustSpeed, WindDir9am, WindDir3pm, WindSpeed9am, WindSpeed3pm, Humidity9am, Humidity3pm, Pressure9am, Pressure3pm, Cloud9am, Cloud3pm, Temp9am, Temp3pm, RainToday, RISK_MM, RainTomorrow], file_type=parquet");
+        let physical_plan_ascii = TestPlanBuilder::default()
+            .build()
+            .await
+            .physical_plan_as_ascii(query, false)
+            .await;
```
gabotechs, line 151: 🤔 I have to say I don't love having two `.await`s here from an ergonomic standpoint.

I think this might be the reason why we were registering the parquet files during physical plan creation on the first place... but i did not remember when you asked.

Sorry for the confusion, but could we put back the parquet file registering to the physical plan creation so that we only have one `.await` in the helpers 🙏 ?

gabotechs: Actually, here's an even better idea:

I'm noticing that `.physical_plan_as_ascii()` is always preceded by a `.build()` call, which makes sense as the `TestPlan` always needs to be built for someone to call `.physical_plan_as_ascii()` on the first place.

What about incorporating `.physical_plan()`, `.physical_plan_as_string()` and `.physical_plan_as_ascii()` also to `TestPlanBuilder` so that we can move from to something like this in the tests:

```rust
       let physical_plan_ascii = TestPlanBuilder::default()
            .physical_plan_as_ascii(query, false)
            .await;
```

gabotechs: That will allow reducing the amount of `.build().await` calls in the tests that do not contribute a lot of information to what's happening in each test case.

gabotechs (approved):
Just left one suggestion that would allow us to trim some boilerplate lines in each test, but otherwise this looks good to me. Nice work @kentkr!

src/distributed_planner/distributed_query_planner.rs
kentkr: @gabotechs no worries, it's a good suggestion. it makes sense I'll update it. before i do just know getting `TestPlan` from `build` will still require an await. I took a long look and can't find a decent way to avoid it

the reason `register_parquet_tables` is in `build` is bc that's where the ctx get's created. otherwise there's technically `TestPlan.ctx` without registered tables until you call one of its methods. Not a fan of that. In fact I think my original [impl missed that and had a bug here](https://github.com/kentkr/datafusion-distributed/blob/53fe40e8aedd6a6f4bc16530083c539505988fd5/src/test_utils/plans.rs#L132-L134)

kentkr: @gabotechs thanks for approving, but can you give it one more look?

gabotechs: > before i do just know getting TestPlan from build will still require an await

Yeah, this is fine, those tests that are just building the plan without displaying it will also need to `.await`, but this is natural as the parquet files also need to be registered there.

Taking another look now

gabotechs (approved):
💯 Thanks @kentkr! this is ready to go

PR #492, opened 2026-06-09
Chunk work unit feeds

gabotechs:
## Summary

- Previously, each work unit from each partition feed was sent as an individual `WorkUnit` message over the gRPC stream. This caused high per-message overhead when many small work units were being dispatched.
- This PR introduces a `WorkUnitBatch` wrapper in the protobuf schema that groups multiple work units together into a single message.
- On the coordinator side, per-partition feeds are now interleaved via `select_all` and then chunked with `ready_chunks(256)`, meaning up to 256 ready work units (potentially from different partitions) are batched into a single `WorkUnitBatch` message before being sent over the wire.
- On the worker side, the receiver iterates over the batch and dispatches each work unit to its respective partition sender.

## Why

Sending one small protobuf message per work unit pays a fixed per-message cost (framing, syscall, etc.) that dominates when work units are cheap. Batching amortises that overhead across up to 256 units per send, reducing the coordinator-to-worker feed latency under high partition counts.

### DIFF
--- a/src/worker/impl_coordinator_channel.rs
+++ b/src/worker/impl_coordinator_channel.rs
@@ -120,22 +120,26 @@ impl Worker {
         })?;
 
         // Continue reading remaining messages (work unit feed data) in the background.
-        let work_unit_senders = remote_work_unit_feed_registry.senders;
+        let mut work_unit_senders = remote_work_unit_feed_registry.senders;
         #[allow(clippy::disallowed_methods)]
         tokio::spawn(async move {
             let mut body = body.map_ok(set_work_unit_received_time);
             while let Some(Ok(msg)) = body.next().await {
-                let Some(Inner::WorkUnit(msg)) = msg.inner else {
+                let Some(Inner::WorkUnitBatch(msg)) = msg.inner else {
                     continue;
                 };
-                let Ok(id) = deserialize_uuid(&msg.id) else {
-                    continue;
-                };
-                let Some(tx) = work_unit_senders.get(&(id, msg.partition as usize)) else {
-                    continue;
-                };
-                if tx.send(Ok(msg)).is_err() {
-                    break; // channel closed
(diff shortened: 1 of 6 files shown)


src/worker/impl_coordinator_channel.rs
```diff
@@ -125,17 +125,19 @@ impl Worker {
...
+                for msg in msg.batch {
+                    let Ok(id) = deserialize_uuid(&msg.id) else {
+                        continue;
+                    };
+                    let Some(tx) = work_unit_senders.get(&(id, msg.partition as usize)) else {
+                        continue;
+                    };
+                    if tx.send(Ok(msg)).is_err() {
+                        break; // channel closed
+                    }
```
LiaCastaneda, line 142: this will break the `for` loop but the `while` will keep ongoing, is that expected?

LiaCastaneda (approved):

src/worker/impl_coordinator_channel.rs
gabotechs: It does not matter much what we do here, as if the channel gets closed, it's because the query ended, and all this will stop anyways as the coordinator will stop streaming work units.

Although you are right that for keeping the same behavior as before, it might be more correct to move this to a `return` rather than `break`, I'l do that 👍

gabotechs: Actually, now that we are here, it would be even more correct to just remove the specific sender from `work_unit_senders` if we see that the receiving end was closed and it will never accept a message again.

I don't think in practice this is going to matter today, but who knows how this will look like in the future, so maybe it's worth being as correct as possible from the beginning

LiaCastaneda: ah yes, the stream would close and the `while` would automatically break. Removing the sender from the registry makes sense to me as well

PR #499, opened 2026-06-16
Pass SessionConfig to Worker::add_on_plan_hook

### LINKED ISSUE #498: Pass `SessionConfig` to `Worker::add_on_plan_hook`
## Background

`Worker::add_on_plan_hook` descends from [#198](https://github.com/datafusion-contrib/datafusion-distributed/pull/198), which introduced a worker-side plan hook specifically to support integrations such as [`datafusion-tracing`](https://github.com/datafusion-contrib/datafusion-tracing).

The current hook only receives the decoded worker plan:

```rust
Fn(Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan>
```

That is enough for static wrapping, but it is not enough when instrumentation needs to be configured from query-scoped session configuration.

## Problem

`datafusion-tracing` is configured through `InstrumentationOptions`, including span custom fields, metric recording, batch preview limits, and preview formatting.

A downstream distributed runtime may build those options from a custom `ConfigExtension` propagated to the worker, for example query/session metadata reconstructed by `WorkerSessionBuilder` from incoming headers.

At the point where `add_on_plan_hook` runs, the worker has already built the per-query `SessionState` and therefore has the final `SessionConfig`. However, the hook only receives the plan. Since hooks are registered on the `Worker` at startup, they cannot capture per-query config values in their closure.

The missing operation looks like this:

```rust
worker.add_on_plan_hook(|plan, session_config| {
    let props = MySessionProperties::from_session_config(session_config)?;
    let rule = instrument_rule(props, props.verbose_tracing);
    rule.optimize(plan, session_config.options())
});
```

## Proposal

Update the existing `Worker::add_on_plan_hook` API so the hook receives the worker `SessionConfig` and can fail:

```rust
pub fn add_on_plan_hook(
    &mut self,
    hook: impl Fn(
            Arc<dyn ExecutionPlan>,
            &SessionConfig,
        ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError>
        + Send
        + Sync
        + 'static,
);
```

The hook would still run in the same lifecycle location: after session construction and physical plan decode, before `TaskData` registration and execution.

## Why `SessionConfig`

The concrete requirement is access to propagated config extensions/options. `SessionConfig` is the narrowest API that provides that.

A broader context object, `SessionState`, `TaskContext`, request headers, or `TaskKey` does not appear necessary for the tracing integration:

- headers should already have been interpreted by `WorkerSessionBuilder`
- `TaskContext` would work, but only because it contains a `SessionConfig`
- task identity is not needed to build `datafusion-tracing` instrumentation options

If a later hook use case needs task identity or other execution context, that can be added with a separate concrete motivation.

## Why change the existing hook

The current hook was added for tracing, but tracing often needs session/request metadata in order to choose span fields and instrumentation behavior. Adding a second hook would leave the tracing-oriented API split between a plan-only hook and a config-aware hook. Updating the existing API makes the intended extension point match its original purpose.

This is a breaking API change, but the migration is small:

```rust
worker.add_on_plan_hook(|plan, _session_config| Ok(plan));
```

## Expected semantics

The returned plan should remain execution-equivalent to the input plan.

Errors returned by the hook should propagate as the worker plan-registration failure.

## Tests

- hook can read a propagated `ConfigExtension` from `SessionConfig`
- hook can apply a fallible `PhysicalOptimizerRule` using `session_config.options()`
- hook errors propagate to the coordinator
- multiple hooks still run in registration order
- worker docs describe hook ordering and plan-equivalence expectations

## Related

- [#198 Add arrow flight endpoint hooks](https://github.com/datafusion-contrib/datafusion-distributed/pull/198)
- [#100 ConfigExtension propagation](https://github.com/datafusion-contrib/datafusion-distributed/pull/100)
- [#340 Passthrough headers](https://github.com/datafusion-contrib/datafusion-distributed/pull/340)

geoffreyclaude:
Closes #498

## What

- Changes `Worker::add_on_plan_hook` so hooks receive the per-query `SessionConfig` and can return `DataFusionError`.
- Keeps the hook at the same lifecycle point: after the worker session is built and the physical plan is decoded, before task registration/execution.
- Propagates hook failures as worker plan-registration failures.
- Documents hook ordering, the execution-equivalent plan expectation, and error propagation.
- Adds an integration test covering propagated config extensions, using a fallible `PhysicalOptimizerRule`, hook ordering, and coordinator-visible hook errors.

## Why

The worker already has the final per-query `SessionConfig` when the plan hook runs, but the hook could only see the decoded plan. That makes tracing/instrumentation integrations awkward when their options come from session config or propagated config extensions reconstructed by `WorkerSessionBuilder`.

This is a breaking API change, but the migration is small:

```rust
worker.add_on_plan_hook(|plan, _session_config| Ok(plan));
```

## Tests

```text
cargo test --features integration --test worker_plan_hook -- --nocapture
```

### DIFF
--- a/tests/worker_plan_hook.rs
+++ b/tests/worker_plan_hook.rs
@@ -1,0 +1,178 @@
+#[cfg(all(feature = "integration", test))]
+mod tests {
+    use arrow::array::Int32Array;
+    use arrow::datatypes::{DataType, Field, Schema};
+    use arrow::record_batch::RecordBatch;
+    use arrow::util::pretty::pretty_format_batches;
+    use datafusion::common::{HashSet, Result, extensions_options, internal_err};
+    use datafusion::config::ConfigExtension;
+    use datafusion::error::DataFusionError;
+    use datafusion::execution::SessionState;
+    use datafusion::physical_plan::ExecutionPlan;
+    use datafusion::prelude::{SessionConfig, SessionContext};
+    use datafusion_distributed::test_utils::in_memory_channel_resolver::start_configured_in_memory_context;
+    use datafusion_distributed::test_utils::session_context::register_temp_parquet_table;
+    use datafusion_distributed::{DistributedExt, Worker, WorkerQueryContext};
+    use std::sync::Arc;
+    use std::sync::Mutex;
+
+    const HOOK_LABEL: &str = "worker-session-value";
+
+    extensions_options! {
+        pub struct PlanHookOptions {
+            pub label: String, default = "".to_string()
+            pub fail_in_hook: bool, default = false
+        }
+    }
+
+    impl ConfigExtension for PlanHookOptions {
+        const PREFIX: &'static str = "plan_hook";
+    }
+
+    async fn build_state(ctx: WorkerQueryContext) -> Result<SessionState, DataFusionError> {
+        Ok(ctx
+            .builder
+            .with_distributed_option_extension_from_headers::<PlanHookOptions>(&ctx.headers)?
+            .build())
+    }
+
+    #[tokio::test]
+    async fn plan_hooks_receive_session_config_and_run_in_order()
+    -> Result<(), Box<dyn std::error::Error>> {
+        let hook_calls = Arc::new(Mutex::new(HookCalls::default()));
+
+        let mut ctx = start_configured_in_memory_context(3, build_state, {
+            let hook_calls = Arc::clone(&hook_calls);
+            move |mut worker| {
+                add_first_hook(&mut worker, Arc::clone(&hook_calls));
+                add_second_hook(&mut worker, Arc::clone(&hook_calls));
+                worker
+            }
+        })
+        .await;
+
+        ctx.set_distributed_option_extension(PlanHookOptions {
+            label: HOOK_LABEL.to_string(),
+            fail_in_hook: false,
+        });
+
+        assert_eq!(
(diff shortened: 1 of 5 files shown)


jayshrivastava (approved):

tests/worker_plan_hook.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        assert_eq!(
+            collect_hook_query(&ctx).await?,
+            "+----+\n\
+             | id |\n\
+             +----+\n\
+             | 2  |\n\
+             | 3  |\n\
+             +----+"
+        );
```
gabotechs, line 67: This should be an `insta` assertion, rather than a raw `assert_eq!`

tests/worker_plan_hook.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let err = collect_hook_query(&ctx)
+            .await
+            .expect_err("plan hook error should propagate to the query");
+
+        let err = err.to_string();
+        assert!(
+            err.contains("plan hook failed for worker-session-value"),
+            "{err}"
+        );
```
gabotechs, line 106: Prefer using upstream `assert_contains!` here rather than a raw `assert!`. That's the one we use in other tests in this codebase.

gabotechs (commented):
Looks good!

PR #501, opened 2026-06-17
Improve worker resolver ergonomics

gabotechs:
Before, users had no way to access their own `WorkerResolver` from the context. This is actually useful for example if they want to retriever the worker URLs in methods like `scale_up_leaf_node`.

Additionally, they might have a specific implementation to which they might want to downcast, which is not possible today because `WorkerResolver` does not implement `Any`.

This PR adds two main things:
- A way for people to retrieve their `WorkerResolver` from the context
- A way for people to downcast to their own specific `WorkerResolver`

### DIFF
--- a/src/distributed_ext.rs
+++ b/src/distributed_ext.rs
@@ -219,16 +219,10 @@ pub trait DistributedExt: Sized {
     ///     .with_distributed_planner()
     ///     .build();
     /// ```
-    fn with_distributed_worker_resolver<T: WorkerResolver + Send + Sync + 'static>(
-        self,
-        resolver: T,
-    ) -> Self;
+    fn with_distributed_worker_resolver<T: WorkerResolver + 'static>(self, resolver: T) -> Self;
 
     /// Same as [DistributedExt::with_distributed_channel_resolver] but with an in-place mutation.
-    fn set_distributed_worker_resolver<T: WorkerResolver + Send + Sync + 'static>(
-        &mut self,
-        resolver: T,
-    );
+    fn set_distributed_worker_resolver<T: WorkerResolver + 'static>(&mut self, resolver: T);
 
     /// This is what tells Distributed DataFusion how to build a Worker gRPC client out of a worker URL.
     ///
@@ -606,10 +600,7 @@ impl DistributedExt for SessionConfig {
         set_distributed_user_codec_arc(self, codec)
     }
 
-    fn set_distributed_worker_resolver<T: WorkerResolver + Send + Sync + 'static>(
-        &mut self,
-        resolver: T,
-    ) {
+    fn set_distributed_worker_resolver<T: WorkerResolver + 'static>(&mut self, resolver: T) {
         set_distributed_worker_resolver(self, resolver);
     }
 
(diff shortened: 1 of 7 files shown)


src/distributed_planner/task_estimator.rs
```diff
@@ -155,9 +155,6 @@ pub struct TaskRoutingContext<'a> {
     pub plan: &'a Arc<dyn ExecutionPlan>,
     /// The number of tasks to be assigned.
     pub task_count: usize,
-    /// Contains a list of URLs representing machines available to receive a task. These URLs are
-    /// sourced at execution time and thus should closely reflect the real state of the cluster.
-    pub available_urls: &'a [Url],
```
JSOD11, line 160: Nice!

src/networking/worker_resolver.rs
```diff
@@ -2,11 +2,12 @@ use crate::DistributedConfig;
 use crate::config_extension_ext::set_distributed_option_extension;
 use datafusion::common::{DataFusionError, exec_err, not_impl_err};
 use datafusion::prelude::SessionConfig;
+use std::any::Any;
 use std::sync::Arc;
 use url::Url;
 
 /// Resolves a list of worker URLs in the cluster available for executing parts of the plan.
-pub trait WorkerResolver {
+pub trait WorkerResolver: Any + Send + Sync {
```
JSOD11, line 10: Why does this change cause trait bounds to change in multiple places? What is unique about allowing access to the worker resolver through the distributed config compared to other actions we take from the distributed config? Any implications we should think about with these added bounds? (Guessing no, but just checking / am curious.)

src/distributed_ext.rs
```diff
@@ -219,16 +219,10 @@ pub trait DistributedExt: Sized {
     ///     .with_distributed_planner()
     ///     .build();
     /// ```
-    fn with_distributed_worker_resolver<T: WorkerResolver + Send + Sync + 'static>(
-        self,
-        resolver: T,
-    ) -> Self;
+    fn with_distributed_worker_resolver<T: WorkerResolver + 'static>(self, resolver: T) -> Self;
```
JSOD11, line 222: Above, it looks like `Send + Sync` were added, but here they are taken away. What is the reasoning there?

JSOD11 (approved):
Great simplification PR that removes a field from the `RoutingContext` struct — I had wanted to do something like this when implementing routing.

Had a couple small questions, but nothing blocking.

src/networking/worker_resolver.rs
gabotechs: > Why does this change cause trait bounds to change in multiple places?

There are two things happening here:
1) Addition of `Any`: this allows people in their own code to downcast to their own specific `WorkerResolver` implementation. You can see this happening in most of `datafusion` traits actually.
2) Addition of `Send` + `Sync`: this is the same as before, just that we were pasting the `Send` + `Sync` one by one in all places that we require a `WorkerResolver` trait, but now we just have it here once. I actually don't have a strong preference about this, but I see that this is how it's done upstream, so I thought that we might as well do the same (e.g., https://github.com/apache/datafusion/blob/7bb6e152bee1df3cd8215176742bfc1cdf300e52/datafusion/physical-plan/src/execution_plan.rs#L96-L96)

src/distributed_ext.rs
gabotechs: From a compiler standpoint, it's the same thing, but as we are adding the `Any` trait additionally, I thought it might be better to follow what upstream does and require `Send + Sync` in the trait definition rather than in the individual usages.

This is how upstream does this:
https://github.com/apache/datafusion/blob/7bb6e152bee1df3cd8215176742bfc1cdf300e52/datafusion/physical-plan/src/execution_plan.rs#L96-L96

PR #506, opened 2026-06-18
metrics: update plan_bytes_sent to be a bytes metric

jayshrivastava:
Instead of printing byte values like `10 GB`, this metric would print `10 B` where `B` is billion. That's confusing bc it looks like "10 bytes". This change fixes that by updating the metric from a counter to a bytes counter.

### DIFF
--- a/src/coordinator/query_coordinator.rs
+++ b/src/coordinator/query_coordinator.rs
@@ -11,9 +11,9 @@ use crate::worker::generated::worker as pb;
 use crate::worker::generated::worker::coordinator_to_worker_msg::Inner;
 use crate::worker::generated::worker::set_plan_request::WorkUnitFeedDeclaration;
 use crate::{
-    DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, DistributedCodec, DistributedConfig,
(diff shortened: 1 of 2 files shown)


src/coordinator/query_coordinator.rs
```diff
@@ -442,3 +440,29 @@ impl CoordinatorToWorkerMetrics {
...
+        let metrics_set = metrics.clone_inner();
+        let metric = metrics_set
+            .iter()
+            .find(|m| m.value().name() == "plan_bytes_sent")
+            .expect("plan_bytes_sent metric should be registered");
+
+        assert_eq!(
+            format!("{}={}", metric.value().name(), metric.value()),
+            "plan_bytes_sent=4.0 GB"
+        );
```
gabotechs: I think this test is a bit overkill for just a visualization thing, if you ask me, we can move forward without it

gabotechs (approved):

src/coordinator/query_coordinator.rs
jayshrivastava: Done.

PR #508, opened 2026-06-20
feat: added a pluggable `WorkerTransport`, with Arrow Flight optional.

### LINKED ISSUE #483: Abstract away gRPC details from the worker protocol
Today, this project is deeply tied to gRPC and protobuf, as the worker protocol is autogenerated from the following protobuf definition:

- https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/worker/worker.proto

Typically, this is not a problem, as most projects are fine with just sticking with the classical gRPC+protobuf stack for inter-worker communication, but there are some cases where people would like to use different transportation mechanism.

Providing custom implementations for the `prost` generated stubs is not acceptable, as that still forces people into protobuf serde when they might be fine with either just in-memory comms or their own proprietary protocols.

The source of truth for all worker interactions is defined in the protobuf specification here:

```proto
service WorkerService {
  // Establishes a bidirectional message stream between a coordinator and a worker, over which messages
  // will be exchanged at any time during a query's lifetime. It's expected to be one coordinator channel
  // per task.
  rpc CoordinatorChannel(stream CoordinatorToWorkerMsg) returns (stream WorkerToCoordinatorMsg);
  // Executes the requested partition range of a subplan previously sent by the coordinator channel.
  // Returns metadata about a worker. Currently only used for worker versioning.
  rpc GetWorkerInfo(GetWorkerInfoRequest) returns (GetWorkerInfoResponse);
}
```

And one possible way would be to build a new `trait` that abstracts away the `WorkerServiceClient` currently autogenerated by `prost`, and exposes an interface similar in shape to the one represented in today's worker.proto, but not coupled to gRPC or protobuf.

A structure that comes to mind can be to declare the protocol definition as a set of well documented traits in:
```
src/
  worker/
    protocol.rs
```

And specific implementations place somewhere else, like:
```
src/
  protocol_impl/
    grpc/
    ...
```

Ideally, that should allow hiding all gRPC-specific deps behind a feature flag, and even omit compiling them if people do not care about gRPC.

mdashti:
## What

This PR puts a pluggable `WorkerTransport` in front of the worker comms, with Arrow Flight as one implementation behind a default-on `flight` feature. It adds an in-memory transport and surfaces partition routing on `NetworkBoundary`.

## Why

The transport is hardcoded to Flight. The coordinator ships plans over gRPC and the consumer reads partitions through an internal, non-swappable connection, so an embedder can't supply its own transport without reaching into crate internals. The crate also can't build without `tonic` / `arrow-flight`.

Closed #483

## How

- A `WorkerTransport` abstraction and a dispatch seam, with Flight as the reference implementation. No behaviour change for the Flight path.
- `route_partition` / `partitions_per_consumer_task` on `NetworkBoundary`, so a transport can place a produced partition without re-deriving the slice formula by hand.
- An in-memory transport: every worker runs in the current process, plans delivered by a function call, partitions read from the local task registry. `start_localhost_context` honours `DATAFUSION_DISTRIBUTED_TEST_TRANSPORT=flight`, so the generic suite runs over either transport.
- The `flight` feature gates `tonic` / `arrow-flight` and the gRPC surface. In-memory is the default when it's off.

## Tests

CI runs the integration suite over Flight, over the in-memory transport, and once more with `--no-default-features`. `clippy` and `rustfmt` are clean in both feature configs.


src/coordinator/distributed.rs
```diff
@@ -133,6 +141,63 @@ impl DistributedExec {
...
     }
+
+    /// Routes and dispatches every stage through the registered transport, then returns the head
+    /// stage for the caller to drive synchronously, with no `execute()` background task.
+    ///
+    /// This is the extension point the in-crate shared-memory transport (and its in-process test) executes
+    /// through: it owns the runtime and drives the head stage itself. The transport's dispatch must
+    /// complete synchronously: a transport that spawns background delivery work, or a plan that
+    /// declares work-unit feeds (which are pumped by that background work), is rejected rather than
+    /// left to stall.
```
stuhood: This references code that will never be upstreamed, right?

src/coordinator/distributed.rs
```diff
@@ -185,20 +250,36 @@ impl ExecutionPlan for DistributedExec {
...
-            self.metrics_store.clone(),
-        );
+        // One token per execution. In-process transports watch it to tear down the moment this
+        // output stream drops, whether the consumer read to the end or abandoned it early (a LIMIT
+        // gather). It rides the dispatch context so every worker fragment shares it.
+        let token = CancellationToken::new();
+        let context = Arc::new(task_ctx_with_extension(
+            &context,
+            DistributedCancellationToken(token.clone()),
+        ));
```
stuhood, line 270: Is this still necessary with the new per-stream cancellation that you implemented?

src/coordinator/query_coordinator.rs
```diff
@@ -44,65 +36,56 @@ use uuid::Uuid;
...
-pub(super) struct QueryCoordinator {
-    task_ctx: Arc<TaskContext>,
+/// The Arrow-Flight plan-delivery side: a [WorkerDispatch] that ships each task's plan over the
+/// bidirectional coordinator-to-worker gRPC stream and wires up the work-unit feed and metrics
+/// back-channel. Per-query state (plan-send metrics, the keep-alive notifier) is created lazily on
+/// the first stage's dispatch and shared across stages. Dropping the dispatcher fires the notifier,
+/// which closes the coordinator->worker streams and propagates EOS to the workers so they can clean
+/// up; the coordinator holds it until the query's result stream is drained.
+#[derive(Default)]
+pub(crate) struct FlightWorkerDispatch {
```
stuhood, line 50: Is this supposed to be feature-flagged for `flight`, or is it not actually specific to that backend?

src/worker/worker_connection_pool.rs
stuhood, line 1: A followup PR should probably split this module (and others with large swaths of feature flags) into transport and non-transport specific modules.

stuhood (approved):
Looks great: thanks!

src/coordinator/distributed.rs
mdashti: Right. Reworded it on its own terms in 6c4373d. It's the extension point an embedded in-process transport drives the head stage through.

src/coordinator/distributed.rs
mdashti: Different scope. This one fires once, when the whole output stream drops, so an in-process transport tears down on an early `LIMIT` gather, not just clean EOF. The per-stream cancel lives in the shm transport, which isn't in this PR.

src/coordinator/query_coordinator.rs
mdashti: It is `flight`-gated. The `#[cfg(feature = "flight")]` sits on `mod query_coordinator;` in `coordinator/mod.rs`, so it's easy to miss from the struct. Added a module note in `6c4373d` so the gating's visible here too.

src/worker/worker_connection_pool.rs
mdashti: Agreed, that's a good followup. I'd rather keep this PR to the abstraction and leave the module split as its own change, to make this PR easier to review, too. The follow-up can be a refactor-only change.

mdashti (commented):
@stuhood Thanks for the comments.

.github/workflows/ci.yml
```diff
@@ -41,12 +43,50 @@ jobs:
       - uses: ./.github/actions/setup
       - run: cargo test --features integration
 
+  # The same suite over the Arrow-Flight gRPC transport. `start_localhost_context` switches to it
+  # when `DATAFUSION_DISTRIBUTED_TEST_TRANSPORT=flight`, so both transports get full coverage.
+  unit-test-flight-transport:
```
gabotechs, line 48: There might be a miss conception here. This project does not use the Arrow-Flight protocol, more info about that decision here:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/375

We still rely on the `arrow-flight` library for IPC decoding, but we don't adhere to the Arrow Flight protocol

src/lib.rs
```diff
@@ -20,11 +20,18 @@ pub mod test_utils;
...
-pub use coordinator::DistributedExec;
+pub use common::{
+    TreeNodeExt, deserialize_uuid, get_distributed_cancellation_token, serialize_uuid,
+};
+pub use config_extension_ext::get_config_extension_propagation_headers;
+pub use coordinator::{
+    CoordinatorToWorkerMetrics, DistributedExec, EncodedTaskPlan, LatencyMetric, MetricsStore,
+    encode_task_plan,
+};
 pub use distributed_ext::DistributedExt;
```
gabotechs, line 31: This does not seem right, I see this is exposing a big amount of internals that are not supposed to be part of the public API.

Thinking about splitting up work for driving this effort forward, a good first step would be to ship a change that has 0 changes to the public API, and evaluate public API changes in isolation.

src/coordinator/distributed.rs
```diff
@@ -133,6 +149,62 @@ impl DistributedExec {
...
     }
+
+    /// Routes and dispatches every stage through the registered transport, then returns the head
+    /// stage for the caller to drive synchronously, with no `execute()` background task.
+    ///
+    /// This is the extension point for an in-process transport that owns the runtime and drives the
+    /// head stage itself. The transport's dispatch must complete synchronously: a transport that
+    /// spawns background delivery work, or a plan that declares work-unit feeds (which are pumped by
+    /// that background work), is rejected rather than left to stall.
+    pub fn prepare_in_process_plan(
```
gabotechs, line 160: Plan preparation should be abstracted from the worker transport implementation.

I don't think we should be seeing methods named after specific transport protocols, instead, transport details should be gathered behind a `trait`, and code whose domain is planning, execution, etc... should not care whether the transport protocol is in-process, gRPC, or something else.

Also, functions here should not be public. `DistributedExec` must behave as a normal `ExecutionPlan` regarding planning and execution. Users should not care about calling foreing methods not part of `datafusion` unless it's strictly necessary.

src/coordinator/query_coordinator.rs
```diff
@@ -44,65 +40,56 @@ use uuid::Uuid;
 /// small batches. See [StreamExt::ready_chunks] docs for more details about how chunking works.
 const WORK_UNIT_FEED_CHUNK_SIZE: usize = 256;
 
-/// Manages communication between coordinator and workers for a single query.
-///
-/// The [QueryCoordinator]'s lifetime is scoped to a single query , and will instantiate independent
-/// [StageCoordinator] scoped to each individual stage.
-pub(super) struct QueryCoordinator {
```
gabotechs, line 51: I don't think we should be dismanteling `QueryCoordinator`. Probably the way to go is making `QueryCoordinator` not coupled to gRPC, but I do think it still serves well as an abstraction.

Thinking about future work, this is still going to be needed for dynamic planning, so I don't think we can just delete it.

src/distributed_planner/network_boundary.rs
```diff
@@ -22,6 +30,34 @@ pub trait NetworkBoundary: ExecutionPlan {
...
     fn producer_head(&self, consumer_tasks: usize) -> ProducerHead;
+
+    /// `P_c`: how many partitions each consumer task reads in the sliced layout
+    /// (`global = P_c * consumer_task + local`) that shuffle and broadcast reads use. Surfaced so a
+    /// transport that has to place a produced partition does not re-derive it from node properties.
+    /// Meaningless for [NetworkCoalesceExec], whose consumers read whole per-producer-task groups.
+    fn partitions_per_consumer_task(&self) -> usize {
+        self.properties().partitioning.partition_count()
+    }
+
```
gabotechs, line 41: This method is not necessary. `NetworkBoundary` already extends `ExecutionPlan`, which has the necessary methods for people to retrieve the partition count.

src/distributed_planner/network_boundary.rs
```diff
@@ -22,6 +30,34 @@ pub trait NetworkBoundary: ExecutionPlan {
...
+    /// Maps a producer output partition to the consumer task and the local partition within that
+    /// task that reads it, for the `global = P_c * consumer_task + local` layout.
+    ///
+    /// Boundaries whose consumers do not read that layout must override this with an error; the
+    /// default would silently misroute them. A zero-partition boundary is a planner bug, so it
+    /// errors instead of routing everything to task `0`.
+    fn route_partition(&self, output_partition: usize) -> Result<PartitionRoute> {
```
gabotechs, line 48: :thinking: I don't fully understand how this change is related to abstracting away the `WorkerTransport`. Is it really related?

If not, I'd suggest to avoid introducing unrelated changes in this PR and focus exclusively on abstracting away the gRPC protocol.

gabotechs (commented):
Such a humongous work! thanks for putting this together :pray:

However, I want to steer the approach in a different direction. I see this PR takes more like a big-bang rewrite rather than focusing on just abstracting away the gRPC protocol, exposing a lot of the internals of the project in the way.

I think this is mainly because there might be better places for choosing the split point between what's a transport detail VS what's actually part of the core project.

I think taking an approach closer to what the original issue describes https://github.com/datafusion-contrib/datafusion-distributed/issues/483#issue-4587508516, has better chances of providing a more elegant abstraction point for introducing customization at the worker level.

However, it's very easy for me to say this if I've not been in the battlefield fighting against this problem, so my expectations might be a bit skewed.

Some traits that I think we should see in a PR that abstracts away the worker protocol is:
- 0 public API change. We can talk about what's worth making public in a follow up, but I think for the first one it's keeping the change scoped to the internals
- No rewrite of the existing abstractions. Rather than rewriting code that has been carefully thought to be future proof, I'd try to just adapt it to not be coupled to gRPC, which should yield very minor or 0 code changes to those pieces.
- Make the split point at the lowest level possible, mirroring the current gRPC spec. The resulting abstraction should look almost identical to what `prost` autogenerates today out of the worker.proto specification, just happening that it's not coupled to gRPC and are normal Rust methods.
- Almost no increase in LOC in the project. As the scope is not really to introduce new concepts, just hiding existing ones behind a trait, I would not expect there to be a huge positive diff in LOC, at most a few hundred.

As it's very easy for me to say this before actually trying something out, I'll actually try to ship something that adheres to this vision and see what blockers I hit in the way, and maybe I find out that I'm wrong in some assumptions.

Either way, thanks so much for this amount of work :pray:

gabotechs:
Gave it a try myself to implement something closer to what the original issue describes https://github.com/datafusion-contrib/datafusion-distributed/issues/483:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/512

I do think an approach closer to that has better chances of being successful, and the disruptions are minimal, both in terms of LOC and new concepts.

src/coordinator/query_coordinator.rs
mdashti: Good point. I'll keep `QueryCoordinator` and just decouple it from gRPC, which is what #512 does. Agreed it earns its place, more so with dynamic planning coming.

mdashti:
Closing in favor of #512, which landed the lower-level `WorkerChannel` cut we agreed on. Thanks @gabotechs.

PR #511, opened 2026-06-24
fix: new session properties should not error in old worker versions

jayshrivastava:
Workers error when trying to deserialize a session property that does not exist. This can happen if the coordinator is on a new version which has a new session property.

This change proposes that workers should ignore any session properties which it does not recognize.

The alternative is to use worker versions to avoid this problem https://github.com/datafusion-contrib/datafusion-distributed/pull/389 but it's a bit annoying to bump the version on every new session property. Ideally you avoid bumping worker versions generally because it limits the pods a query can run on during a mixed-version deploy.

### DIFF
--- a/src/config_extension_ext.rs
+++ b/src/config_extension_ext.rs
@@ -60,21 +60,22 @@ pub(crate) fn set_distributed_option_extension_from_headers<'a, T: ConfigExtensi
         .into_iter()
         .map(|entry| entry.key)
         .collect::<HashSet<_>>();
-    let prefix = format!("{FLIGHT_METADATA_CONFIG_PREFIX}{}.", T::PREFIX);
 
     for (k, v) in headers.iter() {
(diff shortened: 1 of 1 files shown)


src/config_extension_ext.rs
```diff
@@ -53,12 +54,24 @@ pub(crate) fn set_distributed_option_extension_from_headers<'a, T: ConfigExtensi
         None => MutOrOwned::Owned(T::default()),
     };
 
+    let known_keys = result
+        .as_mut()
+        .entries()
+        .into_iter()
+        .map(|entry| entry.key)
+        .collect::<HashSet<_>>();
```
jayshrivastava, line 61: I think we must allocate here. The other options is to have a nested loop which iterates over the entries and checks for existence.

gabotechs (approved):

PR #512, opened 2026-06-29
Abstract worker protocol from gRPC

gabotechs:
From the core logic perspective, excluding `src/protocol/` and the old generated/protobuf locations that were moved under it, this PR removes net **1,455 LOC** from handwritten/core-ish code:

| Scope | Additions | Deletions | Net |
|---|---:|---:|---:|
| Non-protocol handwritten/core-ish paths | `697` | `2,152` | `-1,455` |

The full PR still looks large because a lot of gRPC/protobuf-specific code moved into the new protocol layer, but the core runtime logic got smaller.

Even if this PR looks big, it is mostly just mechanical changes that have been written fully by hand (no AI). The vast majority of this PR is moving files and code blocks around so that we can abstract away gRPC from the worker protocol.

There are just a few non-mechanical changes that were necessary though:

### 1. introduce `ProducerHeadSpec` (protocol level) as intermediary between `ProducerHead` (planner level) and `pb::ProducerHead` (grpc-specific level)

introduce the intermediate `ProducerHeadSpec` (protocol level) as a middle step between `ProducerHead` (planner level) and `pb::ProducerHead` (grpc-specific level). The problem is that the  `ProducerHead` (planner level) contains `Vec<Arc<dyn PhysicalExpr>>`, that need the presence of a `TaskContext` and `PhysicalExtensionCodec`s to be encoded/decoded. This means that people implementing their own transport protocols would need to deal with all these `TaskContext` and `PhysicalExtensionCodec` inside their code for dealing with it, which is not ideal. Instead, we provide protocol implementors with  `ProducerHeadSpec` (protocol level) with an already serialized `Vec<u8>` instead of `Vec<Arc<dyn PhysicalExpr>>`, so that they do not need to worry about encoding those themselves.

### 2. Rework of the `WorkerConnectionPool`

This file contained several details specific to the gRPC protocol that needed to be factored out to the `grpc/` layer.

Here, there was some tricky code that was capable of interleaving multiple `RecordBatch` streams from multiple partitions under the same physical gRPC stream.

For maintaining that, two things need to happen:
- The `WorkerConnectionPool` needs to adopt a structure that allows `WorkerChannel` implementations to batch streams up and interleave them at will, without being coupled to the specific implementation gRPC happened to have.
- All the `RecordBatch` interleaving that we were doing before now is moved almost untouched to the `grpc/` layer.
- The `WorkerConnectionPool` still manages local connection fallback when the target worker is the same as the one in scope.


src/common/time.rs
```diff
@@ -1,9 +1,14 @@
+use num_traits::AsPrimitive;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Nanoseconds elapsed since UNIX epoch.
-pub(crate) fn now_ns() -> u64 {
+pub(crate) fn now_ns<T>() -> T
+where
+    T: 'static + Copy,
+    u128: AsPrimitive<T>,
+{
```
gabotechs, line 9: This now_ns() now needs to be used in places that require `u64` and other places that require a `usize`, so it can work on both now. Thanks `num_traits`

src/distributed_planner/network_boundary.rs
```diff
@@ -61,6 +74,51 @@ impl NetworkBoundaryExt for dyn ExecutionPlan {
 }
 
 impl ProducerHead {
+    pub(crate) fn to_spec(&self, cfg: &SessionConfig) -> Result<ProducerHeadSpec> {
```
gabotechs, line 77: See point 1 in PR description.

src/networking/channel_resolver.rs
```diff
@@ -65,20 +58,6 @@ pub(crate) fn set_distributed_channel_resolver(
...
-// The Tonic channels need to be established and reused under a whole RuntimeEnv scope, not a single
-// TaskContext, which forces us to put the default implementation in a static global variable that
-// stores and reuses tonic channels per RuntimeEnv's pointer address.
-static DEFAULT_CHANNEL_RESOLVER_PER_RUNTIME: LazyLock<
-    moka::sync::Cache<
-        /* Arc<RuntimeEnv> pointer address */ usize,
-        /* ChannelResolver that reuses built channels */ Arc<DefaultChannelResolver>,
-    >,
-> = LazyLock::new(|| moka::sync::Cache::builder().max_capacity(256).build());
-
```
gabotechs, line 81: All the code in this file was not deleted, it was just moved untouched somewhere else

src/protocol/grpc/observability/gen/regen.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -e
+
+repo_root=$(git rev-parse --show-toplevel)
+cd "$repo_root" && cargo run --manifest-path src/protocol/grpc/observability/gen/Cargo.toml
```
gabotechs, line 6: Not new, moved untouched, not sure why this shows as an actual diff.

src/protocol/grpc/observability/service.rs
```diff
@@ -139,11 +142,38 @@ impl ObservabilityServiceImpl {
...
+    fn total_partitions(&self) -> usize {
+        match self.final_plan.get() {
+            Some(Ok(plan)) => plan.output_partitioning().partition_count(),
+            _ => self
+                .base_plan
+                .properties()
+                .output_partitioning()
+                .partition_count(),
+        }
+    }
```
gabotechs: Not new, moved from other place untouched.

src/protocol/grpc/channel_resolver.rs
```diff
@@ -0,0 +1,181 @@
...
+// per TaskContext, as this holds reference to Tonic channels that must outlive a single TaskContext.
+//
+// The Tonic channels need to be established and reused under a whole RuntimeEnv scope, not a single
+// TaskContext, which forces us to put the default implementation in a static global variable that
+// stores and reuses tonic channels per RuntimeEnv's pointer address.
+pub(crate) static DEFAULT_CHANNEL_RESOLVER_PER_RUNTIME: LazyLock<
+    moka::sync::Cache<
+        /* Arc<RuntimeEnv> pointer address */ usize,
+        /* ChannelResolver that reuses built channels */ Arc<DefaultChannelResolver>,
+    >,
```
gabotechs, line 26: Nothing is new in this file, everything was copy pasted untouched.

src/protocol/grpc/worker_client.rs
```diff
@@ -0,0 +1,765 @@
...
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::task::{Context, Poll};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tokio::sync::Notify;
+use tokio::sync::mpsc::UnboundedSender;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_util::sync::CancellationToken;
+use tonic::metadata::MetadataMap;
+use tonic::{Request, Status};
```
gabotechs, line 43: All the code in this file was scattered across other files before, so nothing is really new, but it do is new how it's laid out. I'll post comments method by method explaining what was new VS what is old.

src/protocol/grpc/worker_client.rs
```diff
@@ -0,0 +1,765 @@
...
+
+#[async_trait]
+impl WorkerChannel for pb::worker_service_client::WorkerServiceClient<BoxCloneSyncChannel> {
+    async fn coordinator_channel(
+        &mut self,
+        headers: HeaderMap,
+        c2w_stream: BoxStream<'static, CoordinatorToWorkerMsg>,
+    ) -> Result<BoxStream<'static, Result<WorkerToCoordinatorMsg>>> {
```
gabotechs, line 51: This method is pretty much a worker protocol to gRPC plumbing passthrough. Readers might notice the the `impl_coordinator_channel.rs` file no longer contains any protobuf plumbing. This is because it's not longer coupled to gRPC or protobuf. This method here collects all the protobuf plumbing that was previously scatter around that file here.

src/protocol/grpc/worker_client.rs
```diff
@@ -0,0 +1,765 @@
...
+        Ok(output_stream)
+    }
+
+    async fn execute_task(
+        &mut self,
+        headers: HeaderMap,
+        request: ExecuteTaskRequest,
+        metrics: ExecutionPlanMetricsSet,
+        ctx: &Arc<TaskContext>,
+    ) -> Result<Vec<BoxStream<'static, Result<RecordBatch>>>> {
```
gabotechs, line 77: This method is exactly what `RemoteWorkerConnection` was before. Readers might notice that the code was copy-pasted from there almost unchanged.

The only thing that changed is:

### Before

There was two steps:
1. Build a `RemoteWorkerConnection` with `RemoteWorkerConnection::init()`
2. Execute a specific partition with `RemoteWorkerConnection::execute`

### Now

The contents of 1) and 2) are inlined into the same function, but they are mostly the same.

src/protocol/grpc/worker_client.rs
```diff
@@ -0,0 +1,765 @@
...
+            version: response.version,
+        })
+    }
+}
+
+type WorkerMsg = Result<(FlightData, FlightAppMetadata), Status>;
+
+struct NetworkLatencyMetrics {
+    metrics: ExecutionPlanMetricsSet,
+    values: Option<NetworkLatencyMetricValues>,
```
gabotechs, line 323: The rest of this file is not new code, it's copy-pasted literally as it was

src/protocol/grpc/worker_service.rs
```diff
@@ -0,0 +1,422 @@
...
+use prost::Message;
+use std::sync::Arc;
+use tonic::{Request, Response, Status, Streaming};
+use url::Url;
+
+const RECORD_BATCH_BUFFER_SIZE: usize = 2;
+
+impl Worker {
+    /// Converts this [Worker] into a [`WorkerServiceServer`] with high default message size limits.
+    ///
```
gabotechs, line 36: As before, this file is not new, it contains copy-pasted code that was there in other places.

src/protocol/worker_channel.rs
```diff
@@ -0,0 +1,187 @@
...
+use datafusion::execution::TaskContext;
+use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricsSet};
+use futures::stream::BoxStream;
+use http::HeaderMap;
+use std::sync::Arc;
+use url::Url;
+use uuid::Uuid;
+
+#[async_trait]
+pub trait WorkerChannel: Send + Sync {
```
gabotechs, line 17: This file is the meat of the PR, it's the definition of the Worker protocol in terms of plain traits. Readers might notice that it's an almost 1:1 copy of the `worker.proto` specification, but in plain Rust.

src/worker/gen/regen.sh
```diff
@@ -1,6 +0,0 @@
-#!/usr/bin/env bash
-
-set -e
-
-repo_root=$(git rev-parse --show-toplevel)
```
gabotechs, line 5: Not deleted, just moved

src/worker/impl_coordinator_channel.rs
```diff
@@ -173,41 +162,69 @@ impl Worker {
...
+            if let Some(Ok(plan)) = task_data.final_plan.get() {
+                let d_ctx = DistributedTaskContext {
+                    task_index: key.task_number,
+                    task_count,
+                };
+                let task_data_metrics = &task_data.task_data_metrics;
+                task_data_metrics.mark_execution_finished();
+                send_metrics_via_channel(&task_data.metrics_tx, plan, d_ctx, task_data_metrics);
+            }
+            task_data_entries.invalidate(&key).await
```
gabotechs, line 186: See point 2) in the PR description. This was a bug that this refactor uncovered.

src/worker/impl_coordinator_channel.rs
```diff
@@ -173,41 +162,69 @@ impl Worker {
...
+    plan: &Arc<dyn ExecutionPlan>,
+    dt_ctx: DistributedTaskContext,
+    task_data_metrics: &Arc<TaskDataMetrics>,
+) {
+    let mut pre_order_plan_metrics = vec![];
+    let _ = plan.apply_with_dt_ctx(dt_ctx, |node, _| {
+        pre_order_plan_metrics.push(node.metrics().unwrap_or_default());
+        Ok(TreeNodeRecursion::Continue)
+    });
+
```
gabotechs, line 227: Not new, copy-pasted as is

src/worker/task_data.rs
```diff
@@ -68,72 +69,57 @@ impl TaskDataMetrics {
...
-    /// Returns the total number of partitions in this task.
-    pub(crate) fn total_partitions(&self) -> usize {
-        match self.final_plan.get() {
-            Some(Ok(plan)) => plan.output_partitioning().partition_count(),
-            _ => self
-                .base_plan
-                .properties()
-                .output_partitioning()
-                .partition_count(),
-        }
```
gabotechs, line 115: Moved somewhere else

src/worker/impl_execute_task.rs
```diff
@@ -1,274 +1,67 @@
...
-pub(crate) async fn execute_local_task(
-    task_data_entries: &Arc<TaskDataEntries>,
-    body: ExecuteTaskRequest,
-) -> Result<(Vec<SendableRecordBatchStream>, Arc<TaskContext>)> {
-    let Some(key) = body.task_key.as_ref().cloned() else {
-        return internal_err!("Missing task_key in LocalWorkerConnection");
-    };
-    let Some(producer_head) = body.producer_head.as_ref().cloned() else {
-        return internal_err!("Missing producer_head");
-    };
```
gabotechs, line 45: Diff is very unfortunate in this file...

Not much changed actually, just most of its contents that were gRPC-specific where copy-pasted to `worker_service.rs`.

There do is one thing that got removed related to point 2) in the PR descritpion, I'll add a comment some lines below

src/worker/impl_execute_task.rs
```diff
@@ -1,274 +1,67 @@
...
-                #[allow(clippy::disallowed_methods)]
-                tokio::spawn(async move {
-                    task_data_entries.invalidate(&key).await;
-                });
-                task_data_metrics.mark_execution_finished();
-                if send_metrics {
-                    send_metrics_via_channel(&metrics_tx, &plan, d_ctx, &task_data_metrics);
-                }
-            }
-        });
```
gabotechs: This whole `on_drop_stream` was there before and it was actually removed. It's now on `impl_coordinator_chanel.rs` because of point 2) in the PR description.

src/worker/worker_connection_pool.rs
```diff
@@ -83,475 +63,113 @@ impl WorkerConnectionPool {
...
-
-    first_poll_notify: Arc<Notify>,
-    // Signals the demux task that buffered memory has been freed by a consumer.
-    mem_available_notify: Arc<Notify>,
-
-    // Metrics collection stuff.
-    memory_reservation: Arc<MemoryReservation>,
-    elapsed_compute: Time,
-}
-
```
gabotechs, line 179: Pretty much all this file was moved to `worker_client.rs`. There's just a few bits left that were greatly simplified as it no longer makes sense to have here the `WorkerConnection` abstraction, the worker protocol + gRPC layer provides already all the abstractions we want.

src/worker/worker_service.rs
```diff
@@ -124,60 +110,17 @@ impl Worker {
...
-    /// `max_encoding_message_size(usize::MAX)` to avoid message size limitations for internal
-    /// communication.
-    ///
-    /// You can further customize the returned server by chaining additional tonic methods.
-    ///
-    /// # Example
-    ///
-    /// ```
-    /// # use datafusion_distributed::Worker;
-    /// # use tonic::transport::Server;
```
gabotechs, line 140: Not removed, just copy-pasted somewhere else

stuhood (approved):
This looks great: thanks a lot for picking up the torch on this one: I agree that it ended up better encapsulated.

Do you expect #432 to land before this one, or was basing it atop #432 an exercise to ensure that extracting `trait WorkerChannel` would not hamper landing that one?

@mdashti drafted https://github.com/paradedb/datafusion-distributed/pull/27 atop this one to additionally add an in-process/in-memory worker implementation, which allows the full test suite to run without `grpc`.

How can we help to get this landed?

src/protocol/channel_resolver.rs
```diff
@@ -0,0 +1,98 @@
...
+/// - This is called per gRPC request, so implementors of this trait should make sure that
+///   clients are reused across method calls instead of building a new Worker client every time.
+///
+/// - When implementing `get_worker_client_for_url`, it is recommended to use the
+///   [`create_worker_client`] helper function to ensure clients are configured with
+///   appropriate message size limits for internal communication. This helps avoid message
+///   size errors when transferring large datasets.
+#[async_trait]
+pub trait ChannelResolver {
+    /// For a given URL, get a Worker gRPC client for communicating to it.
```
gabotechs, line 28: Not new, moved as-is to here

src/protocol/grpc/observability/proto/observability.proto
```diff
@@ -24,7 +24,6 @@ message TaskKey {
 // Progress information for a single task
 message TaskProgress {
   TaskKey task_key = 1;
-  uint64 total_partitions = 2;
```
gabotechs, line 27: Was unused, so I removed it

stuhood: Nit: Do you use reserved values when removing fields? https://protobuf.dev/programming-guides/proto3/#fieldreserved

stuhood (approved):
Thanks, still looks great! I'll rebase atop this.

As to this point:
> Here, there was some tricky code that was capable of interleaving multiple RecordBatch streams from multiple partitions under the same physical gRPC stream.

I'm surprised that this ends up being worthwhile! But since the complexity mostly doesn't leak into `trait WorkerChannel`, it seems worth preserving.

src/protocol/grpc/observability/proto/observability.proto
gabotechs: Actually yes, we have already a `reserved` value here, adding this other one as well. Thanks for noticing!

src/protocol/channel_resolver.rs
```diff
@@ -0,0 +1,98 @@
...
+use url::Url;
+
+/// Allows users to customize the way Worker clients are created. A common use case is to
+/// wrap the client with tower layers or schedule it in an IO-specific tokio runtime.
+///
+/// There is a default implementation of this trait that should be enough for the most common
+/// use-cases.
+///
+/// # Implementation Notes
+/// - This is called per gRPC request, so implementors of this trait should make sure that
```
gene-bordegaray: nit: docs are still referring to gRPC specific things

src/protocol/worker_channel.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// Establishes a bidirectional message stream between a coordinator and a worker, over which messages
+    /// will be exchanged at any time during a query's lifetime. It's expected to be one coordinator channel
+    /// per task.
+    async fn coordinator_channel(
+        &mut self,
+        headers: HeaderMap,
```
gene-bordegaray, line 23: is adding something specific to HTTP in the general trait something we want to do here. I know there are use cases where workers are not using these headers. Maybe this should just be general "metadata"

gene-bordegaray (approved):
two nonblockers but this is really nice 💯

src/protocol/channel_resolver.rs
gabotechs: Done. The official docs also refer to gRPC all over the place, but for now I think this is fine as gRPC is still expected to be the default implementation.

src/protocol/worker_channel.rs
gabotechs: Yeah, I would have liked to factor out `http::HeaderMap` out of the protocol. I actually tried, but was not successful for the following reason:

There are some parts of the code outside of the protocol scope that still rely on specifically `http::HeaderMap` to transport things as headers:
- [Passthrough headers](https://github.com/datafusion-contrib/datafusion-distributed/blob/09938a7efbd8e52119ff79c6637c9417ec47f294/src/worker/worker_connection_pool.rs#L124-L124)
- [Config extension propagation machinery](https://github.com/datafusion-contrib/datafusion-distributed/blob/09938a7efbd8e52119ff79c6637c9417ec47f294/src/config_extension_ext.rs#L95-L98)

Factoring these out requires some more non-mechanical changes, and it would get a bit complex without not a lot to gain, as `http::HeaderMap` in the end is just a fancy `HashMap` that helps with serializing/deserializing its KV pairs, but there's not really much coupling to HTTP itself.

That being said, a really clean separation of concerns should not assume that `http::HeaderMap` is going to be present, but there's even more instances of this that we can potentially improve in this abstraction:
- The presence of `http::HeaderMap`.
- The fact that `ExecutionPlan`s are serialized as protobuf bytes. Other implementations (like in-memory) might not want to serialize at all.
- The fact that `ProducerHeadSpec::RepartitionExec` requires the hash expressions to be serialized as protobuf bytes. Again, an in-memory implementation should not need to serialize them.
- Work unit feed messages encoded as protobuf bytes. Again, an in-memory implementation might want to avoid serialization.

I'll do two things:
1. Add a warning at the top of this trait, telling folks that this API is unstable and that it might change in order to account for all of this in the future.
2. Create issues for follow ups on this.

philippemnoel:
Let's go!!! Thank you 🙏

PR #514, opened 2026-06-30
Rearrange if statements in planner

gabotechs:
Just a small rearrangement of the `if` branches in the inject_network_boundaries.rs file so that everything is a bit clearer and hierarchically the same.

### DIFF
--- a/src/distributed_planner/inject_network_boundaries.rs
+++ b/src/distributed_planner/inject_network_boundaries.rs
@@ -286,82 +286,78 @@ async fn _inject_network_boundaries(
     // Cap the reconciled task count by the configured max-per-stage budget.
     task_count = task_count.limit(nb_ctx.max_tasks()?);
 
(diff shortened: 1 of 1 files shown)


src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -286,82 +286,78 @@ async fn _inject_network_boundaries(
     // Cap the reconciled task count by the configured max-per-stage budget.
     task_count = task_count.limit(nb_ctx.max_tasks()?);
 
-    // Upon reaching a hash repartition, we need to introduce a shuffle right above it.
-    if let Some(r_exec) = plan.downcast_ref::<RepartitionExec>() {
```
gabotechs, line 290: Diff is a bit funny, but code is mostly unchanged, it's just about where the different `if` branches are located.

gabotechs:
Due to lack of review power, I'll pull this trivial PR in to let people focus on the bigger ones. If there are any comments please let me know and I'll follow up.

PR #515, opened 2026-06-30
Show rewritten physical plan with --debug in local benchmarks

gabotechs:
Just a small QOL improvement for local benchmarks.


gabotechs:
Due to lack of review power, I'll pull this trivial PR in to let people focus on the bigger ones. If there are any comments please let me know and I'll follow up.

PR #524, opened 2026-07-02
Finalize worker metrics on coordinator channel EOS

### LINKED ISSUE #521: Dynamic planning can deadlock on set-but-never-executed stages
Follow-up from #432 ([review](https://github.com/datafusion-contrib/datafusion-distributed/pull/432#pullrequestreview-4518119541), point 1).

Potential circular wait during dynamic planning:

- `DistributedExec::execute` keeps the end-query guard alive until after `drain_pending_tasks()` (`src/coordinator/distributed.rs:199`, `:218`).
- `send_plan_task` keeps the coordinator-to-worker request stream open via `keep_stream_alive` (`src/coordinator/query_coordinator.rs:184`).
- The worker response stream waits on the metrics stream (`src/worker/impl_coordinator_channel.rs:180`).

If the dynamic planner sets and samples a stage but later never executes it, metrics never completes. `drain_pending_tasks()` then waits for the worker response stream to end, but that stream is kept alive by the coordinator channel, which only closes when the guard drops — and the guard can't drop because execution is blocked inside `drain_pending_tasks()`.

Related: point 10 of the same review suggests a targeted regression test for the planned-but-abandoned stage scenario (planner sends `SetPlanRequest`, worker samples and sends `LoadInfo`, AQE picks a different shape so the stage never reaches `execute_task`, query must still finish without `drain_pending_tasks()` hanging).

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/521

The old lifecycle assumed every producer partition stream would be materialized and dropped. That still tends to hold in the current `main` branch because the WorkerConnectionPool owns eager demux/drop behavior, so there is no natural failing test here.

In the following PR:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/512

The gRPC abstraction refactor changes stream ownership and returns already-demuxed streams through the protocol boundary, which exposes the latent issue: query output can finish while some producer partition streams never reach the final drop path, leaving the metrics oneshot unresolved.

Coordinator-channel EOS is the better task-finalization signal because it represents query-scoped shutdown independently of whether every partition stream was consumed or dropped normally.

Deadlock sequence (WARN: explained by Codex, I still don't understand it 100%)

Why the bug does not reproduce in `main`:

1. Metrics are sent from the worker when the worker-side execution streams for a task finish or are dropped.
2. In `main`, the remote gRPC path owns the full execute-task request lifecycle internally: once `WorkerConnectionPool` initializes a remote connection, the request is driven by the connection/demux machinery.
3. Local self-calls are also effectively per-partition: the local path asks `execute_task` for one partition at a time instead of returning a batch of partition streams where some may never be polled.
4. Because of that, the streams that count toward task finalization are normally materialized and eventually dropped.
5. The metrics oneshot resolves before the coordinator needs the query-scoped coordinator->worker stream to close.
6. `drain_pending_tasks()` can complete, and only after that the query-end guard is dropped. There is no dependency cycle.

Why the bug reproduces in https://github.com/datafusion-contrib/datafusion-distributed/pull/512:

1. Metrics are still tied to the old final stream-drop path.
2. The refactor makes `WorkerChannel::execute_task` return already-demuxed partition streams through the protocol boundary.
3. The gRPC implementation keeps those streams lazy: it does not send the actual `execute_task` RPC until one returned partition stream is first polled.
4. In shapes like `custom_routing_join`, DataFusion may construct some remote partition streams that are never polled.
5. For those streams, the first-poll gate never opens, so the worker never receives the corresponding `execute_task` request.
6. Since the worker never receives that request, it never materializes the worker-side execution streams whose drops would decrement the task-finalization counter.
7. The worker therefore never sends metrics for that task, leaving the worker->coordinator response stream open.
8. The coordinator is waiting in `drain_pending_tasks()` for that response stream to finish, while the query-end guard is still alive.
9. If metrics are moved to coordinator-channel EOS without also dropping the query-end guard before draining, the worker then waits for coordinator->worker EOS, but that EOS is gated by the same query-end guard. That creates the cycle.

The following changes are made here:
- Finalize worker task metrics when the coordinator channel reaches EOS instead of from the last partition stream drop.
- Drop the query-end guard before draining coordinator background tasks so workers can observe the EOS being waited on.
- Keep task invalidation tied to coordinator-channel shutdown.

I think this is something https://github.com/datafusion-contrib/datafusion-distributed/pull/509 also noticed

### DIFF
--- a/src/worker/impl_coordinator_channel.rs
+++ b/src/worker/impl_coordinator_channel.rs
@@ -1,12 +1,14 @@
-use crate::common::deserialize_uuid;
+use crate::common::{TreeNodeExt, deserialize_uuid};
 use crate::execution_plans::SamplerExec;
+use crate::metrics::proto::df_metrics_set_to_proto;
+use crate::protobuf::datafusion_error_to_tonic_status;
 use crate::work_unit_feed::{RemoteWorkUnitFeedRegistry, set_work_unit_received_time};
 use crate::worker::LocalWorkerContext;
 use crate::worker::generated::worker::coordinator_to_worker_msg::Inner;
 use crate::worker::generated::worker::set_plan_request::WorkUnitFeedDeclaration;
 use crate::worker::generated::worker::worker_service_server::WorkerService;
 use crate::worker::generated::worker::{
-    CoordinatorToWorkerMsg, WorkerToCoordinatorMsg, worker_to_coordinator_msg,
+    CoordinatorToWorkerMsg, TaskMetrics, WorkerToCoordinatorMsg, worker_to_coordinator_msg,
 };
 use crate::worker::task_data::TaskDataMetrics;
 use crate::{
@@ -14,7 +16,9 @@ use crate::{
     WorkerQueryContext,
 };
 use datafusion::common::DataFusionError;
+use datafusion::common::tree_node::TreeNodeRecursion;
 use datafusion::execution::SessionStateBuilder;
+use datafusion::physical_plan::ExecutionPlan;
 use datafusion::prelude::SessionConfig;
 use datafusion_proto::physical_plan::AsExecutionPlan;
 use datafusion_proto::protobuf::PhysicalPlanNode;
@@ -23,6 +27,7 @@ use futures::{FutureExt, StreamExt, TryStreamExt};
 use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, OnceLock};
 use tokio::sync::oneshot;
+use tokio::sync::oneshot::Sender;
 use tonic::{Request, Response, Status, Streaming};
 use url::Url;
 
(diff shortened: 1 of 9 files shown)


src/worker/impl_execute_task.rs
```diff
@@ -87,29 +76,6 @@ pub(crate) async fn execute_local_task(
         let stream = plan.execute(partition as usize, Arc::clone(&task_ctx))?;
         let stream_schema = plan.schema();
 
-        let plan = Arc::clone(&plan);
-
-        let task_data_entries = Arc::clone(task_data_entries);
-        let num_partitions_remaining = Arc::clone(&task_data.num_partitions_remaining);
```
jayshrivastava, line 93: I think `num_partitions_remaining` can be removed from the `task_data` struct, right?

jayshrivastava (commented):
This looks right but I need some help filling in some blanks.

Before, `query_coordinator.drain_pending_tasks().await?;` would have to finish before dropping the `_guard`. The reason we deadlock is fundamentally because, after your changes, we don't guarantee that every partition in a task is run. The partition count in the worker would not hit 0, so the worker -> coordinator stream would be kept alive. Since you've deleted this partition counting, the problem goes away. We now drop the guard in the coordinator which kicks off everything, including metrics collection.

I'd like to understand why not all partitions are executed
- The PR desc mentions `custom_routing_join`. What's that? Does that cause some partitions to possibly not be executed? 
- In your original AQE PR you mentioned join order swapping, did you actually implement that? I forgot
- You also mentioned that some tasks may be set but not executed. Ex. if the `SamplerExec` sees no rows.

src/worker/impl_execute_task.rs
gabotechs: We could, but it's still used by the observability service, the service that feeds the console.

gabotechs:
> The PR desc mentions custom_routing_join. What's that? Does that cause some partitions to possibly not be executed?

Ah yeah sorry, I should have stated that more clearly, it's one of the integration test that is showing the problem in the other PR (https://github.com/datafusion-contrib/datafusion-distributed/pull/512). It shows it because that test has a JOIN that drops some partitions of the probe side mid-query (I think).

> In your original AQE PR you mentioned join order swapping, did you actually implement that? I forgot

Nope, I have a very drafty draft (https://github.com/datafusion-contrib/datafusion-distributed/tree/gabrielmusat/aqe), but no, not yet

> You also mentioned that some tasks may be set but not executed. Ex. if the SamplerExec sees no rows.

Not really, I think it mostly happens just on JOINs that decide to drop the probe side early. Today we force-execute all the partitions of the probe side regarding of whether they will be consumed at some point or not, however future implementations of the worker protocol might decide to not force-execute them, which is what happens on the next PR.
(disclaimer: I'm not 100% sure that statement is right, my understanding about this problem is a bit fuzzy)

src/worker/impl_coordinator_channel.rs
jayshrivastava, line 154: > Not really, I think it mostly happens just on JOINs that decide to drop the probe side early. Today we force-execute all the partitions of the probe side regarding of whether they will be consumed at some point or not, however future implementations of the worker protocol might decide to not force-execute them, which is what happens on the next PR.
(disclaimer: I'm not 100% sure that statement is right, my understanding about this problem is a bit fuzzy)

Sure that sounds reasonable. This change looks good. When we drop the guard, the stream closes and this loop terminates.

src/worker/impl_execute_task.rs
jayshrivastava: Ack. But you've removed the `fetch_sub`. I think the `on_drop_steam` can be reinstated just for `num_partitions_remaining.fetch_sub(1, Ordering::SeqCst)`. And we might as well do this too ` task_data_entries.invalidate(&key).await;` to eagerly clear the cache.

jayshrivastava (approved):

src/worker/impl_execute_task.rs
gabotechs: 🤔 You are right here... rather than bringing it back just for the console, I'd be more in favor of just deleting it

gabotechs: It turns out that not even the console was using it, even if it was part of its proto spec, so I just deleted it

gabotechs:
Updated the PR description with a better explanation about what happened.

Issue #542, opened 2026-07-08
Perf: Rewrite physical-plan to reduce network latency


Rich-T-kid:
TBD

Rich-T-kid:
sub issue of  #541

Issue #541, opened 2026-07-08
Perf: Improve network latency for distributed datafusion


Rich-T-kid:
# Why is this important?

No matter how efficient we make DataFusion/arrow-rs, distributed DataFusion is fundamentally constrained by real-world network latency. For high fan-out, high-compute queries that require multiple nodes, which is the whole reason distributed DataFusion exists, we pay a real cost for every byte sent between nodes.
Optimizing and tuning this data-transfer cost has exponential returns as the size of the data being fanned out, and the number of nodes it's fanned out to, grows.

# What exists today

In the past, distributed DataFusion relied on CoalesceBatchesExec to batch record batches together before sending them over the wire via Arrow Flight. The reasoning was straightforward: small record batches pay the fixed overhead cost of network communication for comparatively little payload benefit, so merging batches together before transmission amortizes that cost.

[With CoalesceBatchesExec](https://github.com/datafusion-contrib/datafusion-distributed/issues/386) being deprecated for this purpose, a new config was introduced to replace it:
```
pub shuffle_batch_size: usize, default = 0
```
[This was intended to tune the size of record batches sent between execution stages](https://github.com/datafusion-contrib/datafusion-distributed/pull/407) (i.e., at the network/shuffle boundary).

# What's wrong with the current approach

Currently, `shuffle_batch_size` overwrites the `target_batch_size` for [every operator in the entire query plan](https://github.com/datafusion-contrib/datafusion-distributed/blob/7ab60f155332e2e09d3b434293d3bacac739a909/src/worker/impl_coordinator_channel.rs#L86), not just the ones at network boundaries ([this can be checked by adding print statement's to view the target_size of incoming record batches](https://github.com/datafusion-contrib/datafusion-distributed/blob/7ab60f155332e2e09d3b434293d3bacac739a909/src/worker/impl_execute_task.rs#L96)). This is incorrect for two reasons:
1. **Misleading scope** : The naming and documentation of the config imply that changing it only affects the record batch size of data sent across the network boundary. In practice, it silently overrides `target_batch_size` plan-wide.
2. **Conflated goals** :`target_batch_size` and `shuffle_batch_size` optimize for two different, rarely-aligned concerns:

- `target_batch_size` is tuned to manage memory pressure and spilling behavior within a node, keeping batches small enough to avoid OOM and to control out-of-core spill efficiency.
[...]

PR #543, opened 2026-07-08
decorrelate shuffle size from target_batch_size

Rich-T-kid:
Currently `DistributedConfig.shuffle_batch_size` sets the `target_record_batch_size` for all operators in a datafusion plan. This helps for buffering record batches before being sent over the network but it may cause users to unintentionally change the `target_record_batch`. `target_record_batch` and `shuffle_batch_size` should be independent config values. (see other issue with this in #541)

This PR re-introduces coalesce batching to provide the same buffering of arrow-flight data before being sent over the wire to minimize overhead. the target size for batches is set to `shuffle_batch_size`.

As @gabotechs mentions in [#386](https://github.com/datafusion-contrib/datafusion-distributed/issues/386), coalescing has a cost. It copies rows from source arrays into a new destination buffer. Looking at [push_batch](https://github.com/apache/arrow-rs/blob/4b54fe6f5a03fe196447c95c86cfbde5e66dcc8c/arrow-select/src/coalesce.rs#L325), `in_progress_arrays` briefly holds a pointer to each source array via `set_source()`, and `copy_rows()` is called against it on every push. Despite the name, [copy_rows()](https://github.com/apache/arrow-rs/blob/4b54fe6f5a03fe196447c95c86cfbde5e66dcc8c/arrow-select/src/coalesce/generic.rs#L59) does mostly zero-copy work: it takes a zero-copy `slice()` of the source array and pushes it into a Vec. It's cheap metadata bookkeeping, not a buffer copy.

The actual copy is deferred to `finish_buffered_batch()`, which calls concat over all the buffered slices to write them into a single pre-allocated, contiguous buffer. Because this is one bulk copy into pre-sized memory, we only pay for the writes not re-sizing buffers. Modern CPUs do `memcpy`-style bulk copies extremely fast, especially when the data stays in cache.
The tuning insight is that as `shuffle_batch_size` outpaces `target_batch_size`, the number of separate coalesce-and-send cycles shrinks. 
Consider 12 incoming record batches of 8k rows each. With no coalescing, that's 12 Arrow Flight writes over the wire. If we set `shuffle_batch_size` to 32kb, we instead do 3 Flight writes, at the cost of coalescing 4 arrays into one buffer each time. That's a good trade: a handful of (possibly in-cache) CPU copies is cheaper than the per-message network overhead of sending many small batches over the wire.

**a tradeoff of**  local CPU/copy cost vs. the overhead of many small network packets.

At worst, this performs on par with what main currently does, with the added benefit of being clearer to users. A follow up to this PR is discussed in #542

context:
https://github.com/datafusion-contrib/datafusion-distributed/issues/541

### DIFF
--- a/src/distributed_planner/distributed_config.rs
+++ b/src/distributed_planner/distributed_config.rs
@@ -49,7 +49,7 @@ extensions_options! {
         /// shuffle batch sizes independently of the global `datafusion.execution.batch_size`.
         ///
         /// Set to 0 (the default) to apply no override and inherit `datafusion.execution.batch_size`.
-        pub shuffle_batch_size: usize, default = 0
+        pub shuffle_batch_size: usize, default = 8192 * 4
         /// Maximum tasks that will be assigned per stage during distributed planning.
         /// If set to 0, this value is the number of workers returned by the provided `WorkerResolver`.
         /// It defaults to 0.
--- a/src/worker/impl_execute_task.rs
+++ b/src/worker/impl_execute_task.rs
@@ -8,6 +8,7 @@ use crate::worker::spawn_select_all::spawn_select_all;
 use crate::worker::worker_service::{TaskDataEntries, Worker};
 use arrow_flight::encode::{DictionaryHandling, FlightDataEncoder, FlightDataEncoderBuilder};
 use arrow_flight::error::FlightError;
+use arrow_select::coalesce::BatchCoalescer;
 use arrow_select::dictionary::garbage_collect_any_dictionary;
 use datafusion::arrow::array::{Array, AsArray, RecordBatch, RecordBatchOptions};
 use datafusion::arrow::ipc::CompressionType;
@@ -110,6 +111,7 @@ pub(crate) async fn execute_remote_task(
     };
     let mut flight_streams = Vec::with_capacity(arrow_streams.len());
     for (partition, arrow_stream) in partition_range.zip(arrow_streams) {
+        let arrow_stream = coalesce_flight_stream(arrow_stream, &d_cfg);
         let flight_stream = build_flight_data_stream(arrow_stream, compression)?.map(move |msg| {
             // For each FlightData produced by this stream, mark it with the appropriate
             // partition. This stream will be merged with several others from other partitions,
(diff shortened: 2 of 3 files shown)


src/worker/impl_execute_task.rs
```diff
@@ -205,3 +207,56 @@ fn garbage_collect_arrays(batch: RecordBatch) -> Result<RecordBatch, DataFusionE
...
+
+/// Wraps `stream` in a coalescing adapter that buffers record batches until `target` rows
+/// are accumulated before yielding, reducing the number of Flight messages sent over the wire.
+fn coalesce_flight_stream(
+    stream: SendableRecordBatchStream,
+    d_cfg: &DistributedConfig,
+) -> SendableRecordBatchStream {
+    use futures::StreamExt as FuturesStreamExt;
+
+    if d_cfg.shuffle_batch_size == 0 {
```
Rich-T-kid, line 219: this adds opt out behavior the users that want `target_batch_size` == `shuffle_batch_size`

src/distributed_planner/distributed_config.rs
```diff
@@ -49,7 +49,7 @@ extensions_options! {
         /// shuffle batch sizes independently of the global `datafusion.execution.batch_size`.
         ///
         /// Set to 0 (the default) to apply no override and inherit `datafusion.execution.batch_size`.
-        pub shuffle_batch_size: usize, default = 0
+        pub shuffle_batch_size: usize, default = 8192 * 4
```
Rich-T-kid, line 52: I think this is a sane default, happy to tune it

gabotechs:
Could you share how does this perform over the remote benchmarks?

Rich-T-kid:
> Could you share how does this perform over the remote benchmarks?

@gabotechs 

```
=== Comparing tpch_sf10 results from engine 'datafusion-distributed-main' [prev] with 'datafusion-distributed-decorrelate-shuffle-size' [new] ===
      q1: prev= 421 ms, new= 438 ms, diff=1.04 slower ✖
      q2: prev=1445 ms, new=1393 ms, diff=1.04 faster ✔
      q3: prev=1014 ms, new= 874 ms, diff=1.16 faster ✔
      q4: prev= 438 ms, new= 334 ms, diff=1.31 faster ✅
      q5: prev=1228 ms, new=1183 ms, diff=1.04 faster ✔
      q6: prev= 286 ms, new= 223 ms, diff=1.28 faster ✅
      q7: prev=1144 ms, new=1089 ms, diff=1.05 faster ✔
      q8: prev=1545 ms, new=1632 ms, diff=1.06 slower ✖
      q9: prev=1911 ms, new=1531 ms, diff=1.25 faster ✅
     q10: prev=1219 ms, new=1218 ms, diff=1.00 faster ✔
     q11: prev= 578 ms, new= 678 ms, diff=1.17 slower ✖
     q12: prev= 539 ms, new= 558 ms, diff=1.04 slower ✖
     q13: prev= 782 ms, new= 867 ms, diff=1.11 slower ✖
     q14: prev= 560 ms, new= 558 ms, diff=1.00 faster ✔
     q15: prev= 765 ms, new= 777 ms, diff=1.02 slower ✖
     q16: prev= 443 ms, new= 403 ms, diff=1.10 faster ✔
     q17: prev= 853 ms, new= 864 ms, diff=1.01 slower ✖
     q18: prev=1248 ms, new=1263 ms, diff=1.01 slower ✖
     q19: prev= 442 ms, new= 574 ms, diff=1.30 slower ❌
     q20: prev= 948 ms, new= 923 ms, diff=1.03 faster ✔
     q21: prev=1242 ms, new=1235 ms, diff=1.01 faster ✔
     q22: prev= 365 ms, new= 386 ms, diff=1.06 slower ✖
   TOTAL: prev=19416 ms, new=19001 ms, diff=1.02 faster ✔
```
this doesn't introduce any regressions & the differences seem to be mostly noise.
I also experimented with `(8,16,32,64) * 8192` for shuffle_size and they are all within 1.00-1.08x so pretty minor difference.

gabotechs:
If the difference is just within noise, I'd recommend to not merge this PR.

The problem with this approach is that it implies double-coalescing batches: one time for adhering to standard's `datafusion.execution.batch_size` and another time for adhering to `distributed.shuffle_batch_size`. Note that coalescing is an O(N) operation over the data at worst, as it copies it for coalescing.

Citing the description:

> target_record_batch and shuffle_batch_size should be independent config values

We deliberately made the decision of not making these two values independent, mainly for avoiding double-coalescing and simplifying the setup. We prefer to keep to full pipeline under a single `batch_size`.

If you see that the benchmarks demonstrate that this decision was wrong, this is good to go, but otherwise I'd probably try to look for other perf opportunities.

Rich-T-kid:
ah that make sense.

> The problem with this approach is that it implies double-coalescing batches: one time for adhering to standard's datafusion.execution.batch_size and another time for adhering to distributed.shuffle_batch_size

This was actually the main issue I wanted to address. I spoke with @jayshrivastava offline about this. The goal of this PR was mostly as a building block for #542  to make that PR smaller. 

> If you see that the benchmarks demonstrate that this decision was wrong

With this in mind ill implement some of the optimizations I have in mind for #542 and then open a PR if I see good results.

thank you for the comments @gabotechs ill close this PR now

Rich-T-kid:
> We deliberately made the decision of not making these two values independent, mainly for avoiding double-coalescing and simplifying the setup. We prefer to keep to full pipeline under a single batch_size.

if we can avoid the double-coalescing do you think it would be worth it to make `shuffle_size` and `target_batch_size` independent? Im of the opinion  that these should be distinct since they are tuned for different priorities, i'm curious as to what factors went into the decision.

PR #546, opened 2026-07-09
Fix TopK compute cost estimation

gabotechs:
When there was a Sort(TopK), the compute cost was not taking into account the TopK part, overestimating the compute cost of the operator.

This PR fixes it.


asolimando (approved):
LGTM!

PR #548, opened 2026-07-10
feat: add an in-process worker transport (no gRPC)

mdashti:
## What

This PR makes the `WorkerChannel` abstraction usable by a transport that is not gRPC, and adds the in-process reference transport that proves it.

## Why

#512 abstracted the worker protocol, but gRPC stayed the only implementation and the proto sat behind the `grpc` feature, so nothing built or ran without it.

## How

One commit per concern, each with its own test:

- Moved the prost types out from behind `grpc`. Only the tonic client and server stay gated, and the generator emits those gates.
- Decoupled the `benchmarks` dev-dependency (it re-unified `grpc` into every test build) and added a `unit-test-no-grpc` CI job that runs the whole lib suite with the feature off.
- `InProcessChannelResolver`: routes the three protocol methods to a co-located `Worker`. No gRPC, no IPC, no serialization round trip.
- `metrics_store()` with a public `insert`, and a no-gRPC `decode_task_metrics`: a driver whose transport returns metrics out-of-band files the decoded frames into the store the EXPLAIN rewrite reads. Happy to break this one out into its own PR too if preferred.

## Tests

The in-process transport runs a distributed `GROUP BY` end to end with `grpc` off in CI, and a unit test pins the metrics frame-to-store path.


src/coordinator/distributed.rs
```diff
@@ -64,6 +64,13 @@ impl DistributedExec {
         }
     }
 
+    /// The store where worker task metrics land at runtime, if metrics collection is enabled.
+    /// Exposed for a driver whose transport returns metrics out-of-band; it files decoded frames
+    /// here before the per-task EXPLAIN rewrite.
+    pub fn metrics_store(&self) -> Option<Arc<MetricsStore>> {
+        self.metrics_store.clone()
+    }
```
stuhood: This is technically not related to this PR. _I'm_ fine with it, but would just make it clear in a comment that you're more than willing to break out the relevant commits into separate PRs.

stuhood (approved):
The memory portion looks good, but I would recommend against trying to include the edits to:
* `src/coordinator/query_coordinator.rs`
* `src/distributed_planner/network_boundary.rs`
* `src/dispatch_plan_source.rs`

...in this PR.

mdashti:
Thanks @stuhood. Pulled `dispatch_plan_source.rs`, the `query_coordinator.rs` hook, and the `network_boundary.rs` routing out of this PR; they'll come as a follow-up once this lands. The tip is now the protocol un-gate, the no-gRPC CI job, the in-process transport, and the metrics surface.

src/coordinator/distributed.rs
mdashti: The split is done.

mdashti:
@gabotechs Can you please take a look?

gabotechs (commented):
Before reviewing, I see this PR contains a lot of changes that are unnecessary, for example:
- Tweaks to the CI. I'd leave the CI as it is for now, that can be evaluated separately
- Changes to the public API, like exposes the `metrics_store` method in DistributedExec
- Move some integration tests to the benchmarks

Before reviewing, I'd recommend to really scope the PR just to have an in-memory implementation of the WorkerChannel.

Also, as I don't think there's any external use-case, it might be better to delay adding more `WorkerChannel` implementations to when they are needed. The `WorkerChannel` abstraction is in early stages, and it will likely suffer several breaking changes in the short term as we try to accommodate more features. Having more implementations that this project is not going to use will mean paying the price of maintaining those implementations with no benefit, so I'd recommend to defer this work a bit until there's a valid use case.

For example, when we have AQE re-optimizing queries on the fly, there'll be a use case why people would want to run this project in a fully in-memory fashion, once we get that it might be a good time to start thinking about officially supporting more implementations in this project.

stuhood:
> Before reviewing, I'd recommend to really scope the PR just to have an in-memory implementation of the WorkerChannel.

I agree: I'll attempt to scope this down to only adding the new `WorkerChannel` abstraction. But I do think that that will require one additional CI job that runs unit tests with `--no-default-features`, as that's the only way to prove that things are actually buildable.

> Also, as I don't think there's any external use-case, it might be better to delay adding more `WorkerChannel` implementations to when they are needed. The `WorkerChannel` abstraction is in early stages, and it will likely suffer several breaking changes in the short term as we try to accommodate more features. Having more implementations that this project is not going to use will mean paying the price of maintaining those implementations with no benefit, so I'd recommend to defer this work a bit until there's a valid use case.

I think that having two implementations of `trait WorkerChannel` in the repo is a good thing: any trait with exactly one (nearby) implementation seems much more likely to become unusable by alternative implementations. And the second implementation in this case is ~230 lines (plus 1 CI job)... hopefully not much of a maintenance burden.

gabotechs:
> But I do think that that will require one additional CI job that runs unit tests with --no-default-features, as that's the only way to prove that things are actually buildable.

👍 Sounds good

> I think that having two implementations of trait WorkerChannel in the repo is a good thing: any trait with exactly one (nearby) implementation seems much more likely to become unusable by alternative implementations. And the second implementation in this case is ~230 lines (plus 1 CI job)... hopefully not much of a maintenance burden.

Sounds good also. Let's aim it to be as simple as possible so that the maintenance burden is minimal.

gabotechs (commented):
Is this ready for another review? I see it still contains several changes that I assume are unrelated, like moving some integration tests to the benchmarking crate

stuhood:
> Is this ready for another review? I see it still contains several changes that I assume are unrelated, like moving some integration tests to the benchmarking crate

Most of these changes were necessary: see this point in the description:
> Decoupled the `benchmarks` dev-dependency (it re-unified `grpc` into every test build) and added a `unit-test-no-grpc` CI job that runs the whole lib suite with the feature off.

So yes, this is ready for a review.

gabotechs:
Ah, you are right! at first I thought it was just un-cleaned up stuff, but I see how that's needed now.

I think we should find ways of shuffling things around a bit so that we can make things work without so many "grpc" flag checks scattered across the project and without changing the structure of where the integration tests are located.

For a first PR, I think it's fine to assume that if the "integration" flag is enabled, then the "grpc" flag is also going to be enabled. We can decide how to split the test utils into grpc-scoped and general in a follow up, but I'm overall not too worried about that.

Ideally, we should reach a state where the project almost does not have "grpc" flags scattered across files, besides the one in:

```rust
#[cfg(feature = "grpc")]
pub mod grpc;
```

I see there's a fundamental limitation that is getting in our way:

The `worker.proto` file contains two things:
- protobuf definitions of the vanilla Rust messages present in the protocol <- this should be general-purpose, and not under the grpc scope
- the grpc-specific service definition <- this should live under the grpc scope

One thing we can do, is to separate the `worker.proto` in two:
1. Something like `src/protocol/messages.proto` that just contains proto representations of the existing messages, no grpc service definition
2. The grpc service definition (`service WorkerService { }`), inside a file called something like `src/protocol/grpc/worker_service.proto`, that just imports all the protobuf messages present in the general-purpose `messages.proto`, and provides the service definition.

With `prost`, it's easy to just do this import dance between different files, so that should not be a problem.

stuhood:
@gabotechs : That makes sense. Thanks for your patience!

I've applied the feedback about splitting the proto in https://github.com/datafusion-contrib/datafusion-distributed/pull/573, which I'll be able to edit the description of.

mdashti:
Closed in favour of https://github.com/datafusion-contrib/datafusion-distributed/pull/573

PR #550, opened 2026-07-12
refactor: SessionConfig.extensions instead ConfigOptions.extensions

### LINKED ISSUE #538: Use SessionConfig.extensions instead ConfigOptions.extensions for carrying structs
This project started using `ConfigOptions.extensions` at the beginning because everything was bundled in a `PhysicalOptimizerRule`.

This introduced the necessity of threading a lot of the customization traits as `ConfigOptions`, which was documented as tech debt that should be addressed at some point.

https://github.com/datafusion-contrib/datafusion-distributed/blob/c814a9cfb0cef96d50a4de7de69366c148979e93/src/distributed_planner/distributed_config.rs#L157-L157

However, now that the planner is implemented as a `QueryPlanner`, it should be very easy to transport these structs as normal `SessionConfig.extensions` objects, instead of hacking them into `ConfigOptions`.

Rich-T-kid:
closes #538

# What Changed
Move 4 non-serializable DistributedConfig fields to SessionConfig.extensions

`__private_task_estimator`, `__private_channel_resolver`, `__private_worker_resolver`, and `__private_work_unit_feed_registry` no longer live in `ConfigOptions.extension`s. They are stored directly in `SessionConfig.extensions` via set_extension/get_extension, which removes all the hollow ConfigField + Debug stub impls these types required.

The main structural change is `inject_network_boundaries` taking &SessionConfig instead of &ConfigOptions so it can reach the AnyMap. All other changes are mechanical callsite updates.


src/distributed_planner/distributed_query_planner.rs
```diff
@@ -104,9 +107,9 @@ impl QueryPlanner for DistributedQueryPlanner {
             plan = Arc::new(CoalescePartitionsExec::new(plan));
         }
 
-        let cfg = session_state.config_options();
+        let session_config = session_state.config();
```
Rich-T-kid: naming this `session_config_options` may be better

Rich-T-kid: nvm

src/distributed_planner/task_estimator.rs
```diff
@@ -232,30 +223,40 @@ pub(crate) fn set_distributed_task_estimator(
...
         Some(TaskEstimation::desired(task_count))
     }
+}
+
+impl TaskEstimator for FileScanConfigTaskEstimator {
+    fn task_estimation(
+        &self,
+        plan: &Arc<dyn ExecutionPlan>,
+        cfg: &ConfigOptions,
+    ) -> Option<TaskEstimation> {
```
Rich-T-kid: This never actually gets called in the codebase so I think its fine to leave this here for trait impl, but call `estimate_with_session()`

src/test_utils/work_unit_file_scan.rs
```diff
@@ -366,12 +366,11 @@ impl TaskEstimator for WorkUnitFileScanTaskEstimator {
         plan: &Arc<dyn ExecutionPlan>,
         cfg: &ConfigOptions,
     ) -> Option<TaskEstimation> {
+        const BYTES_PER_PARTITION: usize = 16 * 1024 * 1024;
```
Rich-T-kid: this code exist in test so I think its fine to set constants for this. the alternative is lots of code bloat just for test

src/distributed_ext.rs
```diff
@@ -645,38 +645,43 @@ impl DistributedExt for SessionConfig {
...
     }
 
     fn set_distributed_cardinality_effect_task_scale_factor(
         &mut self,
         factor: f64,
     ) -> Result<(), DataFusionError> {
-        let d_cfg = DistributedConfig::from_config_options_mut(self.options_mut())?;
+        let mut d_cfg = self.get_extension::<DistributedConfig>().map(|a| a.as_ref().clone()).unwrap_or_default();
         d_cfg.cardinality_task_count_factor = factor;
+        self.set_extension(Arc::new(d_cfg));
```
Rich-T-kid: It may seems like unnecessary copying with the current pattern of `get`->`copy`->`set`->`store`, but the struct is only 160 bytes. I think this is negligible.

Rich-T-kid (commented):
left some comments for reviewers on some of my rational in case it helps 🚀

gabotechs:
🤔 I think you might have misunderstood the issue description in https://github.com/datafusion-contrib/datafusion-distributed/issues/538.

The scope is not to move `DistributedConfig` to `SessionConfig.extensions`, the idea is to just move those fields that were transported as `__private_*` fields in `DistributedConfig` as normal `SessionConfig.extensions`.

Citing the issue description

>  it should be very easy to transport these structs as normal SessionConfig.extensions objects, instead of hacking them into ConfigOptions.

Here, `structs` mean `CombinedTaskEstimator`, `ChannelResolverExtension`, `WorkerResolverExtension` and `WorkUnitFeedRegistry`

src/distributed_planner/task_estimator.rs
```diff
@@ -479,9 +465,6 @@ mod tests {
             cfg.execution.target_partitions = 1;
             let d_cfg = DistributedConfig {
                 file_scan_config_bytes_per_partition: 1,
```
Rich-T-kid, line 474: this test needs to mirror what was there before

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -1267,6 +1273,12 @@ mod tests {
         let network_boundaries_ctx = InjectNetworkBoundaryContext {
             cfg: session_config.options(),
             d_cfg: DistributedConfig::from_config_options(session_config.options()).unwrap(),
+            worker_resolver: session_config
+                .get_extension::<WorkerResolverExtension>()
+                .unwrap_or_else(|| Arc::new(WorkerResolverExtension::not_implemented())),
```
Rich-T-kid: sane defaults here would be nice

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -136,11 +139,18 @@ use uuid::Uuid;
...
-    cfg: &ConfigOptions,
+    session_cfg: &SessionConfig,
 ) -> Result<Arc<dyn ExecutionPlan>> {
+    let cfg = session_cfg.options();
     let ctx = InjectNetworkBoundaryContext {
         cfg,
         d_cfg: DistributedConfig::from_config_options(cfg)?,
+        worker_resolver: session_cfg
+            .get_extension::<WorkerResolverExtension>()
+            .unwrap_or_else(|| Arc::new(WorkerResolverExtension::not_implemented())),
```
Rich-T-kid: again sane defaults here would be nice

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -104,9 +108,9 @@ impl QueryPlanner for DistributedQueryPlanner {
             plan = Arc::new(CoalescePartitionsExec::new(plan));
         }
 
-        let cfg = session_state.config_options();
+        let arc_cfg = session_state.config_options();
```
Rich-T-kid:
```suggestion
        let session_config_options = session_state.config_options();
```

src/distributed_planner/distributed_config.rs
```diff
@@ -121,12 +90,12 @@ impl DistributedConfig {
...
     pub fn from_config_options_mut(cfg: &mut ConfigOptions) -> Result<&mut Self, DataFusionError> {
-        let Some(distributed_cfg) = cfg.extensions.get_mut::<DistributedConfig>() else {
-            return plan_err!("DistributedConfig is not in ConfigOptions.extensions");
-        };
-        Ok(distributed_cfg)
+        if cfg.extensions.get::<DistributedConfig>().is_none() {
+            cfg.extensions.insert(DistributedConfig::default());
+        }
+        Ok(cfg.extensions.get_mut::<DistributedConfig>().unwrap())
     }
```
gabotechs, line 99: Any chance to leave this one untouched? or is this change needed for this PR?

src/distributed_planner/distributed_config.rs
```diff
@@ -138,85 +107,8 @@ impl DistributedConfig {
...
-//  extensions in SessionConfig's AnyMap instead of the ConfigOptions. However, we need
-//  to pass this as ConfigOptions as we need these two fields to be present during
-//  planning in the DistributedQueryPlanner, and the signature of the create_physical_plan()
-//  method there accepts a SessionState which only provides ConfigOptions.
-//  The following PR addresses this: https://github.com/apache/datafusion/pull/18168
-//  but it still has not been accepted or merged.
-//  Because of this, all the boilerplate trait implementations below are needed.
-impl ConfigField for ChannelResolverExtension {
-    fn visit<V: Visit>(&self, _: &mut V, _: &str, _: &'static str) {
-        // nothing to do.
```
gabotechs, line 162: ✂️ Nice!

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -1267,6 +1273,12 @@ mod tests {
         let network_boundaries_ctx = InjectNetworkBoundaryContext {
             cfg: session_config.options(),
             d_cfg: DistributedConfig::from_config_options(session_config.options()).unwrap(),
+            worker_resolver: session_config
+                .get_extension::<WorkerResolverExtension>()
+                .unwrap_or_else(|| Arc::new(WorkerResolverExtension::not_implemented())),
+            task_estimator: session_config
+                .get_extension::<CombinedTaskEstimator>()
+                .unwrap_or_else(|| Arc::new(CombinedTaskEstimator::default())),
```
gabotechs: Would be cool to have a `WorkerResolverExtension::from_session_config()` and `CombinedTaskEstimator::from_session_config()` helper methods for this.

gabotechs (commented):
Nice blood red PR!

src/distributed_planner/inject_network_boundaries.rs
Rich-T-kid: I agree, introducing these now.

src/distributed_planner/distributed_config.rs
Rich-T-kid: no its not a must have, reverting this.

Rich-T-kid: actually I think this needs to be updated for this PR.

Rich-T-kid:
> Any chance to leave this one untouched? or is this change needed for this PR?

@gabotechs   

`from_config_options_mut()` took &mut ConfigOptions, which has no access to the SessionConfig-level propagation prefix registry. So any DistributedConfig field set before the worker resolver, things like compression, collect_metrics, broadcast_joins, was never propagated to workers because the prefix was only registered inside `set_distributed_worker_resolver()`. Changing the signature to &mut SessionConfig lets the method register the prefix and insert DistributedConfig in one place, making both orderings safe. 
I added two regression test to catch this behavior in the future.
`distributed_config_header_propagation_scalar_before_resolver()` and `distributed_config_header_propagation_scalar_after_resolver()` in worker_resolver.rs.

examples/custom_worker_url_routing.rs
```diff
@@ -219,8 +219,9 @@ impl TaskEstimator for CachedFileScanConfigTaskEstimator {
     }
 
     fn route_tasks(&self, ctx: &TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>> {
-        let d_cfg = DistributedConfig::from_task_context(&ctx.task_ctx)?;
-        let available_urls = d_cfg.worker_resolver().get_urls()?;
+        let available_urls =
+            datafusion_distributed::get_distributed_worker_resolver(ctx.task_ctx.session_config())?
```
gabotechs: Can we use a normal import statement like the others here?

src/distributed_planner/distributed_config.rs
```diff
@@ -121,12 +92,25 @@ impl DistributedConfig {
...
+            set_distributed_option_extension(cfg, DistributedConfig::default());
+        } else {
+            register_config_extension_prefix(cfg, DistributedConfig::PREFIX);
+        }
+        Ok(cfg
+            .options_mut()
+            .extensions
+            .get_mut::<DistributedConfig>()
+            .unwrap())
     }
```
gabotechs, line 99: This change also seems unrelated to this PR, ca we keep the `from_config_options_mut` method?

gabotechs: Also, that way of injecting `DistributedConfig` in the config.extensions... looks a bit too LLMy, it's overall pretty convoluted and unnecessary.

For ensuring that the `DistributedConfig` is always present in the `SessionConfig`, rather than doing on each `from_session_config_mut`, I'd instead just put in in `SessionStateBuilderExt`, because that's the only mandatory place people should go enable distributed-datafusion:

```rust
impl SessionStateBuilderExt for SessionStateBuilder {
    fn with_distributed_planner(mut self) -> Self {
        let config = self.config().get_or_insert_default().options_mut();
        config
            .optimizer
            .enable_physical_uncorrelated_scalar_subquery = false;
        if config.extensions.get::<DistributedConfig>().is_none() {
            config.extensions.insert(DistributedConfig::default());
        }

        let prev = std::mem::take(self.query_planner());
        self.with_query_planner(Arc::new(DistributedQueryPlanner { prev }))
    }
}
```

But not sure if this is necessary.

Rich-T-kid: Reverted this change. There were some test breaking due to this but I resolved them by adding 
`.with_distributed_option_extension(DistributedConfig::default())` to the sessionState builder

examples/custom_worker_url_routing.rs
```diff
@@ -219,8 +220,8 @@ impl TaskEstimator for CachedFileScanConfigTaskEstimator {
     }
 
     fn route_tasks(&self, ctx: &TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>> {
-        let d_cfg = DistributedConfig::from_task_context(&ctx.task_ctx)?;
-        let available_urls = d_cfg.worker_resolver().get_urls()?;
+        let available_urls =
+            get_distributed_worker_resolver(ctx.task_ctx.session_config())?.get_urls()?;
```
gabotechs: 🤔 Damn... this is actually how people are expected to access their worker resolver, and the API is taking a hit on usability now...

src/distributed_planner/distributed_config.rs
gabotechs: I'm afraid we cannot do that, because that means that in the same way this is blowing up in our tests here, it will blow up in other people's codebase, se we need to find an alternative.

src/distributed_planner/distributed_query_planner.rs
```diff
@@ -116,14 +118,15 @@ impl QueryPlanner for DistributedQueryPlanner {
...
+            .await?;
 
         plan = prepare_network_boundaries(plan)?;
         if !plan.exists(|plan| Ok(plan.is_network_boundary()))? {
             return Ok(original_plan);
         }
 
-        let plan = partial_reduce_below_network_shuffles(plan, cfg)?;
+        let plan = partial_reduce_below_network_shuffles(plan, session_config_opt)?;
         let plan = push_fetch_into_network_coalesce(plan)?;
```
gabotechs: Can we leave this unrelated variable name changes out of the PR? Not that there's a problem with the new names, but if for some reason we need to revert this PR in the future, not changing unrelated things will yield less chances of conflicts with other PRs.

src/distributed_planner/distributed_config.rs
```diff
@@ -121,7 +90,6 @@ impl DistributedConfig {
         Ok(distributed_cfg)
     }
 
-    /// Gets the [DistributedConfig] from the [ConfigOptions]'s extensions.
     pub fn from_config_options_mut(cfg: &mut ConfigOptions) -> Result<&mut Self, DataFusionError> {
         let Some(distributed_cfg) = cfg.extensions.get_mut::<DistributedConfig>() else {
```
gabotechs: Any reason for removing the docs in this function?

src/worker_resolver.rs
```diff
@@ -25,29 +24,20 @@ pub(crate) fn set_distributed_worker_resolver(
...
-                __private_worker_resolver: worker_resolver,
-                ..Default::default()
-            },
-        )
+    cfg.set_extension(Arc::new(WorkerResolverExtension(Arc::new(worker_resolver))));
+    let options = cfg.options_mut();
+    if options.extensions.get::<DistributedConfig>().is_none() {
+        options.extensions.insert(DistributedConfig::default());
     }
 }
```
gabotechs: I'm realizing that setting it like this is not going to be enough, it's not registered as distributed config, it's registered as a normal config, so it will not make it through worker jumps.

gabotechs: We can do here something like this:

```rust
pub(crate) fn set_distributed_worker_resolver(
    cfg: &mut SessionConfig,
    worker_resolver: impl WorkerResolver + 'static,
) {
    DistributedConfig::ensure_in_config(cfg);
    cfg.set_extension(Arc::new(WorkerResolverExtension(Arc::new(worker_resolver))));
}
```

And let the `ensure_in_config` method be like this:

```rust
    pub(crate) fn ensure_in_config(cfg: &mut SessionConfig) {
        if cfg
            .options()
            .extensions
            .get::<DistributedConfig>()
            .is_none()
        {
            set_distributed_option_extension(cfg, DistributedConfig::default())
        }
    }
```

We'll need to call the `ensure_in_config` method here, in `set_distributed_channel_resolver` and in `SessionStateBuilderExt::with_distributed_planner`

examples/custom_worker_url_routing.rs
gabotechs: What comes to mind is, in the same way that we have the `DistributedExt::with_distributed_worker_resolver` and `DistributedExt::set_distributed_worker_resolver`, we can have the `DistributedExt::get_distributed_worker_resolver` for retrieving the worker resolver.

How does that sound?

src/distributed_planner/distributed_config.rs
Rich-T-kid: no, reverted

examples/custom_worker_url_routing.rs
Rich-T-kid: This look's good but it requires adding this stub for  `impl DistributedExt for SessionStateBuilder`

```
 fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError> {
        Err(DataFusionError::Execution(
            "get_distributed_worker_resolver is not available on SessionStateBuilder; use SessionState or SessionConfig".to_string(),
        ))
    }
```

src/distributed_planner/distributed_config.rs
Rich-T-kid: Put the comment back

src/distributed_ext.rs
```diff
@@ -9,7 +9,7 @@ use crate::work_unit_feed::set_distributed_work_unit_feed;
 use crate::worker_resolver::set_distributed_worker_resolver;
 use crate::{
     ChannelResolver, DistributedConfig, TaskEstimator, WorkUnitFeed, WorkUnitFeedProvider,
-    WorkerResolver,
+    WorkerResolver, get_distributed_worker_resolver as get_worker_resolver,
```
Rich-T-kid: renamed this because otherwise 
`impl DistributedExt for SessionConfig` :: get_distributed_worker_resolver() would called `get_distributed_worker_resolver()` which may be confusing.

examples/custom_worker_url_routing.rs
```diff
@@ -219,8 +219,9 @@ impl TaskEstimator for CachedFileScanConfigTaskEstimator {
     }
 
     fn route_tasks(&self, ctx: &TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>> {
-        let d_cfg = DistributedConfig::from_task_context(&ctx.task_ctx)?;
-        let available_urls = d_cfg.worker_resolver().get_urls()?;
+        let available_urls =
+            DistributedExt::get_distributed_worker_resolver(ctx.task_ctx.session_config())?
+                .get_urls()?;
```
gabotechs: We should be fine with accessing the worker resolver like this now:

```rust
        let available_urls = ctx
            .task_ctx
            .session_config()
            .get_distributed_worker_resolver()?
            .get_urls()?;
```

src/distributed_planner/distributed_config.rs
```diff
@@ -138,85 +107,18 @@ impl DistributedConfig {
     pub fn from_task_context(ctx: &Arc<TaskContext>) -> Result<&Self, DataFusionError> {
         Self::from_session_config(ctx.session_config())
     }
-
-    /// Returns the [WorkerResolver] currently in scope for this [DistributedConfig].
-    pub fn worker_resolver(&self) -> &Arc<dyn WorkerResolver> {
-        &self.__private_worker_resolver.0
+    pub(crate) fn ensure_in_config(cfg: &mut SessionConfig) {
```
gabotechs, line 113: Can we separate this one line jump from the function body above, and add a small doc comment about what it does?

src/test_utils/routing.rs
```diff
@@ -291,8 +292,8 @@ impl TaskEstimator for URLEmitterTaskEstimator {
     }
 
     fn route_tasks(&self, routing_ctx: &crate::TaskRoutingContext<'_>) -> Result<Option<Vec<Url>>> {
-        let d_cfg = DistributedConfig::from_task_context(&routing_ctx.task_ctx)?;
-        let mut routed_urls = d_cfg.worker_resolver().get_urls()?;
+        let mut routed_urls =
+            get_distributed_worker_resolver(routing_ctx.task_ctx.session_config())?.get_urls()?;
```
gabotechs: We can also do it here:

```rust
        let mut routed_urls = routing_ctx
            .task_ctx
            .session_config()
            .get_distributed_worker_resolver()?
            .get_urls()?;
```

src/distributed_ext.rs
```diff
@@ -221,6 +221,8 @@ pub trait DistributedExt: Sized {
     /// ```
     fn with_distributed_worker_resolver<T: WorkerResolver + 'static>(self, resolver: T) -> Self;
 
+    fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError>;
```
gabotechs: We are missing a doc comment here. I'd explain what it does and one example of where is typically used.

For example, it can be typically used inside a `TaskEstimator::route_tasks` implementation for retrieving the available URLs at any given moment.

src/distributed_ext.rs
gabotechs: Why would it be confusing? this is in the same situation as set_distributed_task_estimator, set_distributed_work_unit_feed, set_distributed_worker_resolver, set_distributed_channel_resolver, which have the same names than their counterpart methods in DistributedExt.

For consistency with the other functions, I'd just import this one without aliasing.

src/distributed_ext.rs
```diff
@@ -857,6 +863,12 @@ impl DistributedExt for SessionConfig {
 }
 
 impl DistributedExt for SessionStateBuilder {
+    fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError> {
+        Err(DataFusionError::Execution(
+            "get_distributed_worker_resolver is not available on SessionStateBuilder; use SessionState or SessionConfig".to_string(),
+        ))
+    }
+
```
gabotechs, line 870: :thinking: damn, it's true, there's no way of accessing the worker resolver here...

gabotechs: Then, I think it's probably worth it to create another extension trait for getters. I bet that we are going to be in this same situation more times in the future, so it might be better to tackle this now.

Here's an idea:

```rust
pub trait DistributedGetterExt: Sized {
    /// TODO: nice description
    fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError>;
}

impl DistributedGetterExt for SessionConfig {
    fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError> {
        get_distributed_worker_resolver(self)
    }
}

impl DistributedGetterExt for SessionState {
    delegate! {
        to self.config() {
            fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError>;
        }
    }
}

impl DistributedGetterExt for SessionContext {
    delegate! {
        to self.state_ref().read().config() {
            fn get_distributed_worker_resolver(&self) -> Result<Arc<dyn WorkerResolver>, DataFusionError>;
        }
    }
}
```

And expose it in `lib.rs` like this, along with the other DistributedExt:

```rust
pub use distributed_ext::{DistributedExt, DistributedGetterExt};
```

gabotechs: And omit the DistributedGetterExt impl for SessionStateBuilder

gabotechs:
cc @timsaucer, this PR removes that hack we were doing for adding extensions as ConfigOptions. Does this help with the datafusion-python integration?

Rich-T-kid:
I've introduced the changes you mentioned in https://github.com/datafusion-contrib/datafusion-distributed/pull/550/commits/beda7de3ff4ec826171ff77fc517c92005efe075

src/distributed_planner/distributed_config.rs
gabotechs: We are missing the extra space between the function body above, just to give some breathing room to the eye and be consistent with the formatting of the other functions.

gabotechs (approved):
💯 Nice work @Rich-T-kid!

PR #552, opened 2026-07-13
Drop velocity-based estimation in favor of driver path completion estimation

gabotechs:
## Summary

Replaces the **velocity-based** runtime-statistics estimate used by the dynamic task-count planner with a **driver-path completion** estimate.

The sampler on top of each stage feeds statistics (final row/byte counts) back to the coordinator, which uses them to size the stage above. The old estimate extrapolated those final sizes from a sampled throughput multiplied by a fixed, assumed query duration, which systematically overestimated — often by orders of magnitude. The new estimate instead scales what a stage has already produced by how far it is through its *known* total work.

## How it worked before

The `SamplerExec` reported, per partition:

- `*_ready` — rows / bytes already materialized at sampling time, and
- `*_per_second` — a sampled production **velocity** (rows/bytes per second).

`gather_runtime_statistics` extrapolated the final size as:

```text
total_rows  = rows_ready  + rows_per_second  * ESTIMATED_QUERY_TIME_S   // = 10
total_bytes = bytes_ready + bytes_per_second * ESTIMATED_QUERY_TIME_S
```

i.e. *"what's ready now, plus the sampled rate projected over 10 more seconds of execution."*

While doing some benchmarks, I've seen that this approach heavily over-estimates row count.

## How the new estimate avoids it

Instead of projecting a velocity over a made-up duration, it computes an actual **completion fraction** and divides by it:

```text
pct_completed = rows_pulled_from_leaf / estimated_driver_path_leaf_rows
total_rows    = rows_ready  / pct_completed
total_bytes   = bytes_ready / pct_completed
```

- `rows_pulled_from_leaf` — how many rows the stage has actually pulled from its leaf scans so far (now reported by the sampler in place of the velocity fields).
- `estimated_driver_path_leaf_rows` — the total number of rows the stage must pull to finish, obtained by summing the leaf `num_rows` statistics along the **driver path**: the probe-side chain that actually streams to the output. Join build sides are excluded because they are one-off setup work, not output progress (see the new `apply_driver_path` helper). These counts come from real source statistics (e.g. parquet row counts).

## Caveats 

If a stage's leaves expose no row-count statistics, it falls back to a fixed `FALLBACK_ESTIMATED_PCT_SAMPLED = 0.2` (assume ~20 % sampled).

## Benchmarks 

TPCH-SF1: 1.02 faster
TPCH-SF10: 1.04 faster
TPCH-SF100: 1.01 faster
TPC-DS-SF1: 1.01 slower
ClickBench: 1.01 slower

Performance is not impacted because both in `main` and this PR max out on tasks for a cluster of 12 workers.

Regarding q-error on stats estimation, here are the results:

TPCH-SF1: 
QERR P50: prev=1.80x, new=1.80x
QERR P95: prev=6.75x, new=2.29x

TPCH-SF10: 
QERR P50: prev=1.80x, new=1.45x
QERR P95: prev=8.94x, new=2.12x

TPCH-SF100:
QERR P50: prev=2.12x, new=1.73x
QERR P95: prev=9.68x, new=2.16x

TPC-DS-SF1:
QERR P50: prev=1.54x, new=1.45x
QERR P95: prev=4.62x, new=4.01x

ClickBench:
QERR P50: prev=1.42x, new=1.45x
QERR P95: prev=1.42x, new=1.45x

### DIFF
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -3,6 +3,7 @@ use crate::{DistributedTaskContext, NetworkBoundaryExt};
 use datafusion::common::Result;
 use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeIterator, TreeNodeRecursion};
 use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::joins::{CrossJoinExec, HashJoinExec, NestedLoopJoinExec};
 use std::cell::RefCell;
 use std::sync::Arc;
 
@@ -26,12 +27,41 @@ pub(crate) trait TreeNodeExt {
         f: F,
     ) -> Result<TreeNodeRecursion>;
 
+    /// Applies `f` top-down (pre-order) to the nodes on the plan's *driver path*: the operators
+    /// whose ongoing data production feeds, batch for batch, into the root's output. The
+    /// [`TreeNodeRecursion`] returned by `f` steers it as in [`TreeNode::apply`].
+    ///
+    /// It differs from a full traversal only at the pipeline-breaking joins ([`HashJoinExec`],
+    /// [`NestedLoopJoinExec`], [`CrossJoinExec`]): it follows the probe side (`right`) and skips the
+    /// build side (`left`), since a join fully materializes its build side before emitting any
+    /// output row, so build-side rows are setup work rather than output progress. Every other
+    /// operator is transparent.
+    ///
+    /// Used to estimate a running stage's completion: the total rows to pull come from the leaves
+    /// on the driver path (see `estimated_driver_path_leaf_rows`).
+    ///
+    /// ```text
+    ///        ┌────────────┐
+    ///        │ HashJoin   │  visited
+    ///        └─────┬──────┘
+    ///        build │ probe
+    ///      (left)  │  │ (right)
+    ///     ┌────────┘  └────────┐
+    ///     ▼                    ▼
+    /// ┌────────┐          ┌────────┐
+    /// │ Scan B │ SKIPPED  │ Scan P │  visited
+    /// └────────┘          └────────┘
+    /// ```
+    fn apply_driver_path<F: FnMut(&Self) -> Result<TreeNodeRecursion>>(
+        &self,
+        f: F,
+    ) -> Result<TreeNodeRecursion>;
+
     /// Recursively rewrite the tree using `f` in a top-down (pre-order) fashion, propagating
     /// the appropriate [DistributedTaskContext] based on the presence of nodes that can isolate
     /// tasks, like [ChildrenIsolatorUnionExec].
     ///
     /// `f` is applied to the node first, and then its children.
-    #[allow(dead_code)] // Used in follow up work.
     fn transform_down_with_dt_ctx<
         F: FnMut(Self, DistributedTaskContext) -> Result<Transformed<Self>>,
     >(
@@ -62,7 +92,6 @@ pub(crate) trait TreeNodeExt {
     /// count.
     ///
     /// `f` is applied to the node first, and then its children.
-    #[allow(dead_code)] // Used in follow up work.
     fn transform_down_with_task_count<F: FnMut(Self, usize) -> Result<Transformed<Self>>>(
         self,
         task_count: usize,
@@ -102,6 +131,32 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             })
         }
         recurse(self, ctx, &mut f)
+    }
+
+    fn apply_driver_path<F: FnMut(&Self) -> Result<TreeNodeRecursion>>(
+        &self,
+        mut f: F,
+    ) -> Result<TreeNodeRecursion> {
+        fn recurse<F: FnMut(&Arc<dyn ExecutionPlan>) -> Result<TreeNodeRecursion>>(
+            plan: &Arc<dyn ExecutionPlan>,
+            f: &mut F,
+        ) -> Result<TreeNodeRecursion> {
+            f(plan)?.visit_children(|| {
+                if let Some(hash_join) = plan.downcast_ref::<HashJoinExec>() {
+                    recurse(hash_join.right(), f)
+                } else if let Some(nested_loop_join) = plan.downcast_ref::<NestedLoopJoinExec>() {
+                    recurse(nested_loop_join.right(), f)
+                } else if let Some(cross_join) = plan.downcast_ref::<CrossJoinExec>() {
+                    recurse(cross_join.right(), f)
+                } else {
+                    plan.children()
+                        .into_iter()
+                        .apply_until_stop(|child| recurse(child, f))
+                }
+            })
+        }
+
+        recurse(self, &mut f)
     }
 
     fn transform_down_with_dt_ctx<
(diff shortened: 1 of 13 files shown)


benchmarks/src/datasets/common.rs
```diff
@@ -77,12 +77,13 @@ pub async fn register_tables(
...
-            )
-            .await?;
+            let _ = ctx
+                .register_parquet(
+                    table_name,
+                    path.to_str().unwrap(),
+                    ParquetReadOptions::default(),
+                )
+                .await;
         }
```
gabotechs, line 87: Unrelated to this PR, but while running benches I noticed this started failing. I think it's related to when we upgraded to DF 54

benchmarks/src/run.rs
```diff
@@ -92,10 +92,6 @@ pub struct RunOpt {
     #[structopt(long)]
     children_isolator_unions: bool,
 
-    /// Turns on broadcast joins.
-    #[structopt(long = "broadcast-joins")]
-    broadcast_joins: bool,
-
     /// Collects metrics across network boundaries
     #[structopt(long)]
```
gabotechs: Also unrelated to this PR, but Broadcast JOINs are mature enough now to enable them always in the benches at least.

src/execution_plans/sampler.rs
```diff
@@ -28,15 +31,7 @@ use tokio::sync::oneshot;
...
-const READY_CHUNK_LIMIT: usize = 256;
-/// Maximum read of bytes per second allowed to be emitted. Reads greater than this will be
-/// truncated to this max value, as it's assumed that [READY_CHUNK_LIMIT] was hit and no useful
-/// measurement can actually be emitted.
-const MAX_BYTES_PER_SECOND: usize = 512 * 1024 * 1024;
-/// Maximum number of rows per second allowed to be emitted. Reads greater than this will be
-/// truncated to this max value, as it's assumed that [READY_CHUNK_LIMIT] was hit and no useful
-/// measurement can actually be emitted.
-const MAX_ROWS_PER_SECOND: usize = 1024 * 1024;
+const READY_CHUNK_LIMIT: usize = 1024 * 1024;
```
gabotechs, line 34: Increased this value on purpose. The idea is that we should be able to yield as much as possible here syncrhonously, as these are record batches that are anyways already in-memory, so we prefer to sample as much as possible, given that they are already there.

src/execution_plans/sampler.rs
```diff
@@ -348,53 +311,58 @@ impl LoadInfoDropHandler {
...
+                // by the leaf node metrics is not per-partition, is global to the whole node.
+                //
+                // In a perfect scenario, we might be able to map the current SamplerExec partition
+                // in scope to the specific partition of the leaf node, but several things can
+                // happen in between these two nodes that mangle the partitions. For example, the
+                // presence of a RepartitionExec or an InterleaveExec will mess with partition
+                // mapping from SamplerExec to leaf node.
+                //
+                // Just dividing by `partition_count` gives a good approximation, but it's still
+                // not perfect as it will not account for data skews.
```
asolimando, line 351: This is a perfectly reasonable start, I agree.

As a future improvement, I wonder if we can apply sampling similarly to get an idea of the degree of data skew we have across partitions. wdyt?

src/common/recursion.rs
```diff
@@ -348,6 +409,81 @@ mod tests {
...
+    #[test]
+    fn driver_path_jump_skips_subtree() {
+        let child = single(leaf());
+        let plan = single(Arc::clone(&child));
+        assert_snapshot!(
+            trace_driver_path_with(&plan, |p| {
+                if Arc::ptr_eq(p, &child) { TreeNodeRecursion::Jump } else { TreeNodeRecursion::Continue }
+            }),
+            @r"
+        Single
```
asolimando, line 438: Nit: how hard would it be to have an indentation level for children? it's not a big deal as I can read the plan declaration to see what are the children, but if not hard it would have better readability

src/common/recursion.rs
```diff
@@ -664,6 +855,29 @@ mod tests {
...
+    fn trace_driver_path_with<F: FnMut(&Arc<dyn ExecutionPlan>) -> TreeNodeRecursion>(
+        root: &Arc<dyn ExecutionPlan>,
+        mut decide: F,
+    ) -> String {
+        let mut lines = vec![];
+        root.apply_driver_path(|p| {
+            let rec = decide(p);
+            let suffix = match rec {
+                TreeNodeRecursion::Continue => "",
+                TreeNodeRecursion::Jump => " [->jump]",
```
asolimando, line 869: Q: I am not sure I got the "jump" notation, what does this correspond to in a distributed plan? A recursive step to model DAGs?

src/common/recursion.rs
```diff
@@ -104,6 +135,32 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
...
+            plan: &Arc<dyn ExecutionPlan>,
+            f: &mut F,
+        ) -> Result<TreeNodeRecursion> {
+            f(plan)?.visit_children(|| {
+                if let Some(hash_join) = plan.downcast_ref::<HashJoinExec>() {
+                    recurse(hash_join.right(), f)
+                } else if let Some(nested_loop_join) = plan.downcast_ref::<NestedLoopJoinExec>() {
+                    recurse(nested_loop_join.right(), f)
+                } else if let Some(cross_join) = plan.downcast_ref::<CrossJoinExec>() {
+                    recurse(cross_join.right(), f)
```
asolimando, line 150: Q: why do we only follow the right children for these join nodes?

asolimando (commented):
LGTM, I have a few questions on the recursion (jump notation and chasing only the right child for joins), the rest is solid and I agree that removing the temporal notion from the equation (especially with an unknown total duration) will provide more stable estimates.

src/execution_plans/sampler.rs
gabotechs: I think that should be possible. The amount of rows returned in the leaf nodes can be measured per-partition. The main problem is: how do you attribute rows pulled from a leaf node in a specific partition to another specific partition in the sampler?

In the sampled stage, there can be an arbitrary amount of nodes that shuffle the data (e.g, InterleaveExec, or RepartitionExec(RoundRobin)), if there's at least one of these nodes in between the sampler and the leaf node, it'd be impossible to have a 1:1 mapping between partitions at the sampler vs at the leaf.

One future improvement would be to detect if there's one of these leaf nodes that repartition data in between, and if there isn't, then assume the 1:1 partition mapping, which would account for data skews

src/common/recursion.rs
gabotechs: It's something coming from DataFusion's tree recursion API. It's a signal to the methods that recurse a plan, that we don't want to keep recursing down a subtree, but we are still willing to jump to the next subtree without stopping recursion.

gabotechs: From DF docs:

```rust
#[derive(Debug, PartialEq, Clone, Copy)]
pub enum TreeNodeRecursion {
    /// Continue recursion with the next node.
    Continue,
    /// In top-down traversals, skip recursing into children but continue with
    /// the next node, which actually means pruning of the subtree.
    ///
    /// In bottom-up traversals, bypass calling bottom-up closures till the next
    /// leaf node.
    ///
    /// In combined traversals, if it is the `f_down` (pre-order) phase, execution
    /// "jumps" to the next `f_up` (post-order) phase by shortcutting its children.
    /// If it is the `f_up` (post-order) phase, execution "jumps" to the next `f_down`
    /// (pre-order) phase by shortcutting its parent nodes until the first parent node
    /// having unvisited children path.
    Jump,
    /// Stop recursion.
    Stop,
}
```

src/common/recursion.rs
gabotechs: Just gave this a try, and I remember why I did not go for it in the first place when I made this:

The issue is that with these recursion APIs, it's impossible to know at which depth of the tree you are on, and therefore, we do not know how many spaces should we put at the beginning of each line.

src/common/recursion.rs
gabotechs: All these JOIN nodes have one thing in common:

They collect the build side in memory before emitting any data. This means that data production is only tied to probe side, or in other words, the path the drives data production is the probe side (the right node).

asolimando: Now that you mentions this, it makes total sense, I wonder if a little comment could help other folks down the line too. If it feels evident feel free to disregard, it might just be me.

asolimando (approved):
I understood all the pending points I had, the rest LGTM

src/common/recursion.rs
gabotechs: It's actually as a nice doc-comment in the method declaration (at the top of this file). That way, when people hover in the IDEs through the method, they get a nice display with the explanation of what it does.

gabotechs:
I tried to get a new stage-level metric: `stats-q-error`, but it's not trivial to add, and it might be worth shipping that in a separate PR.

I'll do that separately.

Issue #528, opened 2026-07-02
[EPIC] dynamic filtering


jayshrivastava:
Epic for https://github.com/datafusion-contrib/datafusion-distributed/issues/180.

This is meant to outline any pending work to make dynamic filtering work in all datafusion distributed queries

Started working on a formal RFC here: https://github.com/datafusion-contrib/datafusion-distributed/pull/553/changes

PR #553, opened 2026-07-14
rfc: distributed dynamic filtering

jayshrivastava:
Rendered View: https://github.com/datafusion-contrib/datafusion-distributed/blob/js/dynamic-filtering-design/rfcs/1-distributed-dynamic-filtering.md

Informs https://github.com/datafusion-contrib/datafusion-distributed/issues/528


rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,616 @@
...
+                                                                                            │                                    │                                                                                 
+                                                                            ┌──────────────────────────────┐     ┌──────────────────────────────┐                                                                  
+                                                                            │ ┌──────────────────────────┐ │     │ ┌──────────────────────────┐ │                                                                  
+                                                                            │ │     RepartitionExec      │ │     │ │     RepartitionExec      │ │                                                                  
+                                                                            │ │ partitioning=Hash(a, 12) │ │     │ │ partitioning=Hash(a, 12) │ │
```
jayshrivastava, line 469: The partitioning here is hash(a) % 12 but in the above stage it's hash(a) % 4 assuming 4 target_partitions. So, the hash join will generate 4 cases.

It's actually not correct to merge all of the cases. We need to OR within each case.

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,601 @@
...
+                  Task 1 Runs partitions [0,4)                                         Task 2 Runs partitions [4,8)                                        Task 3 Runs partitions [8,12)                  
+```
+
+### Remote Case
+
+The proposed change is to implement dynamic filtering similarly to Trino.
+
+1. During execution, the dynamic filter producers (`HashJoinExec`, `SortExec`, `AggregateExec`) will send dynamic filter updates to the coordinator.
+2. The coordinator will union / merge the dynamic filter updates from all workers.
+3. The coordinator will send the unioned / merged dynamic filter updates to the consumers (`DataSourceExec`).
```
gabotechs, line 272: 👍 Sounds good, with the current infrastructure, I don't think it would be terrible difficult to do this

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,601 @@
...
+`DynamicFilterPhysicalExpr: a@0 >= 1 AND a@0 <= 10 OR IN LIST [12, 13]`   
+
+To this  
+`DynamicFilterPhysicalExpr: a@0 >= 0 AND a@0 <= 10 OR IN LIST [10, 11, 12, 13]`
+
+Pros:  
+- Less overhead than ORing
+
+Cons:  
+- Is brittle. What if we have to support non-range and non-IN-LIST expressions? It would be nice if dynamic filters in vanilla datafusion natively implemented a `merge` or `union` operation
```
gabotechs, line 316: Why do you think it's brittle? For merging, I imagine you can just rely on Arrow comparison kernels for deduping IN LIST expressions.

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,601 @@
...
+
+The design fundamentally relies on distributed datafusion intercepting and modifying dynamic filter updates. `DynamicFilterPhysicalExpr` already offers APIs
+for this:
+```rust
+// Write Updates
+pub fn update(&self, new_expr: Arc<dyn PhysicalExpr>) -> Result<()>
+
+// Read Updates
+pub fn current(&self) -> Result<Arc<dyn PhysicalExpr>>
+pub(crate) fn subscribe(&self) -> DynamicFilterSubscription
```
gabotechs, line 500: 🤔 `subscribe` not being public might be a problem.

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,601 @@
...
+### 1. Getting `&DynamicFilterPhysicalExpr` from `ExecutionPlan`
+
+Requirements
+(a) identify which dynamic filters are producers and consumers in an `ExecutionPlan` tree  
+(b) get access to all `&DynamicFilterPhysicalExpr` in an `ExecutionPlan` (as long as one is holding a reference to a `DynamicFilterPhysicalExpr`, they can read and write updates to it)  
+(c) allow easy extension for custom `ExecutionPlan` implementations  
+
+#### Options 1 & 2: Support new APIs In Vanilla DataFusion
+
+See https://github.com/apache/datafusion/issues/23814
```
gabotechs, line 536: https://github.com/apache/datafusion/issues/23814 seems to be evolving in a good direction, and people seem engaged in the discussion, so I'd lean towards using whatever comes out of that.

gabotechs (commented):
This is looking very good! I really do not have any major comments, this seems to be going in the right direction.

rfcs/1-distributed-dynamic-filtering.md
jayshrivastava: True. I think we can just use this instead https://github.com/apache/datafusion/blob/3a29d6bd8cc9ac2bf5efee9f070dcdeea9f97b32/datafusion/physical-expr/src/expressions/dynamic_filters/mod.rs?plain=1#L320 since we would wait for completion anyways.

rfcs/1-distributed-dynamic-filtering.md
jayshrivastava: Can you expand on this? It's easy to implement a merge operation if we assume the expressions to merge always look like `a@0 >= 0 AND a@0 <= 5 OR IN LIST [10, 11]`. We would downcast to InListExpr and RangeExpr and define a merge operation for those. 

In the future, what if there's other kinds of expressions in there? Then maybe the merge operation would break. 

Maybe @LiaCastaneda has more insights on this. I think she implemented something similar.

rfcs/1-distributed-dynamic-filtering.md
jayshrivastava: Yup I will stay on top of that issue.

rfcs/1-distributed-dynamic-filtering.md
gabotechs: I imagine that not any arbitrary expression can be placed there. Even if it can, you still decide to optimize the most common cases, like IN LISTs or min/max ranges, and if there's anything else besides does two things just leave it OR-ed.

LiaCastaneda: > In the future, what if there's other kinds of expressions in there? 

I think that rather than adding other kinds of expressions, there is probably a higher possibility that the overall dynamic filter shape changes from the current `CASE` statement to something else.
Regarding expression types, I think the other discussed expression that hasn’t been implemented yet is bloom filters pushdown as dynamic filters. In that case, deduping probably becomes awkward, I’m not really sure how you would merge that (or if its possible). But I think it’s a very niche use case, and as Gabriel said, it can be ORed anyway.

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,601 @@
...
+### Remote Case
+
+The proposed change is to implement dynamic filtering similarly to Trino.
+
+1. During execution, the dynamic filter producers (`HashJoinExec`, `SortExec`, `AggregateExec`) will send dynamic filter updates to the coordinator.
+2. The coordinator will union / merge the dynamic filter updates from all workers.
+3. The coordinator will send the unioned / merged dynamic filter updates to the consumers (`DataSourceExec`).
+
+For correctness, the coordinator must wait for all dynamic filter updates from all workers before sending them to the consumers, otherwise, the
+consumers may prune rows incorrectly.
```
barbarj, line 275: Is this 
A) the coordinator must get an update from each worker at least once before sending the updated expression, and then can update eagerly going forward
B) the coordinator must get an update from each worker at least once _per filter generation_ before sending the next generation?
C) or something else?

Relatedly, this is overly strict for `SortExec`'s TopK dynamic filter, as well as `AggregateExec`'s min/max filter, which can safely be updated eagerly. (I've no problem with correctness then optimization as the sequence for shipping this though)

jayshrivastava: Great point. For hash joins, eager updates aren't safe. The filter effectively contains the build side of the join. If the join is on different workers and each has a different build side, we need filter ORed together to get a global filter which represents the entire build side across workers.

However for sort and aggregate, I think eager updates are very safe. For an aggregate that has `MIN(a)` for example the aggregate on any worker can push down `a < 10` for example. We can OR the expressions across workers ex. `a < 10 OR a < 20` and the minimum value, `10` would win. This mechanism is similar for `SortExec`.

For now, I think we should ensure on correctness / safety in the initial implementation, so I'm proposing option A) for now. I think we can roll out eagerness later. This will mean the dynamic filters on `SortExec` and `AggregateExec` will still be pretty useless for now, but we will be able to unlock joins.  

Since you worked on joins recently, maybe you can confirm if what I said makes sense

barbarj: Yeah, after reading up a bit more on `HashJoinExec`'s filter strategy, that makes sense, and A) is a reasonable general approach. 

We'll want to consider how we can track each expression's producer node's type, as that will be necessary info for enabling eagerness correctly later.

jayshrivastava: > We'll want to consider how we can track each expression's producer node's type, as that will be necessary info for enabling eagerness correctly later.

Agreed. At least, maybe some property to indicate if its safe to apply eager updates or not.

I pushed a short section in the RFC on eager updates so this is documented.

rfcs/1-distributed-dynamic-filtering.md
```diff
@@ -0,0 +1,646 @@
...
+worker adds a filter for `a < 20` and another worker adds `a < 10`, either can safely be applied
+to the data source in any order. These expressions can also be OR-ed together `a < 10 OR a < 20`.
+
+Unlike joins, aggregates are only updated and never marked as complete, so datafusion-distributed
+would need to support eager updates in some way for these filters to work.
+
+#### Sorts  
+These operate similarly to Aggregates and can be applied eagerly.
+
+The proposal in this RFC is to take the safest approach and wait for all filters to complete before
```
shinzoxD, line 528: The preceding paragraph says aggregate filters are never marked complete, so a literal wait for *all* discovered filters can never finish for a plan containing `AggregateExec` (and presumably `SortExec`). Could the initial scope explicitly define the required-completion set as join-produced filters only, and state that filter collection must not block query execution? Otherwise an implementation that registers every producer can either never publish these filters or deadlock while awaiting completion. Non-completing producers can then be added when eager updates land.

shinzoxD (commented):
The new eager-update section leaves one liveness boundary to clarify.

PR #556, opened 2026-07-16
Add stats estimation accuracy to benchmarks

gabotechs:
This PR enhances the benchmarks by comparing estimated bytes vs actual bytes in network boundaries during dynamic planninng:

- Compare sampled bytes with actual network-boundary output bytes using q-error.
- Store it per iteration in local and remote benchmarks.
- Report median q-error (`QERR P50`) in benchmark comparisons.

Preliminary step for #552.

---

Sample output:

```
=== Comparing tpch_sf10 results from branch 'stats-accuracy-benches' [prev] with 'stats-accuracy-benches' [new] ===
os:        macos
arch:      aarch64
cpu cores: 16
threads:   2 -> 2
workers:   8 -> 8
===================================================================================================================
tpch_sf10 q1: prev= 327 ms, new= 355 ms, diff=1.09 slower ✖
tpch_sf10 q2: prev= 271 ms, new= 278 ms, diff=1.03 slower ✖
tpch_sf10 q3: prev= 378 ms, new= 346 ms, diff=1.09 faster ✔
tpch_sf10 q4: prev= 132 ms, new= 129 ms, diff=1.02 faster ✔
tpch_sf10 q5: prev= 617 ms, new= 610 ms, diff=1.01 faster ✔
tpch_sf10 q6: prev= 115 ms, new= 109 ms, diff=1.06 faster ✔
tpch_sf10 q7: prev= 765 ms, new= 743 ms, diff=1.03 faster ✔
tpch_sf10 q8: prev= 781 ms, new= 889 ms, diff=1.14 slower ✖
tpch_sf10 q9: prev=1040 ms, new=1010 ms, diff=1.03 faster ✔
tpch_sf10 q10: prev= 448 ms, new= 455 ms, diff=1.02 slower ✖
tpch_sf10 q11: prev= 132 ms, new= 121 ms, diff=1.09 faster ✔
tpch_sf10 q12: prev= 200 ms, new= 205 ms, diff=1.02 slower ✖
tpch_sf10 q13: prev= 233 ms, new= 232 ms, diff=1.00 faster ✔
tpch_sf10 q14: prev= 158 ms, new= 171 ms, diff=1.08 slower ✖
tpch_sf10 q15: prev= 262 ms, new= 241 ms, diff=1.09 faster ✔
tpch_sf10 q16: prev=  92 ms, new=  94 ms, diff=1.02 slower ✖
tpch_sf10 q17: prev= 833 ms, new= 816 ms, diff=1.02 faster ✔
tpch_sf10 q18: prev= 917 ms, new= 946 ms, diff=1.03 slower ✖
tpch_sf10 q19: prev= 187 ms, new= 199 ms, diff=1.06 slower ✖
tpch_sf10 q20: prev= 422 ms, new= 420 ms, diff=1.00 faster ✔
tpch_sf10 q21: prev=1044 ms, new=1022 ms, diff=1.02 faster ✔
tpch_sf10 q22: prev= 101 ms, new= 103 ms, diff=1.02 slower ✖
   TOTAL: prev=9455 ms, new=9494 ms, diff=1.00 slower ✖
QERR P50: prev=n/a, new=19.47x
```

### DIFF
--- a/benchmarks/src/datasets/common.rs
+++ b/benchmarks/src/datasets/common.rs
@@ -77,12 +77,13 @@ pub async fn register_tables(
         let path = entry?.path();
         if path.is_dir() {
             let table_name = path.file_name().unwrap().to_str().unwrap();
-            ctx.register_parquet(
-                table_name,
-                path.to_str().unwrap(),
-                ParquetReadOptions::default(),
-            )
-            .await?;
+            let _ = ctx
+                .register_parquet(
+                    table_name,
+                    path.to_str().unwrap(),
+                    ParquetReadOptions::default(),
+                )
+                .await;
         }
     }
     Ok(())
--- a/benchmarks/src/run.rs
+++ b/benchmarks/src/run.rs
@@ -25,7 +25,7 @@ use datafusion::common::{config_err, exec_err, not_impl_err};
 use datafusion::datasource::source::DataSourceExec;
 use datafusion::error::{DataFusionError, Result};
 use datafusion::execution::SessionStateBuilder;
-use datafusion::physical_plan::collect;
+use datafusion::physical_plan::{ExecutionPlan, collect};
 use datafusion::prelude::*;
 use datafusion_distributed::test_utils::localhost::LocalHostWorkerResolver;
 use datafusion_distributed::test_utils::work_unit_file_scan::{
@@ -37,6 +37,7 @@ use datafusion_distributed::{
     display_plan_ascii, rewrite_distributed_plan_with_metrics,
 };
 use datafusion_distributed_benchmarks::datasets::{clickbench, register_tables, tpcds, tpch};
+use datafusion_distributed_benchmarks::stats::stats_estimation_q_error;
 use std::error::Error;
 use std::fs;
 use std::path::PathBuf;
@@ -87,14 +88,6 @@ pub struct RunOpt {
     /// Task count scale factor for when nodes in stages change the cardinality of the data
     #[structopt(long)]
     cardinality_task_sf: Option<f64>,
-
-    /// Use children isolator UNIONs for distributing UNION operations.
-    #[structopt(long)]
-    children_isolator_unions: bool,
-
-    /// Turns on broadcast joins.
-    #[structopt(long = "broadcast-joins")]
-    broadcast_joins: bool,
 
     /// Collects metrics across network boundaries
     #[structopt(long)]
@@ -216,9 +209,11 @@ impl RunOpt {
                 "none" => None,
                 v => return config_err!("Unknown compression type {v}"),
             })?
-            .with_distributed_children_isolator_unions(self.children_isolator_unions)?
-            .with_distributed_broadcast_joins(self.broadcast_joins)?
-            .with_distributed_metrics_collection(self.collect_metrics || self.debug)?
+            .with_distributed_children_isolator_unions(true)?
+            .with_distributed_broadcast_joins(true)?
+            .with_distributed_metrics_collection(
+                self.collect_metrics || self.debug || self.dynamic,
+            )?
             .with_distributed_max_tasks_per_stage(self.max_tasks_per_stage)?
             .with_distributed_user_codec(WorkUnitFileScanCodec)
             .with_distributed_task_estimator(WorkUnitFileScanTaskEstimator)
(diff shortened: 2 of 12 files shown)


benchmarks/src/datasets/common.rs
```diff
@@ -77,12 +77,13 @@ pub async fn register_tables(
...
-            )
-            .await?;
+            let _ = ctx
+                .register_parquet(
+                    table_name,
+                    path.to_str().unwrap(),
+                    ParquetReadOptions::default(),
+                )
+                .await;
         }
```
gabotechs, line 87: This is something that started happening on DF 54. Upon looking up folders with parquet files, DF 53 does nothing if the folder does not contain any file, but DF 54 fails.

It's fine for us to have folders without parquet files. For example the .results/ folders in which we dump the benchmarking results.

benchmarks/src/run.rs
```diff
@@ -88,14 +88,6 @@ pub struct RunOpt {
...
 
-    /// Use children isolator UNIONs for distributing UNION operations.
-    #[structopt(long)]
-    children_isolator_unions: bool,
-
-    /// Turns on broadcast joins.
-    #[structopt(long = "broadcast-joins")]
-    broadcast_joins: bool,
-
     /// Collects metrics across network boundaries
```
gabotechs, line 92: Now these two are always enabled. I don't think we want them disabled at any moment, as they are pretty stable now.

benchmarks/src/datasets/common.rs
asolimando: Shall we report this to core DF?

gabotechs: I think it's fine. I'd say the behavior is more correct now that it does not fail silently, which is what was happening before, we just happened to be ok with it

benchmarks/src/run.rs
asolimando: Unless tomorrow we want to follow Spark example and have a global AQE knob to turn it off, at that point we can avoid the overhead as there is no consumer. But we can probably do this later

asolimando: We would probably have the same effect by not inserting `SampleExec` nodes, so it's probably enough to do that tomorrow to disable AQE

benchmarks/src/stats.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// Computes the median q-error between the sampled byte estimate and the actual output bytes for
+/// every dynamically sampled stage boundary in `plan`.
+pub fn stats_estimation_q_error(plan: &Arc<dyn ExecutionPlan>) -> Option<f64> {
+    let mut boundary_q_errors = vec![];
+
+    let _ = plan.apply(|node| {
+        if let Some(boundary) = node.as_network_boundary()
+            && let Stage::Local(input_stage) = boundary.input_stage()
+            && let Some(sampled_bytes) = metric_total(&input_stage.metrics_set, "sampled_bytes")
```
asolimando, line 22: I think bytes will be the dominating factor in most cases, but we estimate row count too, I am OK to start with bytes only but if we see that this doesn't correlate with performance benchmarks, we might need to track row counts similarly.

Similarly for median, it's more stable than avg, and I think it's what we need to track, but we might need to add other percentile values possibly if we see that median is not enough.

asolimando (approved):
LGTM, great work!

benchmarks/src/stats.rs
gabotechs: > I think bytes will be the dominating factor in most cases, but we estimate row count too

Yeah, in fact, for future optimizations, rows will start being relevant, as that's what DF optimizer uses for several things.

As today we don't do those optimizations dynamically, and we just care about bytes for sizing the stages, bytes is the only thing that actually use.

When we start doing the other optimizations, I think we should start taking into account output rows as well.

benchmarks/src/stats.rs
```diff
@@ -0,0 +1,54 @@
...
+            && let Some(actual_bytes) = node
+                .metrics()
+                .and_then(|metrics| metric_total(&metrics, "output_bytes"))
+        {
+            boundary_q_errors.push(q_error(sampled_bytes, actual_bytes));
+        }
+        Ok(TreeNodeRecursion::Continue)
+    });
+
+    median(boundary_q_errors)
```
gabotechs: @asolimando I'm not sure if `median` is the best here... it has the chance of swallowing very bad estimations in cases there are enough straight-forward stages in the plan to even that out.

I'm thinking that avg might be a bit better. WDYT?

gabotechs: I'm getting some more meaningful metrics with a `mean` rather than `median` here, I'm going to go for it.

asolimando: I briefly touched upon this [here](https://github.com/datafusion-contrib/datafusion-distributed/pull/556/changes#r3593972133) and I agree that median, being robust to outliers, will "absord" degenerate cases that might result in individual queries to fail.

Being robust to outliers is a good property for this type of measures, avg would suffer from the opposite problem, it won't provide a clear signal for the "regular" case and conflate the two behaviors.

To track also outliers we need to complement median with P90 or P95, exactly as https://vldb.org/pvldb/vol9/p204-leis.pdf does (look at Table 1 page 4 for an example).

Median will already provide a good signal for most cases but as we move forward and refine the approach we will surely need more to avoid regressions for extreme cases.

If that's easy enough to add now, it would be safer, if you don't feel like adding it now, we can wait and see if better q-error median values positively correlates with improvements in benchmark results, otherwise we need extra signals.

gabotechs: 🤔 this is a good call... it's not trivial to do, but it's worth it, because I'm seeing that P50 is swallowing pretty wild overestimation, just because there's a few stages that are trivial to estimate (because 100% of the data was sampled for example)

gabotechs: Even if it's not super trivial, I'm going to go with P95 additionally, even if that will bloat a bit this PR.

gabotechs: Done, P50 and P95 looking very good now, thanks for the suggestion!

PR #559, opened 2026-07-20
Local worker allocations

gabotechs:
Small optimization that allows the dynamic planner to collocate single-tasked stages into the same machine that is acting as coordinator.

During dynamic planning, if all stages are small, they will just use 1 task, and before this PR, there was already some logic for co-locating all stages in a single worker. The challenge:

The single worker was a random worker in the cluster.

This means that if the coordinating context also happens to be in the scope of a worker, there's a missed chance for running everything locally, almost as if it was just single-node DataFusion.

This PRs unlocks this scenario by allowing users to inject their own `LocalWorkerContext` into the coordinating `SessionContext`. That way, the coordinator itself can recognize it's also a worker, and can send to itself all these single-tasked stages, avoiding network hops and data serialization.

## Benchmarks

TPCH-SF1: 1.12 faster ✔
TPCH-SF10: 1.06 faster ✔
TPCH-SF100: 1.01 faster ✔
TPC-DS: 1.08 faster ✔
ClickBench: 1.01 faster ✔

### DIFF
--- a/src/test_utils/localhost.rs
+++ b/src/test_utils/localhost.rs
@@ -61,12 +61,14 @@ where
         });
     }
     tokio::time::sleep(Duration::from_millis(100)).await;
+    let first_worker_url = Url::parse(&format!("http://127.0.0.1:{}", ports[0])).unwrap();
 
     let worker_resolver = LocalHostWorkerResolver::new(ports);
     let state = SessionStateBuilder::new()
         .with_default_features()
         .with_config(SessionConfig::new().with_target_partitions(3))
         .with_distributed_planner()
+        .with_distributed_local_worker_context(workers[0].to_local_worker_context(first_worker_url))
         .with_distributed_worker_resolver(worker_resolver)
         // Test datasets are tiny, so budget one byte per partition: the estimator then asks for far
         // more partitions than exist, which gets capped at the worker count, fanning every scan out
--- a/tests/local_connections.rs
+++ b/tests/local_connections.rs
@@ -1,0 +1,51 @@
+#[cfg(test)]
+mod tests {
+    use datafusion::common::internal_err;
+    use datafusion::common::tree_node::{TreeNode, TreeNodeRecursion};
+    use datafusion::physical_plan::collect;
+    use datafusion_distributed::test_utils::localhost::start_localhost_context;
+    use datafusion_distributed::test_utils::parquet::register_parquet_tables;
+    use datafusion_distributed::{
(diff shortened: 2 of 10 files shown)


tests/local_connections.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        let query =
+            r#"SELECT count(*), "RainToday" FROM weather GROUP BY "RainToday" ORDER BY count(*)"#;
+
+        let df = ctx.sql(query).await?;
+        let plan = df.create_physical_plan().await?;
+        collect(Arc::clone(&plan), ctx.task_ctx()).await?;
+        let format = DistributedMetricsFormat::Aggregated;
+        let plan = rewrite_distributed_plan_with_metrics(plan, format).await?;
+        println!("{}", display_plan_ascii(plan.as_ref(), true));
```
jayshrivastava, line 32: stray println

src/test_utils/localhost.rs
```diff
@@ -61,12 +61,14 @@ where
...
     }
     tokio::time::sleep(Duration::from_millis(100)).await;
+    let first_worker_url = Url::parse(&format!("http://127.0.0.1:{}", ports[0])).unwrap();
 
     let worker_resolver = LocalHostWorkerResolver::new(ports);
     let state = SessionStateBuilder::new()
         .with_default_features()
         .with_config(SessionConfig::new().with_target_partitions(3))
         .with_distributed_planner()
+        .with_distributed_local_worker_context(workers[0].to_local_worker_context(first_worker_url))
```
jayshrivastava, line 71: Nit: Do you think `CoordinatorContext` is more clear than `LocalWorkerContext` because it only exists on the coordinator?

jayshrivastava (approved):

tests/local_connections.rs
gabotechs: Yeah, this one is on purpose, like the ones we have on metrics_collection.rs, for local iteration is nice to see the plan in the output.

src/test_utils/localhost.rs
gabotechs: It also exists on workers actually, during a query, in each remote worker, we populate the SessionContext with this same information:

https://github.com/datafusion-contrib/datafusion-distributed/blob/e240854287c3ca98a110897e4d25730eb12fb95e/src/worker/impl_coordinator_channel.rs#L62-L66

This way, workers can recognize when they need to reach themselves, and fallback to a local connection, rather than a remote one. This is the same, but in the coordinator, in case it also happens to be a worker.

PR #560, opened 2026-07-21
Add stats-q-error to stage level metrics

gabotechs:
Adds a `stats_q_error` metric at the stage level that computes an error rate between the estimated stats VS what actually happened, taken from https://www.vldb.org/pvldb/vol2/vldb09-657.pdf.

I find the current approach to be a bit messy...

It requires a lot of code for wiring this up. There's something special about this metric: is a metric attached to the `Stage` rather than a specific node, and it's collected after runtime finished. 

Until now, all the metrics we stamp at the `Stage` level are known during planning (either static or dynamic). However, this time, we need to wait for the full execution to finish.

I'm very much open to other alternatives, as what I have here, is just too much code for just showing a metric.

@jayshrivastava any ideas on how to make this cleaner?

### DIFF
--- a/src/common/record_batch.rs
+++ b/src/common/record_batch.rs
@@ -1,0 +1,10 @@
+use datafusion::arrow::array::RecordBatch;
+
+/// Returns the logical size of a batch's slices, excluding unused backing-buffer capacity.
+pub(crate) fn logical_record_batch_size(batch: &RecordBatch) -> usize {
+    batch
+        .columns()
+        .iter()
+        .map(|column| column.to_data().get_slice_memory_size().unwrap_or(0))
+        .sum()
+}
--- a/src/metrics/q_error_metric.rs
+++ b/src/metrics/q_error_metric.rs
@@ -1,0 +1,109 @@
+use datafusion::physical_plan::Metric;
+use datafusion::physical_plan::metrics::{CustomMetricValue, MetricValue};
+use std::any::Any;
+use std::borrow::Cow;
+use std::fmt::{Debug, Display, Formatter};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering::Relaxed};
+
+/// A floating-point q-error metric.
+///
+/// When multiple values are aggregated, the largest q-error is retained so a bad estimate is not
+/// hidden. A q-error of `1.0` represents a perfect estimate and is therefore the neutral value.
+#[derive(Clone)]
+pub struct QErrorMetric {
+    value: Arc<AtomicU64>,
+}
+
+impl Default for QErrorMetric {
+    fn default() -> Self {
+        Self::from_value(1.0)
+    }
+}
+
+impl QErrorMetric {
+    pub fn new_metric(name: impl Into<Cow<'static, str>>, value: f64) -> Arc<Metric> {
+        Arc::new(Metric::new(
+            MetricValue::Custom {
+                name: name.into(),
+                value: Arc::new(Self::from_value(value)),
+            },
+            None,
+        ))
+    }
+
+    pub fn from_value(value: f64) -> Self {
+        Self {
+            value: Arc::new(AtomicU64::new(value.to_bits())),
+        }
+    }
+
+    pub fn value(&self) -> f64 {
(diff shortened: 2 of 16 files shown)


src/common/record_batch.rs
```diff
@@ -0,0 +1,10 @@
+use datafusion::arrow::array::RecordBatch;
+
+/// Returns the logical size of a batch's slices, excluding unused backing-buffer capacity.
+pub(crate) fn logical_record_batch_size(batch: &RecordBatch) -> usize {
```
jayshrivastava, line 4: nit: bytes in the name?

src/metrics/q_error_metric.rs
```diff
@@ -0,0 +1,109 @@
...
+impl Debug for QErrorMetric {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("QErrorMetric").field(&self.value()).finish()
+    }
+}
+
+impl Display for QErrorMetric {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let value = format!("{:.2}", self.value());
+        write!(f, "{}x", value.trim_end_matches('0').trim_end_matches('.'))
```
jayshrivastava, line 55: nit: maybe drop the "x". Because a qerror of 2 could mean you're off by 2x or by (1/2)x. Also

src/metrics/task_metrics_rewriter.rs
```diff
@@ -92,6 +94,51 @@ pub async fn rewrite_distributed_plan_with_metrics(
     plan.with_new_children(vec![transformed.data])
 }
 
+fn stage_metrics(
```
jayshrivastava, line 97: Yeah I see what you mean by messy. It's actually not that bad IMO so I won't block, but here's a suggestion:

When do we need to read the q_error? If it's after the query has finished (ie. we don't need it for AQE), we could try to hook up the `ESTIMATED_OUTPUT_BYTES_METRIC` to the `SampleExec` some how (just keep Arc clones of the metric in the `SampleExec`). `SampleExec::metrics()` can lazily calculate the q_error and append it. This way, the  `ESTIMATED_OUTPUT_BYTES_METRIC` and other stage-level metrics will be passed via the standard `drain_pending_tasks` path and available in the plan node metrics. It's not hard to implement and can be tested via the `test_dynamic_stats_q_error_is_collected_at_stage_level` test below.

The only downside is that each `ESTIMATED_OUTPUT_BYTES_METRIC` exists twice. Once in the `LocalStage` as an aggregated value and then N times in several `SamplerExecs`. Maybe we get rid of `LocalStage::metrics_set` if possible and have special casing at display time?

`LocalStage` is a bit confusing. It exists during planning, then during execution it's a `RemoteStage`, and then after the query runs, for displaying, we reconstruct the `LocalStage` but this time with metrics. Maybe we need a `DisplayableStage` variant which has 2 fields, a `LocalStage` and a `metrics_set`.

jayshrivastava (approved):

src/metrics/q_error_metric.rs
gabotechs: Q-error is always >1 so 2x will always mean either the estimation is 2x bigger or 2x smaller.

src/metrics/task_metrics_rewriter.rs
gabotechs: > When do we need to read the q_error?

We need to read it once execution has fully completed, this is what gets challenging. Until now, all the stage-level metrics we have ever reported are known right before execution, but never after execution.

I think this is the broader problem and why I don't love what I'm doing in this PR: we don't have an easy way of registering post-runtime stage-level metrics.

I don't think I'm going to merge this PR as is, seems very little value for the complexity it's adding, I'd prefer first to create the proper infrastructure for registering arbitrary post-runtime stage-level metrics, and then maybe revisit this.

gabotechs:
I think the ratio of value/complexity of this implementation is not worth it, and I'm not planning on investing more time into this for now, so closing indefinitely.

PR #565, opened 2026-07-24
Move children isolator union insertion before distribution

### LINKED ISSUE #539: Move `ChildrenIsolatorUnionExec` injection out of the `inject_network_boundaries` function
There's one architectural decision that has worked really well in this project:

While injecting network boundaries, the single-node plan is already fully formed, good for a single-node execution, and with all the appropriate `ExecutionPlan`s already in place.

This rule is broken by the way we currently inject the `ChildrenIsolatorUnionExec` nodes, which are inserted in one of the top-down recursive passes of the distributed planner.

It would be nice to refactor this, and do something similar that:
1. Injects the `ChildrenIsolatorUnionExec` before the distributed planner injects any network boundary (similar to how we inject `BroadcastExec` beforehand).
2. Make `inject_network_boundaries` react to the presence of a `ChildrenIsolatorUnionExec` and do the per-task children assignment there.

alexanderbianchi:
## What changed

- Add a pre-pass that replaces `UnionExec` with `ChildrenIsolatorUnionExec` before annotation and network-boundary distribution.
- Keep the pre-inserted placeholder single-node-correct by mapping all children into one default task.
- Make annotation react to pre-inserted `ChildrenIsolatorUnionExec` for child task-count propagation without constructing the node there.
- Finalize the per-task child map during `distribute_plan`, where final task counts are available.

## Why

Fixes #539. This keeps union-isolator insertion out of the network-boundary distribution recursion while preserving the existing per-task assignment behavior.

## Validation

- `cargo fmt`
- `cargo test --lib`
- `cargo test --lib insert_children_isolator_union`
- `cargo test --features integration

### DIFF
--- a/src/execution_plans/children_isolator_union.rs
+++ b/src/execution_plans/children_isolator_union.rs
@@ -129,6 +129,15 @@ impl ChildWeight {
 }
 
 impl ChildrenIsolatorUnionExec {
+    /// Creates a single-node union placeholder with every child assigned to the default task.
+    pub(crate) fn new_single_task(
+        children: impl IntoIterator<Item = Arc<dyn ExecutionPlan>>,
+    ) -> Result<Self, DataFusionError> {
+        let children = children.into_iter().collect_vec();
+        let child_count = children.len();
+        Self::from_children_and_weights(children, vec![ChildWeight::desired(1.0); child_count], 1)
+    }
+
     pub(crate) fn from_children_and_weights(
         children: impl IntoIterator<Item = Arc<dyn ExecutionPlan>>,
         children_weights: impl IntoIterator<Item = ChildWeight>,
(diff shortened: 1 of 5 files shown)


src/execution_plans/children_isolator_union.rs
```diff
@@ -129,6 +129,15 @@ impl ChildWeight {
 }
 
 impl ChildrenIsolatorUnionExec {
+    /// Creates a single-node union placeholder with every child assigned to the default task.
+    pub(crate) fn new(
```
gabotechs: I'd probably call this "new_single_task" or something more descriptive.

gabotechs (approved):
Nice! I was expecting this to be a bit more complicated, but it turns out it was simpler. 👍 good work.

gabotechs (approved):
👍 Looks good, thanks @alexanderbianchi!

PR #540, opened 2026-07-07
upgrade to datafusion 55

jayshrivastava:
## Summary 

Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/530

- This PR updates the upstream datafusion SHA to the HEAD of https://github.com/apache/datafusion/commits/branch-55/ (edit: this branch is continuously being updated. I will make sure this PR is at the head before merging)
- Rust upgade to 1.94

## Changes

1. In `src/protobuf/distributed_codec.rs` we now use the `proto_converter` argument during serde 
- We still don't use the `DeduplicatingProtoConverter`, so dynamic filters don't necessarily work. I think this is outside the scope of this PR will be addressed in https://github.com/datafusion-contrib/datafusion-distributed/pull/623, which will be rebased after the upgrade. 

3. `ExecutionPlan::apply_expressions` is added for every custom `ExecutionPlan` in this repo
- Wrapper types (`MetricsWrapperExec`, `WorkUnitFileScanConfig`, `DistributedLeafExec`) delegate to the inner type
- Other plans take`TreeNodeRecursion::Continue` because they have no expressions (ex. `SamplerExec`)
- Note that `apply_expressions` does not need to yield sort or partitioning expressions in the plan properties

3. We migrate from `partition_statistics` to `statistics_from_inputs` for every `ExecutionPlan`. 
- `src/distributed_planner/statistics/plan_statistics.rs` can just use `statistics_from_inputs` directly instead of doing the `StatisticsWrapper` workaround.

5. Range partitioning is now supported.
- CPU costing now includes range-key comparison cost and has a new unit test. See src/distributed_planner/
   statistics/complexity_cpu.rs:238.
- I think there's open questions about range partitioning. I've opened an issue here to make sure it behaves as expected after the upgrade: https://github.com/datafusion-contrib/datafusion-distributed/issues/628#issue-5134085888

6. Peak-memory metrics use the existing gauge wire representation.

   DataFusion added MetricValue::PeakMemoryUsage. It is serialized as the existing named-gauge protobuf variant to avoid a wire-format change. See src/protocol/grpc/
   metrics_proto.rs:124.

   The value and name survive, and aggregation is still additive, but decoding produces a generic Gauge, not PeakMemoryUsage. The practical difference is mainly display formatting: it
   may render as a count rather than human-readable bytes. This is the clearest remaining compromise/risk in the upgrade.

7. File-scan rebalancing changed its discriminator.

[...]

PR #566, opened 2026-07-26
Optional serialization/deserialization for WorkerChannel

gabotechs:
Adds some small changes to the `WorkerChannel` contract so that [de]serialization is optional.  In-memory implementations might not want to pay the overhead of serializing plans.

Details about the changes explained in individual comments guiding reviews.

### DIFF
--- a/src/common/maybe_encoded.rs
+++ b/src/common/maybe_encoded.rs
@@ -1,0 +1,121 @@
+use crate::DistributedCodec;
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::common::{Result, internal_err};
+use datafusion::execution::TaskContext;
+use datafusion::physical_expr::Partitioning;
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion_proto::physical_plan::from_proto::parse_protobuf_partitioning;
+use datafusion_proto::physical_plan::to_proto::serialize_partitioning;
+use datafusion_proto::physical_plan::{
+    AsExecutionPlan, DefaultPhysicalProtoConverter, PhysicalPlanDecodeContext,
+};
+use datafusion_proto::protobuf;
+use datafusion_proto::protobuf::proto_error;
+use prost::Message;
+use std::sync::Arc;
+
+/// A value that a transport may either leave encoded or materialize in memory.
+/// Users are free to pass [MaybeEncoded::Encoded] or [MaybeEncoded::Decoded] at any
+/// moment and Distributed DataFusion's code will internally know how to handle it.
+#[derive(Clone)]
+pub enum MaybeEncoded<T> {
+    Encoded(Vec<u8>),
+    Decoded(T),
+}
+
+impl<T> MaybeEncoded<T> {
+    /// Returns the decoded variant:
+    /// - If in `Decoded` state, it just passes through the content.
+    /// - If in `Encoded` state, it decodes using the provided callback.
+    pub(crate) fn decode_with(self, decode: impl FnOnce(Vec<u8>) -> Result<T>) -> Result<T> {
+        match self {
+            Self::Encoded(encoded) => decode(encoded),
+            Self::Decoded(decoded) => Ok(decoded),
+        }
+    }
+
+    /// Returns the decoded variant:
+    /// - If in `Decoded` state, it just passes through the content.
+    /// - If in `Encoded` state, it throws an error.
+    pub(crate) fn try_decoded(self) -> Result<T> {
+        match self {
+            Self::Encoded(_) => {
+                internal_err!(
+                    "Expected MaybeDecoded::Decoded({}), but got MaybeEncoded::Decoded",
+                    std::any::type_name::<T>()
+                )
+            }
+            Self::Decoded(decoded) => Ok(decoded),
+        }
+    }
+}
+
+impl MaybeEncoded<Arc<dyn ExecutionPlan>> {
+    /// Returns the encoded [ExecutionPlan] as protobuf bytes:
+    /// - If in `Decoded` state, it encodes it using the codecs registered in the [TaskContext].
+    /// - If in `Encoded` state, it just passes through the content.
+    pub fn encode(self, ctx: &Arc<TaskContext>) -> Result<Vec<u8>> {
+        match self {
+            Self::Encoded(encoded) => Ok(encoded),
+            Self::Decoded(plan) => {
+                let codec = DistributedCodec::new_combined_with_user(ctx.session_config());
+                protobuf::PhysicalPlanNode::try_from_physical_plan(plan, &codec)
+                    .map(|v| v.encode_to_vec())
+            }
+        }
+    }
+
+    /// Returns the decoded [ExecutionPlan].
+    /// - If in `Decoded` state, it just passes through the content.
+    /// - If in `Encoded` state, it decodes it using the codecs registered in the [TaskContext].
+    pub(crate) fn decode(self, task_ctx: &TaskContext) -> Result<Arc<dyn ExecutionPlan>> {
+        self.decode_with(|encoded| {
+            let codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
+            let proto_node = protobuf::PhysicalPlanNode::try_decode(encoded.as_ref())?;
+            proto_node.try_into_physical_plan(task_ctx, &codec)
+        })
+    }
+}
+
+impl MaybeEncoded<Partitioning> {
+    /// Returns the encoded [Partitioning] as protobuf bytes:
+    /// - If in `Decoded` state, it encodes it using the codecs registered in the [TaskContext].
+    /// - If in `Encoded` state, it just passes through the content.
+    pub fn encode(self, ctx: &Arc<TaskContext>) -> Result<Vec<u8>> {
+        match self {
+            Self::Encoded(encoded) => Ok(encoded),
+            Self::Decoded(partitioning) => {
+                let codec = DistributedCodec::new_combined_with_user(ctx.session_config());
+                Ok(serialize_partitioning(
+                    &partitioning,
+                    &codec,
+                    // I think nobody cares about this being the default PhysicalProtoConverter.
+                    // If someone does, please open an issue.
+                    &DefaultPhysicalProtoConverter {},
+                )?
+                .encode_to_vec())
+            }
+        }
+    }
+
+    /// Returns the decoded [Partitioning].
+    /// - If in `Decoded` state, it just passes through the content.
+    /// - If in `Encoded` state, it decodes it using the codecs registered in the [TaskContext].
+    pub fn decode(self, schema: SchemaRef, task_ctx: &TaskContext) -> Result<Partitioning> {
+        self.decode_with(|encoded| {
+            let proto_partitioning = protobuf::Partitioning::decode(encoded.as_slice())
+                .map_err(|err| proto_error(err.to_string()))?;
+            let codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
+            let decode_ctx = PhysicalPlanDecodeContext::new(task_ctx, &codec);
+            parse_protobuf_partitioning(
+                Some(&proto_partitioning),
+                &decode_ctx,
+                &schema,
+                // I think nobody cares about this being the default PhysicalProtoConverter.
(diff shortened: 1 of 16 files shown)


src/coordinator/query_coordinator.rs
```diff
@@ -193,15 +187,22 @@ impl<'a> StageCoordinator<'a> {
...
             let mut client = channel_resolver.get_worker_client_for_url(&url).await?;
             let mut worker_to_coordinator_stream = client
-                .coordinator_channel(headers, coordinator_to_worker_stream)
+                .coordinator_channel(
+                    headers,
+                    set_plan_request,
+                    coordinator_to_worker_stream,
+                    &task_ctx,
+                    metrics_set,
+                )
```
gabotechs, line 203: Now we need to pass two extra things here:
1. the `SetPlanRequest`. Before, this was sent as the first message in the stream, this being an invariant of the coordinator->worker channel: the first message always needs to be the `SetPlanRequest`. This was an  artificial invariant imposed by gRPC spec limitations: in gRPC you cannot send both a stream and a unary message at the same time. As this is specific to gRPC, handling this behavior was moved under the `grpc` scope, and the `WorkerChannel` protocol accepts now the two things separately the `CoordinatorToWorkerStream` and the `SetPlanRequest`.
2. An `ExecutionPlanMetricsSet`. The `coordinator_channel` implementations now decide if the inner messages need to be deserialized or can be passed in-memory, so some metrics like `plan_bytes_sent` are now implementation specific (an in-memory implementation will never even serialize to bytes). For this, `ExecutionPlanMetricsSet` is passed to give the chance to implementations to contribute their own metrics.

src/distributed_planner/network_boundary.rs
```diff
@@ -74,49 +69,13 @@ impl NetworkBoundaryExt for dyn ExecutionPlan {
 }
 
 impl ProducerHead {
-    pub(crate) fn to_spec(&self, cfg: &SessionConfig) -> Result<ProducerHeadSpec> {
```
gabotechs, line 77: Now, we can get read of the intermediary `ProducerHeadSpec`, as this existed just for bridging a gap that no longer exists: the original `ProducerHead` can just hold a reference to `MaybeEncoded` and handle both in-memory and serialized formats.

More information about what this was needed before in point 1 of the PR description in https://github.com/datafusion-contrib/datafusion-distributed/pull/512#issue-4768026556.

src/distributed_planner/network_boundary.rs
```diff
@@ -135,9 +94,10 @@ impl ProducerHead {
                 let partitions = input.output_partitioning().partition_count();
                 Arc::new(BroadcastExec::new(input, output_partitions / partitions))
             }
-            ProducerHead::RepartitionExec { partitioning } => {
-                Arc::new(RepartitionExec::try_new(input, partitioning)?)
-            }
+            ProducerHead::RepartitionExec { partitioning } => Arc::new(RepartitionExec::try_new(
+                input,
+                partitioning.try_decoded()?,
```
gabotechs, line 99: I don't love this. This is a runtime failure that was irrepresentable before, but now we need this `try_decoded()?` runtime check because we cannot proceed here if the `ProducerHead` contains a serialized `Partitioning` struct.

It allows to remove so much code that I think it's worth it.

src/protocol/grpc/worker_service.rs
```diff
@@ -188,8 +201,10 @@ fn decode_coordinator_to_worker_msg(
             .inner
             .ok_or_else(missing("CoordinatorToWorkerMsg.inner"))?
         {
-            pb::coordinator_to_worker_msg::Inner::SetPlanRequest(request) => {
-                CoordinatorToWorkerMsg::SetPlanRequest(decode_set_plan_request(request)?)
+            pb::coordinator_to_worker_msg::Inner::SetPlanRequest(_) => {
+                return Err(Status::invalid_argument(
+                    "SetPlanRequest must be the first coordinator message",
+                ));
             }
```
gabotechs, line 208: As the fact that `SetPlanRequest` is the first message in the stream is an invariant imposed by gRPC itself, it's now checked here, instead of the outside of the `grpc` scope

gabotechs:
cc @stuhood and @mdashti, this should be complimentary to https://github.com/datafusion-contrib/datafusion-distributed/pull/548. With this, an in-memory implementation of the `WorkerChannel` no longer needs to pay the price of serialization, keeping everything in-memory.

stuhood:
> cc @stuhood and @mdashti, this should be complimentary to #548. With this, an in-memory implementation of the `WorkerChannel` no longer needs to pay the price of serialization, keeping everything in-memory.

Interesting, thanks. Is there some other motivation for this beyond the in-memory implementation?

While the in-memory implementation is important for test/smoke coverage, I think that all practical implementations will need to serialize: ours certainly does.

Knowing what the other use cases are would help justify the added complexity.

gabotechs:
It's exclusively for the in-memory implementation. For example, if the coordinating context decides that a stage does not need to run in more than one task, it can co-locate it in itself, using the in-memory WorkerChannel instead of paying serialization.

My intention is that the in-memory implementation is not just something that runs for tests, it's part of the production machinery that triggers every time the coordinator or a worker needs to call itself.

src/common/maybe_encoded.rs
```diff
@@ -0,0 +1,121 @@
...
+impl<T> MaybeEncoded<T> {
+    /// Returns the decoded variant:
+    /// - If in `Decoded` state, it just passes through the content.
+    /// - If in `Encoded` state, it decodes using the provided callback.
+    pub(crate) fn decode_with(self, decode: impl FnOnce(Vec<u8>) -> Result<T>) -> Result<T> {
+        match self {
+            Self::Encoded(encoded) => decode(encoded),
+            Self::Decoded(decoded) => Ok(decoded),
+        }
+    }
```
stuhood, line 35: So, this ends up shaped a lot like a `OnceLock`.

Could/should `MaybeEncoded` wrap a `OnceLock` to allow for in-place lazy decoding? A vague concern I have is that because decoding the `MaybeEncoded` doesn't cache the result, different consumers might end up decoding multiple times?

stuhood (approved):
Thanks!

src/common/maybe_encoded.rs
gabotechs: There's one subtle difference with `OnceLock`:

This method consumes the `MaybeEncoded`, and returns the decoded type. This means we are enforcing at compile time that there cannot be multiple consumers, once `decode_with` is called, there's no longer a `MaybeEncoded` enum anymore.

stuhood: Mm, true. Which doesn't allow using it as a struct field. Maybe fine?

src/protocol/grpc/worker_client.rs
```diff
@@ -65,6 +77,10 @@ impl WorkerChannel for pb::worker_service_client::WorkerServiceClient<BoxCloneSy
             .map(|msg| decode_worker_to_coordinator_msg(msg?))
             .boxed();
 
+        MetricBuilder::new(&metrics)
+            .bytes_counter("plan_bytes_sent")
```
jayshrivastava, line 81: Missing the `DISTRIBUTED_DATAFUSION_TASK_ID_LABEL`

src/coordinator/query_coordinator.rs
```diff
@@ -143,28 +143,23 @@ impl<'a> StageCoordinator<'a> {
...
             stage_id: self.stage_id,
             task_number: task_i,
         };
 
-        let msg = CoordinatorToWorkerMsg::SetPlanRequest(SetPlanRequest {
+        let set_plan_request = SetPlanRequest {
             task_key,
             task_count: self.task_count,
-            plan_proto,
+            plan: MaybeEncoded::Decoded(specialized),
```
jayshrivastava, line 158: I think there's a bug with work unit feeds. We may skip serde for `ExecutionPlan` which means we may also skip serde for work unit feeds. This means that we have a local work unit feed both on the coordinator and the worker.

coordinator:
The coordinator `WorkUnitFeed`
https://github.com/datafusion-contrib/datafusion-distributed/blob/7dd7649340a1e5b7506cab268607377b413cc99d/src/coordinator/query_coordinator.rs?plain=1#L308

worker: 
https://github.com/datafusion-contrib/datafusion-distributed/blob/7dd7649340a1e5b7506cab268607377b413cc99d/src/test_utils/work_unit_file_scan.rs?plain=1#L130

Since there's no `RemoteWorkUnitFeed`, there's nothing consuming the coordinator work units, so they get lost.

It would be nice to have test coverage to catch these sort of bugs.

src/common/maybe_encoded.rs
```diff
@@ -0,0 +1,121 @@
...
+                .map_err(|err| proto_error(err.to_string()))?;
+            let codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
+            let decode_ctx = PhysicalPlanDecodeContext::new(task_ctx, &codec);
+            parse_protobuf_partitioning(
+                Some(&proto_partitioning),
+                &decode_ctx,
+                &schema,
+                // I think nobody cares about this being the default PhysicalProtoConverter.
+                // If someone does, please open an issue.
+                &DefaultPhysicalProtoConverter {},
```
jayshrivastava, line 116: This is a bit of a smell. I wonder if we can just represent the producer head as an `ExecutionPlan` which serializes the partitioning rather than passing the `ProducerHead` enum.

src/coordinator/query_coordinator.rs
gabotechs: We do have good test coverage for work unit feeds, this would have been catched in the integration tests, however, this buggy path is not exercised because this PR does not ship code that can trigger that scenario.

In this PR, the option of sending in-memory work units is introduced, however, it's never exercised because for that we'd need to ship this first:

- https://github.com/datafusion-contrib/datafusion-distributed/issues/576

So it's not failing this PR because this PR just makes the changes at the protocol level, but there's still no code capable of of sending `MaybeEncoded::Decoded` work units, as there's still no in-memory implementation of the `WorkerChannel` anywhere.

My bet is that this should be addressed while implementing https://github.com/datafusion-contrib/datafusion-distributed/issues/576. The most naive thing that comes to mind is to do a serde roundtrip to the plan with a `WorkUnitFeed` present, but maybe there's a better option.

jayshrivastava: Okay I see. I'll make a note in #576.

src/common/maybe_encoded.rs
gabotechs: Well, you can use it as a struct field, but your options upon calling a method are:
1. Upon needing to decode the payload, be prepared to consume a full owned reference of the struct containing the field.
2. Clone the field containing the `MaybeEncoded`, and then decode.

In every case, I'd say owned semantics are the correct ones. If this turns out to be a problem (the Clone is too expensive), an easy way out would be to move `Vec<u8>` to `Bytes`, where clones are cheap.

src/common/maybe_encoded.rs
gabotechs: It is... but it's what we are doing right now in the equivalent places:

https://github.com/datafusion-contrib/datafusion-distributed/blob/7dd7649340a1e5b7506cab268607377b413cc99d/src/codec/distributed_codec.rs#L115-L115

The truth is that the way the `PhysicalProtoConverter` argument in the proto functions is weirdly threaded, because typically the places where you'd want to call it are not the places where it's feasable to have one `PhysicalProtoConverter` available.

In this project, we don't care about this argument, and the way this project exposes proto serialization, I cannot imagine a reason why people using would also care about it.

jayshrivastava: I will fix this later. The df55 upgrade adds the converter to `PhysicalExtensionCodec::try_encode/try_decode` so we should be able to plumb it into `parse_stage_proto` and get it here. We want to avoid the default converter to make sure dynamic filters are serialized correctly.

jayshrivastava (approved):

src/common/maybe_encoded.rs
jayshrivastava: In https://github.com/datafusion-contrib/datafusion-distributed/pull/540 after I rebase with main

gabotechs: Nice, thanks!

PR #568, opened 2026-07-27
Define worker plan rewrites in terms of events

gabotechs:
## Stack

1. https://github.com/datafusion-contrib/datafusion-distributed/pull/564
2. https://github.com/datafusion-contrib/datafusion-distributed/pull/567
3. https://github.com/datafusion-contrib/datafusion-distributed/pull/568 <- you are here
4. https://github.com/datafusion-contrib/datafusion-distributed/pull/569
5. https://github.com/datafusion-contrib/datafusion-distributed/pull/570

---

Replaces the worker-side `OnPlanHook` with composable `WorkerPlanRewriteHandler`s. Register handlers in each worker's session-state builder; coordinator registration is not propagated to workers.

```rust
ctx.builder
    .with_distributed_worker_plan_rewrite_handler(|event| {
        Ok(WorkerPlanRewriteEventResponse::new(event.plan))
    })
    .build()
```

PR #571, opened 2026-07-27
migrate event handlers to one generic handler

jayshrivastava:
This change migrates this pattern:
```
SessionStateBuilder::new()
    .with_distributed_event_handler(desired_task_count)
    .with_distributed_event_handler(scale_up_leaf_node)
    .with_distributed_event_handler(route_tasks);
```
to this:
```
SessionStateBuilder::new()
    .with_distributed_event_handler(desired_task_count)
    .with_distributed_event_handler(scale_up_leaf_node)
    .with_distributed_event_handler(route_tasks);
```

The type of handler can be inferred by the argument. Most of the
relevant change are in `src/events`.


src/events/desired_task_count.rs
```diff
@@ -120,37 +104,49 @@ impl TaskCountAnnotation {
...
 {
-    fn handle(&self, ev: DesiredTaskCountEvent) -> Option<DesiredTaskCountEventResponse> {
+    fn handle(&self, ev: DesiredTaskCountEvent<'_>) -> Option<DesiredTaskCountEventResponse> {
         self(ev)
     }
 }
 
-impl DesiredTaskCountHandler for usize {
-    fn handle(&self, ev: DesiredTaskCountEvent) -> Option<DesiredTaskCountEventResponse> {
+impl EventHandler<DesiredTaskCountHandler> for usize {
```
jayshrivastava, line 145: Any usize automatically routes to `EventHandler<DesiredTaskCountHandler>` which is kind of funny.

One day someone could implement `impl EventHandler<MyNewKindOfHandler> for usize` which means rust cannot resolve `with_distributed_event_handler(2)`. You would have to do something like this to get it to work `with_distributed_event_handler::<DesiredTaskCountHandler>(desired_task_count)`

gabotechs: 🤔 This... does not look very good. `with_distributed_event_handler(2)` is too ambiguous. This implementation is only used for tests, if you ask me, I'd just remove it then.

src/events/common.rs
```diff
@@ -1,12 +1,38 @@
...
+    type Response;
 }
 
-impl<H: ?Sized> Default for EventHandlerChain<H> {
+/// Handles [`Event`] data.
+pub trait EventHandler<E: Event>: Send + Sync + 'static {
+    /// Returns `None` when this handler does not accept the event, allowing the next handler in
+    /// the chain to try. Returns `Some` to stop dispatch and select that response.
+    fn handle(&self, ev: E::Data<'_>) -> Option<E::Response>;
+}
```
gabotechs, line 18: There's a challenge with this shape. Some other event handlers, do not need to return an `Option<>`, for example: https://github.com/datafusion-contrib/datafusion-distributed/pull/568/changes#diff-7bd36235787107a21e71a52b4e5511fb7c3fb66c4c6180820a6712c78022c3fe.

One thing that sounds cornering is the fact that all `EventHandler`s need to adhere to the same shape.

Another example is making the `DesiredTaskCountHandler` async:

- https://github.com/datafusion-contrib/datafusion-distributed/pull/569

We'd like that event hander to be `async`, but we want to keep the other handlers synchronous, however, I imagine here either all of them are `async` or none are.

jayshrivastava: Initially, I had 
```
fn handle(&self, ev: E::Data<'_>) -> E::Response;
```
but then that means each `Handler` needs to implement its own `find_map` behavior. It is nice having it here in one spot.
```
impl<E: Event> EventHandlerChain<E> {
    pub(crate) fn handle(&self, ev: E::Data<'_>) -> Option<E::Response> {
        // Give priority to custom handlers registered by users.
        if let Some(res) = self
            .custom
            .iter()
            .find_map(|handler| handler.handle(ev.clone()))
        {
            return Some(res);
        }
        // If no user handler handled the event, use the built ins.
        self.builtin
            .iter()
            .find_map(|handler| handler.handle(ev.clone()))
    }
}
```

The async thing is a bigger problem. Let me think about this

jayshrivastava: Well we could have two traits.
`EventHandler` and `AsyncEventHandler`

gabotechs:
Deferring for now, feel free to reopen if still relevant.

PR #573, opened 2026-07-28
Decouple gRPC from the worker protocol and add an in-process transport

stuhood:
### What

Decouples the `datafusion-distributed` core logic from the `grpc` transport and introduces an in-process transport for single-process distributed task execution.

### Why

Previously, the worker protocol and the `grpc` feature were coupled, which meant any use of the distributed execution abstraction required the `tonic` gRPC stack. Separating the transport layer from the protocol messages allows the framework to be used with alternative transports. The in-process transport provides a mechanism to run and test the abstraction when the `grpc` feature is disabled.

### How

* Split `worker.proto` into `messages.proto` (data structures) and `worker_service.proto` (gRPC service definition in the `worker_grpc` package). This splits the generated code into `worker.rs` (always compiled) and `worker_grpc.rs` (compiled only when the `grpc` feature is enabled).
* Added `InProcessChannelResolver`. It implements `WorkerChannel` by routing the three protocol methods directly to a co-located `Worker` instance in the same process, bypassing IPC, networking, and serialization.
* Build adjustments: 
  * Removed the `benchmarks` dev-dependency to prevent `grpc` from being enabled in every test build.
  * Gated gRPC-coupled test utilities in `test_utils` behind the `grpc` feature.
  * *(Note: The `integration` feature flag still assumes `grpc` is enabled because `test_utils::InMemoryChannelResolver` is used across the integration suite.)*

### Tests
* Added a `unit-test-no-grpc` CI job that verifies the core library and the in-process transport compile and pass unit tests without the `grpc` feature.
* Added an end-to-end test for `InProcessChannelResolver` (a distributed `GROUP BY` across tasks) that runs in the no-gRPC CI job.


console/src/main.rs
```diff
@@ -63,10 +63,10 @@ async fn run_app(
...
         // Check for keyboard input (16ms timeout ~ 60fps responsiveness)
-        if event::poll(Duration::from_millis(16))? {
-            if let Event::Key(key) = event::read()? {
-                input::handle_key_event(app, key);
-            }
+        if event::poll(Duration::from_millis(16))?
+            && let Event::Key(key) = event::read()?
+        {
+            input::handle_key_event(app, key);
         }
```
gabotechs: :thinking: this change is unrelated right? if it is, I'd avoid introducing unrelated changes, mainly for better traceability.

console/src/worker.rs
```diff
@@ -176,32 +176,32 @@ impl WorkerConn {
...
-                                // Maintain bounded size
-                                while self.completed_tasks.len() > MAX_COMPLETED_TASKS {
-                                    self.completed_tasks.pop_back();
-                                }
-
-                                // Remove from first_seen tracking
-                                self.task_first_seen.remove(&key);
+                    if old_task.status == TaskStatus::Running as i32
+                        && let Some(sk) = &old_task.task_key
+                    {
```
gabotechs: Same as above

codegen/src/main.rs
```diff
@@ -4,15 +4,17 @@ use std::fs;
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let repo_root = env::current_dir()?;
 
-    let proto_dir = repo_root.join("src/protocol/grpc");
-    let proto_file = proto_dir.join("worker.proto");
-    let out_dir = repo_root.join("src/protocol/grpc/generated");
+    let protocol_dir = repo_root.join("src/protocol");
+    let messages_proto = protocol_dir.join("messages.proto");
```
gabotechs, line 8: This generator file is under the `grpc` scope, but it's in charge of generating messages.proto which is not part of `grpc`.

Probably copy-pasting this generator script one folder above is not worth it. What about moving it to the root of the project under a codegen/ folder or something like that? we can centralize proto generation there.

src/protocol/generated/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod worker;
+
+#[cfg(feature = "grpc")]
+#[allow(clippy::all)]
+pub mod worker_grpc;
```
gabotechs: If this is specific to `grpc`, ideally it should be generated under the grpc/ folder

src/test_utils/in_memory_channel_resolver.rs
```diff
@@ -105,41 +23,139 @@ impl WorkerResolver for InMemoryWorkerResolver {
...
-        .with_distributed_channel_resolver(channel_resolver)
-        .build();
-    SessionContext::from(state)
-}
+// The channel resolver runs a real `tonic` worker server over a tokio duplex, so it only compiles
+// with the gRPC transport. The worker resolver above stays available without it, so the planner
+// tests that only need worker URLs keep building in a no-gRPC config.
+#[cfg(feature = "grpc")]
+mod channel {
+    use super::InMemoryWorkerResolver;
```
gabotechs: All the test utils that are specific to `grpc` should probably live under the grpc/ folder, that way they are gated automatically.

src/worker/mod.rs
```diff
@@ -3,7 +3,9 @@ mod impl_execute_task;
 mod session_builder;
 mod single_write_multi_read;
 mod task_data;
-#[cfg(any(test, feature = "integration"))]
+// `worker_handles` builds `tonic` servers and Flight channels for the benchmark fixtures, which
+// only compile with the gRPC transport.
+#[cfg(all(feature = "grpc", any(test, feature = "integration")))]
 pub(crate) mod test_utils;
```
gabotechs: This is the pattern we should aim to get rid of in the project: if it's `grpc` related, it should ideally live under grpc/, this means that we probably need to refactor some tests to rely on the in-memory representation.

src/coordinator/metrics_store.rs
```diff
@@ -27,7 +27,7 @@ impl MetricsStore {
         self.rx.borrow().get(key).cloned()
     }
 
-    #[cfg(test)]
+    #[cfg(all(test, feature = "grpc"))]
```
gabotechs, line 30: Seems like this is needed because this code is used in the task_metrics_rewriter.rs tests, and those tests are still coupled to `grpc`.

It'd be nice to decouple that test from `grpc` first so that we can avoid flagging this here.

tests/clickbench_correctness_test.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(all(feature = "integration", feature = "clickbench", test))]
+#[cfg(all(feature = "clickbench", test))]
```
gabotechs, line 1: These tests should really not be under benchmarks. I think I might be missing the reason for these to be here. Do they really need to be here?

src/execution_plans/metrics.rs
```diff
@@ -21,7 +21,7 @@ impl MetricsWrapperExec {
         Self { inner, metrics }
     }
 
-    #[cfg(test)]
+    #[cfg(all(test, feature = "grpc"))]
     pub(crate) fn inner(&self) -> &Arc<dyn ExecutionPlan> {
```
gabotechs, line 25: This is only used in task_metrics_rewriter.rs tests. If we manage to decouple those tests from `grpc`, we can leave this untouched.

src/protocol/in_process.rs
```diff
@@ -0,0 +1,230 @@
...
+use url::Url;
+
+/// A [`ChannelResolver`] backed by a single co-located [`Worker`].
+///
+/// Every URL resolves to that one worker: with no network there is nothing to dial, so the URLs a
+/// [`crate::WorkerResolver`] hands out only label the tasks the planner routes. One worker holds the
+/// state for every task, keyed by [`crate::TaskKey`], the same way the gRPC worker does when several
+/// tasks of a query land on it.
+#[derive(Clone)]
+pub struct InProcessChannelResolver {
```
gabotechs, line 33: For a follow-up PR: this should somehow interact with `channel_resolver.rs` and `LocalWorkerContext` so that, once a request is made to a URL that matches LocalWorkerContext.self_url, the `InProcessChannelResolver` kicks in automatically.

gabotechs (commented):
This is starting to look good!

I see there are different pieces of work that can be factored out into independent PRs that we can ship incrementally. I identify the following pieces of work in the following order:
1. split `worker.proto` into `worker_service.proto` and `messages.proto`, putting some thought into where to place the proto generator script so that we don't copy-paste too much.
2. Ship the `InProcessWorkerChannel`, and wire it up to channel_resolver.rs (or somewhere else that makes sense) so that we can start executing queries without `grpc`. Probably, the interaction of `InProcessWorkerChannel` and `LocalWorkerContext` needs to happen here.
3. Tweak existing tests unit tests that rely on `grpc` so that they just use in-process comms (mainly the tests about rewritting plans with metrics). That will allow to decouple them for gRPC by using the in-process implementation.
 4. Tweak the remaining test utils, moving some to grpc/ and flagging them appropriately:
     - If `grpc` feature is enabled, then gRPC worker channels are used (still with in-process fallbacks for self urls)
    - If `grpc` is disabled, exclusively in-process worker channels are used. This should be transparent to the tests.

stuhood:
Thanks for the feedback: I will apply this tomorrow.

gabotechs:
I have created this epic to track this effort (something I probably should have done before):
- https://github.com/datafusion-contrib/datafusion-distributed/issues/580

I added there, not only the work that has been done already (the `WorkerChannel` abstraction), but also the new pieces of work we've discussed in the PRs review that I think can be shipped independently.

By separating it in independent chunks, we also allow to relief some pressure from you guys in case you want to just drive part of the effort, and not everything.

console/src/main.rs
stuhood: These are a `clippy` failure flagged by [the commit hooks](https://datafusion-contrib.github.io/datafusion-distributed/contributor-guide/02-setup.html). I'll revert them, but we'll need to keep skipping the hooks until a fix goes in.

stuhood:
@gabotechs : I have been traveling the last few days, and Github decided not to render your comment from 3 days ago.

Consequently, I ended up including #576, #578, and #577 in this PR.

Please take a look, and let me know if you'd prefer that I split them out.

gabotechs:
@stuhood, no problem at all.

I think it's worth splitting the changes out and ship them one by one. My intuition tells me that starting with https://github.com/datafusion-contrib/datafusion-distributed/issues/576 would be easier.

PR #574, opened 2026-07-29
Enable broadcast joins by default

gabotechs:
This has proven to be production ready for while now, so let's enable it by default


gene-bordegaray (approved):
I can approve this one confidently 😅

Issue #522, opened 2026-07-01
[Epic][part 2] Adaptive Query Execution, runtime plan optimization


gabotechs:
Follow-up to #377.

#377 laid the foundation for Adaptive Query Execution: stages are now planned dynamically, and runtime statistics gathered by the `SamplerExec` are used to make decisions like the number of tasks assigned to each stage (see #432).

Now that accurate runtime statistics flow through the coordinator as stages execute, the next step is to **re-optimize the not-yet-executed stages on the fly** based on those stats, rather than only sizing tasks. For example:

- Swapping join orders based on observed cardinalities.
- Switching join implementations (e.g. broadcast vs shuffle) once the build-side size is known (relates to the `broadcast_joins` follow-up).
- Coalescing/splitting partitions to handle data skew discovered at runtime.
- Other physical optimizations that become possible once real stats replace planning-time estimates.

Early exploration lives on the `gabrielmusat/aqe` branch.

## Prior art

This kind of execution-time replanning is well established in other distributed engines. Their designs are a good reference for what it unlocks and how it's delivered:

**Apache Spark — Adaptive Query Execution (AQE)** (enabled by default since 3.2). At each shuffle stage boundary, Spark re-runs the optimizer with statistics from completed stages. Three headline optimizations:
- coalescing post-shuffle partitions,
- converting sort-merge joins to broadcast hash joins once a side is known to be small,
- skew-join handling (splitting/replicating skewed partitions into evenly sized tasks).
  - Official docs: https://spark.apache.org/docs/latest/sql-performance-tuning.html#adaptive-query-execution
  - Databricks deep-dive: https://www.databricks.com/blog/2020/05/29/adaptive-query-execution-speeding-up-spark-sql-at-runtime.html

**Trino — Adaptive plan optimizations** (since Trino 457, requires fault-tolerant execution). Because intermediate exchange data is spooled, Trino can reorder partitioned joins based on the actual build/probe sizes observed mid-query, rather than pre-computed connector statistics.
- Adaptive plan optimizations: https://trino.io/docs/current/optimizer/adaptive-plan-optimizations.html
- Fault-tolerant execution (the architecture that enables it): https://trino.io/docs/current/admin/fault-tolerant-execution.html

A recurring theme across both: a materialization/exchange boundary is what makes safe replanning possible, and join strategy/order + partition sizing are the highest-value decisions to defer to runtime — which maps directly onto the stage boundaries this project already has.

[...]

PR #582, opened 2026-07-30
fix: Enforce data distribution requirements and fix plans that break them

### LINKED ISSUE #563: Bug: CollectLeft HashJoin, CrossJoin, and NestedLoopJoin all return incomplete result sets when run over mutliple tasks
A CollectLeft HashJoin, a NestedLoopJoin, and a CrossJoin all require the *complete* build side in every task. `insert_broadcast_execs` only guarantees that for join types that don't emit build-side rows (and only when broadcast joins are enabled), but the task-count logic in `inject_network_boundaries` does not cap the remaining shapes to a single task. The build-side scan then gets sliced across tasks like any other leaf, and each task joins its slice of the build side against its slice of the probe side.

Adding the following test file shows the bug:

```rust
//! Reproducers for joins that collect their build (left) side producing wrong results when
//! placed in a multi-task stage without the build side being broadcast.
//!
//! A CollectLeft HashJoin, a NestedLoopJoin, and a CrossJoin all require the *complete* build
//! side in every task. `insert_broadcast_execs` only guarantees that for join types that don't
//! emit build-side rows (and only when broadcast joins are enabled), but the task-count logic
//! in `inject_network_boundaries` does not cap the remaining shapes to a single task. The
//! build-side scan then gets sliced across tasks like any other leaf, and each task joins its
//! slice of the build side against its slice of the probe side.
//!
//! The tables are laid out so the slicing is visible: `build_side` holds ids 0..100 split
//! sequentially across 4 files, while `probe_side` holds the same ids (each repeated 50 times)
//! rotated one file forward. A task therefore sees *different* ids from each table, and any
//! cross-task match is silently lost.
//!
//! NOTE: these tests assert that the INCORRECT behavior happens (distributed results diverge
//! from single-node results). A passing test means the bug reproduced. When a fix lands, these
//! tests should start failing and be converted into correctness tests.

#[cfg(all(feature = "integration", test))]
mod tests {
    use arrow::array::{Int64Array, RecordBatch};
    use arrow::datatypes::{DataType, Field, Schema};
    use datafusion::error::Result;
    use datafusion::physical_plan::{ExecutionPlan, collect};
    use datafusion::prelude::{ParquetReadOptions, SessionContext};
    use datafusion_distributed::test_utils::in_memory_channel_resolver::start_in_memory_context;
    use datafusion_distributed::test_utils::property_based::compare_result_set;
    use datafusion_distributed::{
        DefaultSessionBuilder, DistributedExec, DistributedExt, display_plan_ascii,
    };
    use parquet::arrow::ArrowWriter;
    use std::fs;
    use std::path::{Path, PathBuf};
    use std::sync::Arc;
    use tokio::sync::OnceCell;

    const NUM_WORKERS: usize = 4;
    const PARTITIONS: usize = 3;
    const FILES_PER_TABLE: i64 = 4;
    const IDS_PER_FILE: i64 = 25;
    const PROBE_DUPLICATES: usize = 50;

    fn data_dir() -> PathBuf {
        Path::new(env!("CARGO_MANIFEST_DIR")).join("target/multi_task_collect_join_repros")
    }

    static INIT: OnceCell<()> = OnceCell::const_new();

    async fn ensure_data() {
        INIT.get_or_init(|| async {
            let dir = data_dir();
            let _ = fs::remove_dir_all(&dir);
            // (table, rows per id, file rotation)
            for (table, duplicates, rotation) in [
                ("build_side", 1usize, 0i64),
                ("probe_side", PROBE_DUPLICATES, 1),
            ] {
                let table_dir = dir.join(table);
                fs::create_dir_all(&table_dir).unwrap();
                for file_idx in 0..FILES_PER_TABLE {
                    let chunk = (file_idx + rotation) % FILES_PER_TABLE;
                    let ids = (chunk * IDS_PER_FILE..(chunk + 1) * IDS_PER_FILE)
                        .flat_map(|id| std::iter::repeat(id).take(duplicates))
                        .collect::<Vec<_>>();
                    write_ids(&table_dir.join(format!("part-{file_idx}.parquet")), &ids);
                }
            }
        })
        .await;
    }

    fn write_ids(path: &Path, ids: &[i64]) {
        let schema = Arc::new(Schema::new(vec![Field::new("id", DataType::Int64, true)]));
        let batch = RecordBatch::try_new(
            schema.clone(),
            vec![Arc::new(Int64Array::from(ids.to_vec()))],
        )
        .unwrap();
        let file = fs::File::create(path).unwrap();
        let mut writer = ArrowWriter::try_new(file, schema, None).unwrap();
        writer.write(&batch).unwrap();
        writer.close().unwrap();
    }

    async fn register_tables(ctx: &SessionContext) -> Result<()> {
        for table in ["build_side", "probe_side"] {
            ctx.register_parquet(
                table,
                data_dir().join(table).to_str().unwrap(),
                ParquetReadOptions::default(),
            )
            .await?;
        }
        Ok(())
    }

    async fn make_distributed_ctx(broadcast_joins: bool) -> Result<SessionContext> {
        let ctx = start_in_memory_context(NUM_WORKERS, DefaultSessionBuilder).await;
        ctx.state_ref()
            .write()
            .config_mut()
            .options_mut()
            .execution
            .target_partitions = PARTITIONS;
        let ctx = ctx
            .with_distributed_file_scan_config_bytes_per_partition(1)?
            .with_distributed_broadcast_joins(broadcast_joins)?;
        register_tables(&ctx).await?;
        Ok(ctx)
    }

    async fn run(
        ctx: &SessionContext,
        query: &str,
        Ok((plan, batches))
    }

    /// Runs `query` on a single-node and a distributed context, asserts the distributed plan
    /// contains `expected_plan_markers` (pinning the unsafe join shape), and asserts the two
    /// result sets DIVERGE — i.e. the bug reproduced.
    async fn assert_bug_reproduces(
        query: &str,
        broadcast_joins: bool,
        expected_plan_markers: &[&str],
    ) -> Result<()> {
        ensure_data().await;

        register_tables(&s_ctx).await?;
        let d_ctx = make_distributed_ctx(broadcast_joins).await?;

        let (_, s_batches) = run(&s_ctx, query).await?;
        let (d_plan, d_batches) = run(&d_ctx, query).await?;

        let display = display_plan_ascii(d_plan.as_ref(), false);
        println!("distributed plan:\n{display}");
        assert!(
            d_plan.is::<DistributedExec>(),
            "query did not get distributed; plan was:\n{display}"
        );
        for marker in expected_plan_markers {
            assert!(
                display.contains(marker),
                "expected the distributed plan to contain `{marker}`; plan was:\n{display}"
            );
        }

        let s_rows: usize = s_batches.iter().map(|b| b.num_rows()).sum();
        let d_rows: usize = d_batches.iter().map(|b| b.num_rows()).sum();
        println!("single-node rows: {s_rows}, distributed rows: {d_rows}");

        let comparison = compare_result_set(&Ok(d_batches), &Ok(s_batches));
        assert!(
            comparison.is_err(),
            "distributed results unexpectedly matched single-node results — bug did not reproduce"
        );
        println!(
            "result sets diverge (this is the bug): {}",
            comparison.unwrap_err()
        );
        Ok(())
    }

    /// Case 1: CollectLeft HashJoin with a build-side-emitting join type (LeftSemi).
    /// Broadcast joins are ON, but `insert_broadcast_execs` skips LeftSemi, and nothing caps
    /// the stage to one task. Every id matches on a single node; distributed, a build id only
    /// survives if its probe rows landed in the same task.
    #[tokio::test]
    async fn repro_collect_left_semi_hash_join() -> Result<()> {
        assert_bug_reproduces(
            "SELECT id FROM build_side WHERE id IN (SELECT id FROM probe_side)",
            true,
            &["HashJoinExec: mode=CollectLeft, join_type=LeftSemi"],
        )
        .await
    }

    /// Case 2: anti join (`NOT IN`). Single-node: every build id exists in probe_side,
    /// so zero rows. Distributed, each task only sees a slice of probe_side, so most build ids
    /// look unmatched and phantom rows are emitted.
    #[tokio::test]
    async fn repro_not_in_anti_hash_join() -> Result<()> {
        assert_bug_reproduces(
            "SELECT id FROM build_side WHERE id NOT IN (SELECT id FROM probe_side)",
            true,
            &["join_type=LeftAnti"],
        )
        .await
    }

    /// Case 3: NestedLoopJoin with a build-side-emitting join type (LeftSemi), produced
    /// by a correlated EXISTS with a non-equi predicate (`p.id > b.id - 1 AND p.id < b.id + 1`
    /// is `p.id = b.id` for integers, but expressed as inequalities so no hash join is possible).
    #[tokio::test]
    async fn repro_nested_loop_left_semi_join() -> Result<()> {
        assert_bug_reproduces(
            "SELECT b.id FROM build_side b WHERE EXISTS ( \
                SELECT 1 FROM probe_side p WHERE p.id > b.id - 1 AND p.id < b.id + 1)",
            true,
            &["NestedLoopJoinExec", "join_type=LeftSemi"],
        )
        .await
    }

    /// Case 4: Full NestedLoopJoin. Emits unmatched rows from BOTH sides, so no
    /// broadcast orientation can ever be correct. Single-node: every row matches, no NULL
    /// padding. Distributed: cross-task matches are lost and spurious NULL-padded rows appear.
    #[tokio::test]
    async fn repro_nested_loop_full_join() -> Result<()> {
        assert_bug_reproduces(
            "SELECT b.id, p.id FROM build_side b FULL JOIN probe_side p \
                ON p.id > b.id - 1 AND p.id < b.id + 1",
            true,
            &["NestedLoopJoinExec", "join_type=Full"],
        )
        .await
    }

    /// Case 5: CrossJoin with broadcast joins DISABLED, so no BroadcastExec exists at
    /// all. Single-node: all 100 x 5000 = 500_000 pairs contribute to the sum. Distributed:
    /// each task only pairs its slice of each side, so most pairs are never produced.
    /// (A bare `count(*)` is folded to a constant from parquet statistics, so sum an
    /// expression the optimizer cannot answer from metadata.)
    #[tokio::test]
    async fn repro_cross_join_broadcast_disabled() -> Result<()> {
        assert_bug_reproduces(
            "SELECT sum(b.id + p.id) AS pair_sum FROM build_side b CROSS JOIN probe_side p",
            false,
            &["CrossJoinExec"],
        )
        .await
    }
}
```

This is a plan correctness invariant that should be validated before shipping the finalized plan, enforcing: "CollectLeft HashJoin, CrossJoin, and NestedLoopJoin nodes with >1 tasks must have their build-side fed by a broadcast". (I'm doing a bit more investigating to see if this invariant can be simplified)

Then the actual plan fix for the 5 cases above:
- HashJoin `CollectLeft` (join types: Left/LeftSemi/LeftAnti/LeftMark/Full) → change `CollectLeft`->`Partitioned` (keys exist; no replication needed)
- HashJoin `CollectLeft`, `null_aware` (join type: LeftAnti) -> cap at 1 task
- NLJ (join types: Left/LeftSemi/LeftAnti/LeftMark) -> call `swap_inputs()`, then broadcast as usual. (makes join shape safe for broadcast).
- NLJ (join type: Full) -> cap at 1 task
- CrossJoin (always safe with broadcast) -> cap at 1 task if broadcasts are disabled.

barbarj:
Closes [#563](https://github.com/datafusion-contrib/datafusion-distributed/issues/563)

The bug in 563 is actually a failure to enforce a plan shape invariant for multi-task stages: for every stage, executing its plan once per task over the per-task input assignment and unioning the outputs must be equivalent to executing it once over all the data.

This PR adds a correctness check after plan finalization (`validate_stages.rs`) and applies fixes for the problematic shapes (`normalize_collect_joins.rs`).

This runs `validate_distributed_stages()` as the final step before returning during physical plan construction. We classify the dataflow across each stage, bottom-up, as partitioned or replicated. Partitioned data flow is safe across multiple tasks. Replicated is not. In cases where the input to a task is replicated data, we (based on our knowledge of node semantics) convert that to partitioned if the join will not emit build-side rows, making the replication (i.e. the broadcast build-side) safe as input and the output meaningfully partitioned. See `validate_stages.rs` for a more in-depth explanation and justification.

There are 5 invalid plan shapes to fix. The following strategies are used:
- HashJoin `CollectLeft` (join types: Left/LeftSemi/LeftAnti/LeftMark/Full) → change
`CollectLeft`->`Partitioned` (keys exist; no replication needed)
- HashJoin`CollectLeft`, `null_aware` (join type: LeftAnti) -> cap at 1 task
- NLJ (join types: Left/LeftSemi/LeftAnti/LeftMark) -> call swap_inputs(), then broadcast as usual. (makes the join shape safe for broadcast).
- NLJ (join type: Full) -> cap at 1 task
- CrossJoin (always safe with broadcast) -> cap at 1 task if broadcasts are disabled.

The `CollectLeft`->`Partitioned` change causes the rewriting of a bunch of plan shapes in the plan tests. The problematic stages run over small data, and thus collapse to a single task during
`prepare_network_boundaries`, avoiding the bug. Due to the shape normalization pass necessarily running before task counts are decided, we can't see that the task consolidation will happen and must modify the plan.

In-repo benchmarks show equivalent or better performance for all queries, save one: TPCH q22. It drops by about 40% (~17ms -> ~24ms on my machine, 8 workers, 2 threads). As far as I can tell, it's losing a dynamic filter due to being cut into more stages.


barbarj:
My apologies for not surfacing the validation design before cutting this PR. If you think that should be done differently, I'm happy to limit this PR to just the shape fixes and do the validation separately.

gabotechs:
Hi @barbarj! thanks so much for this quality PR 🙏.

I'd like to spend some time understanding the problem so that I can better review this and give good suggestions. In the meantime, I think we can start shipping chunks of this PR, for example, I'd like to propose creating a PR with the new `tests/multi_task_collect_join_repros.rs` integration tests (#[ignored] for now as I imagine they would not pass), would you be fine with that?

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,215 @@
...
+/// [inject_network_boundaries] to cap at a single task:
+///
+/// - Null-aware anti joins: their NULL-existence checks require global knowledge held in
+///   process-local shared state, which cannot span tasks in any orientation.
+/// - Full joins: they emit unmatched rows from both sides, so they need global match knowledge
+///   on both sides at once.
+///
+/// [insert_broadcast_execs]: super::insert_broadcast::insert_broadcast_execs
+/// [inject_network_boundaries]: super::inject_network_boundaries::inject_network_boundaries
+pub(super) fn normalize_collect_joins(
```
gabotechs, line 46: Flushing thoughts as I read:

This sounds like something that could be useful if we wanted to dynamically swap JOIN orders at runtime during AQE (Adaptive Query Execution):

https://github.com/datafusion-contrib/datafusion-distributed/issues/522

gabotechs:
Just ran the remote benchmarks against main on a 12 machine cluster:
```
tpch_sf1
   TOTAL: prev=5216 ms, new=4649 ms, diff=1.12 faster ✔
tpch_sf10
   TOTAL: prev=12052 ms, new=11218 ms, diff=1.07 faster ✔
tpch_sf100
   TOTAL: prev=59778 ms, new=59279 ms, diff=1.01 faster ✔
tpcds_sf1
   TOTAL: prev=35592 ms, new=35534 ms, diff=1.00 faster ✔
clickbench_0-100
   TOTAL: prev=28801 ms, new=29263 ms, diff=1.02 slower ✖
```
So performance wise looks good 👍

src/distributed_planner/normalize_collect_joins.rs
gabotechs, line 46: This is pretty high quality work, I think this `normalize_collect_joins.rs` is good to go, for if you want to also open a PR with just this while I review the `validate_stages.rs` piece of code.

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::assert_snapshot;
+    use crate::test_utils::plans::TestPlanBuilder;
+    use datafusion::physical_plan::displayable;
+
+    async fn sql_to_normalized_plan(query: &str, broadcast_enabled: bool) -> String {
```
gabotechs, line 141: We typically place helper methods in tests at the bottom of the `mod tests {};` block. Readers will care more about tests than the helpers, and readers read top to bottom, so putting the "less important" things at the bottom makes life a bit easier for readers.

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[tokio::test]
+    async fn test_left_hash_join_converted_to_partitioned() {
+        let query = r#"
+        SELECT a."MinTemp", b."MaxTemp"
+        FROM weather a LEFT JOIN weather b
+        ON a."RainToday" = b."RainToday"
+        "#;
+        let plan = sql_to_normalized_plan(query, true).await;
+        assert!(plan.contains("HashJoinExec: mode=Partitioned, join_type=Left"));
+        assert_snapshot!(plan);
```
gabotechs, line 163: Can we inline the insta assertions here? like in the other tests?

tests/multi_task_collect_join_repros.rs
```diff
@@ -0,0 +1,229 @@
...
+    const IDS_PER_FILE: i64 = 25;
+    const PROBE_DUPLICATES: usize = 50;
+
+    fn data_dir() -> PathBuf {
+        Path::new(env!("CARGO_MANIFEST_DIR")).join("target/multi_task_collect_join_repros")
+    }
+
+    static INIT: OnceCell<()> = OnceCell::const_new();
+
+    async fn ensure_data() {
```
gabotechs, line 53: Same recommendation as before, for helpers, prefer placing them at the bottom of the `mod tests {};` block.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -273,10 +275,28 @@ async fn _inject_network_boundaries(
...
-        && !broadcast_joins_enabled
+        && (!broadcast_joins_enabled || node.null_aware)
+    {
+        // A CollectLeft join collects its entire build side in every task, so it can only run in
+        // a multi-task stage when [insert_broadcast_execs] broadcast the build side. That is not
+        // possible when broadcast joins are disabled, nor for null-aware anti joins, whose
+        // NULL-existence checks live in process-local shared state that cannot span tasks in any
+        // orientation. Other build-side-emitting join types were already rewritten to Partitioned
+        // by [normalize_collect_joins], so they never reach this arm.
+        task_count = Maximum(1);
```
gabotechs, line 286: Really nice comments here 👍 all of them are clear and make sense. Thanks!

gabotechs (commented):
Flushing some more comments as I keep reviewing.

src/distributed_planner/validate_stages.rs
```diff
@@ -0,0 +1,639 @@
...
+    if node.is_network_boundary() {
+        // NetworkCoalesceExec (or a future boundary type): gathers all partitions into a
+        // single consumer task, so it must never appear in a multi-task stage.
+        return plan_err!(
+            "stage runs {tasks} tasks but contains {}, which requires a single-task \
+             consumer stage",
+            node.name()
+        );
+    }
+    // A DistributedLeafExec resolves to a different slice of the underlying source in every
```
gabotechs, line 128: I don't think this assumption is right. There are valid cases for `NetworkCoalesceExec` to be placed in the middle of the plan.

Take for example this:

```rust
cargo run \
--features integration \
--example custom_distributed_partial_reduction_tree \
'SELECT "RainToday", count(*) FROM weather GROUP BY "RainToday"' \
--show-distributed-plan
```

The runnable example no longer works in this PR, even though it's a valid distributed plan.

src/distributed_planner/validate_stages.rs
```diff
@@ -0,0 +1,639 @@
...
+    // task. (Its per-task variants are children in the plan tree, so check before the leaf
+    // case below.) If the scan preserves a hash partitioning from the storage layout
+    // (hive-style pre-partitioned files with `preserve_file_partitions`), the per-task
+    // slicing follows those same partitions, so equal keys still co-locate cluster-wide.
+    if node.is::<DistributedLeafExec>() {
+        let claims_are_global = matches!(node.output_partitioning(), Partitioning::Hash(..));
+        return Ok(DataFlow::Partitioned { claims_are_global });
+    }
+    // A ChildrenIsolatorUnionExec divides the stage's tasks among its children: child `i`
```
gabotechs, line 137: Why would this be specific to `DistributedLeafExec`? Any user can create their own leaf node and claim a `Partitioned::Hash` output without using `DistributedLeafExec`.

src/distributed_planner/validate_stages.rs
```diff
@@ -0,0 +1,639 @@
...
+        // embedded verbatim in every task's serialized plan: identical, complete data.
+        // NOTE: a volatile leaf (e.g. one backed by a random or time-dependent source)
+        // would break the replication assumption; nothing in the ExecutionPlan API exposes
+        // that.
+        let ev = crate::events::DesiredTaskCountEvent {
+            plan: node,
+            session_config: session_cfg,
+        };
+        let is_task_varying = crate::events::DesiredTaskCountHandlers::handle(ev).is_some();
+        return Ok(if is_task_varying {
```
gabotechs, line 179: :thinking: I'm not sure why this would be needed here. At the moment this function is called, all the task count have already been propagated and they are set in stone, so it matters little whether there are desired task count handlers registered for that node.

gabotechs: I feel like the code that handles `children.is_empty()` and ` node.is::<DistributedLeafExec>()` should be the same, and should not really be taking into account if there are desired task count handlers registered.

src/distributed_planner/validate_stages.rs
```diff
@@ -0,0 +1,639 @@
...
+                    // classes make renamed keys pass; dropped or wrong keys fail. Subset
+                    // satisfaction — partitioned on a subset of the required keys — still
+                    // co-locates equal keys and is accepted, matching what
+                    // EnforceDistribution accepts process-locally (e.g. an aggregate
+                    // grouping on a superset of its input's join keys).
+                    // DataFusion 54 does not export the type `satisfaction` returns, so a
+                    // reference NotSatisfied value (an unknown partitioning can never
+                    // satisfy a hash requirement) stands in for naming the variant.
+                    let eq_properties = child.equivalence_properties();
+                    let not_satisfied = Partitioning::UnknownPartitioning(2).satisfaction(
```
gabotechs, line 232: That hardcoded `2`... there must be a more robust way of doing this.

@gene-bordegaray, I'd love if you can chime in here.

gabotechs (commented):
Left some comments for the validate_stages.rs file. 

The normalize_collect_joins.rs and test from this PR are mostly ready to go, with just some minor cosmetic comments. For shipping those eariler, I'd suggest to split this PR in two:
1. normalize_collect_joins.rs and the tests.
2. validate_stages.rs

I'd like for other people to put some more eyes on 2), but 1) is something we can start shipping sooner. WDYT?

barbarj:
> The normalize_collect_joins.rs and test from this PR are mostly ready to go, with just some minor cosmetic comments. For shipping those eariler, I'd suggest to split this PR in two:
> 
> 1. normalize_collect_joins.rs and the tests.
> 2. validate_stages.rs
> 
> I'd like for other people to put some more eyes on 2), but 1) is something we can start shipping sooner. WDYT?

That works for me and lines up with my confidence level about these changes too. I'll cut a separate PR with just the tests and plan fixes. (edit: Done. See #583)

Will review the comments on 2) after that.

src/distributed_planner/validate_stages.rs
barbarj: I see your point. `NetworkCoalesceExec` maintains it's input partitioning, so I think under this model it'd be safe to return `DataFlow::Partitioned { claims_are_global: true }` here.

src/distributed_planner/validate_stages.rs
barbarj: > 🤔 I'm not sure why this would be needed here. At the moment this function is called, all the task count have already been propagated and they are set in stone, so it matters little whether there are desired task count handlers registered for that node.

This mirrors the how `inject_network_boundaries` decides task counts: 
```rust
// from _inject_network_boundaries()

if plan.children().is_empty() {
        // This is a leaf node, maybe a DataSourceExec, or maybe something else custom from the
        // user. We need to estimate how many tasks are needed for this leaf node, and we'll take
        // this decision into account when deciding how many tasks will be actually used.
        let ev = DesiredTaskCountEvent {
            plan: &plan,
            session_config: nb_ctx.cfg,
        };
        return if let Some(estimate) = DesiredTaskCountHandlers::handle(ev) {
            Ok(nb_ctx.plan_with_task_count(plan, estimate.task_count.limit(nb_ctx.max_tasks()?)))
        } else {
            // We could not determine how many tasks this leaf node should run on, so
            // assume it cannot be distributed and use just 1 task.
            Ok(nb_ctx.plan_with_task_count(plan, Maximum(1)))
        };
    }
```
The actual property we want is to know is if this assumption was applied: 
```
// We could not determine how many tasks this leaf node should run on, so
// assume it cannot be distributed and use just 1 task.
```
~If it was, that count of 1 would propagate up the stage and we'd never get to this code anyways, right? In which case we could just replace this block with an a `plan_err!`.~ Nevermind, this would mean the validator relies on the behavior of the thing it's validating to be correct

src/distributed_planner/validate_stages.rs
barbarj: As I understand it, it's this property:
```
// A DistributedLeafExec resolves to a different slice of the underlying source in every task.
```
which means a `Partitioning::Hash` implies `claims_are_global`.  Though I suppose any node claiming `Partitioning::Hash` implies its output is distributed accordingly.

barbarj: Re: your below comment:
> I feel like the code that handles children.is_empty() and  node.is::<DistributedLeafExec>() should be the same, and should not really be taking into account if there are desired task count handlers registered.

Assuming that I'm right about the task count propagation I mentioned, then this block actually becomes something like:
```rust
if node.is::<DistributedLeafExec>() || node.children().is_empty() {
    let claims_are_global = matches!(node.output_partitioning(), Partitioning::Hash(..));
    return Ok(DataFlow::Partitioned { claims_are_global })
}
```

src/distributed_planner/validate_stages.rs
jayshrivastava: I just started reviewing but here's a random comment

We often have plans where there's a `RepartitionExec Hash(expr, M * N)` at the head of a stage and the stage above has M tasks of N partitions each. It begs the question - is Partitioning=Hash(expr, M x N) the same as Partitioning=Hash(expr, N)? The inter-task partitioning is Partitioning=Hash(expr, N), however the intra-task partitioning is Partitioning=Hash(expr, N*M). All the tasks pass so it's safe but I could see this becoming a problem / annoyance later.

jayshrivastava: > The runnable example no longer works in this PR, even though it's a valid distributed plan.

Should we have CI run all our examples? Or port this example into a test. This seems important.

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,215 @@
...
+    let left = Arc::new(RepartitionExec::try_new(
+        build_input,
+        Partitioning::Hash(left_keys, target_partitions),
+    )?);
+    let right = Arc::new(RepartitionExec::try_new(
+        Arc::clone(join.right()),
+        Partitioning::Hash(right_keys, target_partitions),
+    )?);
+
+    Ok(Arc::new(HashJoinExec::try_new(
```
jayshrivastava, line 121: I think this resets all dynamic filters inside the `HashJoinExec`. I think we need `HashJoinExecBuilder::from`.

Probably worth having a unit test for `collect_left_to_partitioned` which ensures no internal fields are reset.

src/distributed_planner/validate_stages.rs
barbarj: `PartitioningSatisfaction` isn't public in Datafusion 54. It looks like it will be in 55 though.

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,215 @@
...
+    cfg: &ConfigOptions,
+) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    let d_cfg = DistributedConfig::from_config_options(cfg)?;
+    let target_partitions = cfg.execution.target_partitions;
+
+    plan.transform_down(|node| {
+        if let Some(join) = node.downcast_ref::<HashJoinExec>()
+            && join.mode == PartitionMode::CollectLeft
+            && !is_left_broadcast_safe(join.join_type())
+            && join.join_type() != &JoinType::Full
```
jayshrivastava, line 57: I think we need to transform full joins right? 

In your PR desc:
> HashJoin CollectLeft (join types: Left/LeftSemi/LeftAnti/LeftMark/Full) → change
CollectLeft->Partitioned (keys exist; no replication needed)

Right now we exclude full joins from all 3 join layouts (partitioned, broadcast, and single task).

src/distributed_planner/validate_stages.rs
jayshrivastava: > The actual property we want is to know is if this assumption was applied

Still reviewing but the `DesiredTaskCountHandler` is sort of like a hint right? Because the thing that actually scales up leaves is the `ScaleUpLeafNodeHandler`. Ideally we inspect the plan directly.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Oh shoot you're right. Will fix in #583.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Yeah, you're right. Removing the gate and fixing the comments in #583

src/distributed_planner/normalize_collect_joins.rs
barbarj: Addressed in #583.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Addressed in #583.

src/distributed_planner/validate_stages.rs
barbarj: > Still reviewing but the `DesiredTaskCountHandler` is sort of like a hint right? Because the thing that actually scales up leaves is the `ScaleUpLeafNodeHandler`. Ideally we inspect the plan directly.

Hmmm. Yeah, but `ScaleUpLeafNodeHandler` doesn't carry info about intent from what I can tell, just whether or not the scale-up happened. 

And actually, really, what we want to know is "is this leaf's data replicated or partitioned", which using the desired task count is just a heuristic for. I wonder if we can encode that some other way.

barbarj:
Closing. The normalization part of this was accomplished by #583. The validation portion is unsound; I've opened #745 to hopefully get it addressed in the future.

PR #583, opened 2026-07-31
fix: Normalize CollectLeft Joins and similar 

### LINKED ISSUE #563: Bug: CollectLeft HashJoin, CrossJoin, and NestedLoopJoin all return incomplete result sets when run over mutliple tasks

barbarj:
Closes #563 

(This is the plan-fix and tests part of #582)

The bug in 563 is actually a failure to enforce a plan shape invariant for multi-task stages: for every stage, executing its plan once per task over the per-task input assignment and unioning the outputs must be equivalent to executing it once over all the data.

This PR applies fixes for the problematic shapes (normalize_collect_joins.rs). Enforcing this invariant will be handled separately. (See discussion in #582).

There are 5 invalid plan shapes to fix. The following strategies are used:
- HashJoin `CollectLeft` (join types: Left/LeftSemi/LeftAnti/LeftMark/Full) → change
`CollectLeft`->`Partitioned` (keys exist; no replication needed)
- HashJoin`CollectLeft`, `null_aware` (join type: LeftAnti) -> cap at 1 task
- NLJ (join types: Left/LeftSemi/LeftAnti/LeftMark) -> call swap_inputs(), then broadcast as usual. (makes the join shape safe for broadcast).
- NLJ (join type: Full) -> cap at 1 task
- CrossJoin (always safe with broadcast) -> cap at 1 task if broadcasts are disabled.

The `CollectLeft`->`Partitioned` change causes the rewriting of a bunch of plan shapes in the plan tests. The problematic stages run over small data, and thus collapse to a single task during
`prepare_network_boundaries`, avoiding the bug. Due to the shape normalization pass necessarily running before task counts are decided, we can't see that the task consolidation will happen and must modify the plan.

In-repo benchmarks show equivalent or better performance for all queries, save one: TPCH q22. It drops by about 40% (~17ms -> ~24ms on my machine, 8 workers, 2 threads). As far as I can tell, it's losing a dynamic filter due to being cut into more stages.


jayshrivastava:
👋🏽 Left comments in the other PR which are relevant in this one https://github.com/datafusion-contrib/datafusion-distributed/pull/582/

barbarj:
> 👋🏽 Left comments in the other PR which are relevant in this one #582

Comments addressed. I'm now normalizing Full HashExec joins and ensuring the `HashExecJoin`'s internal properties are maintained with a unit test and the use of `HashJoinExecBuilder`.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -1085,6 +1111,32 @@ mod tests {
         ");
     }
 
+    #[tokio::test]
+    async fn test_nested_loop_inner_join_broadcast() {
```
gabotechs, line 1148: Any chance we can add here some more tests that reproduce the other behaviors?

I think we are missing:
- `HashJoinExec` where `null_aware == true`
- `NestedLoopJoinExec` where `node.join_type() == &JoinType::Full`
- `CrossJoinExec`

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,373 @@
...
+        SELECT a."MinTemp", b."MaxTemp"
+        FROM weather a LEFT JOIN weather b
+        ON a."MinTemp" < b."MaxTemp"
+        "#;
+        let plan = sql_to_normalized_plan(query, false).await;
+        assert!(plan.contains("NestedLoopJoinExec: join_type=Left"));
+    }
+
+    #[test]
+    fn collect_left_to_partitioned_preserves_every_field() {
```
gabotechs: This test is a bit out of place compared to how we typically test this `distributed_planner` rules. Seems like it's testing an specific implementation detail rather than a specific outcome.

The original problem was that dynamic filters where getting reseted, is there any chance we can test that following the same pattern as the other tests?

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,373 @@
...
+///
+/// - Null-aware anti joins: their NULL-existence checks ("did the probe side contain any
+///   NULL at all?") are global facts kept in shared state that is only global while a single
+///   build is shared by every probe partition. Per-partition builds lose them, so not even
+///   [PartitionMode::Partitioned] is equivalent — this is a semantic restriction, not a
+///   distribution one.
+/// - Full [NestedLoopJoinExec]s: a NestedLoopJoin only has replication strategies, and a
+///   Full join emits unmatched rows from both sides, so every orientation replicates an
+///   emitting side.
+///
```
gabotechs, line 95: It would be extremely cool to have examples of each case with nicely rendered ASCII tables.

This is a nice to have, not a must, so not a blocker for getting this merged.

gabotechs (commented):
This is looking very good!

Just some minor comments, I'd like to have @jayshrivastava chime in as well as he provided very useful comments here.

src/distributed_planner/inject_network_boundaries.rs
barbarj: These are all covered in `tests/multi_task_collect_join_repros`. These 3 cases are handled by `normalize_collect_joins()`.  The other two are tested here because the fix is to explicitly limit them to one task.

src/distributed_planner/normalize_collect_joins.rs
barbarj: There's not a way I'm aware of to do so via comparing result sets, since dynamic filters are an optimization feature and not a query correctness one. But I've cleaned up this test to just check that the dynamic filter is preserved, since that was the main concern and the use of `HashJoinExecBuilder` largely covers the rest.

src/distributed_planner/inject_network_boundaries.rs
gabotechs: These tests here provide a different kind of coverage: they ensure the reconciled task counts are the expected ones, while the integration tests ensure we don't regress vs single-node.

What I'd be interested to see, is to capture the fact that on joins that we can't distribute, we are effectively stamping a `Maximum(1)` task count, throttling the stage to exactly 1 task, and not just collapsing into 1 task with a `Desried(1)` by luck.

src/distributed_planner/normalize_collect_joins.rs
gabotechs: works for me

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -159,7 +161,7 @@ pub(crate) async fn inject_network_boundaries(
 pub(crate) struct InjectNetworkBoundaryContext<'a> {
     pub(crate) d_cfg: &'a DistributedConfig,
 
-    cfg: &'a SessionConfig,
+    pub(crate) cfg: &'a SessionConfig,
```
gabotechs: :thinking: this does not need to be made public right?

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    let left = Arc::new(RepartitionExec::try_new(
+        build_input,
+        Partitioning::Hash(left_keys, target_partitions),
+    )?);
+    let right = Arc::new(RepartitionExec::try_new(
+        Arc::clone(join.right()),
+        Partitioning::Hash(right_keys, target_partitions),
+    )?);
+
```
gabotechs, line 195: There's a bug here that can be triggered in cases where the original output partition count of `join.right()` is 1.

When output partitioning of something is 1, DataFusion optimizer rules can afford to make some assumptions that are only valid when partitioning is 1, like the fact that sort order is maintained and probably other things.

This code here does not seem to be checking that, and this can end up in situations where we go from 1 partition to `target_partitions` partitions, introducing correctness bugs.

If you add this new integration test you'd get one reproducer of this issue:

```rust
    #[tokio::test]
    async fn converted_full_hash_join_preserves_output_partitioning() -> Result<()> {
        assert_distributed_matches_single_node(
            r"SELECT b.id AS bid, p.id AS pid FROM
                (SELECT id FROM build_side ORDER BY id LIMIT 5) b
                FULL JOIN
                (SELECT id FROM probe_side ORDER BY id LIMIT 1000000) p
                ON b.id = p.id
             ORDER BY bid NULLS LAST, pid NULLS LAST",
            true,
        )
        .await
    }
```

gabotechs (commented):
Flushing a couple more comments as I go

tests/multi_task_collect_join_repros.rs
```diff
@@ -0,0 +1,243 @@
...
+    const PROBE_DUPLICATES: usize = 50;
+
+    static INIT: OnceCell<()> = OnceCell::const_new();
+
+    /// Case 1: CollectLeft HashJoin with a build-side-emitting join type (LeftSemi).
+    /// Broadcast joins are ON, but `insert_broadcast_execs` skips LeftSemi, and nothing caps
+    /// the stage to one task. Every id matches on a single node; distributed, a build id only
+    /// survives if its probe rows landed in the same task.
+    #[tokio::test]
+    async fn capped_collect_left_semi_hash_join_is_correct() -> Result<()> {
```
gabotechs: This test says `capped_`, but it's not capping anything, does it?

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,293 @@
...
+                    .with_new_children(vec![
+                        Arc::clone(coalesce.input()),
+                        Arc::clone(join.right()),
+                    ])?
+                    .downcast_ref::<NestedLoopJoinExec>()
+                    .expect("with_new_children changed the node type")
+                    .swap_inputs()?,
+                None => join.swap_inputs()?,
+            };
+            return Ok(Transformed::yes(swapped));
```
gabotechs: I think this might have the same problem as the hash joins actually, leaving here another integration test that replicates the issue:

```rust
    #[tokio::test]
    async fn swapped_nested_loop_join_preserves_output_partitioning() -> Result<()> {
        assert_distributed_matches_single_node(
            r"SELECT b.id AS bid, p.id AS pid FROM
                (SELECT id FROM build_side WHERE id % 25 = 0) b
                LEFT JOIN
                (SELECT id FROM probe_side ORDER BY id LIMIT 1000000) p
                ON p.id <= b.id
             ORDER BY bid NULLS LAST, pid NULLS LAST",
            true,
        )
        .await
    }
```

src/distributed_planner/inject_network_boundaries.rs
barbarj: Oh yeah, this was leftover from the validation stuff in the other PR. Removing it.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Hmmm good catch. I've added this test and updated the join normalization to skip cases where both sides have single partitions. Single partitions on both side should be safe to broadcast to and don't need to be rewritten. And multi-partition cases won't suffer this correctness bug. (That's my understanding of the behavior, at least.)

src/distributed_planner/normalize_collect_joins.rs
barbarj: Added this test and applied the same fix as the hash join single partition case.

tests/multi_task_collect_join_repros.rs
barbarj: Oh yeah. A couple cases are fixed by capping tasks, but this is a remnant of my first pass at solving this just by capping tasks. I'll fix the test names.

src/distributed_planner/inject_network_boundaries.rs
barbarj: Oh that makes sense. Added!

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,311 @@
...
+        .on()
+        .iter()
+        .map(|(l, r)| (Arc::clone(l), Arc::clone(r)))
+        .unzip();
+
+    let build_input = join
+        .left()
+        .downcast_ref::<CoalescePartitionsExec>()
+        .map_or_else(|| Arc::clone(join.left()), |c| Arc::clone(c.input()));
+
```
gabotechs: :thinking: these `CoalescePartitionExec` nodes can be carrying relevant things within, like `fetch` values that limit the amount of data pulled inside the node itself.

If we just do `c.input()`, we lose that `fetch` value (in case there's some).

This issue is also present in the NLJ path. Here are a couple of new integration test reproducers:

```rust
    #[tokio::test]
    async fn build_side_fetch_is_preserved_by_normalize() -> Result<()> {
        assert_distributed_matches_single_node(
            "SELECT b.id, p.id FROM (SELECT id FROM build_side LIMIT 50) b \
             FULL JOIN probe_side p ON b.id = p.id",
            true,
        )
        .await
    }
```

```rust
    #[tokio::test]
    async fn build_side_fetch_is_preserved_by_broadcast() -> Result<()> {
        assert_distributed_matches_single_node(
            "SELECT b.id, p.id FROM (SELECT id FROM build_side LIMIT 50) b \
             JOIN probe_side p ON b.id = p.id",
            true,
        )
        .await
    }
```

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,311 @@
...
+/// CollectLeft's single-partition requirement) is stripped, as the hash repartition replaces it.
+fn collect_left_to_partitioned(
+    join: &HashJoinExec,
+    target_partitions: usize,
+) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    assert!(
+        join.right().output_partitioning().partition_count() > 1,
+        "Single-partition joins are safe and can actually be made incorrect by repartitioning them"
+    );
+
```
gabotechs: Maybe using an `assert_or_internal_err!` macro is safer than panicking here.

src/distributed_planner/normalize_collect_joins.rs
```diff
@@ -0,0 +1,311 @@
...
+) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+    let d_cfg = DistributedConfig::from_config_options(cfg)?;
+    let target_partitions = cfg.execution.target_partitions;
+
+    plan.transform_down(|node| {
+        if let Some(join) = node.downcast_ref::<HashJoinExec>()
+            && join.mode == PartitionMode::CollectLeft
+            && !is_left_broadcast_safe(join.join_type())
+            && !join.null_aware
+            && !both_sides_use_single_partition(join.left(), join.right())
```
gabotechs, line 115: If `both_sides_use_single_partition` is true, it's fine to not transform the joins, but I think then we should be also throttling them to exactly 1 task no?

Otherwise, they'd run in 1 partition * T tasks == T partitions in total across machines, and we'd still be in trouble.

src/distributed_planner/normalize_collect_joins.rs
barbarj: I didn't know about that! Agreed, switching it.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Yeah, you're right. I'll cap them.

barbarj: Also updated the repro tests to assert the execution plans to make this more visible.

src/distributed_planner/normalize_collect_joins.rs
barbarj: Noted. I've fixed the first example by retaining the `CoalescePartitoinExec` if it has a fetch (same for the NJL case).

The seconds test failure is actually pre-existing. I can replicate it on main. INNER join's are broadcast-safe, so this PR does nothing to them. It definitely appears to be the same missing-fetch problem though.

barbarj: I think it's due to this part of `insert_broadcast.rs` losing the fetch: https://github.com/datafusion-contrib/datafusion-distributed/blob/9dd018a392185c6853f4d32a43d839219b31730d/src/distributed_planner/insert_broadcast.rs#L136
```rust
 let broadcast_input = build_child
     .map_or_else(
         || Arc::clone(build_child),
         |coalesce| Arc::clone(coalesce.input()),
    );
```

src/distributed_planner/normalize_collect_joins.rs
barbarj: Added! Let me know if they need tweaks.

gabotechs: Nice! what you did there is great

src/distributed_planner/normalize_collect_joins.rs
gabotechs: 🤔 You are right. I just opened an issue for this https://github.com/datafusion-contrib/datafusion-distributed/issues/591, this can be addressed in a separate PR as it's another issue.

gabotechs (approved):
Thanks @barbarj for this amazing piece of work 🙏! it was definitely a master lesson on how to do joins correctly.

Once we get the CI green, it should be fine to merge.

barbarj:
> Once we get the CI green, it should be fine to merge.

Thanks @gabotechs! CI should be green now. (I changed `build_side_fetch_is_preserved_by_normalize` to assert counts rather than row contents, since distributed execution of that fetch is non-deterministic in regards to which rows survive).

gabotechs:
Github runners are not in the mood today…

gabotechs:
Seems like the `build_side_fetch_is_preserved_by_normalize` test is failing because of some non-determinism in how parquet files are partitioned?

@barbarj, unless you have a better idea, I think we can omit asserting the plan snapshot there.

barbarj:
> Seems like the `build_side_fetch_is_preserved_by_normalize` test is failing because of some non-determinism in how parquet files are partitioned?
> 
> @barbarj, unless you have a better idea, I think we can omit asserting the plan snapshot there.

I've got nothing better. Removed!

PR #588, opened 2026-08-04
build: make the datafusion parquet and sql features optional

philippemnoel:
## What

This PR makes `parquet` and `sqlparser` optional, default-on features, so we can turn them off in @paradedb. Our build times are growing quite unwieldy, and we don't make use of these features in our embedded use case, so we'd like to turn them off. Hopefully this benefits others and doesn't get in the way for Datadog.

All tests were run and pass, and we merged this in our fork: paradedb/datafusion-distributed#59.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = ["benchmarks", "cli", "console"]
 
 [workspace.dependencies]
 datafusion = { version = "54.0.0", default-features = false }
-datafusion-proto = { version = "54.0.0" }
+datafusion-proto = { version = "54.0.0", default-features = false }
 
 [package]
 name = "datafusion-distributed"
@@ -17,8 +17,6 @@ repository = "https://github.com/datafusion-contrib/datafusion-distributed"
 [dependencies]
 chrono = { version = "0.4.44" }
 datafusion = { workspace = true, features = [
-  "parquet",
-  "sql",
   "unicode_expressions",
   "datetime_expressions",
 ] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -61,7 +59,13 @@ arrow = { version = "58", optional = true, features = ["test_utils"] }
 hyper-util = { version = "0.1.16", optional = true }
 
 [features]
-default = ["grpc"]
+default = ["grpc", "parquet", "sql"]
```
gabotechs, line 62: 👍 Looks good, although I'm not sure why this works:

- https://github.com/datafusion-contrib/datafusion-distributed/blob/ca19679c69281d8cc2b4445c41487adb473be32c/src/events/defaults/file_scan_config.rs

🤔 Is the underlaying upstream code not really coupled to parquet?

src/protocol/grpc/errors/datafusion_error.rs
```diff
@@ -190,7 +193,18 @@ impl DataFusionErrorProto {
...
+                {
+                    DataFusionError::ParquetError(Box::new(err.to_parquet_error()))
+                }
+                // Built without `parquet`, so `DataFusionError::ParquetError` does not exist
+                // here. The peer that sent this may still have it, so keep the message rather
+                // than dropping the error.
+                #[cfg(not(feature = "parquet"))]
+                {
+                    let _ = err;
+                    DataFusionError::Internal("ParquetError from peer".to_string())
```
gabotechs: Maybe we can at least paste the error message in the string.

gabotechs (approved):

Cargo.toml
philippemnoel: It all works on our side at least. I'm not exactly certain about the specifics of this file

src/protocol/grpc/errors/datafusion_error.rs
philippemnoel: Done. It turned out to be a bit more code than expected. If you want, you can revert the commit

gabotechs:
Thanks @philippemnoel!

PR #590, opened 2026-08-06
Expand documentation about `scale_up_leaf_node` guarantees.

stuhood:
The required behavior of nodes which have been scaled up with `scale_up_leaf_node` can be a bit unintuitive. If you don't correctly modify your `ExecutionPlan` and all partitions continue to emit data, then you will have duplicate rows. 

It would be great to find an approach that didn't require the "empty for all but one partition" behavior, but I couldn't think of one. So this is a PR to expand the docs, rather than a ticket discussing changing this behavior.

### DIFF
--- a/docs/source/advanced/05-custom-distributed-plans.md
+++ b/docs/source/advanced/05-custom-distributed-plans.md
@@ -82,7 +82,10 @@ automatic path. You only place the boundaries; the leaves are scaled for you.
 
 This means a custom leaf node still needs its desired task-count and leaf-scale handlers (or
 `DistributedTaskContext`-based dispatch) registered, just as it would for automatic planning — you do
-**not** need to hand-build `DistributedLeafExec` in your boundary-injection rule.
+**not** need to hand-build `DistributedLeafExec` in your boundary-injection rule. Regardless of how
+`ScaleUpLeafNodeHandler` splits the leaf (via `DistributedLeafExec` variants or a single plan), each node
+must handle `execute(partition, context)` for any partition index probed by upstream stage operators
+(such as `RepartitionExec`), returning an empty stream for unassigned partitions to prevent duplicate rows.
 
 ## Example: a progressive partial-reduction tree
 
--- a/docs/source/user-guide/04-distribute-custom-plan.md
+++ b/docs/source/user-guide/04-distribute-custom-plan.md
@@ -178,11 +178,18 @@ that task's worker:
 ```
 
 ```{note}
+When a task runs on a worker, upstream stage operators (such as `RepartitionExec`)
+may call `execute(partition, context)` for every partition index in the stage's
+requested partition range. Any leaf node returned by a `ScaleUpLeafNodeHandler` —
+whether wrapped in per-task `DistributedLeafExec` variants or as a single plan that
+dispatches internally — must return an empty record batch stream (such as
+`EmptyRecordBatchStream`) for any unassigned `partition` index. If unassigned
+partitions emit data instead of empty streams, duplicate rows will be produced.
+
 If your node dispatches internally — reading
 `DistributedTaskContext::from_ctx(&ctx).task_index` inside `execute()` to decide
 what to produce — you can skip `DistributedLeafExec` and return a single prepared
-plan directly. The runnable `numbers(start, end)` example linked below takes this
-route; the choice is yours.
+plan directly (as in the `numbers(start, end)` example linked below).
 ```
 
 ## Putting it together
--- a/examples/custom_execution_plan.rs
+++ b/examples/custom_execution_plan.rs
@@ -184,6 +184,12 @@ impl ExecutionPlan for NumbersExec {
         _partition: usize,
         context: Arc<TaskContext>,
     ) -> Result<SendableRecordBatchStream> {
+        // Upstream stage operators (such as RepartitionExec) may call execute() for any
+        // partition index in the stage. Leaf plans returned by scale_up_leaf_node handlers
+        // must return an empty stream for unassigned partition indices to avoid emitting
+        // duplicate rows. NumbersExec exposes a single partition (UnknownPartitioning(1)),
+        // so _partition is 0 here.
+        //
         // Get the distributed task context to determine which subset of numbers
         // this task should generate
         let dist_ctx = DistributedTaskContext::from_ctx(&context);
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -164,6 +164,11 @@ fn cached_file_scan_desired_task_count_handler(
     Some(DesiredTaskCountEventResponse::desired(usize::MAX))
 }
 
+/// Assigns each file to a task slot by hashing its path so the same file always lands in the
+/// same variant slot. Returns a `DistributedLeafExec` containing the per-task variants.
+/// Any leaf node returned by `scale_up_leaf_node` must handle `execute(partition, context)` for
+/// any partition index probed by upstream stage operators (such as `RepartitionExec`), returning
+/// an empty stream for unassigned partitions to prevent duplicate rows.
 fn cached_file_scan_scale_up_leaf_node_handler(
     ev: ScaleUpLeafNodeEvent,
 ) -> Option<Result<ScaleUpLeafNodeEventResponse>> {
--- a/src/distributed_ext.rs
+++ b/src/distributed_ext.rs
@@ -614,6 +614,11 @@ pub trait DistributedExt: Sized {
     /// Registers a handler that can replace leaf nodes with distributed variants once the
     /// distributed planner has decided a final task count.
     ///
+    /// Any leaf node returned by a leaf scaling handler (whether as per-task variants in
+    /// [DistributedLeafExec] or a single plan) must handle `execute(partition, context)` for
+    /// all partition indices in the stage's requested range. Returning data on unassigned
+    /// partitions causes duplicate rows; unassigned partition requests must return an empty stream.
+    ///
     /// A function with the following signature can be provided as argument:
     ///
     /// ```rust
--- a/src/execution_plans/distributed_leaf.rs
+++ b/src/execution_plans/distributed_leaf.rs
@@ -60,6 +60,16 @@ use std::sync::Arc;
 ///
 /// This way, the different workers get to execute different versions of the same plan, each
 /// handling its own range of non-overlapping data.
+///
+/// ## Leaf Execution Contract
+/// When a scaled-up leaf node (or a per-task variant in `DistributedLeafExec`) is executed on a
+/// worker, upstream stage operators (such as `RepartitionExec`) may call
+/// `leaf.execute(partition, context)` for every partition index `partition` in the stage's
+/// requested partition range. Leaf plans or variants that map to a specific physical partition
+/// must return an empty record batch stream (such as `EmptyRecordBatchStream`) when `partition`
+/// does not match their assigned partition, allowing multi-partition stage pipelines to pull and
+/// repartition data without failing. If unassigned partitions emit data instead of returning an
+/// empty stream, duplicate rows will be produced.
 #[derive(Debug)]
 pub struct DistributedLeafExec {
     pub(crate) original: Arc<dyn ExecutionPlan>,


docs/source/advanced/05-custom-distributed-plans.md
```diff
@@ -82,7 +82,10 @@ automatic path. You only place the boundaries; the leaves are scaled for you.
 
 This means a custom leaf node still needs its desired task-count and leaf-scale handlers (or
 `DistributedTaskContext`-based dispatch) registered, just as it would for automatic planning — you do
-**not** need to hand-build `DistributedLeafExec` in your boundary-injection rule.
+**not** need to hand-build `DistributedLeafExec` in your boundary-injection rule. Regardless of how
+`ScaleUpLeafNodeHandler` splits the leaf (via `DistributedLeafExec` variants or a single plan), each node
+must handle `execute(partition, context)` for any partition index probed by upstream stage operators
+(such as `RepartitionExec`), returning an empty stream for unassigned partitions to prevent duplicate rows.
```
gabotechs, line 88: 🤔 I don't fully understand this part:

> each node must handle `execute(partition, context)` for any partition index probed by upstream stage operators (such as `RepartitionExec`)

This is not really something specific to this project, it's more a pure DataFusion concept: if your node exposes P partitions, your node needs to be prepared to be called with `execute(0..N, context)` and return a valid stream.

docs/source/user-guide/04-distribute-custom-plan.md
```diff
@@ -181,11 +178,18 @@ that task's worker:
 ```
 
 ```{note}
+When a task runs on a worker, upstream stage operators (such as `RepartitionExec`)
+may call `execute(partition, context)` for every partition index in the stage's
+requested partition range. Any leaf node returned by a `ScaleUpLeafNodeHandler` —
+whether wrapped in per-task `DistributedLeafExec` variants or as a single plan that
+dispatches internally — must return an empty record batch stream (such as
+`EmptyRecordBatchStream`) for any unassigned `partition` index. If unassigned
+partitions emit data instead of empty streams, duplicate rows will be produced.
```
gabotechs, line 187: Ok, I think I understand what's happening here.

You guys seem be doing something like this:

1. You have a data source that  adheres to `datafusion.execution.target_partitions` number of partitions, let's assume this value is 6 because you are in a 6 CPU machine:

```
┌────────────────────────────────────┐
│             DataSource             │
│                                    │
│┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐│
││ p0 ││ p1 ││ p2 ││ p3 ││ p4 ││ p5 ││
│└────┘└────┘└────┘└────┘└────┘└────┘│
└────────────────────────────────────┘
```

2. You wrap copies of the same data source with untrimmed partitions as variants of a DistributedLeafExec

```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                DistributedLeafExec                                                 │
│                                                                                                                    │
│ ┌────────────────────────────────────┐┌────────────────────────────────────┐┌────────────────────────────────────┐ │
│ │             DataSource             ││             DataSource             ││             DataSource             │ │
│ │                                    ││                                    ││                                    │ │
│ │┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐││┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐││┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐│ │
│ ││ p0 ││ p1 ││ p2 ││ p3 ││ p4 ││ p5 ││││ p0 ││ p1 ││ p2 ││ p3 ││ p4 ││ p5 ││││ p0 ││ p1 ││ p2 ││ p3 ││ p4 ││ p5 ││ │
│ │└────┘└────┘└────┘└────┘└────┘└────┘││└────┘└────┘└────┘└────┘└────┘└────┘││└────┘└────┘└────┘└────┘└────┘└────┘│ │
│ └────────────────────────────────────┘└────────────────────────────────────┘└────────────────────────────────────┘ │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

3. During execution, you then return empty streams for partitions that are not meant to be handled by each specific task index:

```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                DistributedLeafExec                                                 │
│                                                                                                                    │
│ ┌────────────────────────────────────┐┌────────────────────────────────────┐┌────────────────────────────────────┐ │
│ │             DataSource             ││             DataSource             ││             DataSource             │ │
│ │                                    ││                                    ││                                    │ │
│ │┌────┐┌────┐┌ ─ ─ ┌ ─ ─ ┌ ─ ─ ┌ ─ ─ ││┌ ─ ─ ┌ ─ ─ ┌────┐┌────┐┌ ─ ─ ┌ ─ ─ ││┌ ─ ─ ┌ ─ ─ ┌ ─ ─ ┌ ─ ─ ┌────┐┌────┐│ │
│ ││ p0 ││ p1 │  p2 │  p3 │  p4 │  p5 │││  p0 │  p1 ││ p2 ││ p3 │  p4 │  p5 │││  p0 │  p1 │  p2 │  p3 ││ p4 ││ p5 ││ │
│ │└────┘└────┘└ ─ ─ └ ─ ─ └ ─ ─ └ ─ ─ ││└ ─ ─ └ ─ ─ └────┘└────┘└ ─ ─ └ ─ ─ ││└ ─ ─ └ ─ ─ └ ─ ─ └ ─ ─ └────┘└────┘│ │
│ └────────────────────────────────────┘└────────────────────────────────────┘└────────────────────────────────────┘ │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

If this is what you are doing, then you probably should not be using `DistributedLeafExec`.

The point of `DistributedLeafExec` is that the different contained variants have all `datafusion.execution.target_partitions` partitions, and each variant handles a complete different set of files.

Following the example above, imagine this, each `f*` being a file to read (assuming the data source is reading files):

```
p0: f0, f1, f2, f3, f4, f5, f6, f7, f8, f9
p1: f10, f11, f12, f13, f14, f15, f16, f17, f18, f19
p3: f20, f21, f22, f23, f24, f25, f26, f27, f28, f29
p3: f30, f31, f32, f33, f34, f35, f36, f37, f38, f39
p4: f40, f41, f42, f43, f44, f45, f46, f47, f48, f49
p5: f50, f51, f52, f53, f54, f55, f56, f57, f58, f59
```

You should not be stamping all p0, p1, p2, p3, p4, p5 in the task with index 0, as you'll be sending a plan over the network with a lot of useless information irrelevant for task 0.

```
p0: f0, f1, f2, f3, f4, f5, f6, f7, f8, f9 <- useful for task 0
p1: f10, f11, f12, f13, f14, f15, f16, f17, f18, f19 <- useful for task 0
p3: f20, f21, f22, f23, f24, f25, f26, f27, f28, f29 <- useless for task 0, the price of sending this over the wire was paid with no benefit
p3: f30, f31, f32, f33, f34, f35, f36, f37, f38, f39 <- useless for task 0, the price of sending this over the wire was paid with no benefit
p4: f40, f41, f42, f43, f44, f45, f46, f47, f48, f49 <- useless for task 0, the price of sending this over the wire was paid with no benefit
p5: f50, f51, f52, f53, f54, f55, f56, f57, f58, f59 <- useless for task 0, the price of sending this over the wire was paid with no benefit
```

The correct way of using `DistributedLeafExec` in this scenario would be to re-distribute files across variants, where each variant will be executed on `datafusion.execution.target_partitions`. For example:

```
p0: f0, f1, f2, f3
p1: f4, f5, f6
p3: f7, f8, f9
p3: f10, f11, f12, f13
p4: f14, f15, f16
p5: f17, f18, f19 
p6: f20, f21, f22, f23, 
p7: f24, f25, f26, 
p8: f27, f28, f29 
p9: f30, f31, f32, f33,
p10: f34, f35, f36, 
p11: f37, f38, f39
p12: f40, f41, f42, f43,
p13: f44, f45, f46,
p14: f47, f48, f49  
p15: f50, f51, f52, f53,
p16: f54, f55, f56,
p17:  f57, f58, f59
```

And then creating variants that handle those partitions:

```
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                DistributedLeafExec                                                 │
│                                                                                                                    │
│ ┌────────────────────────────────────┐┌────────────────────────────────────┐┌────────────────────────────────────┐ │
│ │             DataSource             ││             DataSource             ││             DataSource             │ │
│ │                                    ││                                    ││                                    │ │
│ │┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐││┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐││┌────┐┌────┐┌────┐┌────┐┌────┐┌────┐│ │
│ ││ p0 ││ p1 ││ p2 ││ p3 ││ p4 ││ p5 ││││ p6 ││ p7 ││ p8 ││ p9 ││p10 ││p11 ││││p12 ││p13 ││p14 ││p15 ││p16 ││p17 ││ │
│ │└────┘└────┘└────┘└────┘└────┘└────┘││└────┘└────┘└────┘└────┘└────┘└────┘││└────┘└────┘└────┘└────┘└────┘└────┘│ │
│ └────────────────────────────────────┘└────────────────────────────────────┘└────────────────────────────────────┘ │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

gabotechs: This is pretty much the only point of `DistributedLeafExec`. If for some reason your setup does not allow to re-distribute files like this, you probably should be either:
- Using a `PartitionIsolatorExec` (no longer maintained in this project, but people can still recover it and use it in their codebases) assuming you are willing to pay the price of unnecessarily serializing partitions that are going to end up being ignored
- Just reacting to the presence of a `DistributedTaskContext` present in `Arc<TaskContext>` during execution.

docs/source/advanced/05-custom-distributed-plans.md
stuhood: The critical (and unique) bit is this part:
> returning an empty stream for unassigned partitions to prevent duplicate rows.

docs/source/user-guide/04-distribute-custom-plan.md
stuhood: > If this is what you are doing, then you probably should not be using `DistributedLeafExec`.

Correct, that is what we are doing.

> You should not be stamping all p0, p1, p2, p3, p4, p5 in the task with index 0, as you'll be sending a plan over the network with a lot of useless information irrelevant for task 0.

Our metadata is very small, but yea: I get your point.

----

I believe that some of the disconnect might be in this sentence:
> Every variant must expose the same schema and partition count (`try_new` enforces this) — that is what keeps the node transparent to the network boundaries above it.

When constructing variants under a `DistributedLeafExec`, the variants must have the same number of partitions as one another, BUT, they:
* don't need to have a partition count equal to the original `ExecutionPlan`
* don't need to have a partition count equal to the number of tasks
* don't need to have a partition count equal to the target partitions (but might want to)

So I think that in our case (where we always aim to have 1 task per worker, because we use a single-threaded-runtime), the simplest thing to do would be to always create single partition variants.

The above may seem obvious to you, but it's surprising to me, because my assumption was definitely that our leaf needed the same partition count before and after being "scaled up". That may have been based on a misunderstanding about when `scale_up_leaf_node` runs... it must be running _before_ e.g. `EnforceDistribution`, for example?

gabotechs: > The above may seem obvious to you, but it's surprising to me, because my assumption was definitely that our leaf needed the same partition count before and after being "scaled up".

Probably the clarifications should go in this direction then. This is indeed not a real constraint, variants need to agree on the same partition count, but they might diverge from the partition count of the original plan.

> So I think that in our case (where we always aim to have 1 task per worker, because we use a single-threaded-runtime), the simplest thing to do would be to always create single partition variants.

Yes, this sounds right

> it must be running before e.g. EnforceDistribution, for example?

It runs after EnforceDistribution actually. Running it before EnforceDistribution might be problematic, for example, if you indeed choose one partition per variant, EnforceDistribution will attempt to remove repartitions and other operators that are necessary for a multi-partitioned setup, believing that it's running under just 1 partition, but it's in truth running under 1 * T partitions, where T is the number of tasks.

stuhood:
Opened https://github.com/datafusion-contrib/datafusion-distributed/pull/619 to replace this. Thanks!

Issue #532, opened 2026-07-02
[dynamic filtering] 4. implement remote dynamic filtering


jayshrivastava:
Dynamic filters won't work out of the box when the producer is remote from the consumer. We need to add the necessary plumbing to make this work.

I have thought about this quite a bit and the way to do it is something like
1. Identify tasks which produce dynamic filters.
2. Identify tasks which consume those dynamic filters. 
3. On execution, producers should send dynamic filters to the coordinator
4. The coordinator should "merge" the dynamic filters. Currently dynamic filters have a CASE expr for each partition id. We need to merge all the cases together into one filter.
5. The coordinator must push the merged filter to all the consumer tasks

This is very similar to how dynamic filters are implemented in vanilla datafusion. The consumers (data sources) don't know which partition certain rows will be in when they reach the producer node, so the CASEs are used to resolve this in the consumers. We have to do something similar in distributed datafusion, except we must be able to "merge" the cases together.

<img width="610" height="331" alt="Image" src="https://github.com/user-attachments/assets/1f530d67-25c4-4a90-81df-3f4edcfdccdc" />

Note the partitioning of each Data Source here is different than (a) the partitioning of the task and (b) the partitioning of the dynamic filter producer (HashJoinExec). There may be several layers of shuffles, repartitions etc. etc between the producers and consumers. Vanilla datafusion solves this problem using the CASE method described above. That seems to be the best approach.

PR #592, opened 2026-08-06
feat: add dynamic filter store in coordinator

jayshrivastava:
This PR adds a registry in the query coordinator for collecting dynamic filter expressions by expression ID. These can be collected during planning once `apply_expressions` becomes available.

This store counts received expressions (so we know when we've collected all the filters from workers in a task) and supports the merge() operation which ORs them, following the [RFC](https://github.com/datafusion-contrib/datafusion-distributed/pull/553).

Informs: https://github.com/datafusion-contrib/datafusion-distributed/issues/532

Remaining work
- implmement dynamic filter update collection and propagation from worker <-> coordinator

### DIFF
--- a/src/coordinator/distributed.rs
+++ b/src/coordinator/distributed.rs
@@ -191,6 +191,8 @@ impl ExecutionPlan for DistributedExec {
             Arc::clone(&context),
             &self.metrics,
             self.metrics_store.clone(),
+            // Dynamic-filter discovery will supply the query's expression IDs here.
+            std::iter::empty(),
         );
 
         let mut builder = RecordBatchReceiverStreamBuilder::new(self.schema(), 1);
--- a/src/coordinator/dynamic_filters.rs
+++ b/src/coordinator/dynamic_filters.rs
@@ -53,6 +53,9 @@ impl DynamicFilterStore {
     /// The merge is non-destructive: subsequent calls see the same expressions plus any that have
     /// arrived since the previous call. The expression tree is balanced to avoid creating a deep
     /// left- or right-associated tree when a filter has many producers.
+    ///
+    /// DataFusion may eventually merge dynamic filter expressions natively; see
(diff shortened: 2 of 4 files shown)


src/coordinator/distributed.rs
```diff
@@ -191,6 +191,8 @@ impl ExecutionPlan for DistributedExec {
             Arc::clone(&context),
             &self.metrics,
             self.metrics_store.clone(),
+            // Dynamic-filter discovery will supply the query's expression IDs here.
+            std::iter::empty(),
```
jayshrivastava, line 195: Eventually, I will implement a plan traversal which collects dynamic filters using `apply_expressions`

src/coordinator/dynamic_filters.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Identifies all instances of the same dynamic filter within a query.
+///
+/// Equivalent to [`PhysicalExpr::expression_id`].
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub(crate) struct ExpressionId(pub(crate) u64);
+
+/// Collection of completed dynamic filter expressions received from producers.
+pub(super) struct DynamicFilterStore {
+    expressions: Mutex<HashMap<ExpressionId, Vec<Arc<dyn PhysicalExpr>>>>,
+}
```
gabotechs, line 16: It's probably too soon to start speculating about how this will look like.

For example, I'm certain that this is not going to be enough, and that we should be using a Tokio `Watch` for broadcasting updates.

I recommend to defer this work until we can actually verify that dynamic filters work end to end, otherwise we will be merging code that I'm pretty sure we'll need to heavily change in the future.

One thing we can start doing is building on top of a branch that cherry picks all necessary work upstream, and creating these new extra structs in this project for having dynamic filters working end to end.

jayshrivastava: No problem! I'll close this out

PR #594, opened 2026-08-09
[Iceberg] Commit taxi iceberg dataset

gabotechs:
Commits an Iceberg dataset to LFS for future testing. This is the same approach we use today for parquet files, but for Iceberg in this case, the upper PR in this stack will use it for integration testing.

---


shinzoxD (changes requested):
One factual fixture/documentation mismatch needs correction: I inspected all 91 Parquet column chunks (7 files × 13 columns) with PyArrow 25; each reports `UNCOMPRESSED` and `parquet-rs 58.3.0`, while the README calls the files Zstandard-compressed and shows `COMPRESSION zstd`. Please either regenerate the committed files with ZSTD or update the README/extraction description to match the actual fixture. The other integrity checks passed: 175,000 rows total, 25,000 per date partition, positive-value filters, field IDs 1–13, and Iceberg v2 manifest paths/counts/sizes all agree.

PR #595, opened 2026-08-09
[Iceberg] Add basic iceberg integration

gabotechs:
Closes https://github.com/datafusion-contrib/datafusion-distributed/issues/600

> [!NOTE]
> The PR is large because it contains a lot of boilerplate for a basic Iceberg integration, but it was crafted by hand, and it contains just whatever is minimal to get some tests e2e running

Adds a very basic Iceberg integration as a new crate that depends on `datafusion`, `datafusion-distributed` and `iceberg-rust`.

This is mostly a rewrite of https://github.com/apache/iceberg-rust/tree/main/crates/integrations/datafusion, with a few exceptions for helpers like expression transpilation.

It heavily relies on `WorkUnitFeed`s for streaming `FileScanTask`s at runtime, following a pattern that looks like this today in single-node:

```rust
┌────────────────────────────────────────────┐
│             IcebergDataSource              │
│                                            │
│┌──────────────────────────────────────────┐│
││           IcebergWorkUnitFeed            ││
││┌────────────┐┌────────────┐┌────────────┐││
│││   Feed 0   ││   Feed 1   ││   Feed 2   │││
││└──────┬─────┘└──────┬─────┘└──────┬─────┘││
│└───────┼─────────────┼─────────────┼──────┘│
│  .─────▼─────. .─────▼─────. .─────▼─────. │
│ (FileScanTask (FileScanTask (FileScanTask )│
│  .───────────. `─────┬─────' .───────────. │
│ (FileScanTask )      │      (FileScanTask )│
│  `─────┬─────'       │       .───────────. │
│        │             │      (FileScanTask )│
│        │             │       `─────┬─────' │
│        │             │             │       │
│ ┌──────▼─────┐┌──────▼─────┐┌──────▼─────┐ │
│ │Partition 0 ││Partition 1 ││Partition 2 │ │
│ │ArrowReader ││ArrowReader ││ArrowReader │ │
│ └──────┬─────┘└──────┬─────┘└──────┬─────┘ │
│        │             │             │       │
│  .─────▼─────.       │       .─────▼─────. │
│ ( RecordBatch ).─────▼─────.( RecordBatch )│
│  `─────┬─────'( RecordBatch ).───────────. │
│        │       `─────┬─────'( RecordBatch )│
│        │             │       `───────────' │
└────────┼─────────────┼─────────────┼───────┘
         ▼             ▼             ▼
```

And that will look like this once distribution is supported:

```rust
[...]

PR #597, opened 2026-08-09
[Iceberg] Add Iceberg crate dependencies

gabotechs:
Introduces the empty `datafusion-distributed-iceberg` crate, its dependency set, and the corresponding lockfile updates.

This is the first layer of the Iceberg integration stack. The implementation follows in #595.

### DIFF
--- a/src/execution_plans/sampler.rs
+++ b/src/execution_plans/sampler.rs
@@ -255,6 +255,7 @@ impl PartitionSampler {
                 // The input produced nothing at all: the stream is exhausted, so this partition
                 // reached EOS with zero rows.
                 reporter.load_info.reached_eos = true;
+                reporter.report(&peek);
                 return Ok(peek.chain(input_stream).boxed());
             };
             let _guard = elapsed_compute.timer();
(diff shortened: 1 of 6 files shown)


shinzoxD (changes requested):
The new workspace crate is not covered by the current CI commands. Because the workspace root is itself a package and there is no `default-members` entry, `cargo metadata --no-deps` reports only `datafusion-distributed` as a default member. The workflow runs `cargo build`, `cargo test --features integration`, and `cargo clippy --all-targets --all-features` without `--workspace` or `-p datafusion-distributed-iceberg`, so the green checks do not compile or lint this crate (or validate its dependency/feature combination); only `cargo fmt --all` reaches it.

Please add explicit CI coverage for `datafusion-distributed-iceberg` (or include it in the intended default-member/workspace commands) as part of introducing the crate. I also attempted the missing `cargo check -p datafusion-distributed-iceberg --all-targets` locally, but its fresh dependency build exceeded a three-minute bounded run, so I am not claiming a local compiler result.

src/execution_plans/sampler.rs
```diff
@@ -255,6 +255,7 @@ impl PartitionSampler {
                 // The input produced nothing at all: the stream is exhausted, so this partition
                 // reached EOS with zero rows.
                 reporter.load_info.reached_eos = true;
+                reporter.report(&peek);
```
gabotechs, line 258: Needed for the Rust toolchain upgrade

PR #617, opened 2026-08-09
feat: allow fractional desired task counts

### LINKED ISSUE #562: Allow DesiredTaskCountHandler to return f64 as desired task counts
Implementations might choose to declare fractions of tasks, so that when they get aggregated together (e.g., during `ChildrenIsolatorUnionExec`s), they do not request a higher task count than needed.

When resolving to the final task count, the ceil of the fractional value should be taken into account.

shinzoxD:
## Summary

- allow `DesiredTaskCountHandler` responses to use fractional `f64` task-count hints
- preserve fractions through hint merging and union aggregation, then round up when resolving an integer task count
- migrate built-in handlers, examples, public docs, and the 3.0 upgrade guide

Closes #562

## Validation

- `cargo check --lib`
- `cargo clippy --lib -- -D warnings`
- `cargo test --features integration --lib -- --skip protocol::grpc::channel_resolver::tests::fails_establishing_connection` (289 passed, 0 failed, 1 ignored, 1 filtered out)
- `cargo check --example custom_worker_url_routing --features integration`

The library suite and example check passed locally with the benchmark-only dev-dependency edge temporarily excluded to avoid its unrelated vendored-OpenSSL Windows build. The snapshot path filter was normalized only for the Windows test run, and the Windows-only connection-failure assertion was skipped because its socket error text differs by platform. `Cargo.toml`, `Cargo.lock`, and the snapshot helper were restored with no diff before commit.

Assisted-by: OpenAI Codex

### DIFF
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -169,9 +169,7 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(
-        usize::MAX as f64,
-    )))
+    Some(Ok(DesiredTaskCountEventResponse::desired(f64::MAX)))
 }
 
 fn cached_file_scan_scale_up_leaf_node_handler(
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -169,7 +169,9 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(usize::MAX)))
+    Some(Ok(DesiredTaskCountEventResponse::desired(
+        usize::MAX as f64,
+    )))
 }
 
 fn cached_file_scan_scale_up_leaf_node_handler(
--- a/src/coordinator/prepare_dynamic_plan.rs
+++ b/src/coordinator/prepare_dynamic_plan.rs
@@ -107,10 +107,13 @@ pub(super) async fn prepare_dynamic_plan(
                     let (stats, new_metrics) =
                         gather_runtime_statistics(load_info_rxs, &input_stage.plan).await?;
                     metrics.extend(new_metrics);
-                    // returning Desired(1) here is our way to tell the planner that we don't care
+                    // returning Desired(0) here is our way to tell the planner that we don't care
                     // about the task count assigned to the network boundary in the consumer stage,
-                    // and we don't want it to affect other task count decisions.
-                    (Some(Arc::new(stats)), Desired(1.0))
+                    // and we don't want it to affect other task count decisions. A 0.0 hint does
+                    // not mask smaller fractional values via max-merge, and does not inflate
+                    // UNION sums when this boundary sits under a child of a
+                    // ChildrenIsolatorUnionExec.
+                    (Some(Arc::new(stats)), Desired(0.0))
                 };
 
                 // Capture the output partitioning of the (rescaled, sampler-wrapped) input plan
--- a/src/coordinator/prepare_dynamic_plan.rs
+++ b/src/coordinator/prepare_dynamic_plan.rs
@@ -68,7 +68,7 @@ pub(super) async fn prepare_dynamic_plan(
                 .clamp(1, nb_ctx.max_tasks()?);
             let task_count = nb_ctx
                 .task_count(&input_stage.plan)?
-                .merge(Desired(compute_based_task_count));
+                .merge(Desired(compute_based_task_count as f64));
 
             // Propagate the final task_count inferred based on runtime statistics and compute cost.
             // Here is where leaf nodes are scaled up by ScaleUpLeafNodeHandler, and the
@@ -110,7 +110,7 @@ pub(super) async fn prepare_dynamic_plan(
                     // returning Desired(1) here is our way to tell the planner that we don't care
                     // about the task count assigned to the network boundary in the consumer stage,
                     // and we don't want it to affect other task count decisions.
-                    (Some(Arc::new(stats)), Desired(1))
+                    (Some(Arc::new(stats)), Desired(1.0))
                 };
 
                 // Capture the output partitioning of the (rescaled, sampler-wrapped) input plan
--- a/src/distributed_planner/inject_network_boundaries.rs
+++ b/src/distributed_planner/inject_network_boundaries.rs
@@ -790,8 +790,8 @@ mod tests {
             .broadcast_joins(false);
         let annotated = annotate_test_plan(test_plan_builder, query).await;
         assert_snapshot!(annotated, @"
-        HashJoinExec: task_count=Desired(2)
-          NetworkShuffleExec: task_count=Desired(2)
+        HashJoinExec: task_count=Desired(1.3333333333333333)
+          NetworkShuffleExec: task_count=Desired(1.3333333333333333)
             RepartitionExec: task_count=Desired(2)
               ProjectionExec: task_count=Desired(2)
                 AggregateExec: task_count=Desired(2)
@@ -801,7 +801,7 @@ mod tests {
                         FilterExec: task_count=Desired(4)
                           RepartitionExec: task_count=Desired(4)
                             DistributedLeafExec: task_count=Desired(4)
-          NetworkShuffleExec: task_count=Desired(2)
+          NetworkShuffleExec: task_count=Desired(1.3333333333333333)
             RepartitionExec: task_count=Desired(2)
               ProjectionExec: task_count=Desired(2)
                 AggregateExec: task_count=Desired(2)
@@ -850,8 +850,8 @@ mod tests {
             .broadcast_joins(false);
         let annotated = annotate_test_plan(test_plan_builder, query).await;
         assert_snapshot!(annotated, @r"
-        AggregateExec: task_count=Desired(3)
-          NetworkShuffleExec: task_count=Desired(3)
+        AggregateExec: task_count=Desired(2.6666666666666665)
+          NetworkShuffleExec: task_count=Desired(2.6666666666666665)
             RepartitionExec: task_count=Desired(4)
               AggregateExec: task_count=Desired(4)
                 DistributedLeafExec: task_count=Desired(4)
@@ -1003,8 +1003,8 @@ mod tests {
             .desired_task_count_handler(repartition_max_one_desired_task_count_handler);
         let annotated = annotate_test_plan(test_plan_builder, query).await;
         assert_snapshot!(annotated, @r"
-        AggregateExec: task_count=Desired(1)
-          NetworkShuffleExec: task_count=Desired(1)
+        AggregateExec: task_count=Desired(0.6666666666666666)
+          NetworkShuffleExec: task_count=Desired(0.6666666666666666)
             RepartitionExec: task_count=Desired(1)
               AggregateExec: task_count=Desired(1)
                 DistributedLeafExec: task_count=Desired(1)
--- a/src/distributed_planner/inject_network_boundaries.rs
+++ b/src/distributed_planner/inject_network_boundaries.rs
@@ -790,8 +790,8 @@ mod tests {
             .broadcast_joins(false);
         let annotated = annotate_test_plan(test_plan_builder, query).await;
         assert_snapshot!(annotated, @"
-        HashJoinExec: task_count=Desired(1.3333333333333333)
-          NetworkShuffleExec: task_count=Desired(1.3333333333333333)
+        HashJoinExec: task_count=Desired(1.33)
+          NetworkShuffleExec: task_count=Desired(1.33)
             RepartitionExec: task_count=Desired(2)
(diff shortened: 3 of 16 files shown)


examples/custom_worker_url_routing.rs
```diff
@@ -161,7 +161,9 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(usize::MAX)))
+    Some(Ok(DesiredTaskCountEventResponse::desired(
+        usize::MAX as f64,
+    )))
```
gabotechs: This is a bit weird, we could move this to `f64::MAX` instead

src/events/desired_task_count.rs
```diff
@@ -184,3 +215,14 @@ impl DesiredTaskCountHandlers {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn fractional_desired_task_count_rounds_up_when_resolved() {
+        assert_eq!(Desired(0.5).as_usize(), 1);
+        assert_eq!(Desired(1.25).as_usize(), 2);
+    }
+}
```
gabotechs: This test is really just testing that the `.ceil()` method in the standard Rust library works, I'd remove it

src/events/desired_task_count.rs
```diff
@@ -146,6 +164,19 @@ where
...
+    ) -> Option<Result<DesiredTaskCountEventResponse>> {
+        ev.plan
+            .children()
+            .is_empty()
+            .then(|| Ok(DesiredTaskCountEventResponse::desired(*self as f64)))
+    }
+}
+
+#[async_trait]
+impl DesiredTaskCountHandler for f64 {
```
gabotechs: I understand that this is following the previous existing pattern, which is a good practice in general, but I'd like to ask to avoid doing it as it leaks into the public API and I'd say it's not really a good pattern.

This is used in `test_union_all_sums_fractional_task_counts_before_rounding`, can we just have a normal event handler function definition there instead?

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -883,6 +883,32 @@ mod tests {
...
+        let annotated = annotate_test_plan(test_plan_builder, query).await;
+        assert_snapshot!(annotated, @r"
+        ChildrenIsolatorUnionExec: task_count=Desired(1)
+          FilterExec: task_count=Maximum(1)
+            RepartitionExec: task_count=Maximum(1)
+              DistributedLeafExec: task_count=Maximum(1)
+          ProjectionExec: task_count=Maximum(1)
+            FilterExec: task_count=Maximum(1)
+              RepartitionExec: task_count=Maximum(1)
+                DistributedLeafExec: task_count=Maximum(1)
```
gabotechs, line 916: 🤔 I'm not sure why there's a `Maximum(1)` getting stamped in the two separate branches.

How it is that we are not seeing a `Desired(0.5)` instead?

gabotechs (commented):
This looks mostly great, thanks @shinzoxD! just left some minor comments

examples/custom_worker_url_routing.rs
shinzoxD: Changed the sentinel to `f64::MAX` in `d32a4cf`.

src/events/desired_task_count.rs
shinzoxD: Removed the direct `ceil()` test in `d32a4cf`.

src/events/desired_task_count.rs
shinzoxD: Removed the public `DesiredTaskCountHandler for f64` implementation. The union regression now registers a normal test-local event handler that returns `desired(0.5)` for leaf nodes.

src/distributed_planner/inject_network_boundaries.rs
shinzoxD: The `Maximum(1)` annotations are the final per-child allocations: after the union resolves to one task, `propagate_task_count_until_network_boundaries` stamps each allocated child count as a hard cap. The union's `Desired(1)` is the key assertion that the two `0.5` hints were summed before rounding; early rounding would produce `Desired(2)`. I added that explanation beside the snapshot.

gabotechs: Sounds good. Can we add another test that double checks what happens when the two leaves of the union report 0.0 task count?

src/events/defaults/file_scan_config.rs
```diff
@@ -30,7 +30,9 @@ pub(crate) fn file_scan_config_desired_task_count(
         .div_ceil(d_cfg.file_scan_config_bytes_per_partition)
         .div_ceil(cfg.target_partitions());
 
-    Some(Ok(DesiredTaskCountEventResponse::desired(task_count)))
+    Some(Ok(DesiredTaskCountEventResponse::desired(
+        task_count as f64,
+    )))
```
gabotechs: Here, we should no longer be doing a `.div_ceil()`, we should be doing normal f64 divisions so that we get true fractional numbers.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -790,8 +790,8 @@ mod tests {
             .broadcast_joins(false);
         let annotated = annotate_test_plan(test_plan_builder, query).await;
         assert_snapshot!(annotated, @"
-        HashJoinExec: task_count=Desired(2)
-          NetworkShuffleExec: task_count=Desired(2)
+        HashJoinExec: task_count=Desired(1.3333333333333333)
```
gabotechs: For the Debug implementation, it might be enough to just show 2 decimal numbers.

shinzoxD:
CI exposed three stale inline snapshot expectations that should now remain fractional. I updated only those expectations in 849f66d; the broad local library suite passes with 289 tests passed, 0 failed, 1 ignored, and the known Windows-only socket-text assertion filtered out. The replacement workflow is waiting for external-contributor approval: https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/31325189612

gabotechs:
@shinzoxD let me know if you still want to work on this, otherwise we can have other contributors help here

shinzoxD:
Yes, still working on this. The remaining review items are:

1. a union-of-zero-task-count leaves test
2. dropping `.div_ceil()` in `file_scan_config` so the default handler returns a true fractional `f64`
3. two-decimal `Debug` for the task-count annotation

I'll push those shortly.

src/distributed_planner/inject_network_boundaries.rs
shinzoxD: Added in 01c6441 as `test_union_all_zero_task_count_leaves`. Two `Desired(0.0)` leaves still sum before rounding, then `ChildrenIsolatorUnionExec` rejects the resulting zero-task stage as an internal planning error rather than silently collapsing to one task.

src/events/defaults/file_scan_config.rs
shinzoxD: Done in 01c6441. The default file-scan handler now does `total_bytes as f64 / bytes_per_partition / target_partitions` (and the work-unit scan helper matches). `as_usize()` still ceils only when the planner resolves an integer task count.

src/distributed_planner/inject_network_boundaries.rs
shinzoxD: `TaskCountAnnotation::Debug` now prints whole numbers as `Desired(4)` and fractions with two decimals (`Desired(1.33)`). Snapshot expectations in this file were updated to match.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -912,6 +912,32 @@ mod tests {
...
+            .num_workers(4)
+            .distributed_planner(false)
+            .broadcast_joins(false)
+            .desired_task_count_handler(zero_leaf_desired_task_count_handler);
+        // Two 0.0 leaf hints sum to Desired(0) before rounding. The isolator
+        // then rejects a zero-task union as an internal planning error rather
+        // than silently collapsing to a single task.
+        let err = annotate_test_plan_result(test_plan_builder, query)
+            .await
+            .expect_err("zero-task union should fail planning");
```
gabotechs: This should not error out. A `Desired(0)` is a valid outcome (e.g. all datasets in the leaves of the union are empty). Sounds like something we need to fix.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -1379,16 +1414,22 @@ mod tests {
...
     async fn annotate_test_plan(test_plan_builder: TestPlanBuilder, query: &str) -> String {
+        annotate_test_plan_result(test_plan_builder, query)
+            .await
+            .expect("failed to annotate plan")
+    }
+
+    async fn annotate_test_plan_result(
+        test_plan_builder: TestPlanBuilder,
+        query: &str,
+    ) -> datafusion::error::Result<String> {
```
gabotechs: We never want to assert errors here, I'd rollback this change.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -109,7 +109,7 @@ pub(super) async fn prepare_dynamic_plan(
                     // returning Desired(1) here is our way to tell the planner that we don't care
                     // about the task count assigned to the network boundary in the consumer stage,
                     // and we don't want it to affect other task count decisions.
-                    (Some(Arc::new(stats)), Desired(1))
+                    (Some(Arc::new(stats)), Desired(1.0))
```
gabotechs: 🤔 I think we should be returning 0.0 here. The dynamic planner might now decide to that maybe 0.2 tasks are enough, and this might get added between children of a UNION, so we would not want the 0.2 to be masked by the greater 1.0 in this case, we want the lower fractional values to be summed across children in a UNION.

src/events/defaults/file_scan_config.rs
```diff
@@ -26,9 +26,9 @@ pub(crate) fn file_scan_config_desired_task_count(
         }
     }
 
-    let task_count = total_bytes
-        .div_ceil(d_cfg.file_scan_config_bytes_per_partition)
-        .div_ceil(cfg.target_partitions());
+    let bytes_per_partition = d_cfg.file_scan_config_bytes_per_partition.max(1) as f64;
+    let target_partitions = cfg.target_partitions().max(1) as f64;
+    let task_count = total_bytes as f64 / bytes_per_partition / target_partitions;
```
gabotechs, line 31: 👍 nice!

src/distributed_planner/inject_network_boundaries.rs
shinzoxD: Addressed in 84bacd09dd3237f945ae9bf3d8d17ddb2fa4828c: Desired(0) is now a valid empty-stage outcome. Two zero-weight union leaves plan as Desired(0) / Maximum(0) instead of an internal error.

src/distributed_planner/inject_network_boundaries.rs
shinzoxD: Addressed in 84bacd09dd3237f945ae9bf3d8d17ddb2fa4828c: rolled back `annotate_test_plan_result`. The helper is success-only again; the zero-task case now snapshots the valid Desired(0) plan instead of asserting an error.

src/coordinator/prepare_dynamic_plan.rs
shinzoxD: Addressed in 84bacd09dd3237f945ae9bf3d8d17ddb2fa4828c: the dynamic planner now returns Desired(0.0) for the consumer-side network-boundary hint so small fractional values are not masked by 1.0 and can be summed across UNION children.

shinzoxD:
Rebased onto `main` (now at `ee561b5`) to resolve the conflict.

The only conflict was in `src/distributed_planner/inject_network_boundaries.rs`. Resolution keeps main's newer structure from #655 (unified leaf handling, `NetworkCoalesceExec` above leaves) and this PR's fractional task-count behavior:

- Union children are summed as `f64` before rounding
- Absent non-leaf hints default to `Desired(0.0)` so they do not inflate fractional children via max-merge
- `Desired(0)` remains a valid empty-stage outcome
- Dynamic planner still returns `Desired(0.0)` so a 0.0 hint does not mask smaller fractional values or inflate UNION sums

New HEAD: `fe459198204b1bed843f36e882651af077ed8b3c`

shinzoxD:
Remaining review items from that thread are in: union zero-task leaves, fractional file-scan, two-decimal Debug, Desired(0) valid, and dynamic planner 0.0. PR is mergeable on current main.

shinzoxD:
Rebased onto latest `main` (includes DF 55 / 4.0.0 and #661). Remaining review items are still in; not merging.

Fractional behavior is unchanged: union children sum as `f64` before rounding, absent non-leaf hints are `Desired(0.0)`, file-scan uses true `f64` division, `Desired(0)` remains a valid empty-stage outcome. The `Desired(f64)` migration note moved to `docs/upgrade/5.0.0.md` now that 4.0.0 is released.

shinzoxD:
Rebase onto main pulled in `build_side_fetch_is_preserved_by_broadcast` (#618). Fractional file-scan hints were giving that broadcast producer 4 tasks, so LIMIT 50 applied per task (3750 vs 2500). The producer stage is now capped at 1 when the subtree has a fetch/limit.

shinzoxD:
Rebased onto `main` after #670. Dropped the 1-task producer cap — that was papering over the missing topology. #670 is the right fix (fetch-bearing coalesce stays below `BroadcastExec`).

PR #618, opened 2026-08-09
fix: preserve broadcast build-side fetch

### LINKED ISSUE #591: BroadcastExec insertion ignores CoalescePartitionsExec fetch value
`insert_broadcast_execs` reuses the *input* of a build-side `CoalescePartitionsExec` and
rebuilds a fresh one above the `BroadcastExec`:

https://github.com/datafusion-contrib/datafusion-distributed/blob/9dd018a392185c6853f4d32a43d839219b31730d/src/distributed_planner/insert_broadcast.rs#L136

```rust
let broadcast_input = build_child
    .map_or_else(
        || Arc::clone(build_child),
        |coalesce| Arc::clone(coalesce.input()),
    );

let broadcast: Arc<dyn ExecutionPlan> = Arc::new(BroadcastExec::new(broadcast_input, 1));
let new_build_child: Arc<dyn ExecutionPlan> =
    Arc::new(CoalescePartitionsExec::new(broadcast));
```

`CoalescePartitionsExec` carries a `fetch`, which is where `LimitPushdown` parks a `LIMIT`.
Neither `coalesce.input()` nor `CoalescePartitionsExec::new()` keeps it, so the limit is
dropped and the join runs against the whole build side. Silent wrong results, no error.

Repro: `build_side` = ids 0..100 over 4 files, `probe_side` = same ids x50, 4 workers,
`target_partitions = 3`, broadcast joins on.

```sql
SELECT b.id, p.id FROM (SELECT id FROM build_side LIMIT 50) b
JOIN probe_side p ON b.id = p.id
```

Single node has `CoalescePartitionsExec: fetch=50` under the join and returns 2500 rows.
Distributed has a plain `CoalescePartitionsExec` above the `BroadcastExec` and returns 3750.

Pre-existing on `main` (`9d51c04f`). `normalize_collect_joins::collect_left_to_partitioned`
and the `NestedLoopJoinExec` arm next to it (added in #583) do the same strip, so a fix should
cover all three.

Keeping the coalesce below the `BroadcastExec` isn't enough — the fetch would then apply once
per producer task instead of once globally.

shinzoxD:
## Summary

- preserve the `fetch` value when replacing a build-side `CoalescePartitionsExec` during broadcast insertion
- add an integration regression that compares distributed and single-node results for a limited build side

## Testing

- `cargo test --features integration --test multi_task_collect_join_repros` (8 passed, 2 ignored)
- `cargo test distributed_planner::insert_broadcast::tests --lib` (7 passed)
- `cargo test distributed_planner::normalize_collect_joins::tests --lib` (7 passed)
- `cargo test --features integration --lib -- --skip protocol::grpc::channel_resolver::tests::fails_establishing_connection` (288 passed, 1 ignored, 1 filtered)
- `cargo check --lib`
- `cargo clippy --lib -- -D warnings`
- `git diff --check`

Closes #591

Assisted-by: OpenAI Codex

### DIFF
--- a/src/distributed_planner/insert_broadcast.rs
+++ b/src/distributed_planner/insert_broadcast.rs
@@ -133,17 +133,17 @@ pub(super) fn insert_broadcast_execs(
             return Ok(Transformed::no(node));
         };
 
-        let broadcast_input = build_child
+        let (broadcast_input, coalesce_fetch) = build_child
             .downcast_ref::<CoalescePartitionsExec>()
             .map_or_else(
-                || Arc::clone(build_child),
-                |coalesce| Arc::clone(coalesce.input()),
+                || (Arc::clone(build_child), None),
+                |coalesce| (Arc::clone(coalesce.input()), coalesce.fetch()),
             );
 
         // consumer_task_count=1 is a placeholder and will be corrected during optimizer rule.
         let broadcast: Arc<dyn ExecutionPlan> = Arc::new(BroadcastExec::new(broadcast_input, 1));
         let new_build_child: Arc<dyn ExecutionPlan> =
-            Arc::new(CoalescePartitionsExec::new(broadcast));
+            Arc::new(CoalescePartitionsExec::new(broadcast).with_fetch(coalesce_fetch));
 
         let mut new_children: Vec<Arc<dyn ExecutionPlan>> = children.into_iter().cloned().collect();
         new_children[0] = new_build_child;
--- a/tests/multi_task_collect_join_repros.rs
+++ b/tests/multi_task_collect_join_repros.rs
@@ -343,6 +343,20 @@ mod tests {
         .unwrap();
     }
 
+    /// A build-side `LIMIT` is carried by the `CoalescePartitionsExec` that a
+    /// broadcast rewrite replaces. The replacement must preserve that fetch or
+    /// the join observes every build row instead of the requested 50.
+    #[tokio::test]
+    async fn build_side_fetch_is_preserved_by_broadcast() {
+        assert_distributed_matches_single_node(
+            "SELECT count(*) FROM (SELECT id FROM build_side LIMIT 50) b \
+             JOIN probe_side p ON b.id = p.id",
+            true,
+        )
+        .await
+        .unwrap();
+    }
+
     fn data_dir() -> PathBuf {
         Path::new(env!("CARGO_MANIFEST_DIR")).join("target/multi_task_collect_join_repros")
     }


shinzoxD:
CI is awaiting maintainer approval for this external-contributor workflow. The regression is reproduced on current `main`, and the focused integration target plus the 288-test library suite pass locally.

gabotechs (approved):
Looks good, thanks @shinzoxD!

gabotechs:
@shinzoxD it seems like the test might be a bit flaky?

https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/31358856004/job/93363560852

stuhood:
@gabotechs: It looks like this change might violate the conditions that @barbarj was planning to validate:
> `src/distributed_planner/normalize_collect_joins.rs` establishes that a fetch-less CoalescePartitionsExec (a single-node partitioning artifact) should be stripped, but a fetch-bearing CoalescePartitionsExec must be retained on the input below the boundary so the limit is evaluated in the producer stage before distribution.

gabotechs:
@stuhood 🤔 but is that a problem? without fetch propagation there the query is incorrect.

I see this code handles it gracefully:

https://github.com/datafusion-contrib/datafusion-distributed/blob/e2b60948c04a16dc148e2915d7f1dd95684b6457/src/distributed_planner/normalize_collect_joins.rs#L175-L186

And tests are green.

Although I wonder if it would be better to propagate the `fetch` to the left RepartitionExec instead.

barbarj:
> @shinzoxD it seems like the test might be a bit flaky?
> 
> https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/31358856004/job/93363560852

This looks like the same problem the other fetch-preserving test had. Which row survives is not deterministic, so you need to just assert row counts instead.

barbarj:
> Although I wonder if it would be better to propagate the `fetch` to the left RepartitionExec instead.

@gabotechs `RepartitionExec` doesn't support a fetch though, or am I missing something?

gabotechs:
@barbarj yeah, indeed it doesn't.

shinzoxD:
Thanks — I don't think this failure is only nondeterministic row selection. This test already selects `count(*)`, so each result set contains one aggregate row; CI failed because the aggregate values were `3750` and `2500`.

The fixture has 100 unique build IDs and exactly 50 probe rows for every ID. Any valid `LIMIT 50` therefore joins to exactly 2,500 rows regardless of *which* 50 IDs survive. A value of 3,750 means one execution effectively joined 75 build rows. Comparing only the number of aggregate rows would be vacuous (`1 == 1`) and hide that mismatch.

This suggests the fetch moved above `BroadcastExec` is being applied after fan-out / per consumer rather than preserving the original global limit. I'll prepare a focused follow-up that keeps a cardinality assertion and fixes the placement unless you prefer this tracked as an issue first.

barbarj:
Ah gotcha. @stuhood 's comment is right then. That does sound like it's violating the invariant I made an attempt at validating in #582

shinzoxD:
Thanks for confirming. The focused follow-up is #624. It retains a fetch-bearing `CoalescePartitionsExec` as the input to `BroadcastExec`, while still stripping fetch-less coalesces. The regression passed locally once with the corrected plan and then in five repeated runs; the full upstream CI matrix is now running.

PR #623, opened 2026-08-11
coordinator: display consumer dynamic filters after execution

### LINKED ISSUE #529: [dynamic filtering] 2. collect and display dynamic filters in plans
Dynamic filters must be collected from workers and propagated up to the coordinator for displaying. This is a good way to help test that they are working. It's also a good way to test that collection works.

We need some sort of primitive to collect dynamic filters from plans (for displaying but also for when we need to propagate changes across the network). 

See https://github.com/apache/datafusion/issues/21207#issuecomment-4857069773. `apply_expressions` was useful for this but it got reverted. We can try re-adding something like that OR have some sort of solution involving downcasting in datafusion-distributed. The downcasting option is less ideal because it hurts extensibility. If we downcast, we will probably need a way for users to register their custom `ExecutionPlan` nodes into the downcast chain.

jayshrivastava:
## Stack

1. https://github.com/datafusion-contrib/datafusion-distributed/pull/623 <- you are here

Closes: https://github.com/datafusion-contrib/datafusion-distributed/issues/529

## Problem

Post df-55 upgrade, dynamic filters should work in the worker-local case. There's no way to observe them working other than looking at metrics. 
```
  ┌───── Stage 2 ── tasks=1
  │ AggregateExec: Final COUNT(*)
  │   [Stage 1] => NetworkCoalesceExec
    ┌───── Stage 1 ── tasks=2
    │ HashJoinExec: orders.customer_id = selected_customers.customer_id
    │   DistributedLeafExec:
    |     ...
    │   DistributedLeafExec:
    │     t0: DataSourceExec: predicate=DynamicFilter [ empty ]
    │     t1: DataSourceExec: predicate=DynamicFilter [ empty ]
    └────────────────────────────────────────────────
```

Ideally we want the final filters visible when displaying plans.

## Solution

This PR adds a new protocol which is basically identical to the metrics protocol. Even the `MetricsStore` is now just `Store` and is generic over `TaskMetrics` and `TaskCompletedDynamicFilters` (contains completed dynamic filters for a task).
```rust
pub(crate) type MetricsStore = Store<TaskMetrics>;
pub(crate) type CompletedDynamicFilterStore = Store<TaskCompletedDynamicFilters>;
```

Similar to the metrics protocol, workers now collect completed dynamic filters and send them back to the coordinator.

```
Coordinator                                               Worker
-----------                                               ------
       Create independent display copies
                    |
                    +-- SetPlan(task 0, filter IDs) -------> Decode plan
                    |                                       |
                    |                                       | execute
                    |                                       |
                    |                                       |
                    |                                       |
                    |                                       |
                    |                                       | task finishes
                    |                                       v
                    |<----- TaskDynamicFilters ----- Serialize completed filters from the consumers
                    |
                    v
```

Then, at display time, we call `apply_reports_to_distributed_leaves` which traverses the `plan_for_viz` and updates the dynamic filters for all the variants:
```
DistributedLeafExec
  task 0: DynamicFilter [ key@0 >= 1 AND key@0 <= 10 ]
  task 1: DynamicFilter [ empty ]
```

### Duplicate RPC Messages

We will eventually have more dynamic filter RPCs which manage the worker -> coordinator -> merge -> worker flow mentioned in https://github.com/datafusion-contrib/datafusion-distributed/pull/553.

In theory, the coordinator will know at `merge` time what the completed filters are, making the `TaskCompletedDynamicFilters` and final worker -> coordinator message in this PR irrelevant. 

However, I think having these mechanisms be separate is good because a) it helps us validate that the dynamic filter coordinator -> worker flow work using external "oracle", and b) there's no guarantee that the coordinator -> worker propagation happens before the query is done (ex. the `DataSourceExec` may not block execution waiting for dynamic filters), so it's good to have a separate way to know if the final `DataSourceExec` applied a filter or not.

### `AND true` and empty filters

```
DynamicFilter [ sr_returned_date_sk@0 >= 2451545 AND sr_returned_date_sk@0 <= 2451910 AND true ] AND DynamicFilter [ empty ]
```
In this filter `AND true` occurs because of https://github.com/apache/datafusion/issues/24277. The first `DynamicFilter` is active but we lose the `HashTableLookupExpr` when serializing it to send back to the coordinator.

The 2nd filter is `DynamicFilter [ empty ]` because this is a dynamic filter produced by a remote producer, which does not get propagated to this node yet. This will be fixed later.

### Displaying Dynamic Filters

Protocol is as similar to the metrics protocol as possible. Due to double wrapping (`MetricsWrapperExec` wraps `DistributedLeafExec`, it's tricky to do the dynamic filter rewrite after doing the metrics rewrite. So `rewrite_distributed_plan_with_dynamic_filters` has to be called **first**.

```rust
let plan = rewrite_distributed_plan_with_dynamic_filters(plan).await?;
let plan = rewrite_distributed_plan_with_metrics(plan, DistributedMetricsFormat::Aggregated).await?;
println!("{}", display_plan_ascii(plan.as_ref(), true));
```

## Testing
- Tests in `tests/dynamic_filtering.rs`


stuhood:
Thanks for working on this: this will be very useful!

One quick thought: the dynamic filter will sometimes be much, much larger than what you would actually want to display in an EXPLAIN plan (a large `InList` or `Hash`). That suggests that rather than sending the whole filter, what would actually make sense to send back is some sort of human readable summary of the filter?

Also, I have a draft of a related change on our codebase, and it seemed like the easiest mechanism for transferring this kind of information back is via metrics... but the most natural/obvious thing that seemed to be missing in that case was essentially a "string" metric type (we would use it to display a chosen strategy/enum from a scan). Do you think that that might be worth pursuing upstream?

jayshrivastava:
> One quick thought: the dynamic filter will sometimes be much, much larger than what you would actually want to display in an EXPLAIN plan (a large InList or Hash). That suggests that rather than sending the whole filter, what would actually make sense to send back is some sort of human readable summary of the filter?

> Also, I have a draft of a related change on our codebase, and it seemed like the easiest mechanism for transferring this kind of information back is via metrics... but the most natural/obvious thing that seemed to be missing in that case was essentially a "string" metric type (we would use it to display a chosen strategy/enum from a scan). Do you think that that might be worth pursuing upstream?

Serializing them as a string is reasonable. Rather than a metric, I think we can implement a `PhysicalExpr` which just wraps a string and inject it into the play for display using `DynamicFilterPhysicalExpr::update(string_expr)`. @gabotechs what do you think?

gabotechs:
🤔 I'm not sure if I'm understanding the suggestion. Updating a filter with `DynamicFilterPhysicalExpr::update` is not really related to visualization, it's how you actually update the filter no?

jayshrivastava:
> 🤔 I'm not sure if I'm understanding the suggestion. Updating a filter with DynamicFilterPhysicalExpr::update is not really related to visualization, it's how you actually update the filter no?

The suggestion is basically to consider returning a string representation from the worker to the coordinator for displaying rather than a serialized `PhysicalExpr`. We will need to do some hacking to make that work.

I think we can consider this later. The serialization in this PR is simpler and uses the native `PhysicalExpr` serialization.

docs/source/user-guide/05-metrics.md
```diff
@@ -26,8 +26,11 @@ channel, so they are not lost even if the result stream is dropped early (for ex
...
 
-Two functions, both exported from the crate root, do the work:
+These functions, all exported from the crate root, do the work:
 
+- `rewrite_distributed_plan_with_dynamic_filters(plan)` — folds the completed dynamic filters
+  reported by each worker task into an isolated copy of the plan. When displaying both dynamic
+  filters and metrics, apply the dynamic-filter rewrite first.
 - `rewrite_distributed_plan_with_metrics(plan, format)` — folds every task's metrics back into the
   coordinator's copy of the plan. It waits for all worker metrics to arrive, so the result is always
   complete. The `format` is a `DistributedMetricsFormat`:
```
gabotechs: 🤔 From a public API design standpoint this is not great (It was not great before already).

This is the first thing I encountered in the PR, I'll keep reading, but so far this does seem already like something we should address here.

src/protocol/worker_channel.rs
```diff
@@ -5,6 +5,7 @@ use datafusion::common::Result;
 use datafusion::execution::TaskContext;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricsSet};
+use datafusion_proto::protobuf::PhysicalExprNode;
```
gabotechs: The worker protocol should not be coupled to protobuf, this should depend on either vanilla `datafusion` types or just raw `Vec<u8>`.

src/protocol/worker_channel.rs
```diff
@@ -120,12 +121,29 @@ pub enum WorkerToCoordinatorMsg {
...
+    /// Final expressions keyed by their DataFusion physical-expression ID. The TaskKey is
+    /// implicit from the coordinator channel that carried this message.
+    pub filters: Vec<TaskDynamicFilter>,
+}
+
+#[derive(Clone, Debug)]
+pub struct TaskDynamicFilter {
+    pub expression_id: u64,
+    /// A `DynamicFilterPhysicalExpr` proto containing its final predicate and completion state.
+    pub expression: PhysicalExprNode,
```
gabotechs: Same as the other `datafusion` related dependencies in this file, this should be an `MaybeEncoded<Arc<dyn PhysicalExpr>>` instead, and not be coupled to protobuf specifically.

Implementations of the `WorkerChannel` trait might not want to build a protobuf message here at all, or they might want to just pass an in-memory `Arc<dyn PhysicalExpr>`.

I think this change should be trivial.

src/worker/impl_coordinator_channel.rs
```diff
@@ -167,7 +173,10 @@ impl Worker {
                 if let Some(metrics_tx) = metrics_tx {
                     send_metrics_via_channel(metrics_tx, plan, d_ctx, task_data_metrics);
                 }
+                dynamic_filters = build_task_completed_dynamic_filters(plan, &task_data.task_ctx)
+                    .unwrap_or_default();
             }
```
gabotechs: Aren't we swallowing the error here? if you think that's fine, we can just leave a comment explaining why is it fine.

src/worker/impl_coordinator_channel.rs
```diff
@@ -190,10 +199,46 @@ impl Worker {
...
+                    dynamic_filters,
+                ))
+            },
+        );
+
+        Ok(futures::stream::select(
+            load_info_stream,
+            futures::stream::select(metrics_stream, dynamic_filters_stream),
+        )
+        .map(Ok)
```
gabotechs: There's the `futures::stream::select_all` function that would work nicely here.

src/worker/impl_coordinator_channel.rs
```diff
@@ -190,10 +199,46 @@ impl Worker {
...
+        .boxed())
+    }
+}
+
+fn build_task_completed_dynamic_filters(
+    plan: &Arc<dyn ExecutionPlan>,
+    task_ctx: &Arc<datafusion::execution::TaskContext>,
```
gabotechs: Prefer normal `use` statements rather than fully qualified imports, like in the rest of this file.

src/dynamic_filtering/discovery.rs
```diff
@@ -0,0 +1,235 @@
...
+/// against.
+#[derive(Clone)]
+pub(crate) struct DiscoveredDynamicFilter {
+    pub(crate) id: u64,
+    pub(crate) expression: Arc<dyn PhysicalExpr>,
+    pub(crate) input_schema: SchemaRef,
+}
+
+/// Finds dynamic-filter consumers in `plan`, deduplicated by expression ID.
+pub(crate) fn discover_dynamic_filter_consumers(
```
gabotechs, line 19: So the way of discovering dynamic filter consumers is:
1. Get the dynamic filter producers with `.dynamic_expressions_produced()`
2. Get all the dynamic filters
3. Exclude from the collected dynamic filters those who have an idea seen in the producer dynamic filters from  `.dynamic_expressions_produced()`

Seems a bit... convoluted, but this seems to be the only way, so all good :+1:

src/coordinator/query_coordinator.rs
```diff
@@ -116,6 +121,8 @@ impl QueryCoordinator {
 /// - Building tasks that communicate a serialized plan to multiple workers for further execution.
 /// - Building tasks that stream partition feeds from local [WorkUnitFeedExec] nodes to their
 ///   remote counterparts.
+type SpecializedTaskPlan = (Arc<dyn ExecutionPlan>, Vec<WorkUnitFeedDeclaration>);
+
 pub(super) struct StageCoordinator<'a> {
```
gabotechs: I think we can live without this small `SpecializedTaskPlan` refactor. It does not seem to be contributing much, and it's even placing the doc comment above in the wrong entity.

src/coordinator/store.rs
```diff
@@ -0,0 +1,77 @@
...
+
+type StoreMap<T> = HashMap<TaskKey, T>;
+
+/// Stores task-scoped values and notifies waiters when entries change.
+#[derive(Debug, Clone)]
+pub(crate) struct Store<T> {
+    tx: watch::Sender<StoreMap<T>>,
+    rx: watch::Receiver<StoreMap<T>>,
+}
+
```
gabotechs, line 17: Nice! pretty elegant

src/coordinator/store.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub(crate) fn from_entries(entries: impl IntoIterator<Item = (TaskKey, T)>) -> Self {
+        let map: HashMap<_, _> = entries.into_iter().collect();
+        let (tx, rx) = watch::channel(map);
+        Self { tx, rx }
+    }
+}
+
+pub(crate) type MetricsStore = Store<TaskMetrics>;
+pub(crate) type CompletedDynamicFilterStore = Store<TaskCompletedDynamicFilters>;
+
```
gabotechs: To be honest, I think I could even live without these type aliases.

Seeing in the other parts of the code explicitly that both `Store<TaskMetrics>` and `Store<TaskCompletedDynamicFilters>` are just specializations for the same `Store` struct is actually good information for readers.

src/protocol/grpc/worker.proto
```diff
@@ -39,9 +39,33 @@ message WorkerToCoordinatorMsg {
     LoadInfo load_info = 2;
 
     bool load_info_eos = 3;
+
+    // Final dynamic filters used by dynamic-filter consumer execution-plan nodes.
+    //
+    // Filters are deduplicated by expression_id because consumers with the same ID share
+    // logical filter state within a task. For example, this plan includes one entry:
```
gabotechs, line 46: Couldn't there be situations where the dynamic filters reported by the different tasks are actually different? even having the same `expression_id`?

src/coordinator/distributed.rs
```diff
@@ -33,18 +33,26 @@ pub struct DistributedExec {
...
     ///   non-distributed plan.
     base_plan: Arc<dyn ExecutionPlan>,
-    /// Resulting [ExecutionPlan] after execution ready for visualization purposes.
-    /// - If the plan was distributed statically, this is equal to the base plan.
-    /// - If the plan is going to be distributed dynamically during execution, this is the resulting
-    ///   plan re-calculated based on runtime statistics.
-    plan_for_viz: Arc<Mutex<Option<Arc<dyn ExecutionPlan>>>>,
-    /// The head stage meant to be executed locally on [DistributedExec::execute].
-    head_stage: Arc<Mutex<Option<Arc<dyn ExecutionPlan>>>>,
+    prepared_execution: Arc<Mutex<Option<PreparedExecution>>>,
```
gabotechs: This could use a refactor into an explicit state machine (not for this PR). That's essentially what it is: state that changes at runtime during the different operations of the query.

For this PR is well handled already.

src/coordinator/distributed.rs
```diff
@@ -82,58 +90,67 @@ impl DistributedExec {
...
-            .map_err(|e| internal_datafusion_err!("Failed to lock head stage: {}", e))?
-            .clone()
-            .ok_or_else(|| internal_datafusion_err!("No head stage found. Was execute() called?"))
+        Ok(self.prepared_execution()?.head_stage)
+    }
+
+    pub(crate) fn task_ctx(&self) -> Result<Arc<TaskContext>> {
+        Ok(self.prepared_execution()?.task_ctx)
+    }
+
```
gabotechs: This does not seem right. a `TaskContext` is not something an `ExecutionPlan` implementation should own, it something that should be provided from the outside.

src/coordinator/dynamic_filters.rs
```diff
@@ -0,0 +1,196 @@
...
+            .variants()
+            .iter()
+            .map(|variant| {
+                let proto = PhysicalPlanNode::try_from_physical_plan_with_converter(
+                    Arc::clone(variant),
+                    &codec,
+                    &converter,
+                )?;
+                proto.try_into_physical_plan_with_converter(task_ctx, &codec, &converter)
+            })
```
gabotechs: Why do we need this proto roundtrip here?

src/coordinator/distributed.rs
```diff
@@ -82,58 +90,67 @@ impl DistributedExec {
...
+    /// Builds a non-executable visualization result while preserving the state needed by a
+    /// subsequent rewrite. Dynamic filters must be rewritten before metrics.
+    pub(crate) fn with_rewritten_plan(
+        &self,
+        plan_for_viz: Arc<dyn ExecutionPlan>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let mut prepared_execution = self.prepared_execution()?;
+        prepared_execution.plan_for_viz = Arc::clone(&plan_for_viz);
+        Ok(Arc::new(Self {
+            base_plan: plan_for_viz,
```
gabotechs: This seems to be wrong. The base plan is supposed to be the base plan that came after physical optimization, not the plan meant for visualization.

src/coordinator/distributed.rs
```diff
@@ -33,18 +33,26 @@ pub struct DistributedExec {
...
-    head_stage: Arc<Mutex<Option<Arc<dyn ExecutionPlan>>>>,
+    prepared_execution: Arc<Mutex<Option<PreparedExecution>>>,
     /// DataFusion metrics.
     metrics: ExecutionPlanMetricsSet,
     /// Storage where metrics collected from workers at runtime will place their results as they
     /// finish their respective remote tasks.
     pub(crate) metrics_store: Option<Arc<MetricsStore>>,
+    /// Storage for the completed dynamic filters reported by each worker task.
+    pub(crate) completed_dynamic_filter_store: Arc<CompletedDynamicFilterStore>,
+}
```
gabotechs: In the same way that metrics collection is optional, it'd be also consistent to make the completed dynamic filters optional. If there's people with a reason for not collecting metrics, for that same reason they might also not want to collect dynamic filters.

src/coordinator/dynamic_filters.rs
```diff
@@ -0,0 +1,196 @@
...
+use datafusion_proto::protobuf::PhysicalPlanNode;
+use std::sync::Arc;
+
+/// Rewrites an executed distributed plan with the completed dynamic filters reported by its
+/// worker tasks.
+///
+/// When composing this with [`crate::rewrite_distributed_plan_with_metrics`], dynamic filters must
+/// be rewritten first.
+pub async fn rewrite_distributed_plan_with_dynamic_filters(
+    plan: Arc<dyn ExecutionPlan>,
```
gabotechs: Why do we need to rewrite the plan here?

I imagine just mutating the dynamic filters in-place should be enough, we don't need to create a new plan.

src/coordinator/distributed.rs
gabotechs: I see this is needed because `rewrite_distributed_plan_with_dynamic_filters` requires it, but I don't think it should be needed there.

There are two places where it's used:
- The proto roundtrip in `isolate_distributed_leaf_variants_for_display`, which I don't really understand why we need it.
- The `apply_reports_to_distributed_leaves` function, which needs to decode the physical expression. This one is needed because the `WorkerChannel` protocol declares a `datafusion_proto::PhysicalExprNode` as part of the contract, which should not be the case (I surfaced this in another comment). Solving that in the protocol specification should automatically solve the issue here.

docs/source/user-guide/05-metrics.md
gabotechs: Just read a bit more, I don't have a good suggestion for this PR, so it should be fine, but I think we should improve this API so that this is handled on behalf of users before the 5.0.0 release.

src/coordinator/dynamic_filters.rs
```diff
@@ -0,0 +1,196 @@
...
+    apply_reports_to_distributed_leaves(&plan_for_viz, &reports, &task_ctx);
+    distributed_exec.with_rewritten_plan(plan_for_viz)
+}
+
+/// Replaces the variants in the visualization plan with independent per-task copies.
+pub(super) fn isolate_distributed_leaf_variants_for_display(
+    plan: Arc<dyn ExecutionPlan>,
+    task_ctx: &Arc<TaskContext>,
+) -> Result<Arc<dyn ExecutionPlan>> {
+    let codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
```
gabotechs: :thinking: I don't think I understand why we need this function. If there's a `DistributedLeafExec` present in the plan, it means that the leaf variants where already isolated

gabotechs (commented):
It's looking very good! love the new `Store` and how that generalizes to dyn filtering.

Also, really good choice having this PR shipped first before the other ones.

Left a first round of comments:

docs/source/user-guide/05-metrics.md
jayshrivastava: Filed https://github.com/datafusion-contrib/datafusion-distributed/issues/678. I agree. For now, I thought it would be best to align with metrics so we can migrate both things to something better.

src/coordinator/store.rs
jayshrivastava: Ty!

src/coordinator/query_coordinator.rs
jayshrivastava: Done. I think clippy caused this to fail to some point but doesn't seem to anymore.

src/coordinator/store.rs
jayshrivastava: Removed.

src/protocol/grpc/worker.proto
jayshrivastava: This message / proto is within a singular task.

But adding a note to myself that maybe we should keep the one with the latest generation or something. That's slightly better maybe.

It's always going to be a bit racy though so maybe taking the last generation is not super worthwhile. In theory this could happen:
1. A task finishes its work
2. A dynamic filter update arrives
3. We read the dynamic filter and send it to the coordinator.

I think this race exists somewhat in vanilla datafusion as well.
1. Leaf finishes execution
2. Producer updates the filter
3. The leaf and producer display the updated filter.

src/dynamic_filtering/discovery.rs
jayshrivastava: This is pretty much the way we settled on upstream: https://github.com/apache/datafusion/issues/23814. I agree it's a bit convoluted.

docs/source/user-guide/05-metrics.md
gabotechs: Makes sense, thanks!

src/dynamic_filtering/discovery.rs
gabotechs: It's fine, not super ergonomic but not terrible

src/worker/impl_coordinator_channel.rs
jayshrivastava: Filed https://github.com/datafusion-contrib/datafusion-distributed/issues/686. Leaving a TODO.

src/coordinator/distributed.rs
jayshrivastava: Done. Pretty much copied the exact same code.

src/worker/impl_coordinator_channel.rs
jayshrivastava: Done.

src/coordinator/dynamic_filters.rs
jayshrivastava: I've moved this to `src/dynamic_filtering/display.rs` and added a lot of explanation. I'm going to  resolve the related comments you left and leave this one open for discussion.

TLDR - there's shared state everywhere. All the variants share the same filter in memory, so updating one will update all variants. So, we need to deepcopy each variant. Each variant is basically just a `DataSourceExec`, so not too expensive hopefully.


Let me know if it's clear.

src/protocol/worker_channel.rs
jayshrivastava: Migrated to `MaybeEncoded`

src/coordinator/distributed.rs
jayshrivastava: We basically need the proto roundtrip to deep clone the variants, so the `TaskContext` is needed.

jayshrivastava:
@gabotechs This is ready for another round. I've left some of the conversations open.

I've refactored things quite a bit, mainly moving source code to `src/dynamic_filtering` and moving tests to `tests/dynamic_filtering`. I've also added more tests for different topologies.

The main thing that's still up for discussion is dynamic filter serialization. We **have** to do this to basically deep copy them to break in-memory relations. I've left a lot of comments in `/src/dynamic_filtering/display.rs` to hopefuly make it more clear.

src/coordinator/query_coordinator.rs
```diff
@@ -258,8 +265,23 @@ impl<'a> StageCoordinator<'a> {
...
+            // completition. Since not all tasks report dynamic filters, add placeholders here.
+            //
+            // Also, note that completed_dynamic_filter_store will be None if collection is
+            // disabled.
+            if let Some(store) = &completed_dynamic_filter_store
+                && store.get(&task_key).is_none()
+            {
+                store.insert(task_key, TaskCompletedDynamicFilters::default());
+            }
         });
```
gabotechs: It feels a bit strange that we need to do this special case here. I'll keep reading.

src/coordinator/query_coordinator.rs
```diff
@@ -387,7 +409,11 @@ impl<'a> StageCoordinator<'a> {
 
             Ok(Transformed::no(plan))
         })?;
-        Ok((transformed.data, work_unit_feed_declarations))
+        let plan = maybe_roundtrip_plan_to_sever_in_memory_dynamic_filter_relationships(
```
gabotechs, line 401: :joy: I really like this long ugly name.

src/dynamic_filtering/discovery.rs
```diff
@@ -0,0 +1,281 @@
...
+use datafusion::physical_expr::expressions::DynamicFilterPhysicalExpr;
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::Arc;
+
+/// A dynamic-filter consumer discovered in an execution plan along with the schema it is evaluated
+/// against.
+#[derive(Clone)]
+pub(crate) struct DiscoveredDynamicFilter {
+    pub(crate) id: u64,
+    pub(crate) expression: Arc<dyn PhysicalExpr>,
```
gabotechs: Wouldn't it be convenient to store here the actual downcasted version?

```rust
    pub(crate) expression: Arc<DynamicFilterPhysicalExpr>,
```

If you downcast it like this I think you should be able to retrieve it:

```rust
                let Ok(dyn_expr) = Arc::downcast::<DynamicFilterPhysicalExpr>(Arc::clone(expression) as _) else {
                    return Ok(TreeNodeRecursion::Continue);
                };
```

src/protocol/grpc/worker_service.rs
```diff
@@ -107,6 +107,14 @@ impl pb::worker_service_server::WorkerService for Worker {
...
 
         let set_plan_request = decode_set_plan_request(set_plan_request)?;
+        let task_key = set_plan_request.task_key;
+        // Dynamic-filter reports may carry decoded physical expressions. Retain the worker's
+        // task data so the gRPC boundary can encode them with its configured codecs, even after
+        // the completed task has been removed from the worker cache.
+        let task_data_entry = self
+            .task_data_entries
+            .get_with(task_key, async { Default::default() })
+            .await;
```
gabotechs, line 118: Even if this is allowed, it feels like the `grpc` layer is access an internal detail of the `Worker` by accessing `task_data_entries`.

If we were to move the `grpc` module to a specific crate, this will start giving problems. Just something to keep in mind, I don't think we need to solve this now, what you have here is good enough.

src/worker/impl_coordinator_channel.rs
```diff
@@ -167,6 +184,15 @@ impl Worker {
                 if let Some(metrics_tx) = metrics_tx {
                     send_metrics_via_channel(metrics_tx, plan, d_ctx, task_data_metrics);
                 }
+                if dynamic_filters_tx.is_some() {
+                    // TODO(#686): handle error
+                    dynamic_filters =
+                        Some(build_task_completed_dynamic_filters(plan).unwrap_or_default());
+                }
+            }
```
gabotechs: Why cannot just send it here like this?

```rust
                if let Some(dynamic_filters_tx) = dynamic_filters_tx {
                    let dyn_filter_result = build_task_completed_dynamic_filters(plan);
                    let _ = dynamic_filters_tx.send(dyn_filter_result.unwrap_or_default());
                }
```

src/coordinator/distributed.rs
```diff
@@ -33,18 +33,26 @@ pub struct DistributedExec {
...
+/// for post-execution work such as plan rewrites to display metrics and dynamic filters.
+#[derive(Debug, Clone)]
+struct PreparedExecution {
+    /// Resulting plan reconstructed after static or dynamic planning.
+    plan_for_viz: Arc<dyn ExecutionPlan>,
+    /// The head stage actually executed locally by the coordinator.
+    head_stage: Arc<dyn ExecutionPlan>,
+    /// The task context of the [`DistributedExec`]. Useful for decoding protobufs.
+    task_ctx: Arc<TaskContext>,
 }
```
gabotechs: I do think this still has chanced of not aging well. Even if more annoying, I'd be in favor of removing it from here and just explicitly requiring it in the `rewrite_distributed_plan_with_dynamic_filters` function:

```rust
pub async fn rewrite_distributed_plan_with_dynamic_filters(
    plan: Arc<dyn ExecutionPlan>,
    ctx: &Arc<TaskContext>,
) -> Result<Arc<dyn ExecutionPlan>>;
```

I think we are going to be able to retire both `rewrite_distributed_plan_with_dynamic_filters` and `rewrite_distributed_plan_with_metrics` with a more native and integrated API anyways (probably reusing the `TaskContext` we already store in `QueryCoordinator` or something).

src/dynamic_filtering/display.rs
```diff
@@ -0,0 +1,357 @@
...
+///        DistributedLeafExec:
+///          t0: DataSourceExec:  predicate=DynamicFilter [ f_dkey@2 >= A AND f_dkey@2 <= A AND f_dkey@2 IN (SET) ([<values>]) ] <- unique filter
+///          t1: DataSourceExec:  predicate=DynamicFilter [ f_dkey@2 >= B AND f_dkey@2 <= B AND f_dkey@2 IN (SET) ([<values>]) ] <- unique filter
+/// ```
+///
+/// This is done by deep-copying every leaf variant so we don't have to
+/// worry about any shared state.
+///
+/// [`update()`]: DynamicFilterPhysicalExpr::update()
+pub(crate) fn sever_dynamic_filter_relationships_in_plan_for_display(
```
gabotechs, line 73: You did a pretty good job design these functions so that they are isolated from the main code paths. I still see all these as "hacks" we need to do for workarounding the fact that shared stated is being baked into the nodes and expressions.

I think this is a broader topic worth rising upstream (if not already), and all the workarounds we need to do in this project are a strong signal that shared state, in general, should be scoped to a `TaskContext`, not to individual `ExecutionPlan`s or `PhysicalExpr`s.

Ballista and other distribution layers are going to have the same issues, so I don't think this is a problem only this project will suffer.

src/dynamic_filtering/display.rs
```diff
@@ -0,0 +1,357 @@
...
+            .map(|variant| {
+                // Proto roundtrip is used to deep copy the variant.
+                let proto = PhysicalPlanNode::try_from_physical_plan_with_converter(
+                    Arc::clone(variant),
+                    &codec,
+                    &converter,
+                )?;
+                let variant =
+                    proto.try_into_physical_plan_with_converter(task_ctx, &codec, &converter)?;
+                // The variant can have a SortExec
```
gabotechs: We have a `roundtrip_pb` function here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/26297c85ea6b3491d4d841db371281813c5f5756/src/coordinator/query_coordinator.rs#L446-L446

That btw I'm just realizing it's serializing to bytes and back, but I think it should not be necessary.

For all proto roundtrips, what do you think about having a `roundtrip_pb` here:
https://github.com/datafusion-contrib/datafusion-distributed/blob/730266f6a6ee7790d70f01fa4fa267b957e8c580/src/codec/physical_plan.rs

And reusing it everywhere we need to roundtrip a plan?

gabotechs (commented):
Still an incomplete review, but flushing today's comments, I'll resume review tomorrow. I'm mainly trying to invest some time in coming up with suggestions for improving the more sketchy parts, but for most of the things I'm not finding better alternatives to what you already have here.

Given the cards we are dealt with, this is looking like pretty solid work!

src/dynamic_filtering/display.rs
```diff
@@ -0,0 +1,357 @@
...
+        let isolated = proto.try_into_physical_plan_with_converter(task_ctx, &codec, &converter)?;
+        let isolated = isolated.replace_children(vec![input], recompute)?;
+
+        Ok(Transformed::yes(isolated))
+    })
+    .map(|transformed| transformed.data)
+}
+
+/// Applies successful worker reports only to the matching task-local visualization variants.
+pub(super) fn apply_reports_to_distributed_leaves(
```
gabotechs: This function can be made private:

```suggestion
fn apply_reports_to_distributed_leaves(
```

src/dynamic_filtering/display.rs
```diff
@@ -0,0 +1,357 @@
...
+        let leaf = isolated.downcast_ref::<DistributedLeafExec>().unwrap();
+        let task_0 = displayable(leaf.variants()[0].as_ref())
+            .one_line()
+            .to_string();
+        let task_1 = displayable(leaf.variants()[1].as_ref())
+            .one_line()
+            .to_string();
+        assert!(task_0.contains("DynamicFilter [ a@0 > 10 ]"));
+        assert!(task_1.contains("DynamicFilter [ empty ]"));
+        Ok(())
```
gabotechs: If this assertion starts failing at some point, it's going to be hard to debug what's going wrong, as the test will just report something like:

```
expected `task_0.contains("DynamicFilter [ a@0 > 10 ]")` to be `true` but it was `false`
```

For these, an `insta` based assertion can help a bit more.

src/dynamic_filtering/display.rs
```diff
@@ -0,0 +1,357 @@
...
+                .to_string()
+                .contains("filter=[a@0 > 20]")
+        );
+        assert!(
+            displayable(isolated.as_ref())
+                .one_line()
+                .to_string()
+                .contains("filter=[a@0 > 10]")
+        );
+        Ok(())
```
gabotechs: Same as above.

src/coordinator/distributed.rs
```diff
@@ -82,58 +99,72 @@ impl DistributedExec {
...
+
+    pub(crate) async fn wait_for_dynamic_filters(
+        &self,
+    ) -> Result<Option<HashMap<TaskKey, TaskCompletedDynamicFilters>>> {
+        let Some(store) = &self.completed_dynamic_filter_store else {
+            return Ok(None);
+        };
+        let plan = self.plan_for_viz()?;
+        Ok(Some(store.wait_for(&task_keys_for_plan(&plan)).await))
     }
```
gabotechs, line 107: Nit: For consistency, it's probably worth it to make the signature and behavior of this function match the one above (`wait_for_metrics`). Unless you think there's a reason to diverge.

I think we need to clean up all this API for collecting metrics, dyn filters, etc... so this might go away at that point anyways.

src/coordinator/distributed.rs
```diff
@@ -153,7 +184,12 @@ impl ExecutionPlan for DistributedExec {
...
 
     fn children(&self) -> Vec<&Arc<dyn ExecutionPlan>> {
-        vec![&self.base_plan]
+        vec![
+            self.prepared_execution
+                .get()
+                .map(|prepared| &prepared.plan_for_viz)
+                .unwrap_or(&self.base_plan),
+        ]
     }
```
gabotechs: :thinking: this is a behavior change that can blow up on other people's faces...

It will not blow up at Datadog, because this fits how we use DistributedExec, but I wonder if other people are actually relying on retrieving the original children of the DistributedExec after execution.

If you see opportunities for avoiding this change, I'd try to do it.

src/coordinator/distributed.rs
```diff
@@ -167,12 +203,16 @@ impl ExecutionPlan for DistributedExec {
         self: Arc<Self>,
         children: Vec<Arc<dyn ExecutionPlan>>,
     ) -> Result<Arc<dyn ExecutionPlan>> {
+        let child = require_one_child(&children)?;
+        if self.prepared_execution.get().is_some() {
+            return self.with_plan_for_viz(child);
+        }
         Ok(Arc::new(DistributedExec {
```
gabotechs, line 195: All the conditional logic about different behaviors for when there's a `prepared_execution` and when there's not is getting a bit messy...

I think there are some opportunities for improving this, but probably worth exploring in another PR.

tests/dynamic_filtering/config.rs
```diff
@@ -0,0 +1,23 @@
+#[cfg(test)]
+mod tests {
+    use crate::common::TestQuery;
+    use datafusion::common::Result;
+
+    #[tokio::test]
+    async fn completed_filter_collection_can_be_disabled() -> Result<()> {
+        TestQuery::new(
```
gabotechs, line 8: This test is not really asserting that dynamic filters are not getting collected no?

gabotechs (commented):
This is looking very good! I think this is my last round of comments before approval.

src/dynamic_filtering/display.rs
gabotechs, line 73: Here's one idea that is comes to mind:

What if we add yet another preparatory step for the plan in `distributed_planner/`, something like `insert_broadcast` or `normalize_collect_joins`, that severs all dynamic filter connections for good?

This would imply that dynamic filters will never be able to work through normal upstream mechanisms, and they should always be updated passing through the coordinator, even in the local case, but I do imagine this can simplify the overall approach, specially for future PRs.

src/dynamic_filtering/display.rs
jayshrivastava: 🫡 Good point.

src/coordinator/distributed.rs
jayshrivastava: Done.

tests/dynamic_filtering/config.rs
jayshrivastava: It happens in `execute()`. We assert that the plan isn't rewritten because there's no dynamic filters to collect. I'll add a comment.

src/dynamic_filtering/discovery.rs
jayshrivastava: Seems good

src/worker/impl_coordinator_channel.rs
jayshrivastava: 👍🏽 Makes sense

src/dynamic_filtering/display.rs
jayshrivastava: I would like to leave this for now. 

The tricky part is that we want to sever some relationships but not all of them. For example, if there's local dynamic filters, a producer should be able to atomically update them in memory.

In future PRs, I detect local vs remote dynamic filters during static and dynamic planning, so we can think about severing some relationships then.

src/coordinator/distributed.rs
jayshrivastava: Reverted. I think that's fair. 

There's a weird lifecycle that the metrics rewrite does that I'm just going to preserve for now.

`children()` is always the base plan, but `rewrite_distributed_plan_with_dynamic_filters` and `rewrite_distributed_plan_with_metrics` replaces both the `children` and `plan_for_viz`.
- It replaces the `children()` so users can iterate over the `DistributedExec` / `Arc<dyn ExecutionPlan` and extract metrics
- `plan_for_viz` is used for `display_plan_ascii` in this repo

I opened for 2 separate APIs
- `with_new_children` replaces the children
- `with_plan_for_viz` replaces `plan_for_viz`

src/coordinator/distributed.rs
```diff
@@ -43,28 +50,19 @@ pub struct DistributedExec {
     pub(crate) completed_dynamic_filter_store: Option<Arc<Store<TaskCompletedDynamicFilters>>>,
 }
 
-/// Execution state produced by distributed planning (static or dynamic) retained
-/// for post-execution work such as plan rewrites to display metrics and dynamic filters.
 #[derive(Debug, Clone)]
-struct PreparedExecution {
```
jayshrivastava: Now that `TaskCtx` is gone, we can now use `PreparedPlan` directly

src/coordinator/distributed.rs
```diff
@@ -89,72 +87,66 @@ impl DistributedExec {
...
-            return;
-        };
-        task_metrics.wait_for(&task_keys_for_plan(&plan)).await;
+    /// Waits until all worker tasks have reported their metrics back via the coordinator channel
+    /// if metrics collection is enabled.
+    pub async fn wait_for_metrics(&self) -> Option<HashMap<TaskKey, TaskMetrics>> {
+        let task_metrics = self.metrics_store.as_ref()?;
+        let plan = &self.prepared_plan.get()?.plan_for_viz;
+        Some(task_metrics.wait_for(&task_keys_for_plan(plan)).await)
     }
```
jayshrivastava, line 96: I decided this is the cleanest API: "wait until all task datas are present and then return them". It lets us remove the whole `get()` API on the store and just work with HashMap<...> directly

src/coordinator/distributed.rs
jayshrivastava: Removed the logic. I opted to reset the `prepared_plan` though. I felt that it doesn't make sense to change the children and leave the old `prepared_plan` / `plan_for_viz`.

src/metrics/task_metrics_rewriter.rs
```diff
@@ -89,7 +89,46 @@ pub async fn rewrite_distributed_plan_with_metrics(
...
-    distributed_exec.with_plan_for_viz(transformed.data)
+    let plan = distributed_exec.with_plan_for_viz(Arc::clone(&transformed.data))?;
+    plan.replace_children(
+        vec![transformed.data],
+        ReplaceChildrenOptions::new(ChildrenPropertiesMode::Recompute),
+    )
+}
+
+/// Gathers metrics that belong to a task as a whole rather than to an execution-plan node.
+fn stage_metrics(
```
jayshrivastava, line 100: Basically a copy paste of `gather_stage_header_metrics` removed below

src/coordinator/distributed.rs
jayshrivastava: https://github.com/datafusion-contrib/datafusion-distributed/pull/623#discussion_r3919847054

src/protocol/grpc/worker_service.rs
jayshrivastava: Sure, let's clean this up whenever it becomes a problem.

src/dynamic_filtering/display.rs
jayshrivastava: 👍🏽 Will look out for ways to improve this

jayshrivastava:
@gabotechs Addressed the comments in the last few commits. I left the nontrivial comments open. I can check CI and rebase tomorrow 🫡

tests/multi_task_collect_join_repros.rs
```diff
@@ -295,18 +295,18 @@ mod tests {
           ┌───── Stage 1 ── tasks=4, partitions=8
           │ SortExec: TopK(fetch=5), expr=[id@0 ASC NULLS LAST], preserve_partitioning=[true]
           │   DistributedLeafExec:
-          │     t0: DataSourceExec: file_groups={2 groups: [[/target/multi_task_collect_join_repros/build_side/part-0.parquet:<int>..<int>], [/target/multi_task_collect_join_repros/build_side/part-2.parquet:<int>..<int>]]}, projection=[id], file_type=parquet, predicate=DynamicFilter [ empty ], sort_order_for_reorder=[id@0 ASC NULLS LAST], dynamic_rg_pruning=eligible
```
jayshrivastava, line 298: We lose these fields when doing the proto roundtrip before displaying. So, this means we actually lose these fields during execution since we serialize this plan before executing.

In a way, roundtripping the plan before displaying gives us a better picture of what's actually happening during execution.

I've addressed this here: https://github.com/apache/datafusion/pull/24930, so these optimizations will come back in df56.

src/coordinator/distributed.rs
gabotechs: :+1: yeap, this does look clean indeed, thanks!

src/coordinator/distributed.rs
gabotechs: Very well handled!

gabotechs:
benchmarks run tpch/sf100

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/623#issuecomment-5537796870).

## Benchmark results

**Compared:** [PR base `417f80977d7f`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) → [PR head `7f006309f5b8`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) · [View exact source diff](https://github.com/datafusion-contrib/datafusion-distributed/compare/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94...7f006309f5b8664ca838251a0ac0053559dbeb0c)

<pre>=== Comparing tpch/sf100 results 'datafusion-benchmark-base' [prev] with 'datafusion-benchmark-head' [new] ===
TASKS: prev=849.0, new=849.0, diff=no change (sum of per-query averages)
TOTAL: prev=82049 ms, new=81259 ms, diff=1.01 faster ✔</pre>

<details>
<summary>Show full query output</summary>

<pre>      q1: prev=3251 ms, new=2890 ms, diff=1.12 faster ✔, tasks: prev=18.0, new=18.0, diff=no change
      q2: prev=3485 ms, new=3668 ms, diff=1.05 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
      q3: prev=3397 ms, new=2968 ms, diff=1.14 faster ✔, tasks: prev=48.0, new=48.0, diff=no change
      q4: prev=1441 ms, new=1437 ms, diff=1.00 faster ✔, tasks: prev=38.0, new=38.0, diff=no change
      q5: prev=4820 ms, new=4810 ms, diff=1.00 faster ✔, tasks: prev=52.0, new=52.0, diff=no change
      q6: prev=1510 ms, new=1758 ms, diff=1.16 slower ✖, tasks: prev=12.0, new=12.0, diff=no change
      q7: prev=10115 ms, new=10316 ms, diff=1.02 slower ✖, tasks: prev=53.0, new=53.0, diff=no change
      q8: prev=5518 ms, new=5374 ms, diff=1.03 faster ✔, tasks: prev=73.0, new=73.0, diff=no change
      q9: prev=9982 ms, new=10005 ms, diff=1.00 slower ✖, tasks: prev=77.0, new=77.0, diff=no change
     q10: prev=7865 ms, new=7664 ms, diff=1.03 faster ✔, tasks: prev=40.0, new=40.0, diff=no change
     q11: prev=3471 ms, new=3581 ms, diff=1.03 slower ✖, tasks: prev=24.0, new=24.0, diff=no change
     q12: prev=2239 ms, new=2062 ms, diff=1.09 faster ✔, tasks: prev=44.0, new=44.0, diff=no change
     q13: prev=1991 ms, new=1959 ms, diff=1.02 faster ✔, tasks: prev=32.0, new=32.0, diff=no change
     q14: prev=1852 ms, new=1829 ms, diff=1.01 faster ✔, tasks: prev=24.0, new=24.0, diff=no change
     q15: prev=4271 ms, new=4101 ms, diff=1.04 faster ✔, tasks: prev=30.0, new=30.0, diff=no change
     q16: prev= 826 ms, new= 972 ms, diff=1.18 slower ✖, tasks: prev=47.0, new=47.0, diff=no change
     q17: prev=5002 ms, new=5125 ms, diff=1.02 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
     q18: prev=5771 ms, new=5682 ms, diff=1.02 faster ✔, tasks: prev=68.0, new=68.0, diff=no change
     q19: prev=2179 ms, new=2209 ms, diff=1.01 slower ✖, tasks: prev=24.0, new=24.0, diff=no change
     q20: prev=3063 ms, new=2849 ms, diff=1.08 faster ✔, tasks: prev=65.0, new=65.0, diff=no change
q21: Previously failed, and now also failed ❌
q22: Previously failed, and now also failed ❌</pre>

</details>

<details>
<summary>Verification and run details</summary>

Job `46` captured both immutable revisions when the request was queued. The bot fetched and checked out each full commit SHA in detached HEAD, then built and deployed the `datafusion-distributed-benchmarks --bin worker` target from that checkout.

| Identity | PR base | PR head |
| --- | --- | --- |
| Source commit | [`417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) | [`7f006309f5b8664ca838251a0ac0053559dbeb0c`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) |

| Phase | Base | PR head |
| --- | ---: | ---: |
| Build and deployment | 1m 43s | 1m 23s |
| All benchmarks | 8m 39s | 8m 28s |
| Benchmark `tpch/sf100` | 8m 39s | 8m 28s |

**Workload:** `tpch/sf100` · all queries · 1 warmup + 5 measured iterations per query

**Capacity:** 12 `c5n.2xlarge` nodes for both revisions

**Other timings:** Queue 1s · Dataset validation 0s · Total 20m 21s

</details>

gabotechs (approved):
💯 let's go!

jayshrivastava:
benchmarks run tpch/sf100

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/623#issuecomment-5542240070).

## Benchmark results

**Compared:** [PR base `417f80977d7f`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) → [PR head `7f006309f5b8`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) · [View exact source diff](https://github.com/datafusion-contrib/datafusion-distributed/compare/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94...7f006309f5b8664ca838251a0ac0053559dbeb0c)

<pre>=== Comparing tpch/sf100 results 'datafusion-benchmark-base' [prev] with 'datafusion-benchmark-head' [new] ===
TASKS: prev=849.0, new=849.0, diff=no change (sum of per-query averages)
TOTAL: prev=80526 ms, new=81697 ms, diff=1.01 slower ✖</pre>

<details>
<summary>Show full query output</summary>

<pre>      q1: prev=2826 ms, new=2797 ms, diff=1.01 faster ✔, tasks: prev=18.0, new=18.0, diff=no change
      q2: prev=3287 ms, new=3455 ms, diff=1.05 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
      q3: prev=3139 ms, new=3044 ms, diff=1.03 faster ✔, tasks: prev=48.0, new=48.0, diff=no change
      q4: prev=1512 ms, new=1420 ms, diff=1.06 faster ✔, tasks: prev=38.0, new=38.0, diff=no change
      q5: prev=4618 ms, new=4973 ms, diff=1.08 slower ✖, tasks: prev=52.0, new=52.0, diff=no change
      q6: prev=1419 ms, new=1564 ms, diff=1.10 slower ✖, tasks: prev=12.0, new=12.0, diff=no change
      q7: prev=9901 ms, new=9935 ms, diff=1.00 slower ✖, tasks: prev=53.0, new=53.0, diff=no change
      q8: prev=5211 ms, new=5675 ms, diff=1.09 slower ✖, tasks: prev=73.0, new=73.0, diff=no change
      q9: prev=9713 ms, new=10151 ms, diff=1.05 slower ✖, tasks: prev=77.0, new=77.0, diff=no change
     q10: prev=7519 ms, new=8038 ms, diff=1.07 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
     q11: prev=3489 ms, new=3642 ms, diff=1.04 slower ✖, tasks: prev=24.0, new=24.0, diff=no change
     q12: prev=2223 ms, new=2076 ms, diff=1.07 faster ✔, tasks: prev=44.0, new=44.0, diff=no change
     q13: prev=1917 ms, new=1946 ms, diff=1.02 slower ✖, tasks: prev=32.0, new=32.0, diff=no change
     q14: prev=2150 ms, new=1783 ms, diff=1.21 faster ✔, tasks: prev=24.0, new=24.0, diff=no change
     q15: prev=4503 ms, new=4086 ms, diff=1.10 faster ✔, tasks: prev=30.0, new=30.0, diff=no change
     q16: prev= 837 ms, new= 863 ms, diff=1.03 slower ✖, tasks: prev=47.0, new=47.0, diff=no change
     q17: prev=5190 ms, new=5221 ms, diff=1.01 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
     q18: prev=5891 ms, new=5917 ms, diff=1.00 slower ✖, tasks: prev=68.0, new=68.0, diff=no change
     q19: prev=2130 ms, new=2126 ms, diff=1.00 faster ✔, tasks: prev=24.0, new=24.0, diff=no change
     q20: prev=3051 ms, new=2985 ms, diff=1.02 faster ✔, tasks: prev=65.0, new=65.0, diff=no change
q21: Previously failed, and now also failed ❌
q22: Previously failed, and now also failed ❌</pre>

</details>

<details>
<summary>Verification and run details</summary>

Job `47` captured both immutable revisions when the request was queued. The bot fetched and checked out each full commit SHA in detached HEAD, then built and deployed the `datafusion-distributed-benchmarks --bin worker` target from that checkout.

| Identity | PR base | PR head |
| --- | --- | --- |
| Source commit | [`417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) | [`7f006309f5b8664ca838251a0ac0053559dbeb0c`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) |

| Phase | Base | PR head |
| --- | ---: | ---: |
| Build and deployment | 2m 28s | 1m 24s |
| All benchmarks | 8m 28s | 9m 2s |
| Benchmark `tpch/sf100` | 8m 28s | 9m 2s |

**Workload:** `tpch/sf100` · all queries · 1 warmup + 5 measured iterations per query

**Capacity:** 12 `c5n.2xlarge` nodes for both revisions

**Other timings:** Queue 1s · Dataset validation 0s · Total 21m 31s

</details>

jayshrivastava:
benchmarks run tpch/sf10

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/623#issuecomment-5542525675).

## Benchmark results

**Compared:** [PR base `417f80977d7f`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) → [PR head `7f006309f5b8`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) · [View exact source diff](https://github.com/datafusion-contrib/datafusion-distributed/compare/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94...7f006309f5b8664ca838251a0ac0053559dbeb0c)

<pre>=== Comparing tpch/sf10 results 'datafusion-benchmark-base' [prev] with 'datafusion-benchmark-head' [new] ===
TASKS: prev=562.0, new=560.0, diff=2.0 fewer (0.4%) (sum of per-query averages)
TOTAL: prev=18866 ms, new=20422 ms, diff=1.08 slower ✖</pre>

<details>
<summary>Show full query output</summary>

<pre>      q1: prev= 344 ms, new= 630 ms, diff=1.83 slower ❌, tasks: prev=18.0, new=16.0, diff=2.0 fewer (11.1%)
      q2: prev= 816 ms, new= 887 ms, diff=1.09 slower ✖, tasks: prev=6.0, new=6.0, diff=no change
      q3: prev= 728 ms, new= 970 ms, diff=1.33 slower ✖, tasks: prev=28.0, new=28.0, diff=no change
      q4: prev= 326 ms, new= 586 ms, diff=1.80 slower ❌, tasks: prev=30.0, new=30.0, diff=no change
      q5: prev= 937 ms, new=1022 ms, diff=1.09 slower ✖, tasks: prev=32.0, new=32.0, diff=no change
      q6: prev= 526 ms, new= 341 ms, diff=1.54 faster ✅, tasks: prev=12.0, new=12.0, diff=no change
      q7: prev=1294 ms, new=1346 ms, diff=1.04 slower ✖, tasks: prev=33.0, new=33.0, diff=no change
      q8: prev=1156 ms, new=1352 ms, diff=1.17 slower ✖, tasks: prev=54.0, new=54.0, diff=no change
      q9: prev=1510 ms, new=1595 ms, diff=1.06 slower ✖, tasks: prev=57.0, new=57.0, diff=no change
     q10: prev=1196 ms, new=1239 ms, diff=1.04 slower ✖, tasks: prev=20.0, new=20.0, diff=no change
     q11: prev= 587 ms, new= 560 ms, diff=1.05 faster ✔, tasks: prev=6.0, new=6.0, diff=no change
     q12: prev= 452 ms, new= 681 ms, diff=1.51 slower ❌, tasks: prev=30.0, new=30.0, diff=no change
     q13: prev= 592 ms, new= 627 ms, diff=1.06 slower ✖, tasks: prev=10.0, new=10.0, diff=no change
     q14: prev= 479 ms, new= 398 ms, diff=1.20 faster ✔, tasks: prev=21.0, new=21.0, diff=no change
     q15: prev=1074 ms, new=1035 ms, diff=1.04 faster ✔, tasks: prev=30.0, new=30.0, diff=no change
     q16: prev= 409 ms, new= 487 ms, diff=1.19 slower ✖, tasks: prev=3.0, new=3.0, diff=no change
     q17: prev= 769 ms, new= 743 ms, diff=1.03 faster ✔, tasks: prev=37.0, new=37.0, diff=no change
     q18: prev=1172 ms, new=1184 ms, diff=1.01 slower ✖, tasks: prev=45.0, new=45.0, diff=no change
     q19: prev= 427 ms, new= 569 ms, diff=1.33 slower ✖, tasks: prev=21.0, new=21.0, diff=no change
     q20: prev= 755 ms, new= 810 ms, diff=1.07 slower ✖, tasks: prev=32.0, new=32.0, diff=no change
     q21: prev=3020 ms, new=3054 ms, diff=1.01 slower ✖, tasks: prev=28.0, new=28.0, diff=no change
     q22: prev= 297 ms, new= 306 ms, diff=1.03 slower ✖, tasks: prev=9.0, new=9.0, diff=no change</pre>

</details>

<details>
<summary>Verification and run details</summary>

Job `48` captured both immutable revisions when the request was queued. The bot fetched and checked out each full commit SHA in detached HEAD, then built and deployed the `datafusion-distributed-benchmarks --bin worker` target from that checkout.

| Identity | PR base | PR head |
| --- | --- | --- |
| Source commit | [`417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94`](https://github.com/datafusion-contrib/datafusion-distributed/commit/417f80977d7fa5b6c4be0e4d1fa7f30864c5ca94) | [`7f006309f5b8664ca838251a0ac0053559dbeb0c`](https://github.com/datafusion-contrib/datafusion-distributed/commit/7f006309f5b8664ca838251a0ac0053559dbeb0c) |

| Phase | Base | PR head |
| --- | ---: | ---: |
| Build and deployment | 1m 22s | 1m 22s |
| All benchmarks | 2m 3s | 2m 13s |
| Benchmark `tpch/sf10` | 2m 3s | 2m 13s |

**Workload:** `tpch/sf10` · all queries · 1 warmup + 5 measured iterations per query

**Capacity:** 12 `c5n.2xlarge` nodes for both revisions

**Other timings:** Queue 1s · Dataset validation 0s · Total 7m 9s

</details>

jayshrivastava:
benchmarks run clickbench/0-100-date32 --base e5a4a779d191206c1acd5cbfb31f3fed87c57934

gabot-0:
@jayshrivastava Invalid base `e5a4a779d191206c1acd5cbfb31f3fed87c57934`; only `main` is supported.

<details>
<summary>How to use the benchmark bot</summary>

Post a comment whose first non-empty line is:

`benchmarks run <suite>/<variant>... [--instance-type <type>] [--nodes <count>] [--base main] [--config <key=value>]...`

For example:

`benchmarks run tpch/sf10 tpch/sf100 --instance-type m5.2xlarge --nodes 24 --base main --config distributed.collect_dynamic_filters=false`

**Currently available datasets:** `clickbench/0-100`, `tpcds/sf1`, `tpch/sf1`, `tpch/sf10`, `tpch/sf100`. Request one or more, without duplicates. The bot validates availability before provisioning and runs every query in each dataset.

| Option | Default | Supported values and behavior |
| --- | --- | --- |
| `--instance-type <type>` | `c5n.4xlarge` | One of the supported instance types listed below, subject to availability within the cluster's availability zones and quota. |
| `--nodes <count>` | `12` | An integer from 1 to 60. The deployment uses one benchmark worker per node. |
| `--base main` | PR base | Compare against a snapshot of `main`; no other explicit base is supported. |
| `--config <key=value>` | none | Apply a safe DataFusion session setting to the PR head only. Repeat for distinct keys; spaces and shell syntax are not supported. |

**Supported instances and per-worker limits:**

| Instance type | EC2 capacity | Worker requests and limits |
| --- | ---: | ---: |
| `c5n.2xlarge` | 8 vCPU, 21 GiB | 7 vCPU, 17Gi |
| `c5n.4xlarge` | 16 vCPU, 42 GiB | 15 vCPU, 38Gi |
| `m5.2xlarge` | 8 vCPU, 32 GiB | 7 vCPU, 28Gi |
| `m5.4xlarge` | 16 vCPU, 64 GiB | 15 vCPU, 60Gi |
| `r5.2xlarge` | 8 vCPU, 64 GiB | 7 vCPU, 60Gi |
| `r5.4xlarge` | 16 vCPU, 128 GiB | 15 vCPU, 124Gi |

Each node runs one worker. The worker allocation reserves 1 vCPU and 4 GiB for Kubernetes and system processes, then makes the rest of the selected instance available to the benchmark.

**Limits:** Only authorized users can enqueue jobs. Jobs run serially, the queue holds 20 active jobs, and each requester may have 3. Query selection and iteration overrides are not supported; every query uses 1 warmup and 5 measured iterations for both revisions.

</details>

PR #634, opened 2026-08-13
feat: plan distributed dynamic filters

jayshrivastava:
## Stack

2. https://github.com/datafusion-contrib/datafusion-distributed/pull/634 <- you are here

## Goal

The coordinator should know what dynamic filters exist and where to route updates.

### 1. Dynamic Filter Registry

```
  QueryCoordinator
  └── DynamicFilterRegistry
      └── filters: Map<expression_id, PlannedDynamicFilter>
```

Each `PlannedDynamicFilter` stores
- the producers and their stage/tasks
- the consumers and their stage/tasks

This will be used in future PRs to store incoming dynamic filter updates from workers and determine how/where to forward the updates.

#### Implementation

In the `StageCoordinator`, we send every task to the registry and extract dynamic filters.

### 2. Network Anchors

In this situation, the hash join does an `execute()`-time check to determine if it should update its dynamic filter. It checks to see if the filter is used by any children using `apply_expressions` (before `apply_expressions` was added upstream, it was an Arc pointer strong count check to see if there were multiple references).
```
worker 1
HashJoinExec  (dynamic_filter_predicate)
    NetworkShuffleExec

worker 2
DataSourceExec (dynamic_filter_predicate)
```
The join sees that no plan nodes below it use the filter, so it decides not to update it.

Ideally, the hash join decides at optimization time, before distributed planning. I've opened a discussion here about it: https://github.com/apache/datafusion/issues/18856#issuecomment-5375571998. While that issue is being resolved, I propose this workaround:

We create an "anchor" to make it seem like the `NetworkShuffleExec` uses the filter.
```
worker 1
HashJoinExec  (dynamic_filter_predicate)
    NetworkShuffleExec  (anchor: dynamic_filter_predicate)

worker 2
DataSourceExec (dynamic_filter_predicate)
```

#### Network Anchors Implementation

The implementation adds serialization overhead but is simpler. In static and dynamic planning, we recursively propagate all anchors upwards in the plan to all the network boundaries. We can revisit this implementation in  future iterations. This recursive implementation is in `inject_network_boundaries`.
```
stage3:
    HashJoinExec <- producer of filter1
        NetworkShuffleExec  (anchors: filter1, filter2)

stage2:
    RepartitionExec
        AggregateExec <- producer of filter2
            NetworkShuffleExec   (anchors: filter1, filter2)

stage1:
        DataSourceExec (consumer: filter1, filter2)
```
This means we serialize 8 filters in total.

However, the minimal anchors you need are like this:
```
stage3:
    HashJoinExec <- producer #1
        NetworkShuffleExec  (anchors: filter2)

stage2:
    RepartitionExec
        AggregateExec <- producer #2
            NetworkShuffleExec   (anchors: filter2)

stage1:
        DataSourceExec (consumer: filter1, filter2)
```
In this plan, we would serialize 6 filters.

For 1 dynamic filter, the minimum filters you need to serialize are 1 (producer) + N (consumers) + 1 (network boundary). In this implementation, we serialize 1 (producer) + N (consumers) + M (all network boundaries above the consumer)

### Other Notes

See https://github.com/datafusion-contrib/datafusion-distributed/issues/528. During dynamic planning, the sampler on the probe side of a hash join may overreport rows / cost because dynamic filters aren't being applied yet.

### DIFF
--- a/src/coordinator/dynamic_filter_registry.rs
+++ b/src/coordinator/dynamic_filter_registry.rs
@@ -1,0 +1,213 @@
+use crate::TaskKey;
+use crate::dynamic_filtering::{
+    discover_dynamic_filter_consumers, discover_dynamic_filter_producers,
+};
+use datafusion::common::{HashMap, HashSet, Result};
+use datafusion::physical_plan::ExecutionPlan;
+use std::sync::{Arc, Mutex};
+
+#[derive(Default)]
+pub(super) struct PlannedDynamicFilter {
+    // Producer and consumer tasks for a dynamic filter.
+    //
+    // Note that it is not guaranteed that every task within a stage produces / consumes dynamic filters. For
+    // example, a distributed union may prevent a dynamic filter from appearing in all tasks. So, we
+    // store task keys rather than stage ids.
+    pub(super) producer_tasks: HashSet<TaskKey>,
+    pub(super) consumer_tasks: HashSet<TaskKey>,
+}
+
+#[derive(Default)]
+pub(super) struct DynamicFilterRegistryState {
+    pub(super) filters: HashMap<u64, PlannedDynamicFilter>,
+}
+
+/// Query-scoped hub for distributed dynamic filtering. It stores the locations
+/// of dynamic filters and runtime state, informing the coordinator where dynamic filter
+/// updates are coming from, how/if they should be merged, where they need to be forwarded.
+#[derive(Default)]
+pub(crate) struct DynamicFilterRegistry {
+    pub(super) state: Mutex<DynamicFilterRegistryState>,
+}
+
+impl DynamicFilterRegistry {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Adds any dynamic filter producers and consumers found in `plan` to the registry.
+    pub(crate) fn register_task(
+        &self,
+        plan: &Arc<dyn ExecutionPlan>,
+        task_key: TaskKey,
+    ) -> Result<()> {
+        let producers = discover_dynamic_filter_producers(plan)?;
+        // We can safely ignore anchors because they are not evaluated by network boundaries. This
+        // means they do not need updates forwarded to them.
+        let consumers = discover_dynamic_filter_consumers(plan)?.consumers;
+
+        let mut state = self.state.lock().expect("dynamic filter registry poisoned");
+        for producer in producers {
+            state
+                .filters
+                .entry(producer.id)
+                .or_default()
+                .producer_tasks
+                .insert(task_key);
+        }
+        for consumer in consumers {
+            state
+                .filters
+                .entry(consumer.id)
+                .or_default()
+                .consumer_tasks
+                .insert(task_key);
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::execution_plans::NetworkShuffleExec;
+    use datafusion::arrow::datatypes::{DataType, Field, Schema};
+    use datafusion::common::tree_node::TreeNodeRecursion;
+    use datafusion::execution::{SendableRecordBatchStream, TaskContext};
+    use datafusion::physical_expr::expressions::{Column, DynamicFilterPhysicalExpr, lit};
+    use datafusion::physical_expr::{Partitioning, PhysicalExpr};
+    use datafusion::physical_plan::empty::EmptyExec;
+    use datafusion::physical_plan::repartition::RepartitionExec;
+    use datafusion::physical_plan::{
+        DisplayAs, DisplayFormatType, PlanProperties, apply_expression_roots,
+    };
+    use std::fmt::Formatter;
+    use uuid::Uuid;
+
+    // Test that we correctly register producers and consumers while ignoring anchors.
+    #[test]
+    fn registers_dynamic_filters_by_expression_id() -> Result<()> {
+        let schema = Arc::new(Schema::new(vec![Field::new("a", DataType::Int32, false)]));
+        let dynamic_filter = Arc::new(DynamicFilterPhysicalExpr::new(
+            vec![Arc::new(Column::new("a", 0))],
+            lit(true),
+        )) as Arc<dyn PhysicalExpr>;
+        let id = dynamic_filter.expression_id().unwrap();
+
+        let producer_occurrence =
+            Arc::clone(&dynamic_filter).with_new_children(vec![Arc::new(Column::new("a", 0))])?;
+        assert!(!Arc::ptr_eq(&dynamic_filter, &producer_occurrence));
+
+        let input = Arc::new(EmptyExec::new(Arc::clone(&schema))) as Arc<dyn ExecutionPlan>;
+        let repartition = Arc::new(RepartitionExec::try_new(
+            input,
+            Partitioning::Hash(vec![Arc::new(Column::new("a", 0))], 1),
+        )?) as Arc<dyn ExecutionPlan>;
+        let boundary = Arc::new(
+            NetworkShuffleExec::try_new(repartition, 1)?
+                .with_dynamic_filter_anchors(vec![Arc::clone(&dynamic_filter)]),
+        ) as Arc<dyn ExecutionPlan>;
+        let producer = Arc::new(ExpressionExec::new(
+            boundary,
+            producer_occurrence,
+            Some(Arc::clone(&dynamic_filter)),
+        )) as Arc<dyn ExecutionPlan>;
+        let producer_task = task_key(0);
+
+        let consumer = Arc::new(ExpressionExec::new(
+            Arc::new(EmptyExec::new(schema)),
+            dynamic_filter,
+            None,
+        )) as Arc<dyn ExecutionPlan>;
+        let consumer_task = task_key(1);
+
+        let registry = DynamicFilterRegistry::new();
+        registry.register_task(&producer, producer_task)?;
+        registry.register_task(&consumer, consumer_task)?;
+
+        let state = registry.state.lock().unwrap();
+        let filter = state.filters.get(&id).unwrap();
+        assert_eq!(filter.producer_tasks, HashSet::from([producer_task]));
+        assert_eq!(filter.consumer_tasks, HashSet::from([consumer_task]));
+        Ok(())
+    }
+
+    fn task_key(task_number: usize) -> TaskKey {
+        TaskKey {
+            query_id: Uuid::nil(),
+            stage_id: 3,
+            task_number,
+        }
+    }
+
+    #[derive(Debug)]
+    struct ExpressionExec {
+        input: Arc<dyn ExecutionPlan>,
+        expression: Arc<dyn PhysicalExpr>,
+        produced_expression: Option<Arc<dyn PhysicalExpr>>,
+    }
+
+    impl ExpressionExec {
+        fn new(
+            input: Arc<dyn ExecutionPlan>,
+            expression: Arc<dyn PhysicalExpr>,
+            produced_expression: Option<Arc<dyn PhysicalExpr>>,
+        ) -> Self {
+            Self {
+                input,
+                expression,
+                produced_expression,
+            }
+        }
+    }
+
+    impl DisplayAs for ExpressionExec {
+        fn fmt_as(&self, _: DisplayFormatType, f: &mut Formatter) -> std::fmt::Result {
+            write!(f, "ExpressionExec")
+        }
+    }
+
+    impl ExecutionPlan for ExpressionExec {
+        fn name(&self) -> &str {
+            "ExpressionExec"
+        }
+
+        fn properties(&self) -> &Arc<PlanProperties> {
+            self.input.properties()
+        }
+
+        fn children(&self) -> Vec<&Arc<dyn ExecutionPlan>> {
+            vec![&self.input]
+        }
+
+        fn dynamic_expressions_produced(&self) -> Vec<Arc<dyn PhysicalExpr>> {
+            self.produced_expression.iter().cloned().collect()
+        }
+
+        fn apply_expressions(
+            &self,
+            f: &mut dyn FnMut(&Arc<dyn PhysicalExpr>) -> Result<TreeNodeRecursion>,
+        ) -> Result<TreeNodeRecursion> {
+            apply_expression_roots([&self.expression], f)
+        }
+
+        fn with_new_children(
+            self: Arc<Self>,
+            mut children: Vec<Arc<dyn ExecutionPlan>>,
+        ) -> Result<Arc<dyn ExecutionPlan>> {
+            Ok(Arc::new(Self::new(
+                children.remove(0),
+                Arc::clone(&self.expression),
+                self.produced_expression.clone(),
+            )))
+        }
+
+        fn execute(
+            &self,
+            partition: usize,
+            context: Arc<TaskContext>,
+        ) -> Result<SendableRecordBatchStream> {
+            self.input.execute(partition, context)
+        }
+    }
+}
--- a/src/coordinator/dynamic_filter_registry.rs
+++ b/src/coordinator/dynamic_filter_registry.rs
@@ -22,9 +22,12 @@ pub(super) struct DynamicFilterRegistryState {
     pub(super) filters: HashMap<u64, PlannedDynamicFilter>,
 }
 
-/// Query-scoped hub for distributed dynamic filtering. It stores the locations
-/// of dynamic filters and runtime state, informing the coordinator where dynamic filter
-/// updates are coming from, how/if they should be merged, where they need to be forwarded.
+/// Query-scoped hub for distributed dynamic filtering.
+///
+/// It stores the locations of dynamic filters and their runtime state. Informs the coordinator
+/// - where dynamic filter updates are coming from
+/// - how/if dynamic filter updates should be merged
+/// - where dynamic filter updates should be forwarded
 #[derive(Default)]
 pub(crate) struct DynamicFilterRegistry {
     pub(super) state: Mutex<DynamicFilterRegistryState>,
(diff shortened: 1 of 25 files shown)


src/worker/impl_coordinator_channel.rs
```diff
@@ -234,15 +246,80 @@ impl Worker {
...
+            loop {
+                // `wait_update()` uses a Tokio watch channel, so multiple generations can
+                // naturally coalesce while this observer is busy. The stream promises the latest
+                // observed state rather than one message per generation; completion is awaited
+                // separately because it does not advance the generation.
+                let completed = tokio::select! {
+                    _ = dynamic_filter.wait_update() => false,
+                    _ = dynamic_filter.wait_complete() => true,
+                    _ = cancel_rx.wait_for(|cancelled| *cancelled) => return None,
+                };
```
jayshrivastava: This is up for discussion. We went an update across the network whenever there's an update. 

We need this because certain dynamic filter producers such as sort and aggregate rely on these updates for early pruning and don't actually mark their filters as complete.

The tradeoff is that it's expensive to send a network request on every update. 

Note that any updates which happen between calls to `wait_update()` are deduped, but there still may be an update ready every time we hit this line. 

The alternative I was considering is adding a delay between polls to avoid too many updates across the network.

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -0,0 +1,210 @@
...
+    // that every task within a stage produces / consumes dynamic filters so we specifically store
+    // task keys rather than stage ids. For example, a distributed union may prevent a dynamic filter
+    // consumer from appearing in all tasks.
+    pub(super) producer_tasks: HashSet<TaskKey>,
+    pub(super) consumer_tasks: HashSet<TaskKey>,
+}
+
+#[derive(Default)]
+pub(super) struct DynamicFilterRegistryState {
+    pub(super) filters: HashMap<u64, PlannedDynamicFilter>,
```
jayshrivastava, line 22: In future PRs, we add more mutex protected state in this struct.

src/dynamic_filtering/mod.rs
```diff
@@ -34,14 +40,53 @@ pub(crate) use display::sever_dynamic_filter_relationships_in_plan_for_display;
...
+// Stage 2 Task 1
+// HashJoinExec <- Dynamic Filter Produced: (foo != 150)
+//   DataSourceExec <- consumer
+// ```
+//
+// Since both producers and both consumers are located on the same worker, they all share
+// one in-memory dynamic filter. This ends up being a race between two writers and two readers.
+// For a partitioned hash join, a producer may update its task-local consumer in memory, but
+// updates must not cross task boundaries.
 pub(crate) fn maybe_roundtrip_plan_to_sever_in_memory_dynamic_filter_relationships(
```
jayshrivastava, line 80: I added more failure modes here and an integration test, `colocated_local_dynamic_filters` to cover this. I think it's safest to roundtrip any plan that has dynamic filters.

jayshrivastava:
benchmarks run tpch/sf100

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/634#issuecomment-5595173868).

Benchmark job 50 failed for `tpch/sf100`. Full details are available in the controller journal.

gabotechs:
benchmarks run tpch/sf100

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/634#issuecomment-5597428735).

## Benchmark results

**Compared:** [PR base `e5bca18b8016`](https://github.com/datafusion-contrib/datafusion-distributed/commit/e5bca18b801652f4c79928f1536aed2e1d48e21f) → [PR head `820f92a02296`](https://github.com/datafusion-contrib/datafusion-distributed/commit/820f92a02296ef4e378b92a6c4ff74c2cfdd3f1a) · [View exact source diff](https://github.com/datafusion-contrib/datafusion-distributed/compare/e5bca18b801652f4c79928f1536aed2e1d48e21f...820f92a02296ef4e378b92a6c4ff74c2cfdd3f1a)

<pre>=== Comparing tpch/sf100 results 'datafusion-benchmark-base' [prev] with 'datafusion-benchmark-head' [new] ===
TASKS: prev=1286.0, new=1286.0, diff=no change (sum of per-query averages)
TOTAL: prev=73065 ms, new=75487 ms, diff=1.03 slower ✖</pre>

<details>
<summary>Show full query output</summary>

<pre>      q1: prev=3030 ms, new=2677 ms, diff=1.13 faster ✔, tasks: prev=24.0, new=24.0, diff=no change
      q2: prev=1679 ms, new=1686 ms, diff=1.00 slower ✖, tasks: prev=94.0, new=94.0, diff=no change
      q3: prev=3205 ms, new=3232 ms, diff=1.01 slower ✖, tasks: prev=56.0, new=56.0, diff=no change
      q4: prev=1339 ms, new=1530 ms, diff=1.14 slower ✖, tasks: prev=48.0, new=48.0, diff=no change
      q5: prev=4114 ms, new=4267 ms, diff=1.04 slower ✖, tasks: prev=83.0, new=83.0, diff=no change
      q6: prev=1538 ms, new=1564 ms, diff=1.02 slower ✖, tasks: prev=12.0, new=12.0, diff=no change
      q7: prev=4935 ms, new=4995 ms, diff=1.01 slower ✖, tasks: prev=83.0, new=83.0, diff=no change
      q8: prev=5106 ms, new=5299 ms, diff=1.04 slower ✖, tasks: prev=100.0, new=100.0, diff=no change
      q9: prev=6699 ms, new=6790 ms, diff=1.01 slower ✖, tasks: prev=102.0, new=102.0, diff=no change
     q10: prev=5237 ms, new=5664 ms, diff=1.08 slower ✖, tasks: prev=69.0, new=69.0, diff=no change
     q11: prev=1076 ms, new=1127 ms, diff=1.05 slower ✖, tasks: prev=65.0, new=65.0, diff=no change
     q12: prev=2024 ms, new=2187 ms, diff=1.08 slower ✖, tasks: prev=48.0, new=48.0, diff=no change
     q13: prev=1746 ms, new=1817 ms, diff=1.04 slower ✖, tasks: prev=44.0, new=44.0, diff=no change
     q14: prev=2054 ms, new=2026 ms, diff=1.01 faster ✔, tasks: prev=28.0, new=28.0, diff=no change
     q15: prev=4273 ms, new=4212 ms, diff=1.01 faster ✔, tasks: prev=50.0, new=50.0, diff=no change
     q16: prev= 824 ms, new= 844 ms, diff=1.02 slower ✖, tasks: prev=53.0, new=53.0, diff=no change
     q17: prev=4979 ms, new=5123 ms, diff=1.03 slower ✖, tasks: prev=40.0, new=40.0, diff=no change
     q18: prev=5782 ms, new=5858 ms, diff=1.01 slower ✖, tasks: prev=68.0, new=68.0, diff=no change
     q19: prev=1976 ms, new=2372 ms, diff=1.20 slower ✖, tasks: prev=28.0, new=28.0, diff=no change
     q20: prev=2767 ms, new=3077 ms, diff=1.11 slower ✖, tasks: prev=65.0, new=65.0, diff=no change
     q21: prev=7862 ms, new=8238 ms, diff=1.05 slower ✖, tasks: prev=86.0, new=86.0, diff=no change
     q22: prev= 820 ms, new= 902 ms, diff=1.10 slower ✖, tasks: prev=40.0, new=40.0, diff=no change</pre>

</details>

<details>
<summary>Verification and run details</summary>

Job `52` captured both immutable revisions when the request was queued. The bot fetched and checked out each full commit SHA in detached HEAD, then built and deployed the `datafusion-distributed-benchmarks --bin worker` target from that checkout.

| Identity | PR base | PR head |
| --- | --- | --- |
| Source commit | [`e5bca18b801652f4c79928f1536aed2e1d48e21f`](https://github.com/datafusion-contrib/datafusion-distributed/commit/e5bca18b801652f4c79928f1536aed2e1d48e21f) | [`820f92a02296ef4e378b92a6c4ff74c2cfdd3f1a`](https://github.com/datafusion-contrib/datafusion-distributed/commit/820f92a02296ef4e378b92a6c4ff74c2cfdd3f1a) |

| Phase | Base | PR head |
| --- | ---: | ---: |
| Build and deployment | 1m 21s | 1m 20s |
| All benchmarks | 7m 36s | 8m 0s |
| Benchmark `tpch/sf100` | 7m 36s | 8m 0s |

**Workload:** `tpch/sf100` · all queries · 1 warmup + 5 measured iterations per query

**Capacity:** 12 `c5n.2xlarge` nodes for both revisions

**Other timings:** Queue 1s · Dataset validation 0s · Total 18m 24s

</details>

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -0,0 +1,213 @@
...
+    use datafusion::physical_plan::repartition::RepartitionExec;
+    use datafusion::physical_plan::{
+        DisplayAs, DisplayFormatType, PlanProperties, apply_expression_roots,
+    };
+    use std::fmt::Formatter;
+    use uuid::Uuid;
+
+    // Test that we correctly register producers and consumers while ignoring anchors.
+    #[test]
+    fn registers_dynamic_filters_by_expression_id() -> Result<()> {
```
gabotechs, line 89: This test seems a bit too big and verbose, and I seems a bit coupled to the specific implementation detail of the `DynamicFilterRegistry`.

I think we should be able to cover this with an integration test, that tests this just through the public API instead.

src/dynamic_filtering/discovery.rs
```diff
@@ -179,26 +261,58 @@ mod tests {
...
+        assert_eq!(
+            discovered.consumers[0].id,
+            dynamic_filter.expression_id().unwrap()
+        );
         Ok(())
     }
 
     #[test]
-    fn identifies_a_producer_without_a_local_consumer() -> Result<()> {
+    fn distinguishes_consumers_from_network_boundary_anchors() -> Result<()> {
```
gabotechs, line 274: This is one of the issues with testing at the private module level: as the code evolves, the test needs to evolve with it, introducing maintenance burden and making the tests inefective for future contributions.

As dynamic filter integration evolves, it could make sense to think about moving these to just integration tests that stay stable across contributions.

src/execution_plans/network_broadcast.rs
```diff
@@ -122,6 +122,7 @@ pub struct NetworkBroadcastExec {
     pub(crate) properties: Arc<PlanProperties>,
     pub(crate) input_stage: Stage,
     pub(crate) worker_connections: WorkerConnectionPool,
+    pub(crate) dynamic_filter_anchors: Vec<Arc<dyn PhysicalExpr>>,
```
gabotechs, line 125: I see the anchors are present at every time, however, I think it might make sense to move them to `Stage::Remote { .., dynamic_filter_anchors }` instead.

The reason is that, if the `input_stage` is `Stage::Local`, execution plan traversal will work as normal, and you should not need to resource to artifical anchors for discovering dynamic filters.

However, if `input_stage == Stage::Remote`, the plan got broken down into fragments, and the fragment below is not recursible. This seems like the only moment we really need to resource to dynamic filter anchors as a workaround for discovering dynamic filters.

gabotechs: I bet that placing `dynamic_filter_anchors` would also provide a cleaner implementation, with less repetition, and better encapsulation of the fact that we are capturing the anchors specifically because of the fact that we are transitioning a `Stage` to `Stage::Remote`, losing access to the plan below.

src/coordinator/query_coordinator.rs
```diff
@@ -157,6 +162,9 @@ impl<'a> StageCoordinator<'a> {
             stage_id: self.stage_id,
             task_number: task_i,
         };
+        self.dynamic_filter_registry
+            .register_task(&specialized, task_key)?;
+
```
gabotechs, line 167: :+1: very naturally handled

src/coordinator/dynamic_filter_registry.rs
gabotechs: Given that `PlannedDynamicFilter` is mostly just a placeholder implementation for now, I think we should be good with just removing this test.

src/test_utils/routing.rs
```diff
@@ -386,3 +389,33 @@ impl PhysicalExtensionCodec for URLEmitterExtensionCodec {
             .map_err(|e| proto_error(format!("Failed to encode URLEmitterExec: {e}")))
     }
 }
+
+/// Colocates all tasks on the same worker by choosing a URL once and caching it.
```
jayshrivastava, line 393: This is new, as a part of the rebase. Previously, this was in `tests/dynamic_filtering/common.rs`

src/coordinator/dynamic_filter_registry.rs
jayshrivastava: Deleted 👍🏽

src/dynamic_filtering/discovery.rs
jayshrivastava: Great point. I pushed a change which I think you will like.

The tests in `src/dynamic_filtering/discovery.rs` are now gone in favor of a test in `tests/dynamic_filtering/discovery.rs` where we use SQL and snapshot the plan.

src/execution_plans/network_broadcast.rs
jayshrivastava: I've pushed this in the last commit. Let me know what you think. The difference in lines of code is about the same.

I agree that it makes sense for the anchors to live in `Stage::Remote` though. That way we keep the `NetworkBoundary` implementations clean.

 I haven't updated the PR description yet in case we would like to revert that.

jayshrivastava:
@gabotechs This is ready for another review.

src/execution_plans/network_broadcast.rs
jayshrivastava: One side effect is that the insertion of anchors happens in prepare static/dynamic plan now instead of `inject_network_boundaries`

src/coordinator/dynamic_filter_registry.rs
gabotechs: More? I already see one `Mutex` in `DynamicFilterRegistry.state`, do you think we'll need more than that?

src/dynamic_filtering/discovery.rs
```diff
@@ -6,20 +7,45 @@ use datafusion::physical_expr::expressions::DynamicFilterPhysicalExpr;
...
+    // Real consumers, ordered by expression id.
+    pub consumers: Vec<DiscoveredDynamicFilter>,
+    // Artificial consumers. Dynamic filters in network boundaries. Also ordered by expression id.
+    pub anchors: Vec<DiscoveredDynamicFilterAnchor>,
+}
+
+/// Finds dynamic-filter consumers and network-boundary anchors in `plan`, deduplicated by
+/// expression ID within each category.
+pub fn discover_dynamic_filter_consumers(
     plan: &Arc<dyn ExecutionPlan>,
```
gabotechs, line 45: I see the public exposure here is getting a bit out of hand. Do we really need all these to be part of the public API?

My intuition tells me that none of these should be part of the public API

src/dynamic_filtering/mod.rs
```diff
@@ -4,44 +4,138 @@ mod display;
...
+/// will not share state with plan nodes outside of this plan, such as in parent stages.
+///
+/// # Correctness: Task-Local Dynamic Filters are Safe to Apply
+///
+/// Claim: It is correct for a producer to *always* update consumers within the same task.
+///
+/// Proof:
+///
+/// Invariant: If the task-local consumer filters out a row, removing that row must not change the output
+/// of the task-local producer.
```
gabotechs: This doc comment is a bit too ong and LLM-y, If we can shorten it out a bit that'd be great.

src/dynamic_filtering/mod.rs
```diff
@@ -4,44 +4,138 @@ mod display;
...
     task_ctx: &Arc<TaskContext>,
 ) -> Result<Arc<dyn ExecutionPlan>> {
-    if has_nonlocal_dynamic_filter_relationships(&plan)? {
+    let has_producers = !discover_dynamic_filter_producers(&plan)?.is_empty();
+    let has_consumers = !discover_dynamic_filter_consumers(&plan)?
+        .consumers
+        .is_empty();
+
+    if has_producers || has_consumers {
         roundtrip_pb(plan, task_ctx)
```
gabotechs, line 177: This `roundtrip_pb` function is serializing to bytes and decoding those bytes back. The byte serialization is unnecessary, it's enough to just convert to prost-generated struct and back, without paying for serialization here.

src/test_utils/routing.rs
```diff
@@ -386,3 +389,33 @@ impl PhysicalExtensionCodec for URLEmitterExtensionCodec {
...
+        let url = {
+            let mut cached = self.cached.lock().await;
+            if let Some(url) = cached.as_ref() {
+                url.clone()
+            } else {
+                let urls = match ev.worker_resolver.get_urls() {
+                    Ok(urls) => urls,
+                    Err(error) => return Some(Err(error)),
+                };
+                let Some(url) = urls.into_iter().next() else {
```
gabotechs:
```suggestion
                let urls = ok_or_some_err!( ev.worker_resolver.get_urls());
```

tests/dynamic_filtering/discovery.rs
```diff
@@ -0,0 +1,284 @@
...
+                    .anchors
+                    .into_iter()
+                    .map(|anchor| anchor.id)
+                    .collect(),
+                producers: discover_dynamic_filter_producers(plan)?
+                    .into_iter()
+                    .map(|producer| producer.id)
+                    .collect(),
+            };
+            render(plan.as_ref(), 1, &discovered, &mut normalizer, &mut output)?;
```
gabotechs: I don't think we want to expose this API to the public:
- `discover_dynamic_filter_consumers`
- `DiscoveredDynamicFilterConsumers`
- `DiscoveredDynamicFilterAnchor`
- `DiscoveredDynamicFilter`
- `DiscoveredDynamicFilterProducer`

These are internal struct that we invented inside `datafusion-distributed`, and users should not be able to interact with them.

I get the feeling that the previous unit tests where moved here as "integration" tests, but these tests are still coupled to internal implementation details of the dyn filter discovery machinery.

Some options here:
- Move this tests to src/dynamic_fitlering/discovery.rs, and remove all the public exposure of the internal structs
- Create a test_utils/dynamic_filters.rs helper that exposes a properly encapsulated `display_dynamic_filter_discovery()` function, keeping private all the other internal structs.
- Keep the tests here as real integration tests, using exclusively the public API

src/dynamic_filtering/mod.rs
jayshrivastava: I actually wrote it myself 😭 

I'll shorten it out though. I just wanted to have this written down somewhere

src/dynamic_filtering/discovery.rs
jayshrivastava: These are actually gated by `#[cfg(any(feature = "integration", test))]`

The reason is because I moved to integration style tests in `/home/bits/datafusion-distributed/tests/`, so these helpers need to be pub.

However, I think we can probably move those integration-style tests to this file to avoid the import hacks.

src/dynamic_filtering/mod.rs
jayshrivastava: I added diagrams, which is more lines but is better than a wall of text.

src/dynamic_filtering/mod.rs
jayshrivastava: Done. Good observation

tests/dynamic_filtering/discovery.rs
jayshrivastava: Done

jayshrivastava:
Ready for another review! I'll rebase before merging. I don't want to rebase and overwrite the history so that my diffs are easy to review.

src/codec/physical_plan.rs
```diff
@@ -49,8 +49,13 @@ pub(crate) fn roundtrip_pb(
...
 ) -> Result<Arc<dyn ExecutionPlan>> {
-    let encoded = encode_execution_plan(plan, task_ctx)?;
-    decode_execution_plan(&encoded, task_ctx)
+    let encode_codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
+    let encode_converter = new_proto_converter();
+    let proto = encode_converter.execution_plan_to_proto(&plan, &encode_codec)?;
+    let decode_codec = DistributedCodec::new_combined_with_user(task_ctx.session_config());
+    let decode_ctx = PhysicalPlanDecodeContext::new(task_ctx, &decode_codec);
+    let decode_converter = new_proto_converter();
+    decode_converter.proto_to_execution_plan(&proto, &decode_ctx)
```
gabotechs, line 58: 👍 nice

gabotechs (approved):
👍 Looks good!

PR #635, opened 2026-08-13
feat: forward remote dynamic filter updates to coordinator

jayshrivastava:
## Stack

3. https://github.com/datafusion-contrib/datafusion-distributed/pull/635 <- you are here

## Problem

The `QueryCoordinator` needs to receive partial dynamic filter updates from workers.

## Solution
We introduce a new `WorkerToCoordinatorMsg` which 
```
message ProducedDynamicFilter {
  uint64 expression_id = 1;
  // Serialized datafusion.proto.PhysicalExprNode.
  bytes expression_proto = 2;
}
```

In this PR makes each worker unconditionally send updates (via `wait_update()` and `wait_complete()`) to the coordinator for any `dynamic_filter_remote_producer_ids` in the `SetPlanRequest`. The purpose of `dynamic_filter_remote_producer_ids` is to exclude any dynamic filters who only have local consumers - these don't need to be forwarded to the coordinator.

### DIFF
--- a/src/coordinator/query_coordinator.rs
+++ b/src/coordinator/query_coordinator.rs
@@ -497,11 +497,13 @@ impl<'a> StageCoordinator<'a> {
     }
 }
 
-/// Returns producer IDs whose consumers cross a network boundary in this task plan.
+/// Returns producer IDs with at least one remote consumer.
 ///
-/// Consumers in the same task share their dynamic-filter expression with the producer and are
-/// updated directly by DataFusion. Network-boundary expressions are metadata anchors installed by
-/// distributed planning; their IDs identify the producers whose updates must cross the coordinator.
+/// If a producer ID is present in the dynamic-filter anchors of any [`NetworkBoundary`], the plan
+/// contains at least one remote consumer and the producer's updates must be forwarded to the
+/// coordinator.
+///
+/// [`NetworkBoundary`]: crate::NetworkBoundary
 fn dynamic_filter_remote_producer_ids(plan: &Arc<dyn ExecutionPlan>) -> Result<Vec<u64>> {
     let producer_ids: HashSet<_> = discover_dynamic_filter_producers(plan)?
         .into_iter()
--- a/src/coordinator/query_coordinator.rs
+++ b/src/coordinator/query_coordinator.rs
@@ -1,11 +1,11 @@
-use crate::codec::roundtrip_pb;
+use crate::codec::{encode_execution_plan, roundtrip_pb};
 use crate::common::{TreeNodeExt, now_ns, task_ctx_with_extension};
 use crate::config_extension_ext::get_config_extension_propagation_headers;
 use crate::coordinator::DynamicFilterRegistry;
 use crate::coordinator::Store;
 use crate::coordinator::latency_metric::LatencyMetric;
 use crate::dynamic_filtering::{
-    is_dynamic_filtering_enabled,
+    discover_dynamic_filter_producers, is_dynamic_filtering_enabled,
     maybe_roundtrip_plan_to_sever_in_memory_dynamic_filter_relationships,
 };
 use crate::events::{
@@ -19,15 +19,16 @@ use crate::work_unit_feed::{build_work_unit_batch_msg, set_work_unit_send_time};
 use crate::{
     CoordinatorToWorkerMsg, DISTRIBUTED_DATAFUSION_TASK_ID_LABEL, DistributedGetterExt,
     DistributedTaskContext, DistributedWorkUnitFeedContext, LoadInfo, LocalWorkerContext,
-    MaybeEncoded, SetPlanRequest, TaskCompletedDynamicFilters, TaskKey, TaskMetrics,
-    WorkUnitFeedDeclaration, WorkerToCoordinatorMsg, get_distributed_channel_resolver,
+    MaybeEncoded, NetworkBoundaryExt, SetPlanRequest, TaskCompletedDynamicFilters, TaskKey,
+    TaskMetrics, WorkUnitFeedDeclaration, WorkerToCoordinatorMsg, get_distributed_channel_resolver,
 };
 use datafusion::common::Result;
 use datafusion::common::instant::Instant;
 use datafusion::common::runtime::JoinSet;
-use datafusion::common::tree_node::{Transformed, TreeNodeRecursion};
-use datafusion::common::{DataFusionError, internal_err};
+use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeRecursion};
+use datafusion::common::{DataFusionError, HashSet, internal_err};
 use datafusion::execution::TaskContext;
+use datafusion::physical_expr::expressions::DynamicFilterPhysicalExpr;
 use datafusion::physical_expr_common::metrics::{ExecutionPlanMetricsSet, Label, MetricBuilder};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::metrics::Count;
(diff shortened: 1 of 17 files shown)


src/coordinator/query_coordinator.rs
```diff
@@ -278,6 +293,9 @@ impl<'a> StageCoordinator<'a> {
                             store.insert(task_key, filters);
                         }
                     }
+                    // Runtime dynamic-filter reports are accepted by this transport change. A
+                    // later change in the stack will retain and merge them.
+                    WorkerToCoordinatorMsg::ProducedDynamicFilter(_) => {}
```
jayshrivastava, line 295: In this PR, we do nothing when receiving dynamic filter updates. In later PRs, we will consume these updates.

src/worker/impl_coordinator_channel.rs
```diff
@@ -231,6 +263,58 @@ impl Worker {
...
+            let dynamic_filter = expression
+                .downcast_ref::<DynamicFilterPhysicalExpr>()
+                .expect("producer discovery returns DynamicFilterPhysicalExpr");
+
+            loop {
+                // `wait_update()` uses a Tokio watch channel, so multiple generations can
+                // naturally coalesce while this observer is busy. The stream promises the latest
+                // observed state rather than one message per generation; completion is awaited
+                // separately because it does not advance the generation.
+                let completed = tokio::select! {
```
jayshrivastava: This PR introduces some overhead. We forward these messages even though they aren't used yet. They will be used in future PRs.

jayshrivastava:
benchmarks run tpch/sf100

gabot-0:
Requested by [this comment](https://github.com/datafusion-contrib/datafusion-distributed/pull/635#issuecomment-5595247422).

Benchmark job 51 failed for `tpch/sf100`. Full details are available in the controller journal.

src/coordinator/query_coordinator.rs
```diff
@@ -126,6 +129,12 @@ impl QueryCoordinator {
 /// - Building tasks that communicate a serialized plan to multiple workers for further execution.
 /// - Building tasks that stream partition feeds from local [WorkUnitFeedExec] nodes to their
 ///   remote counterparts.
+struct SpecializedTaskPlan {
+    plan: Arc<dyn ExecutionPlan>,
+    work_unit_feed_declarations: Vec<WorkUnitFeedDeclaration>,
+    dynamic_filter_remote_producer_ids: Vec<u64>,
+}
+
```
gabotechs, line 137: For this struct, It might be worth to:
- Move it below in this file, so that the comment above still applies to `StageCoordinator`
- Name it `TaskSpacializedPlan` to match the method's name

src/coordinator/query_coordinator.rs
```diff
@@ -488,6 +528,75 @@ impl CoordinatorToWorkerMetrics {
...
+        let remote_filter = dynamic_filter();
+        let repartition = Arc::new(RepartitionExec::try_new(
+            empty(&schema),
+            Partitioning::Hash(vec![Arc::new(Column::new("a", 0))], 1),
+        )?) as Arc<dyn ExecutionPlan>;
+        let remote_probe = Arc::new(
+            NetworkShuffleExec::try_new(repartition, 1)?
+                .with_dynamic_filter_anchors(vec![remote_filter.clone()]),
+        ) as Arc<dyn ExecutionPlan>;
```
gabotechs, line 560: I see this being a pattern with the tests: plans get constructed manually by chaining operators together. 

It's pretty hard to see what's happening and what's getting chained with what to a human eye. Do you think there's a chance we can just use normal SQL for these tests? 

The main reason for having the weather and flights datasets committed to the codebase is so that we can just use SQL for building plans, rather than constructing them manually.

src/worker/impl_coordinator_channel.rs
```diff
@@ -272,3 +333,72 @@ fn send_metrics_via_channel(
...
+    async fn streams_dynamic_filter_updates_and_completion() -> Result<()> {
+        let dynamic_filter = Arc::new(DynamicFilterPhysicalExpr::new(
+            vec![Arc::new(Column::new("a", 0))],
+            lit(true),
+        ));
+        let expression_id = dynamic_filter.expression_id().unwrap();
+        let expression = Arc::clone(&dynamic_filter) as Arc<dyn PhysicalExpr>;
+        let (_cancel_tx, cancel_rx) = watch::channel(false);
+        let task_ctx = SessionContext::new().task_ctx();
+        let mut stream = produced_dynamic_filter_stream(expression_id, expression, cancel_rx);
```
gabotechs, line 354: I really think we should change the approach to testing in general in this PR.

The `produced_dynamic_filter_stream` is a private function of this module, and therefore an implementation detail subject to change. By testing at this level, we are testing the implementation detail rather than the overall functionality, and the tests tend to be very verbose.

Here's one idea for change the way we approach testing in this PR: for anything related to dynamic filter collection and over-the-wire transfer, we can add some nice DataFusion metrics at the coordinator level, and during integration testing, we can perform assertion on those metrics. 

With that, we'd win two things:
- Reliable tests that survive changes to the implementation details.
- Runtime metrics that give visibility around what happened.

src/coordinator/query_coordinator.rs
jayshrivastava: Done.

src/coordinator/query_coordinator.rs
jayshrivastava: Done. I migrated the function `dynamic_filter_remote_producer_ids` to `src/dynamic_filtering/discovery.rs` and we now test it in the snapshots in `tests/dynamic_filtering/discovery.rs`

src/worker/impl_coordinator_channel.rs
jayshrivastava: I added a metric to track how many updates are received. I think this is good enough for now, we just assert that they are nonzero. Eventually, once remote dynamic filtering works, we will assert that the filters show up in plans and that the results match with dyn filtering on vs off.

jayshrivastava:
This is ready for another review.

src/worker/impl_coordinator_channel.rs
```diff
@@ -285,3 +346,24 @@ fn send_metrics_via_channel(
...
     });
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::physical_expr::expressions::{Column, lit};
+
+    #[tokio::test]
+    async fn cancellation_stops_dynamic_filter_updates() {
```
gabotechs: This test is essentially just testing that `tokio::select!` inside a private function works. This is one of type of tests that is subject to change as implementation details change.

A more complete coverage could be to run a relatively long query (e.g. TPCH Q1), abort it mid-way, and make sure the query ends immediately, without dyn filters getting any update.

tests/dynamic_filtering/common.rs
```diff
@@ -146,6 +156,7 @@ async fn execute_query_and_display(
     sql: &str,
     expected_rows: usize,
     collect_dynamic_filters: bool,
+    expect_dynamic_filter_updates: Option<bool>,
```
gabotechs: Why would this function accept `None` instead of just a `false`?

gabotechs (approved):
:+1: nice! nothing major

src/worker/impl_coordinator_channel.rs
jayshrivastava: I guess what we want is to assert the streams all close (which we don't today). In `tests/stateful_data_cleanup.rs`, we only test that the task data entries are evicted. I'm going to add a little test-only metric that let's us observe streams getting cleaned up.

PR #636, opened 2026-08-13
coordinator: merge partial dynamic filters 

jayshrivastava:
## Stack

4. https://github.com/datafusion-contrib/datafusion-distributed/pull/636 <- you are here

## Details

This PR adds machinery around merging dynamic filters in the `DynamicFilterRegistry`. 

```
    register_task(stage1, task1)                                                                                              
    register_task(stage1, task2)       ───────────┐                                                                              
    register_task(stage1, task3)                  │          ┌───────────────────────┐      merge() when                         
         seal_stage(stage1)                       ├─────────▶│ DynamicFilterRegistry │───▶  - stage is sealed; and               
                                                  │          └───────────────────────┘      - there's enough partial filter      
                                                  │                                         updates                              
record_dynamic_filter_update(stage1, task1)   ────┘                                                                              
record_dynamic_filter_update(stage2, task1)                                                                                           
record_dynamic_filter_update(stage3, task1)                                                                                           
                                                                                                                              
```

The query coordinator calls `register_task` for each task in a stage. Concurrently, any running task from any stage can send a dynamic filter update to the registry via `record_dynamic_filter_update`. The registry needs to detect when all the updates are present and `merge()` the partial dynamic filters. To help detect this, the coordinator is responsible for calling `seal_stage` once all the tasks have commenced so we know that no tasks will be added in the future.

The PR implements the above. In the next 2 PRs, we will actually forward the merged filters to consumers.


src/coordinator/dynamic_filter_registry.rs
```diff
@@ -1,27 +1,55 @@
...
+    pub(super) merge_mode: Option<DynamicFilterMergeMode>,
     // Producer and consumer tasks for a dynamic filter.
     //
     // Note that it is not guaranteed that every task within a stage produces / consumes dynamic filters. For
     // example, a distributed union may prevent a dynamic filter from appearing in all tasks. So, we
     // store task keys rather than stage ids.
     pub(super) producer_tasks: HashSet<TaskKey>,
     pub(super) consumer_tasks: HashSet<TaskKey>,
+    /// Latest accepted snapshot from each producer task.
+    pub(super) producer_filters: HashMap<TaskKey, PhysicalDynamicFilterNode>,
```
gabotechs, line 43: Isn't this `producer_filters` redundant with `producer_tasks`? I imagine the keys present in both need to be exactly the same.

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -54,19 +82,65 @@ impl DynamicFilterRegistry {
...
+            // at the coordinator before being applied.
+            let merge_mode = if node
+                .downcast_ref::<HashJoinExec>()
+                .is_some_and(|join| matches!(join.partition_mode(), PartitionMode::CollectLeft))
+            {
+                DynamicFilterMergeMode::FirstProducerComplete
+            } else if node.downcast_ref::<SortExec>().is_some()
+                || node.downcast_ref::<AggregateExec>().is_some()
+            {
+                DynamicFilterMergeMode::Incremental
```
gabotechs, line 104: Just a bit shorter:

```suggestion
            } else if node.is::<SortExec>() || node.is::<AggregateExec>() {
```

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -54,19 +82,65 @@ impl DynamicFilterRegistry {
         plan: &Arc<dyn ExecutionPlan>,
         task_key: TaskKey,
     ) -> Result<()> {
-        let producers = discover_dynamic_filter_producers(plan)?;
+        let mut producers = vec![];
+
+        plan.apply(|node| {
```
gabotechs, line 87: :thinking: here, this is registering the plan for a specific task_key, however, we are recursing the full plan, even for unreachable paths (e.g. ChildrenIsolatorUnionExec).

Should we be using `apply_with_dt_ctx` instead? it's similar, but it omits recursing unreachable branches.

gabotechs: A wait! this is already the task-specialized plan. Then my suggestion is to just rename the argument from `plan` to `task_specialized_plan`.

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -1,27 +1,55 @@
...
     //
     // Note that it is not guaranteed that every task within a stage produces / consumes dynamic filters. For
     // example, a distributed union may prevent a dynamic filter from appearing in all tasks. So, we
     // store task keys rather than stage ids.
     pub(super) producer_tasks: HashSet<TaskKey>,
     pub(super) consumer_tasks: HashSet<TaskKey>,
+    /// Latest accepted snapshot from each producer task.
+    pub(super) producer_filters: HashMap<TaskKey, PhysicalDynamicFilterNode>,
+    /// Full dynamic filter containing the merged predicate and its completion state.
+    pub(super) merged: Option<PhysicalDynamicFilterNode>,
```
gabotechs, line 45: I see this is working in terms of protobuf messages rather than raw DataFusion expressions.

This means that proto conversion would happen even in a fully in-memory setup. Is it easy to avoid and just work with normal DataFusion types here rather than protobuf messages?

gabotechs, line 45: I see this is working in terms of protobuf messages rather than raw DataFusion expressions.

This means that proto conversion would happen even in a fully in-memory setup. Is it easy to avoid and just work with normal DataFusion types here rather than protobuf messages?

gabotechs (commented):
Looking good! left some small comments, nothing structural.

src/coordinator/dynamic_filter_registry.rs
```diff
@@ -78,4 +152,126 @@ impl DynamicFilterRegistry {
...
+        {
+            return;
+        }
+        filter.producer_filters.insert(task_key, *dynamic_filter);
+        Self::try_merge(&mut state, report.expression_id);
+    }
+
+    /// Merges partial dynamic filters together for the provided dynamic filter
+    /// id only if there are enough updates present.
+    fn try_merge(state: &mut DynamicFilterRegistryState, id: u64) -> bool {
```
gabotechs, line 207: By convention, all methods starting with `try_` return `Result<>`. As this method is infallible, what about calling it just `merge()`, or `attempt_merge()`

jayshrivastava: Done

src/coordinator/dynamic_filter_registry.rs
jayshrivastava: Yup! Renamed.

src/coordinator/dynamic_filter_registry.rs
jayshrivastava: Done

src/coordinator/dynamic_filter_registry.rs
jayshrivastava: The main reason is honestly that `is_complete` isn't public on `DynamicFilterPhysicalExpr`. So there's actually no way to know if it's complete or not here.

We could detect it on the worker [here](https://github.com/datafusion-contrib/datafusion-distributed/blob/fa0a1a203dcceab489e010d0ad3e8e5afcbcd05c/src/worker/impl_coordinator_channel.rs?plain=1#L322) and send a `is_complete` bit in the `ProducedDynamicFilter. Do you think making that change is worth it?

src/coordinator/dynamic_filter_registry.rs
jayshrivastava: They were slightly different but can be represented as one struct. `producer_tasks` is the static list of tasks which we expect producer updates from. `producer_filters` contained the latest update for each filter.

I changed it to just `pub(super) producers: HashMap<TaskKey, Option<PhysicalDynamicFilterNode>>`. We populate the value as updates come in.

Issue #629, opened 2026-08-13
[EPIC] Add other benchmarking datasets


gabotechs:
Right now we can run benchmarks on three different datasets:
- TPCH
- TPC-DS
- ClickBench

However, it would be cool to add some more, for example:

- Variants of existing benchmark, like sorted ClickBench or sorted TPCH.
- Other datasets, like h2o or imdb.

We have good foundations here for holding more datasets easily, the process typically is:
1. Create the Rust code that prepares the dataset in https://github.com/datafusion-contrib/datafusion-distributed/tree/main/benchmarks/src under a `prepare_<dataset>.rs` name.
2. Create the .sh script that calls the Rust code with the appropriate config and generates the dataset in https://github.com/datafusion-contrib/datafusion-distributed/tree/main/benchmarks under a `gen-<dataset>.sh` name, under different scale factors or configs: 
    - If the dataset allows scale factors, the dataset should be generated in testdata/<dataset>/sf<sf>/...all dirs with parquet tables...
    - If not, we can just do testdata/<dataset>/...all dirs with parquet tables...
3. Place the SQL queries used for benchmarks in https://github.com/datafusion-contrib/datafusion-distributed/tree/main/testdata under `<dataset>/queries/q<number>.sql`
4. Wire up the new dataset in https://github.com/datafusion-contrib/datafusion-distributed/blob/main/benchmarks/src/run.rs (`queries_for_dataset`)

The result is that running a command like this:

```
WORKERS=8 ./benchmarks/run.sh --threads 2 --dataset <dataset>[/<sf100>]
```

Should end up benchmarking the new dataset

The task is big and will need to be broken down in several sub issues that people can contribute to independently:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/631
- https://github.com/datafusion-contrib/datafusion-distributed/issues/633
- https://github.com/datafusion-contrib/datafusion-distributed/issues/632

shinzoxD:
Taking the remaining dataset from this epic (h2o groupby; not 631/632/633). I'll add prepare/gen scripts, the 10 official groupby queries, and wire it up so `./benchmarks/run.sh --dataset h2o/sf<scale-factor>` works.

shinzoxD:
[...]

PR #646, opened 2026-08-15
feat: add a sorted TPC-H benchmark dataset

### LINKED ISSUE #632: [Benchmark datasets] Add a sorted TPC-H benchmark dataset
Part of #629.

Add a deterministic TPC-H dataset variant whose Parquet tables are sorted by explicitly documented key columns. This should make physical data ordering a controlled benchmark dimension while retaining the standard TPC-H query workload.

Scope:
- Add the Rust preparation code and `gen-tpch-sorted.sh` generation entry point.
- Support the same scale-factor model as TPC-H and write data below `testdata/tpch-sorted/sf<scale-factor>`.

For a more detailed explanation of how to wire this up, see https://github.com/datafusion-contrib/datafusion-distributed/issues/629

shinzoxD:
## Summary

Adds a deterministic `tpch/sorted_sf1` benchmark dataset variant so physical data ordering is a controlled dimension while keeping the standard TPC-H query workload. Part of #629, closes #632.

Uses the same `tpchgen` generators and Arrow writer as unsorted TPC-H, recording table-specific Parquet `sorting_columns` metadata in one pass (`prepare-tpch --sorted`).

## Sort keys

| Table | Sort columns |
| --- | --- |
| region | `r_regionkey` |
| nation | `n_nationkey` |
| customer | `c_custkey` |
| supplier | `s_suppkey` |
| part | `p_partkey` |
| partsupp | `ps_partkey` |
| orders | `o_orderkey` |
| lineitem | `l_orderkey`, `l_linenumber` |

`partsupp` records only `ps_partkey` because `ps_suppkey` wraps within each part.

Data is written to `testdata/tpch/sorted_sf<scale-factor>/`.

```bash
# Default SF1, 16 files per table
./benchmarks/gen-tpch-sorted.sh

# Tiny smoke-scale dataset
SCALE_FACTOR=0.01 ./benchmarks/gen-tpch-sorted.sh

WORKERS=8 ./benchmarks/run.sh --threads 2 --dataset tpch/sorted_sf1
```

Queries stay in `testdata/tpch/queries`. `run_local` extracts the `tpch` suite from `tpch/sorted_sf1`, so `queries_for_dataset` is unchanged. `register_tables` is unchanged.

## Implementation

- `benchmarks/src/datasets/tpch.rs`: optional `sorting_columns` on the existing writer
- `prepare-tpch --sorted` instead of a separate async rewrite command
- `benchmarks/gen-tpch-sorted.sh` writes `testdata/tpch/sorted_sf<scale-factor>`

SF1 was not generated in this PR (heavy, and blocked here by the OpenSSL build). Use `SCALE_FACTOR=0.01` to smoke the path, then `./benchmarks/gen-tpch-sorted.sh` for SF1.


shinzoxD:
CI was failing while compiling `datafusion-distributed-benchmarks`:

```
error[E0382]: use of moved value: `df`
--> benchmarks/src/datasets/tpch_sorted.rs:163:10
    let mut stream = df.execute_stream().await?;
    drop(df);  // used after move
```

`DataFrame::execute_stream` takes `self`, so the follow-up `drop(df)` was a use-after-move. Clippy also treated the unused `std::sync::Arc` import as an error under `-D warnings`.

Fixed in 077e2242f95645413597e14df97fc735b0ec0618: drop the finished stream (and session) so file handles are released before the table directory is replaced, and move the unused import plus test-only sort helpers into the test module.

benchmarks/src/datasets/tpch_sorted.rs
```diff
@@ -0,0 +1,473 @@
...
+///
+/// Files are written under `data_dir/<table>/<part>.parquet`, matching the
+/// unsorted TPC-H layout.
+pub async fn generate_tpch_sorted_data(
+    data_dir: &Path,
+    sf: f64,
+    parts: usize,
+) -> Result<(), Box<dyn std::error::Error>> {
+    tpch::generate_tpch_data(data_dir, sf, parts)?;
+    rewrite_tables_globally_sorted(data_dir).await
```
gabotechs: tpchgen-rs already emits lineitem in strict (l_orderkey, l_linenumber) order across contiguous numeric part ranges; I verified 120,515 rows across the 16 generated files with no violations. Apache DataFusion’s sorted TopK benchmark reuses standard TPC-H files and attaches ordering metadata at registration time instead of rewriting them. Please follow that model: add an opt-in sorted-registration path for tpch/sf... and a small generator-order test, rather than maintaining a duplicated, re-sorted dataset. Limit the initial declared ordering to lineitem unless each additional table’s generator order is separately verified.

gabotechs: This extra read/sort/rewrite pass should not be needed. It should be enough to implement the sorted layout as a small variant of the existing `datasets/tpch.rs`: reuse the same tpchgen generators and Arrow writer, passing table-specific `SortingColumn`s to `WriterProperties` when each Parquet file is created. I verified that the pinned tpchgen emits lineitem ordered by `(l_orderkey, l_linenumber)` and orders/customer/supplier/part by their generated keys. For partsupp, only `ps_partkey` is naturally ordered because `ps_suppkey` wraps, so declare only that prefix. The existing `prepare-tpch` command can take a sorted-variant flag instead of adding this rewrite implementation and a separate asynchronous preparation path.

benchmarks/src/run.rs
```diff
@@ -131,7 +131,7 @@ pub struct RunOpt {
 
 fn queries_for_dataset(dataset: &str) -> Result<Vec<(String, String)>, DataFusionError> {
     match dataset {
-        "tpch" => tpch::get_queries()
+        "tpch" | "tpch-sorted" => tpch::get_queries()
```
gabotechs: We can model this as the TPC-H variant `tpch/sorted_sf1`, rather than a new top-level `tpch-sorted` dataset. `run_local` already extracts `tpch` as the suite before calling `queries_for_dataset`, and `dataset_path` resolves the suffix to `testdata/tpch/sorted_sf1`, so the existing `"tpch"` branch works unchanged. The generation script should write to that location and this additional match arm can be removed.

benchmarks/src/datasets/tpch_sorted.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+fn count_parquet_files(dir: &Path) -> Result<usize, Box<dyn std::error::Error>> {
+    Ok(fs::read_dir(dir)?
+        .filter_map(|e| e.ok())
+        .filter(|e| e.path().extension().and_then(|ext| ext.to_str()) == Some("parquet"))
+        .count())
+}
+
+#[cfg(test)]
```
gabotechs: We can remove this test module, including its test-only helpers and imports. This is benchmark data-preparation code rather than production code, and these tests add roughly 240 lines of maintenance overhead to what should be a small Parquet-metadata variant of the existing TPC-H generator.

shinzoxD: Implemented in 26040dc. Removed the test module, helpers, and `tpch_sorted.rs`.

benchmarks/src/datasets/tpch_sorted.rs
shinzoxD: Implemented in 26040dc. Sorted TPC-H now reuses the existing tpchgen generators and Arrow writer in `datasets/tpch.rs`, writing table-specific `SortingColumn` metadata in one pass. `partsupp` records only `ps_partkey`. `--sorted` is a flag on `prepare-tpch`; the rewrite path and `prepare-tpch-sorted` command are gone.

benchmarks/src/run.rs
shinzoxD: Implemented in 26040dc. The variant is now `tpch/sorted_sf1`, resolved to `testdata/tpch/sorted_sf1`. `queries_for_dataset` keeps the existing `"tpch"` branch; `gen-tpch-sorted.sh` writes to that path.

benchmarks/gen-tpch-sorted.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCALE_FACTOR=${SCALE_FACTOR:-1}
```
gabotechs: It should be fine to use reuse the existing `gen-tpch.sh` script with a simple `SORTED={true|false}` flag populated by an env variable.

shinzoxD: Implemented in f31267d. Sorted generation is now `SORTED=true ./gen-tpch.sh`; `gen-tpch-sorted.sh` is removed. Default remains unsorted (`SORTED=false`).

gabotechs (approved):
This one looks good. Thanks @shinzoxD!

PR #648, opened 2026-08-15
fix: preserve range partitions when scaling file scans

### LINKED ISSUE #628: tests and benchmarks on range partitioned data
See https://github.com/datafusion-contrib/datafusion-distributed/blob/a6cff424261623dbdd1e2b5cbe83e5273ebc22de/tests/join.rs?plain=1#L120

It's not valid in datafusion to claim your data source is distributed by hash(expr), so this test isn't useful. We should migrate it to a distribution which maps to range partitioning.

There's several open questions that can only really be solved with tests
- `file_scan_config_scale_up_leaf_node` in `src/events/defaults/file_scan_config.rs` if the file scan is claiming range partitioning and we rearrange files, that can cause correctness issues right? Should we call `FileScanConfig::repartitioned`
- Say target_partitions is `10` and our data source has `100` range partitions. Does this distribute correctly to 10 leaf tasks? Does datafusion introduce range partitions? 
- Are all instances of `scale_partitioning` correct when it comes to range partitioned data? https://github.com/datafusion-contrib/datafusion-distributed/blob/a6cff424261623dbdd1e2b5cbe83e5273ebc22de/src/execution_plans/common.rs?plain=1#L17

shinzoxD:
﻿## Summary

Fixes #628. Depends on #540 (this branch is based on `branch-55` because `Partitioning::Range` does not exist on DataFusion 54).

A listing scan cannot honestly claim `Hash(expr)`. The join suite now registers the hive-layout testdata with declared `Range` output partitioning, so the optimizer can skip shuffles for a real distribution instead of an invalid hash claim.

The tests also answer the questions on that issue:

1. **Rearranging range file groups is incorrect.** `FileScanConfig::repartitioned` already returns `None` when `output_partitioning` is set. `file_scan_config_scale_up_leaf_node` was flattening files and round-robin assigning them, which moved e.g. partition 2 into slot 1. It now keeps group `i` as partition `i` and gives each group to a single task.
2. **`target_partitions` does not collapse range partitions.** With 8 range partitions and `target_partitions=2`, DataFusion still builds 8 file groups. Scaling to 2 leaf tasks keeps those 8 slots; even partitions go to task 0 and odd ones to task 1. The plan stays `SinglePartitioned` on the range key (no extra shuffle).
3. **`scale_partitioning` cannot change a Range's partition count.** The count is the split-point width. Identity scaling keeps `Range`; `NetworkCoalesceExec` multiplying by input tasks degrades to `UnknownPartitioning`. Stage display treats Range like Hash (shared partition indices across tasks).

## Validation

```bash
cargo test -p datafusion-distributed --lib scale_
cargo test -p datafusion-distributed --lib events::defaults::file_scan_config
cargo test -p datafusion-distributed --features integration --test join --test range_partitioned
```

### DIFF
--- a/src/events/defaults/file_scan_config.rs
+++ b/src/events/defaults/file_scan_config.rs
@@ -40,33 +40,38 @@ pub(crate) fn file_scan_config_scale_up_leaf_node(
     let file_scan = dse.data_source().downcast_ref::<FileScanConfig>()?;
     let partition_count = ev.plan.output_partitioning().partition_count();
 
-    let rebalanced = if file_scan.output_partitioning.is_some() {
-        let all_partitioned_files = file_scan
-            .file_groups
-            .iter()
-            .flat_map(|file_group| file_group.iter().cloned())
-            .collect::<Vec<_>>();
-        rebalance_round_robin(all_partitioned_files, partition_count * ev.task_count)
+    let file_scans = if file_scan.output_partitioning.is_some() {
+        // File groups are the declared hash/range partitions. Moving a file to a
+        // different group would invalidate that mapping — DataFusion's
+        // FileScanConfig::repartitioned also refuses to do so. Keep group i as
+        // partition i on every task, and assign each group to a single task so
+        // co-partitioned joins and single-partitioned aggregates stay correct.
+        scale_declared_partition_file_groups(&file_scan.file_groups, ev.task_count)
             .into_iter()
-            .map(FileGroup::new)
+            .map(|file_groups| {
+                let mut cfg = file_scan.clone();
+                cfg.file_groups = file_groups;
+                cfg
+            })
             .collect::<Vec<_>>()
     } else {
-        FileGroupPartitioner::new()
+        let rebalanced = FileGroupPartitioner::new()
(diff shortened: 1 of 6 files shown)


shinzoxD:
Clippy failed on `tests/range_partitioned.rs` (`clippy::ptr_arg`: `&PathBuf` instead of `&Path`).

Pushed 8c6d2e3: `register_range_table` now takes `&Path`. Format-check and tpch-plans-test were already green on the previous SHA; waiting on the new CI run for clippy/unit-test and the remaining integration jobs.

gabotechs:
@shinzoxD I see you opened several PRs while there were already some in-flight. These PRs look AI generated without supervision, as the approach to most of them shows several questionable traits of one-off behaviors, unnecessary special cases and unnecessarily verbose testing approach.

In order to allow other people to contribute, I'm going to close without a review some of your PRs, leaving just the first 3 you opened. Please, upon investing time in this project, prefer quality over quantity, and prefer investing time in getting just a small number of PRs up to a high standard, rather than a big quantity of purely AI generated PRs.

As always, thanks for your contributions, and see you in the other PRs!

PR #650, opened 2026-08-15
feat: add h2o groupby benchmark dataset

shinzoxD:
﻿## Summary

Adds the [h2oai/db-benchmark](https://github.com/h2oai/db-benchmark) groupby dataset so physical aggregation is a first-class benchmark suite alongside TPC-H, TPC-DS, and ClickBench. Part of #629.

This is the remaining dataset called out in that epic after the sorted TPC-H / sorted ClickBench / TPC-DS scale-factor issues.

## Layout

SF1 is 10 million rows, matching official `G1_1e7_1e2_0_0` (N=1e7, K=100, no NAs, unsorted). Data is written to `testdata/h2o/sf<scale-factor>/x/`.

The table schema and column distributions follow [groupby-datagen.R](https://github.com/h2oai/db-benchmark/blob/master/_data/groupby-datagen.R): `id1`/`id2`/`id4`/`id5` have K distinct values, `id3`/`id6` have N/K, `v1` is in `[1, 5]`, `v2` is in `[1, 15]`, and `v3` is `round(U(0, 100), 6)`. Generation uses the published seed `108`.

Queries are the ten official groupby statements, in `testdata/h2o/queries/`. Join and window variants are not included.

```bash
# Default SF1, 16 parquet files
./benchmarks/gen-h2o.sh

# Tiny smoke-scale dataset (100k rows)
SCALE_FACTOR=0.01 ./benchmarks/gen-h2o.sh

# Official "medium" size (100 million rows)
SCALE_FACTOR=10 ./benchmarks/gen-h2o.sh

WORKERS=8 ./benchmarks/run.sh --threads 2 --dataset h2o/sf1
```

## Implementation

- `benchmarks/src/datasets/h2o.rs`: generate the `x` table and load the query files
- `benchmarks/src/prepare_h2o.rs` + `prepare-h2o` CLI
- `benchmarks/gen-h2o.sh`
- `queries_for_dataset` treats `h2o/sf<scale-factor>` as the h2o suite
- `register_tables` is unchanged (`x/` is registered as parquet table `x`)

## Tests / verification

- Unit tests in `h2o.rs`: scale-factor mapping, partitioned write (row count + column ranges), and that generated data answers groupby q1
- `cargo fmt --all`
- Full `cargo test -p datafusion-distributed-benchmarks --lib datasets::h2o` did not complete locally: vendored `openssl-sys` fails to build on this Windows machine (Git perl is missing `Locale::Maketext::Simple`). CI on Ubuntu should run the crate tests if/when they are invoked; the new tests compile as part of the benchmarks lib

Do not generate a full SF1 in review unless needed. `SCALE_FACTOR=0.01 ./benchmarks/gen-h2o.sh` is the cheap smoke.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,8 @@ testdata/tpcds/*
 !testdata/tpcds/README.md
 testdata/clickbench/*
 !testdata/clickbench/queries
+testdata/h2o/*
+!testdata/h2o/queries
+!testdata/h2o/README.md
 codegen/target/*
 codegen/Cargo.lock
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2167,6 +2167,7 @@ dependencies = [
  "object_store",
  "openssl",
  "parquet",
+ "rand 0.9.4",
  "reqwest",
  "serde",
  "serde_json",
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -31,6 +31,7 @@ sketches-ddsketch = "0.3"
 object_store = { version = "0.13", features = ["aws"] }
 openssl = { version = "0.10", features = ["vendored"] }
 mimalloc = "0.1"
+rand = "0.9"
 
 [dev-dependencies]
 criterion = "0.5"
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -15,6 +15,9 @@ For example, `tpch/sf1` is stored in `testdata/tpch/sf1`.
 
 # TPC-DS (only SCALE_FACTOR=1 is supported)
 ./gen-tpcds.sh
+
+# h2o groupby (default: SCALE_FACTOR=1 = 10 million rows)
+./gen-h2o.sh
 ```
 
 ### Running Benchmarks in single-node mode
--- a/benchmarks/gen-h2o.sh
+++ b/benchmarks/gen-h2o.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+SCALE_FACTOR=${SCALE_FACTOR:-1}
+PARTITIONS=${PARTITIONS:-16}
+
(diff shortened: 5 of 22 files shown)


shinzoxD:
Pushed a clippy fix to this branch (`32dae45`).

**clippy** (`neg-cmp-op-on-partial-ord` at `benchmarks/src/datasets/h2o.rs`):
- Replaced `!(sf > 0.0)` with `!sf.is_finite() || sf <= 0.0`
- Same error message; also rejects NaN/Inf up front instead of relying on a negated `PartialOrd` comparison

**unit-test** (`tests::build_side_fetch_is_preserved_by_broadcast` in `tests/multi_task_collect_join_repros.rs`):
- Failed with result-set sizes 3750 vs 2500
- Pre-existing flake on main, already tracked/fixed by #624
- This PR does not touch that file; CI re-run after this push will pick it up. If it flakes again, leave it.

shinzoxD:
Clippy is green on `32dae45` (`neg-cmp-op-on-partial-ord` in `rows_for_scale_factor`).

`unit-test` failed again on the same pre-existing main flake, not this PR:

```
tests::build_side_fetch_is_preserved_by_broadcast
  left (distributed): 3750
  right (single-node): 2500
```

This PR does not touch `tests/multi_task_collect_join_repros.rs` or the broadcast planner. The 3750-vs-2500 mismatch is the global-fetch bug #624 is fixing (LIMIT 50 applied per producer task instead of once). Re-running the failed job; if it flakes again we will leave it to #624 rather than mix planner changes into this dataset PR.

gabotechs:
@shinzoxD I see you opened several PRs while there were already some in-flight. These PRs look AI generated without supervision, as the approach to most of them shows several questionable traits of one-off behaviors, unnecessary special cases and unnecessarily verbose testing approach.

In order to allow other people to contribute, I'm going to close without a review some of your PRs, leaving just the first 3 you opened. Please, upon investing time in this project, prefer quality over quantity, and prefer investing time in getting just a small number of PRs up to a high standard, rather than a big quantity of purely AI generated PRs.

As always, thanks for your contributions, and see you in the other PRs!

PR #653, opened 2026-08-16
gabrielmusat/dynamic-stage-built-events-2

gabotechs:
Adds DynamicStageBuiltHandler, called whenever dynamic planning builds a stage. A handler can inspect cost and the outgoing plan, rewrite it, or abort execution.

config.with_distributed_dynamic_stage_built_handler(|event| {
    Some(Ok(DynamicStageBuiltEventResponse::new(
        Arc::clone(event.plan),
    )))
});

### DIFF
--- a/src/events/dynamic_stage_built.rs
+++ b/src/events/dynamic_stage_built.rs
@@ -1,0 +1,67 @@
+use crate::events::common::EventHandlerChain;
+use datafusion::common::{Result, stats::Precision};
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::prelude::SessionConfig;
+use std::sync::Arc;
+
+/// Estimated cost associated to the fragment of the plan. Calculated based on the amount of data
+/// that will flow through each node and the static cpu, memory and network complexity of the
+/// different nodes based on the algorithms that will run inside them.
+#[derive(Default, Debug, Clone, Copy)]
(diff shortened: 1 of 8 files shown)


src/events/dynamic_stage_built.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub session_config: &'a SessionConfig,
+}
+
+pub struct DynamicStageBuiltEventResponse {
+    /// A potentially modified plan (e.g., optimizations applied).
+    pub plan: Arc<dyn ExecutionPlan>,
+}
+
+impl DynamicStageBuiltEventResponse {
+    pub fn new(plan: Arc<dyn ExecutionPlan>) -> Self {
```
gabotechs: The stage task count is computed before this response is applied (`prepare_dynamic_plan` computes it from the original plan, then later launches that many workers). A handler is documented as being able to return a modified plan, but a rewrite such as wrapping the plan in `CoalescePartitionsExec` can reduce its output partition count; workers past partition 0 will then fail with `partition … not available`. Either recompute/reconcile the task count after applying the response, or make this contract require the returned plan to preserve the stage topology and output partitioning, and enforce it.

gabotechs:
This was just a test for trying out the new review guidance, closing it.

PR #661, opened 2026-08-20
fix: look through ProjectionExec when pushing fetch

### LINKED ISSUE #659: Avoid redundant LocalLimitExec when fetch is separated by a ProjectionExec
## Problem

After the DataFusion 55 optimizer changes, a `ProjectionExec` can be placed above a fetch-bearing operator. `ProjectionExec` preserves row count, but the distributed fetch pushdown currently does not look through it. As a result, plans such as ClickBench q17 gain an unnecessary per-task `LocalLimitExec`.

## Desired behavior

When determining whether to propagate a fetch into a `NetworkCoalesceExec` input stage, traverse through a `ProjectionExec` and evaluate the fetch on its child. Preserve the existing handling for `GlobalLimitExec`, including its skip adjustment.

## Acceptance criteria

- A fetch separated by a `ProjectionExec` is handled without adding the redundant `LocalLimitExec`.
- Add a focused regression test for this plan shape.

## Context

Follow-up from https://github.com/datafusion-contrib/datafusion-distributed/pull/540#discussion_r3811037920

mehrdad3301:
## Summary
- When pushing a fetch into a `NetworkCoalesceExec` producer stage, walk through `ProjectionExec` (it does not change row count) and apply the bound to the child.
- Skip inserting a redundant `LocalLimitExec` when that child already has a sufficient fetch, for example `SortExec: TopK`.
- `GlobalLimitExec` still pushes `skip + fetch`. Unbounded producers such as ClickBench q17 still get a per-task `LocalLimitExec`.

Closes #659.

### DIFF
--- a/src/distributed_planner/push_fetch_into_network_coalesce.rs
+++ b/src/distributed_planner/push_fetch_into_network_coalesce.rs
@@ -108,11 +108,13 @@ fn fetch_required_from_child(plan: &Arc<dyn ExecutionPlan>) -> Option<usize> {
 mod tests {
     use std::sync::Arc;
 
-    use datafusion::arrow::datatypes::Schema;
+    use datafusion::arrow::datatypes::{DataType, Field, Schema};
+    use datafusion::physical_expr::expressions::Column;
     use datafusion::physical_plan::ExecutionPlan;
     use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
     use datafusion::physical_plan::empty::EmptyExec;
     use datafusion::physical_plan::limit::{GlobalLimitExec, LocalLimitExec};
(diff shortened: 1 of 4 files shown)


mehrdad3301:
Personally, I think it's better to drop the recursive function call and just repeat the logic. There are no real recursions here, and it makes things much easier to understand. Let me know what you think

gabotechs:
👍 sounds good

gabotechs (approved):
Everything looks good, no comments. Thanks @mehrdad3301!

gene-bordegaray (commented):
fwiw also a 👍

gene-bordegaray (approved):

PR #667, opened 2026-08-24
Add chaos test

gabotechs:
Adds an integration test that randomly introduces different sets of network errors, and expects the system to behave correctly with the appropriate retrying and re-scheduling mechanisms.

This just adds the integration test for qualifying further improvements, but the test is ignored as we still don't have a good mechanism for retries.

### DIFF
--- a/tests/chaos.rs
+++ b/tests/chaos.rs
@@ -1,0 +1,276 @@
+#[cfg(all(feature = "tpch", feature = "chaos-tests", test))]
+mod tests {
+    use datafusion::common::Result;
+    use datafusion::common::runtime::JoinSet;
+    use datafusion::error::DataFusionError;
+    use datafusion::execution::{SessionState, SessionStateBuilder};
+    use datafusion::prelude::SessionContext;
+    use datafusion_distributed::test_utils::localhost::LocalHostWorkerResolver;
+    use datafusion_distributed::{
+        DistributedExt, SessionStateBuilderExt, Worker, WorkerQueryContext,
+    };
+    use datafusion_distributed_benchmarks::datasets::{register_tables, tpch};
+    use moka::future::FutureExt;
+    use std::collections::VecDeque;
+    use std::convert::Infallible;
+    use std::fs;
+    use std::future::Future;
+    use std::ops::Range;
+    use std::path::Path;
+    use std::pin::Pin;
+    use std::sync::{Arc, Mutex};
+    use std::task::{Context, Poll};
+    use std::time::Duration;
+    use tokio::net::TcpListener;
+    use tokio::sync::{OnceCell, Semaphore};
+    use tonic::Status;
+    use tonic::body::Body;
+    use tonic::transport::Server;
+    use tower::{Layer, Service};
+    use url::Url;
+
+    const NUM_WORKERS: usize = 8;
+    const MAX_IN_FLIGHT: usize = 1000;
+    const TOTAL_QUERIES: usize = 100;
+    const CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE: Range<usize> = 1..4;
+    const CONCURRENT_CLIENTS: usize = 10;
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "Still no good retrying mechanism that make this test pass"]
+    async fn chaos() -> Result<()> {
+        let cfg = ChaosClusterConfig {
+            num_workers: NUM_WORKERS,
+            max_in_flight: MAX_IN_FLIGHT,
+        };
+        let (ctx, _guard) = chaos_localhost_cluster(cfg).await;
+        let data_dir = ensure_tpch_data().await;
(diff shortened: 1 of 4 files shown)


tests/chaos.rs
```diff
@@ -0,0 +1,276 @@
...
+        let (ctx, _guard) = chaos_localhost_cluster(cfg).await;
+        let data_dir = ensure_tpch_data().await;
+        register_tables(&ctx, &data_dir).await?;
+        let query = tpch::get_query("q20")?;
+
+        let mut queries = VecDeque::new();
+
+        for _ in 0..TOTAL_QUERIES {
+            let mut batch = vec![];
+            for _ in 0..rand::random_range(CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE) {
```
jayshrivastava: I think we should generate a seed and log it at the beginning of the test so we can reproduce deterministically.

tests/chaos.rs
```diff
@@ -0,0 +1,276 @@
...
+    #[ignore = "Still no good retrying mechanism that make this test pass"]
+    async fn chaos() -> Result<()> {
+        let cfg = ChaosClusterConfig {
+            num_workers: NUM_WORKERS,
+            max_in_flight: MAX_IN_FLIGHT,
+        };
+        let (ctx, _guard) = chaos_localhost_cluster(cfg).await;
+        let data_dir = ensure_tpch_data().await;
+        register_tables(&ctx, &data_dir).await?;
+        let query = tpch::get_query("q20")?;
```
jayshrivastava: Consider randomizing the query as well.

tests/chaos.rs
```diff
@@ -0,0 +1,276 @@
...
+    const CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE: Range<usize> = 1..4;
+    const CONCURRENT_CLIENTS: usize = 10;
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "Still no good retrying mechanism that make this test pass"]
+    async fn chaos() -> Result<()> {
+        let cfg = ChaosClusterConfig {
+            num_workers: NUM_WORKERS,
+            max_in_flight: MAX_IN_FLIGHT,
+        };
```
jayshrivastava, line 63: Why not put all these in the config?
```
  const NUM_WORKERS: usize = 8;
    const MAX_IN_FLIGHT: usize = 1000;
    const TOTAL_QUERIES: usize = 100;
    const CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE: Range<usize> = 1..4;
    const CONCURRENT_CLIENTS: usize = 10;
```

tests/chaos.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        for result in concurrent_clients.join_all().await {
+            result?;
+        }
+
+        let first = rx.recv().await.expect("No result returned");
+        let first = arrow::util::pretty::pretty_format_batches(&first)?;
+        while let Some(next_result) = rx.recv().await {
+            let next_result = arrow::util::pretty::pretty_format_batches(&next_result)?;
```
jayshrivastava: nit import `arrow::util::pretty::pretty_format_batches` at the top rather than fully qualifying

tests/chaos.rs
```diff
@@ -0,0 +1,276 @@
...
+                    ChaosAction::Delay(duration) => {
+                        tokio::time::sleep(duration).await;
+                        response.await
+                    }
+                    ChaosAction::Timeout(duration) => {
+                        tokio::time::sleep(duration).await;
+                        Ok(Status::deadline_exceeded("chaos: worker timed out").into_http())
+                    }
+                    ChaosAction::Unavailable(duration) => {
+                        tokio::time::sleep(duration).await;
```
jayshrivastava, line 192: import at top

jayshrivastava: Actually can you add this to the https://github.com/datafusion-contrib/datafusion-distributed/blob/main/AGENTS.md? That we should import at the top rather than using full qualifiers?

jayshrivastava (approved):
LGTM but please take a look at the comments before merging

tests/chaos.rs
gabotechs: This is just the cluster config, and holds just the values relevant for the localhost servers.

The other constants are client-side config, and they are not relevant for the cluster setup.

tests/chaos.rs
gabotechs: I actually did this one on purpose. I want to make clear that this is `tokio`'s sleep, not `std`'s sleep. All the `tokio` sleeps in this project are actually fully qualified.

gabotechs: Added to the AGENTS.md

tests/chaos.rs
gabotechs: Given that this uses multiple threads concurrently, I don't think we can ever do this deterministically.

I added this, but results are still not fully deterministic because of this.

Issue #608, opened 2026-08-09
[Iceberg] Populate planning statistics from table metadata


gabotechs:
Part of #599.

Populate planning-time statistics for `IcebergDataSource` from the Iceberg table and selected snapshot.

Essentially, implementing this TODO comment in the code:

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/data_source.rs#L252-L255

The snapshot summary and file metadata may provide row-count and byte-size estimates without planning all scan tasks.

gabotechs:
Addressed in https://github.com/datafusion-contrib/datafusion-distributed/pull/676.

Opened this as a follow up:
- https://github.com/datafusion-contrib/datafusion-distributed/issues/681

gabotechs:
Thanks for addressing this @sandugood!

PR #676, opened 2026-08-25
feat[iceberg]: runtime statistics

sandugood:
Tackles #608.

Added runtime statistics that can be extracted from the current snapshot metadata.

`total-records` and `total-files-size` which are being used in the `partition_statistics` function.


sandugood:
Seems like CI failure is not related to the current PR scope

iceberg/src/data_source.rs
```diff
@@ -297,3 +300,98 @@ impl DataSource for IcebergDataSource {
...
+        .unwrap_or(0);
+    let total_byte_size = props
+        .get("total-files-size")
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(0);
+
+    Ok(Arc::new(Statistics {
+        num_rows: Precision::Exact(num_rows),
+        total_byte_size: Precision::Exact(total_byte_size),
+        column_statistics: vec![],
```
gabotechs: This is the reason why the CI is failing (https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/32901579585/job/97976700594?pr=676).

The schema might declare, for example, 3 columns, but here we are returning a `column_statistics` vec of 0 columns, and DataFusion, while propagating statistics to upstream nodes, will try to access index 0, 1 and 2 of this empty `vec![]`.

It would be cool if we can actually get some per-column statistics. Do you think that would be possible?

iceberg/src/data_source.rs
```diff
@@ -297,3 +300,98 @@ impl DataSource for IcebergDataSource {
...
+        return Ok(Arc::new(Statistics::new_unknown(schema)));
+    };
+    let props = &snap.summary().additional_properties;
+
+    let num_rows = props
+        .get("total-records")
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(0);
+    let total_byte_size = props
+        .get("total-files-size")
```
gabotechs: 🤔 If future upgrades of `iceberg-rust` change these names, this can silently start falling back to 0.

Can we add a link to the place in the `iceberg-rust` code that populates these metrics?

iceberg/src/data_source.rs
```diff
@@ -297,3 +300,98 @@ impl DataSource for IcebergDataSource {
...
 }
+
+/// Getting statistics from the provided snapshot.
+fn stats_from_snapshot(
+    snapshot: Option<&SnapshotRef>,
+    schema: &SchemaRef,
+) -> Result<Arc<Statistics>> {
+    let Some(snap) = snapshot else {
+        return Ok(Arc::new(Statistics::new_unknown(schema)));
+    };
```
gabotechs, line 324: So, if current snapshot is `None`, we get no statistics. This will be pretty bad, as these stats are essentially what will inform the distributed planner how much to distribute.

Is there any chance of getting the stats from somewhere else that we know it's always going to be present?

iceberg/src/data_source.rs
```diff
@@ -297,3 +300,98 @@ impl DataSource for IcebergDataSource {
...
+                .build(),
+        )
+    }
+
+    #[test]
+    fn no_snapshot_returns_unknown_stats() {
+        let s = stats_from_snapshot(None, &schema()).unwrap();
+        assert!(matches!(s.num_rows, Precision::Absent));
+        assert!(matches!(s.total_byte_size, Precision::Absent));
+    }
```
gabotechs: Nice tests! although it would be cool to add a new integration test to `tests/` that performs real queries against real iceberg catalogs, getting the stats out of them.

I think it's worth porting these tests to integration tests to something like `tests/statistics.rs`, where we issue real queries, and we assert the presence of some stats in there.

gabotechs (commented):
Thanks @sandugood! looking very good.

iceberg/src/data_source.rs
sandugood: I have changed this piece of code and now populate everything with zeroes in that case.

Why do I think that this is correct? If the table was created, then we would have an entry for that table in the catalog (REST, Glue, HMS etc.). However, if no data files were committed means `None` for snapshot.

If table wasn't even created we would get error earlier, though.

iceberg/src/data_source.rs
sandugood: Moved them under `/tests `to have integration tests with the taxi dataset.

iceberg/src/data_source.rs
sandugood: I have added a link to the Iceberg spec itself (about these fields) and also to lines in the` iceberg-rust` repo.

It would be very nice if we could just reimport them to keep everything intact, however they are private :disappointed:

iceberg/src/data_source.rs
sandugood: Fixed by returning `ColumnStatistics::new_unknown()` for each column of the projected schema. Per-column stats are possible from the manifest entries (null_value_counts, lower_bounds, upper_bounds per data file), but that needs async manifest reads cached at construction, so I'd propose doing it in a follow-up PR.

iceberg/src/data_source.rs
gabotechs: 🤔 Yeap, your reasoning sounds right. I don't know enough about Iceberg but what you say makes sense, so let's stick to it.

If you happen to have a link to some docs explaining this, cool, otherwise, it's fine to leave it like this.

iceberg/tests/statistics.rs
```diff
@@ -0,0 +1,133 @@
...
+        assert_eq!(stats.column_statistics.len(), 2);
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn explain_shows_statistics_on_the_iceberg_source() -> Result<()> {
+        let harness = IcebergTestHarness::new().await?;
+        let plan = harness.physical_plan("SELECT vendor_id FROM taxi").await?;
+        let display = displayable(plan.as_ref())
+            .set_show_statistics(true)
```
gabotechs, line 81: Nice! I didn't now this was an option.

gabotechs (approved):
Amazing, great work @sandugood!

gabotechs:
A follow up to this one might be to have some nice per-column statistics, but that depends more on `iceberg-rust` capabilities for exposing that kind of information.

gabotechs:
Added a comment in the original issue about this.

iceberg/src/data_source.rs
sandugood: https://iceberg.apache.org/spec/#table-metadata-fields
`current-snapshot-id` is an Optional parameter (for all v1-v4 of Iceberg)


Also I've tried to confirm it with a simple Iceberg table creation with Spark (without any data inserted)
After running `DESCRIBE TABLE EXTENDED` and checking `metadata.json`

tg_image_3160661423

alexanderbianchi: I think that within a snapshot, these stats are optional
 ```text
   Option<Snapshot>
   └── Snapshot
       └── summary: Map<String, String>
           ├── optional "total-records"
           └── optional "total-files-size"
 ```
 
 ```rust
   let props = &snap.summary().additional_properties;

   let num_rows = props
       .get(TOTAL_RECORDS)
       .and_then(|v| v.parse().ok())
       .unwrap_or(0);
 ```

 Here, snap exists. Only props.get(...) returns None.

 The same thing happens when the value exists but cannot be parsed:

 ```text
   .get(...)                 => None, or Some("invalid")
   .and_then(parse)          => None
   .unwrap_or(0)             => 0
   Precision::Exact(value)   => Exact(0)
 ```

The Iceberg spec calls these Optional Snapshot Summary Fields, including total-records and total-files-size: Iceberg specification.
https://iceberg.apache.org/spec/#optional-snapshot-summary-fields

iceberg/src/data_source.rs
```diff
@@ -297,3 +308,34 @@ impl DataSource for IcebergDataSource {
...
+    let props = &snap.summary().additional_properties;
+
+    let num_rows = props
+        .get(TOTAL_RECORDS)
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(0);
+    let total_byte_size = props
+        .get(TOTAL_FILE_SIZE)
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(0);
```
alexanderbianchi, line 334: https://github.com/datafusion-contrib/datafusion-distributed/pull/690/changes

PR #682, opened 2026-08-26
fix: preserve network boundary equivalence groups

### LINKED ISSUE #679: Preserve equivalence properties across network-boundary serialization
`NetworkShuffleExec` serialization drops `EquivalenceProperties`, which can make a valid sorted window plan fail after it is sent to a worker.

### Reproduction

A physical plan contains:

```text
BoundedWindowAggExec, mode=Sorted
  SortExec: [partition_expr@1 ASC, modified_at@0 DESC]
    RepartitionExec: Hash([partition_expr__temp__0@2])
      ProjectionExec:
        partition_expr@1 = __common_expr_9@0
        partition_expr__temp__0@2 = __common_expr_9@0
```

The two projected columns are equivalent because they are both derived from the same input expression. Before serialization, DataFusion correctly uses that equivalence to establish that `partition_expr__temp__0@2` is ordered by the `SortExec` key `partition_expr@1`.

After `NetworkShuffleExec` is serialized and decoded on the worker, it reconstructs its properties with:

```rust
EquivalenceProperties::new(schema)
```

The codec carries the schema, partitioning, and input stage, but not the equivalence group. The worker therefore retains the sort order on `@1` but loses the proof that `@1 == @2`.

When `BoundedWindowAggExec` starts its sorted execution path, it fails with:

```text
internal: Expects PARTITION BY expression to be ordered
```

### Expected behavior

A plan that is valid before distributed serialization should remain valid after deserialization. In particular, value-equivalence relationships required to interpret a child ordering should survive network-boundary serialization.

### Proposed fix

Extend the distributed network-boundary wire format to preserve the relevant `EquivalenceProperties`—at least the equivalence group—for `NetworkShuffleExec`. The same audit should cover `NetworkCoalesceExec` and `NetworkBroadcastExec`, which currently reconstruct empty equivalence properties as well.

The fix should not preserve an input ordering across a shuffle, since that would be unsound. This issue only requires preserving the fact that two output expressions have equal values.

WilliamK112:
## Summary

- add a backward-compatible protobuf field for value-equivalence classes on distributed network boundaries
- restore equivalence groups when decoding `NetworkShuffleExec`, `NetworkCoalesceExec`, and `NetworkBroadcastExec`
- keep ordering properties out of the wire format so shuffle input ordering is not propagated unsafely
- cover all three boundary types with a focused round-trip regression test

## Testing

- `cargo test --lib` (297 passed, 1 ignored)
- `cargo clippy --all-targets --all-features -- -D warnings`
- `git diff --check`

Fixes #679

### DIFF
--- a/src/codec/distributed_codec.rs
+++ b/src/codec/distributed_codec.rs
@@ -187,7 +187,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
                 let schema = Arc::new(schema);
                 let equivalence_properties = parse_equivalence_properties(
                     equivalence_classes,
-                    schema.clone(),
+                    schema,
                     &decode_ctx,
                     proto_converter,
                 )?;
--- a/src/codec/distributed_codec.rs
+++ b/src/codec/distributed_codec.rs
@@ -154,7 +154,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
                 let schema = Arc::new(schema);
                 let equivalence_properties = parse_equivalence_properties(
                     equivalence_classes,
-                    schema.clone(),
+                    schema,
                     &decode_ctx,
                     proto_converter,
                 )?;
--- a/src/codec/distributed_codec.rs
+++ b/src/codec/distributed_codec.rs
@@ -15,6 +15,7 @@ use datafusion::common::Result;
 use datafusion::error::DataFusionError;
 use datafusion::execution::TaskContext;
 use datafusion::physical_expr::EquivalenceProperties;
+use datafusion::physical_expr::equivalence::{EquivalenceClass, EquivalenceGroup};
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::union::UnionExec;
 use datafusion::physical_plan::{ExecutionPlan, Partitioning, PlanProperties};
@@ -102,6 +103,7 @@ impl PhysicalExtensionCodec for DistributedCodec {
                 schema,
                 partitioning,
                 input_stage,
+                equivalence_classes,
             }) => {
                 let schema: Schema = schema
                     .as_ref()
@@ -116,10 +118,17 @@ impl PhysicalExtensionCodec for DistributedCodec {
                     proto_converter,
                 )?
                 .ok_or(proto_error("NetworkShuffleExec is missing partitioning"))?;
+                let schema = Arc::new(schema);
+                let equivalence_properties = parse_equivalence_properties(
+                    equivalence_classes,
+                    schema.clone(),
+                    &decode_ctx,
+                    proto_converter,
+                )?;
 
-                Ok(Arc::new(new_network_hash_shuffle_exec(
+                Ok(Arc::new(new_network_hash_shuffle_exec_with_properties(
                     partitioning,
-                    Arc::new(schema),
+                    equivalence_properties,
                     parse_stage_proto(input_stage, inputs)?,
                 )))
             }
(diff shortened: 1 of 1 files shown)


src/codec/distributed_codec.rs
```diff
@@ -467,14 +560,27 @@ pub struct ChildIdxWithTaskContextProto {
     task_count: u64,
 }
 
+#[cfg(test)]
 fn new_network_hash_shuffle_exec(
```
gabotechs: If the function is only used in tests, it would be nice to just remove it.

What I would do is to just keep a single function with this signature:

```rust
fn new_network_hash_shuffle_exec(
    partitioning: Partitioning,
    equivalence_properties: EquivalenceProperties,
    input_stage: Stage,
) -> NetworkShuffleExec;
```

And use that everywhere.

src/codec/distributed_codec.rs
```diff
@@ -495,16 +601,31 @@ pub struct NetworkCoalesceExecProto {
     partitioning: Option<protobuf::Partitioning>,
     #[prost(message, optional, tag = "3")]
     input_stage: Option<StageProto>,
+    #[prost(message, repeated, tag = "4")]
+    equivalence_classes: Vec<EquivalenceClassProto>,
 }
 
+#[cfg(test)]
 fn new_network_coalesce_tasks_exec(
```
gabotechs, line 595: Let's aim for not having custom functions just used in the tests

gabotechs (commented):
Looks good, thanks @WilliamK112! left suggestions mostly about the tests.

src/codec/distributed_codec.rs
WilliamK112: Thanks! Removed both test-only wrappers and kept one constructor per boundary taking `EquivalenceProperties`. The round-trip tests now call the same constructors as decoding, including the equivalence regression test. `cargo fmt --all -- --check`, `cargo check --tests`, and `git diff --check` pass; the update is in 92e6b7d.

src/codec/distributed_codec.rs
```diff
@@ -116,17 +118,25 @@ impl PhysicalExtensionCodec for DistributedCodec {
                     proto_converter,
                 )?
                 .ok_or(proto_error("NetworkShuffleExec is missing partitioning"))?;
+                let schema = Arc::new(schema);
+                let equivalence_properties = parse_equivalence_properties(
+                    equivalence_classes,
+                    schema.clone(),
```
gabotechs:
```suggestion
                    schema,
```

src/codec/distributed_codec.rs
```diff
@@ -141,17 +151,25 @@ impl PhysicalExtensionCodec for DistributedCodec {
                     proto_converter,
                 )?
                 .ok_or(proto_error("NetworkCoalesceExec is missing partitioning"))?;
+                let schema = Arc::new(schema);
+                let equivalence_properties = parse_equivalence_properties(
+                    equivalence_classes,
+                    schema.clone(),
```
gabotechs:
```suggestion
                    schema,
```

src/codec/distributed_codec.rs
```diff
@@ -166,10 +184,17 @@ impl PhysicalExtensionCodec for DistributedCodec {
                     proto_converter,
                 )?
                 .ok_or(proto_error("NetworkBroadcastExec is missing partitioning"))?;
+                let schema = Arc::new(schema);
+                let equivalence_properties = parse_equivalence_properties(
+                    equivalence_classes,
+                    schema.clone(),
```
gabotechs:
```suggestion
                    schema,
```

gabotechs (approved):
Looks good! just left some small suggestions, thanks @WilliamK112!

PR #683, opened 2026-08-26
feat: serialize Iceberg file scan work units

### LINKED ISSUE #604: [Iceberg] Serialize FileScanTask work units
Part of #599.

Implement a wire representation for `FileScanTaskMessage` so Iceberg scan tasks can be streamed from the coordinator to workers.

Essentially, implementing this TODO comment in the code:

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/codec.rs#L30-L34

I'm not sure is what's the best path forward for this, things that come to mind are:
- Take advantage of the serde capabilities already present in the FileScanTask
- Create a separate protobuf struct that mirrors FileScanTask, can convert to it

alexanderbianchi:
## Summary

Implements a wire representation for Iceberg `FileScanTask` work units so they can be serialized by the distributed work-unit transport.

The implementation uses a protobuf envelope containing a MessagePack payload. It validates file ranges, projected field IDs, partition metadata, and canonical partition-value encodings while reconstructing the Iceberg task on the consumer.

Closes #604.

## Commits

The PR is split into two independently working commits so the context-reuse trade-off is easy to review:

1. `d08b3b1` — adds a stateless, production-ready encoder where every work unit contains its complete task context.
2. `93a76dd` — adds per-feed context reuse and the stateful decoder described below.

## Shared-context decoder

Iceberg file tasks commonly repeat the same schema, projection, predicate, partition spec, name mapping, and case-sensitivity setting. Although several of those values use `Arc` in memory, serializing an `Arc<T>` still writes the complete `T` for every work unit.

The second commit assigns repeated contexts an ID. The first work unit in each partition feed defines the context, and later work units reference it by ID. A decoder owned by that partition stream retains the definitions while reconstructing tasks. State is isolated per feed and per query; it is not shared across partitions, distributed tasks, or queries.

This reduces repeated serialization, allocation, and network bytes at the cost of an order-dependent per-feed protocol. A distributed transport test verifies that context definitions and references remain ordered through coordinator batching, partition routing, and gRPC transport.

## Taxi fixture measurement

Using the seven real file scan tasks from the checked-in taxi table, routed over its four partition feeds:

- repeated complete context: **6,552 bytes**
- shared per-feed context: **4,425 bytes**
- reduction: **2,127 bytes (32.5%)**

This is a deterministic encoded-size regression test, not a runtime benchmark. The taxi fixture is conservative because it has only seven files spread over four feeds; larger tables normally reuse each feed context across many more files.

## Validation

- `cargo test --features integration --test work_unit_feed`
- targeted Clippy checks for the Iceberg library and distributed work-unit integration test
- formatting and diff checks


iceberg/src/work_unit_wire_tests.rs
```diff
@@ -0,0 +1,331 @@
+mod tests {
```
gabotechs: :thinking: Is there any specific reason why these tests are defined in a separate file? (go-style). In Rust, unless there's a good reason, unit tests are placed as a test module right below the code that's getting tested.

This would mean having this `mod tests {};` block right at the bottom of the `work_unit_wire.rs` file.

src/test_utils/test_work_unit_feed.rs
```diff
@@ -34,6 +34,10 @@ use std::time::Duration;
 pub struct RowGeneratorWorkUnit {
     #[prost(uint64, tag = "1")]
     n_rows: u64,
+    #[prost(uint64, tag = "2")]
+    context_id: u64,
+    #[prost(uint64, optional, tag = "3")]
+    context: Option<u64>,
 }
```
gabotechs: Are the changes in this file intentional? they do not seem related with the changes in this PR.

tests/work_unit_feed.rs
```diff
@@ -81,6 +81,31 @@ mod tests {
         Ok(())
     }
 
+    /// Stateful work units define per-partition context once and reference it from later
+    /// messages. This verifies that coordinator batching and gRPC routing preserve each
+    /// partition's message order through the remote feed.
+    #[tokio::test]
+    async fn preserves_stateful_work_unit_context_per_partition()
+    -> Result<(), Box<dyn std::error::Error>> {
```
gabotechs: This test is just testing that the test util in `test_work_unit_feed.rs` is implemented as intended, but it does not really add any new coverage that is not handled by existing tests already.

I think we can drop the changes in this file and `test_work_unit_feed.rs`, as I don't see them really contributing any new coverage.

iceberg/src/work_unit_wire.rs
```diff
@@ -0,0 +1,371 @@
...
+use prost::Message;
+use serde::{Deserialize, Serialize};
+
+/// Wire representation of one Iceberg file scan task.
+#[derive(Clone, PartialEq, Message)]
+pub struct FileScanTaskMessage {
+    #[prost(bytes = "vec", tag = "1")]
+    payload: Vec<u8>,
+}
+
```
gabotechs: I see two big downsides with the approach taken here:

- Serialization is not optional: there are multiple situations where WorkUnits are not transferred across the wire, for example: the coordinator and the destination worker are co-located in the same machine, or the query is not big enough to be distributed, therefore it collapses into a single-node plan. In these cases, we are paying the price of [de]serialization, where we could just transfer the payloads in-memory.
- Easy to break protocol: it's going to be important to always maintain structs sent over the wire backwards compatible. Protobuf handles this nicely with numbered fields and automatic defaults. With the current approach, It's going to be easy to just introduce a new field in a random place in these structs and break the protocol.

Have you considered using protobuf instead of message pack for this? this is what the rest of the project uses every time we need to send something over the wire so the tooling is already there. This would solve the two problems above.

One solution that could be quite elegant is to define any serializable Iceberg payload in an `iceberg.proto` file, and tweak the project-level generator script to automatically generate the Rust structs for you:

https://github.com/datafusion-contrib/datafusion-distributed/blob/3b8a26635352320c11296c6c37ae93413630c2a5/codegen/src/main.rs

iceberg/src/work_unit_wire_tests.rs
alexanderbianchi: no specific reason besides I clearly am very go-brained

iceberg/src/proto/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub(crate) mod generated {
+    include!("generated/iceberg.rs");
```
gabotechs: For exposing generated code, I'd follow instead the same pattern than in the rest of the project, rather than using the `include!` macro. You have an example here:

https://github.com/datafusion-contrib/datafusion-distributed/blob/986781486a500fd02ffa5ae63ecad054ee3f6b62/src/protocol/grpc/generated

iceberg/src/work_unit_wire.rs
```diff
@@ -0,0 +1,1526 @@
...
+                task.partition.as_ref(),
+                task.partition_spec.as_deref(),
+                &task.schema,
+            )?,
+        }),
+    })
+}
+
+#[derive(Debug, PartialEq)]
+struct TaskContext {
```
gabotechs: I'd name this something more descriptive, as the current name collides with `datafusion`'s `TaskContext`, which is widely used in this project.

iceberg/src/work_unit_wire.rs
```diff
@@ -0,0 +1,1526 @@
...
+    schema: SchemaRef,
```
gabotechs: A bit annoying how iceberg-rust decided to choose the same names for some things than `arrow-rs`.

For avoiding confusions, it'd be nice if we could fully qualify the types in these situations:

```rust
iceberg::spec::SchemaRef
```

iceberg/src/proto/iceberg.proto
```diff
@@ -0,0 +1,270 @@
+syntax = "proto3";
+package iceberg;
+
+message FileScanTask {
+  uint32 context_id = 1;
+  FileScanTaskContext context = 2;
+  FileScanTaskBody task = 3;
+}
+
+message FileScanTaskContext {
```
gabotechs, line 10: This encoding context is at the edge of being a premature optimization, but after seeing how gracefully handled it is, and given that it really does not introduce much complexity, having it now looks like a good call :+1:

iceberg/src/work_unit_wire.rs
```diff
@@ -0,0 +1,1526 @@
...
+    }
+
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: WireType,
+        buf: &mut impl Buf,
+        ctx: DecodeContext,
+    ) -> std::result::Result<(), DecodeError> {
+        let FileScanTaskPayload::Proto(proto) = &mut self.payload else {
```
gabotechs, line 154: Here, if there was a `Native` variant stored, we are discarding it without merging its inner proto no?

Actually, if `self` is in `Native` state, I think it should even be safe to just error out, I don't think we're ever going to reach that situation, and if we do, it would be weird. So maybe we can just throw an error here for now and care about it in the future.

iceberg/src/work_unit_wire.rs
```diff
@@ -0,0 +1,1526 @@
...
+        };
+        let task = body_from_proto(required(proto.task, "task")?, context)?;
+        if let Some(definition) = definition {
+            self.contexts.entry(proto.context_id).or_insert(definition);
+        }
+        Ok(task)
+    }
+}
+
+impl Message for FileScanTaskMessage {
```
gabotechs, line 139: Interesting (and smart) trick here.

I wonder if this unlocks an opportunity for reusing the existing serialization built into `FileScanTask` (that struct implements `Serialialize` an `Deserialize` natively) by just storing it as an opaque `Vec<u8>`.

That will remove the possibility of doing the encoder/decoder trick though.

I think what we have here in this PR is fine, but do you think we are missing out on something by not going down that path?

gabotechs (approved):
Left some comments, but giving a +1 for unblocking after the comments are resolved.

Really nice work!

Looks like scary amount of plumbing, but after reviewing it's just that: plumbing, so I don't think this is ever going to be a mantainability hazard, just the typical thing where an LLM can one shot any fix or addition if needed.

iceberg/src/work_unit_wire.rs
alexanderbianchi: I played around with ways to do it but FileScanTask cannot currently serialize all of its fields.
We could go in the middle like:
 ```protobuf
   message FileScanTask {
     bytes serde_payload = 1;
     PartitionValues partition = 2;
     PartitionSpec partition_spec = 3;
     NameMapping name_mapping = 4;
   }
 ```
But
 - The protocol becomes split
 - Compatibility depends on iceberg-rust’s Serde representation
 - Coordinator/worker versioning becomes more difficult
 - Context deduplication is difficult.

IMO more complex then this. If we go down that path we should make FileScanTask fully serializable upstream in iceberg-rust first.

gabotechs: 👍 Makes sense, what you did here sounds best then.

I wonder why they decided to just serialize some of the fields 🤔. Looks like the fields that are not getting serialized are actually important, so I wonder why implement any serialization at all.

Issue #605, opened 2026-08-09
[Iceberg] Estimate desired task counts from table metadata


gabotechs:
Part of #599.

Implement the desired task count handler for `IcebergDataSource`. It should inspect the inner Iceberg table and determine a useful desired number of distributed tasks.

Essentially, implementing these TODO comments in the code:

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/distributed_desired_task_count_handler.rs#L7-L8

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/distributed_desired_task_count_handler.rs#L18-L18

The table metadata and selected snapshot should provide enough information to estimate the amount of scan work.

alexanderbianchi:
take

alexanderbianchi:
take

PR #684, opened 2026-08-26
feat: serialize Iceberg data source plans

### LINKED ISSUE #603: [Iceberg] Serialize IcebergDataSource plans for distributed execution
Part of #599.

Implement `IcebergCodec` so physical plans containing `IcebergDataSource` can be encoded on the coordinator and decoded on workers.

Essentially, implementing this TODO comment in the code:

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/codec.rs#L14-L14

Some roundtrip unit test can also be shipped for ensuring no data is lost during [de]serialization

alexanderbianchi:
## Summary

Implements `IcebergCodec` so physical plans containing `IcebergDataSource` can be serialized by the coordinator and reconstructed on workers.

The generated protobuf preserves the source schema, work-unit feed, DataFusion partitioning, fetch limit, and storage properties. Decoding rebuilds `FileIO` using the worker-local storage factory and Iceberg runtime. Local statistics retain the existing current-snapshot behavior; remote sources report unknown statistics rather than serializing planning-only information.

The Iceberg session integration registers both the plan codec and work-unit feed getter. A harness-backed codec round-trip test covers schema, partitioning, fetch, feed identity, storage properties, and remote unknown statistics.

Closes #603.

## Scope

Based on `iceberg-0.10`, which now includes #683's file scan task serialization and #690's unknown-statistics fix.

Per review, distributed harness setup and end-to-end distributed tests are deferred until #605 supplies the real task-count handler. This PR does not add a temporary fixed-count handler or a separate distributed harness mode.

Explicit non-empty storage-property round-trip coverage is tracked in draft #703, based on #700's harness builder. This avoids a hidden test dependency on default fixture options.

## Validation

- `cargo clippy -p datafusion-distributed-iceberg --lib -- -D warnings`
- `git diff --check origin/iceberg-0.10...HEAD`


iceberg/src/data_source.rs
```diff
@@ -126,8 +127,8 @@ pub struct IcebergDataSource {
     partitioning: Partitioning,
     fetch: Option<usize>,
     metrics: ExecutionPlanMetricsSet,
-    current_snapshot: Option<SnapshotRef>,
-    iceberg_file_io: iceberg::io::FileIO,
+    statistics: Arc<Statistics>,
```
gabotechs: Storing `statistics` here can be problematic. It's at risk of becoming stale as other properties of this node get mutated.

I see this is here just for the sake of having statistics access during remote execution in workers, but I doubt that's going to be any useful there, so I think it's worth it to just drop it here.

In the `partition_statistics` method we can just do something like:

```rust
    fn partition_statistics(&self, _partition: Option<usize>) -> Result<Arc<Statistics>> {
    let Some(feed) = self.feed.inner() else {
        return Ok(Arc::new(Statistics::new_unknown(&self.schema)));
    };

    Ok(stats_from_snapshot(
        feed.iceberg_table.metadata().current_snapshot(),
        &self.schema,
    ))
}
```

Which means that stats will just be unknown after proto roundtrips, but this is fine, stats are only needed at planning time.

gabotechs: With this, we should even be able to drop the `current_snapshot` field for free.

iceberg/src/data_source.rs
```diff
@@ -181,9 +182,42 @@ impl IcebergDataSource {
...
+    pub(crate) fn from_remote(
+        schema: SchemaRef,
+        partitioning: Partitioning,
+        fetch: Option<usize>,
+        statistics: Arc<Statistics>,
+        iceberg_file_io: FileIO,
+        iceberg_runtime: iceberg::Runtime,
+        feed: WorkUnitFeed<IcebergWorkUnitFeed>,
+    ) -> Self {
+        Self {
```
gabotechs: At this point, I'd just remove this proxy function and build the struct in-line in `codec.rs`. This is just 1 param away of violating a clippy rule, and it does not really help much to have it here, as it does nothing.

iceberg/src/codec.rs
```diff
@@ -1,24 +1,258 @@
...
+struct IcebergDataSourceProto {
+    #[prost(message, optional, tag = "1")]
+    schema: Option<protobuf::Schema>,
+    #[prost(message, optional, tag = "2")]
+    feed: Option<WorkUnitFeedProto>,
+    #[prost(uint64, tag = "3")]
+    partitions: u64,
+    #[prost(enumeration = "PartitioningKind", tag = "4")]
+    partitioning: i32,
+    #[prost(uint64, optional, tag = "5")]
```
gabotechs: This proto message design is likely not going to age well. As soon as Range or Hash partitioning are supported, new fields for holding the `SplitPoint`s and the hashing expressions will need to be added, and consistency across fields will not be granted at compile time.

Take `NetworkShuffleExec` as an example:

https://github.com/datafusion-contrib/datafusion-distributed/blob/main/src/codec/distributed_codec.rs#L427

You can just rely on upstream functions for [de]serializaing `Partitioning`

iceberg/src/codec.rs
```diff
@@ -1,24 +1,258 @@
...
+    fetch: Option<u64>,
+    #[prost(bytes = "vec", tag = "6")]
+    storage_config: Vec<u8>,
+    #[prost(uint64, tag = "7")]
+    num_rows: u64,
+    #[prost(uint64, tag = "8")]
+    total_byte_size: u64,
+}
```
gabotechs: As mentioned in the other message, we should be good without this.

iceberg/src/codec.rs
```diff
@@ -1,24 +1,258 @@
...
+    async fn roundtrips_data_source_plan() -> Result<()> {
+        let harness = IcebergTestHarness::new().await?;
+        let plan = harness.physical_plan("SELECT * FROM taxi LIMIT 10").await?;
+        let source_plan = iceberg_plan(&plan)?;
+        let source = iceberg_source(&source_plan)?;
+        let codec = IcebergCodec::new(
+            Arc::new(FixtureStorageFactory::default()),
+            iceberg::Runtime::current(),
+        );
+        let mut bytes = Vec::new();
```
gabotechs: Ideally, the codec should be injected by the testing harness, not manually in each test.

iceberg/tests/distributed_read.rs
```diff
@@ -0,0 +1,73 @@
...
+        ctx.set_iceberg_integration(integration_options());
+        ctx.set_distributed_desired_task_count_handler(iceberg_test_task_count);
+        ctx.sql(&format!(
+            "CREATE EXTERNAL TABLE taxi STORED AS ICEBERG \
+             LOCATION '{FIXTURE_URI}/metadata/v1.metadata.json'"
+        ))
+        .await?
+        .collect()
+        .await?;
+
```
gabotechs: Here, we should be using the `IcebergTestHarness` rather than constructing the context manually.

iceberg/src/test_utils/harness.rs
```diff
@@ -70,7 +75,7 @@ impl IcebergTestHarness {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct FixtureStorageFactory {
+pub struct FixtureStorageFactory {
```
gabotechs: I see this was made public so that the integration tests can use it, but the integration tests should be using the `IcebergTestHarness` itself instead.

The main purpose of the `IcebergTestHarness` is to reduce all the boilerplate and preparation code from each individual test. If you see yourself needing to prepare a DataFusion `SessionContext` manually inside a test, it means that the harness should instead be used.

This means that maybe you need to extend a bit the harness to hold more functionality, like spawning a fleet of localhost workers while enriching the context with distributed extensions for example. This is fine, it's part of the expected natural evolution of `IcebergTestHarness`

gabotechs (commented):
Nice! this is going in the right direction, left some comments that should make things a bit easier.

iceberg/src/test_utils/harness.rs
```diff
@@ -25,27 +37,44 @@ const WAREHOUSE_URI: &str = "s3://iceberg-test/warehouse/";
 
 pub struct IcebergTestHarness {
     ctx: SessionContext,
+    #[cfg(feature = "integration")]
+    _worker_guard: Option<datafusion::common::runtime::JoinSet<()>>,
 }
```
gabotechs: :thinking: all the feature checks in this file are getting a bit hard to manage.  Probably the whole harness should be hidden behind the `integration` flag, not just pieces of it.

iceberg/src/test_utils/harness.rs
```diff
@@ -62,6 +91,41 @@ impl IcebergTestHarness {
...
+        storage_factory: Arc::new(FixtureStorageFactory::default()),
+        iceberg_runtime: iceberg::Runtime::current(),
+    }
+}
+
+#[cfg(feature = "integration")]
+fn iceberg_test_task_count(
+    event: DesiredTaskCountEvent,
+) -> Option<Result<DesiredTaskCountEventResponse>> {
+    event
```
gabotechs: I can imagine how this will disappear in the near future:

After https://github.com/datafusion-contrib/datafusion-distributed/issues/605, we'd have one of these handlers that has a configured "bytes_per_partition"-like setting that we can set in the `IcebergConfig` to something unreasonably low for forcing distribution.

iceberg/src/test_utils/harness.rs
```diff
@@ -25,27 +37,44 @@ const WAREHOUSE_URI: &str = "s3://iceberg-test/warehouse/";
...
-                iceberg_runtime: iceberg::Runtime::current(),
-            })
+            .with_iceberg_integration(integration_options())
             .build();
-        let ctx = SessionContext::new_with_state(state);
+        Self::from_context(SessionContext::new_with_state(state)).await
+    }
+
+    #[cfg(feature = "integration")]
+    pub async fn new_distributed() -> Result<Self> {
```
gabotechs: :thinking: I can imagine how we'd want the distributed planner to be enabled in every test, whether the test ends up distributing or not.

gabotechs (approved):
I think it's worth deferring changes to the test harness until we ship https://github.com/datafusion-contrib/datafusion-distributed/issues/605. I see the new functions in the harness are only used by one small test, so I think we can just remove it from this PR, and add a bunch of more tests in an follow up once we get the DesiredTaskCountHandler in place.

The rest looks pretty good! so +1

iceberg/src/codec.rs
```diff
@@ -1,24 +1,197 @@
...
+            source.partitioning.to_string(),
+            decoded.partitioning.to_string()
+        );
+        assert_eq!(source.fetch, decoded.fetch);
+        assert_eq!(source.feed.to_proto(), decoded.feed.to_proto());
+        assert_eq!(
+            source.iceberg_file_io.config().props(),
+            decoded.iceberg_file_io.config().props()
+        );
+        assert!(!source.iceberg_file_io.config().props().is_empty());
```
alexanderbianchi: This line depends on the update to the harness which is a really tricky link, I don't like that and would rather remove it for now and wait for https://github.com/datafusion-contrib/datafusion-distributed/pull/700

alexanderbianchi:
Originally there were some conflicts inbound with https://github.com/datafusion-contrib/datafusion-distributed/pull/687 - i've stripped this PR to be more focused (good practice anyways) and @sandugood should have an easier time now. I do want to make sure a couple things don't fall through the cracks so I opened 
https://github.com/datafusion-contrib/datafusion-distributed/pull/702
https://github.com/datafusion-contrib/datafusion-distributed/pull/703
Which build off https://github.com/datafusion-contrib/datafusion-distributed/pull/700 to add test coverage for this PR and 687

PR #687, opened 2026-09-02
feat[iceberg]: table statistics suite

### LINKED ISSUE #681: [Iceberg] Add per-column statistics to the IcebergDataSource
Follow up on https://github.com/datafusion-contrib/datafusion-distributed/issues/608.

We now have global row count and byte size, but it would be nice to add some per-column statistics for DataFusion to make better planning decisions.

sandugood:
Follow-up to #676 (closes #681)

What was changed/added:
1. Statistics are being computed once at the very beginning of the scan and not recomputed
2. Everything is being computed from `ManifestList` -> `ManifestFile` -> `Manifest` (from where we get `DataFile` information). Now we are getting datafile level stats directly by mapping `field-ids` to actual columns in `.parquet` files.
3. Casting from iceberg's `Datum`

### DIFF
--- a/iceberg/src/data_source.rs
+++ b/iceberg/src/data_source.rs
@@ -417,7 +417,6 @@ pub async fn compute_column_stats(
 
             for entry in manifest.entries().iter().filter(|e| e.is_alive()) {
                 let df = entry.data_file();
-                println!("data file: {:?}", df);
 
                 for (i, &id) in fields_ids.iter().enumerate() {
                     let cs = &mut col_stats[i];
@@ -431,8 +430,6 @@ pub async fn compute_column_stats(
                     if let Some(scalar) = df.upper_bounds().get(&id).and_then(datum_to_scalar) {
                         cs.max_value = cs.max_value.max(&Precision::Inexact(scalar));
                     }
-
-                    // println!("col stat: {:?}", cs);
                 }
             }
 
--- a/iceberg/src/data_source.rs
+++ b/iceberg/src/data_source.rs
@@ -281,12 +281,10 @@ impl DataSource for IcebergDataSource {
         let mut stats = stats_from_snapshot(self.current_snapshot.clone(), &self.schema)?;
 
         if let Some(col_stats) = &self.column_stats {
-            if col_stats.len() == stats.column_statistics.len() {
-                for (i, cs) in col_stats.iter().enumerate() {
-                    stats.column_statistics[i] = cs.clone();
-                }
-            } else {
-                println!("placeholder");
+            // Lengths should always match
+            debug_assert_eq!(col_stats.len(), stats.column_statistics.len());
+            for (i, cs) in col_stats.iter().enumerate() {
+                stats.column_statistics[i] = cs.clone();
             }
         }
 
--- a/iceberg/src/data_source.rs
+++ b/iceberg/src/data_source.rs
@@ -16,23 +16,19 @@ use datafusion::physical_plan::metrics::{BaselineMetrics, ExecutionPlanMetricsSe
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::{DisplayFormatType, SortOrderPushdownResult};
 use datafusion::prelude::Expr;
+use datafusion::scalar::ScalarValue;
 use datafusion_distributed::WorkUnitFeed;
 use futures::{StreamExt, TryStreamExt};
 use iceberg::arrow::ArrowReaderBuilder;
 use iceberg::io::FileIO;
-use iceberg::spec::SnapshotRef;
+use iceberg::spec::{
+    Datum, Manifest, ManifestContentType, ManifestList, PrimitiveLiteral, PrimitiveType,
+};
+use iceberg::table::Table;
 
 use crate::common::{convert_filters_to_predicate, df_err, iceberg_err};
 use crate::work_unit_wire::FileScanTaskDecoder;
 use crate::{IcebergConfig, IcebergWorkUnitFeed};
-
-/// Snapshot summary keys defined by the Iceberg table spec:
-/// https://iceberg.apache.org/spec/#optional-snapshot-summary-fields
-///
-/// iceberg-rust defines them privately:
-/// https://github.com/apache/iceberg-rust/blob/4168a0b2950dc5f85588e5cb3ab6796e5228b309/crates/iceberg/src/spec/snapshot_summary.rs#L46-L47
-const TOTAL_RECORDS: &str = "total-records";
-const TOTAL_FILE_SIZE: &str = "total-files-size";
 
 /// Consumes a stream of [iceberg::scan::FileScanTask]s per partition and reads the underlying
 /// files into an Arrow stream.
@@ -127,6 +123,7 @@ pub struct IcebergDataSource {
     pub(crate) partitioning: Partitioning,
     pub(crate) fetch: Option<usize>,
     pub(crate) metrics: ExecutionPlanMetricsSet,
+    pub(crate) table_stats: Arc<Statistics>,
     pub(crate) iceberg_file_io: FileIO,
     pub(crate) iceberg_runtime: iceberg::Runtime,
     pub(crate) feed: WorkUnitFeed<IcebergWorkUnitFeed>,
@@ -144,12 +141,12 @@ pub(crate) struct IcebergDataSourceOptions<'a> {
 
 impl IcebergDataSource {
     /// Creates a new [`IcebergDataSource`] object.
-    pub(crate) fn new(
+    pub(crate) async fn new(
         table: iceberg::table::Table,
         schema: SchemaRef,
         partitioning: Partitioning,
-        opts: IcebergDataSourceOptions,
-    ) -> Self {
+        opts: IcebergDataSourceOptions<'_>,
+    ) -> iceberg::Result<Self> {
         let output_schema = match opts.projection {
             None => schema.clone(),
             Some(projection) => Arc::new(schema.project(projection).unwrap()),
@@ -160,9 +157,11 @@ impl IcebergDataSource {
                 .collect::<Vec<String>>()
         });
 
+        let table_stats = Arc::new(compute_table_stats(&table).await?.project(opts.projection));
+
         let predicates = convert_filters_to_predicate(opts.filters);
 
-        Self {
+        Ok(Self {
             schema: output_schema,
             iceberg_file_io: table.file_io().clone(),
             partitioning: partitioning.clone(),
@@ -179,7 +178,8 @@ impl IcebergDataSource {
                 partitioning,
                 sync_manager: Default::default(),
             }),
-        }
+            table_stats,
+        })
     }
 }
 
@@ -259,13 +259,7 @@ impl DataSource for IcebergDataSource {
     }
 
     fn partition_statistics(&self, _partition: Option<usize>) -> Result<Arc<Statistics>> {
-        let Some(feed) = self.feed.inner() else {
-            return Ok(Arc::new(Statistics::new_unknown(&self.schema)));
-        };
-        stats_from_snapshot(
-            feed.iceberg_table.metadata().current_snapshot(),
-            &self.schema,
-        )
+        Ok(self.table_stats.clone())
     }
 
     fn with_fetch(&self, fetch: Option<usize>) -> Option<Arc<dyn DataSource>> {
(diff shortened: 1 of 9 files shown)


sandugood:
cc @gabotechs 

By the way, on the current `main` of `iceberg-rust`(rev = `1ecd497`) there is a nice API of `ManifestReader`, which would strip the `table.file_io()` part 
However I think that staying with `0.10.0` is also alright. Wdyt?

gabotechs:
🤔 I'd say let's stick to 0.10 until `iceberg-rust` releases a new version to crates.io that uses datafusion@55. Once that's on crates.io we can start thinking about consuming all those new APIs.

It does look promising though.

iceberg/src/data_source.rs
```diff
@@ -125,7 +121,7 @@ pub struct IcebergDataSource {
     partitioning: Partitioning,
     fetch: Option<usize>,
     metrics: ExecutionPlanMetricsSet,
-    current_snapshot: Option<SnapshotRef>,
+    table_stats: Arc<Statistics>,
```
gabotechs: Rather than storing the statistics as a field, I'd just recompute them on every `partition_statistics`, same as we do now. There are several reasons for that:
- If the node gets mutated, this stats here will become stale.
- We need to reconstruct this node over the wire across a protobuf boundary, which will become difficult if these stats are in the middle.

It should be fine to just recompute the stats in the in each `partition_statistics` call.

gabotechs: Damn, I just saw that computing stats is `async`.

iceberg/src/data_source.rs
```diff
@@ -143,12 +139,12 @@ pub(crate) struct IcebergDataSourceOptions<'a> {
...
     /// Creates a new [`IcebergDataSource`] object.
-    pub(crate) fn new(
+    pub(crate) async fn new(
         table: iceberg::table::Table,
         schema: SchemaRef,
         partitioning: Partitioning,
-        opts: IcebergDataSourceOptions,
-    ) -> Self {
+        opts: IcebergDataSourceOptions<'_>,
+    ) -> iceberg::Result<Self> {
```
gabotechs: As this is a DataFusion implementation, we should be returning a `datafusion::common::Result` rather than an `iceberg::Result` here.

I think it's anyways worth it to strip out the stats gathering logic outside of the constructor, leaving it like a normal synchronous constructor method:

```rust
    pub(crate) fn new(
        table: iceberg::table::Table,
        schema: SchemaRef,
        partitioning: Partitioning,
        opts: IcebergDataSourceOptions,
    ) -> Self;
```

And create a separate builder method that is `async` and enriches the struct with statistics:

```rust
        pub(crate) fn with_column_statistics(mut self, table: iceberg::table::Table) -> Self;
```

That will give us some flexibility for enabling/disabling column stats collection.

iceberg/src/data_source.rs
gabotechs: I think it's worth avoiding storing the `Statistics` object as a field. We can compose the `Statistics` object at `partition_statistics` call time joining two different pieces of information:
1. total byte size and row count, essentially what you did in the other PR. We can leave this pretty much untouched.
2. column stats, something that can optionally be collected at the expense of planning time, and the populates a field here  in `IcebergDataSource` that is as minimal as possible, and that we use at `partition_statistics` call time for enriching the `Statistics` struct with this extra column-specific info.

iceberg/src/data_source.rs
```diff
@@ -309,33 +305,127 @@ impl DataSource for IcebergDataSource {
...
+    for manifest_file in manifest_list.entries() {
+        if manifest_file.content != ManifestContentType::Data {
+            continue;
+        }
+        let m_bytes = table
+            .file_io()
+            .new_input(&manifest_file.manifest_path)?
+            .read()
+            .await?;
+        let manifest = Manifest::parse_avro(&m_bytes)?;
```
gabotechs: I'm afraid doing this sequentially will be terribly slow. A trick for issuing concurrent IO for reading each manifest is to store a bunch of boxed futures in an `Vec`, and then call `try_join_all`.

sandugood: I've changed it to tokio's `JoinSet`

iceberg/src/data_source.rs
sandugood: Got rid of the idea of adding `Statistics` as a field to the `IcebergDataSource`

Now getting stats by calling `partition_statistics`  (so no precomputed values at the creation)

iceberg/tests/statistics.rs
```diff
@@ -110,6 +182,27 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn exact_row_count_lets_count_star_skip_the_scan_w_col_stats() -> Result<()> {
+        // With Precision::Exact(num_rows) the AggregateStatistics optimizer
+        // rule answers COUNT(*) from metadata without reading any data file.
+        let harness = test_harness_with_column_stats().await?;
+        let (plan, batches) = harness.query("SELECT count(*) FROM taxi").await?;
```
gabotechs, line 207: Nice!

iceberg/tests/statistics.rs
```diff
@@ -7,14 +7,21 @@ mod tests {
...
+    // Under `snapshots` key in the JSON
     const TAXI_ROWS: usize = 175_000;
     const TAXI_BYTES: usize = 4_480_382;
     const TAXI_COLUMNS: usize = 13;
 
+    async fn test_harness_with_column_stats() -> Result<IcebergTestHarness> {
+        let mut harness = IcebergTestHarness::new().await?;
+        harness.ctx.set_iceberg_column_stats_enabled(true);
+        Ok(harness)
+    }
```
gabotechs: We can skip this function, it's really just saving one line of code in each test, and it would be nice to see how this is explicitly enabled in each test that relies on column stats.

iceberg/src/data_source.rs
```diff
@@ -125,6 +131,8 @@ pub struct IcebergDataSource {
     partitioning: Partitioning,
     fetch: Option<usize>,
     metrics: ExecutionPlanMetricsSet,
+    column_stats: Option<Arc<HashMap<i32, ColumnStatistics>>>,
```
gabotechs: Why do we need to store this `projected_field_ids`? If it's just for remapping the `column_stats` to the correct position in the Arrow schema, I think we can just store the `column_stats` as something like this no?

```rust
    column_stats: Option<Vec<ColumnStatistics>>,
```

With already the correct position relative to the DataFusion schema.

iceberg/src/data_source.rs
```diff
@@ -181,8 +203,19 @@ impl IcebergDataSource {
...
             current_snapshot,
+            projected_field_ids: Arc::new(projected_field_ids),
+            column_stats: None,
         }
     }
+
+    pub(crate) async fn with_column_statistics(
+        mut self,
+        table: Table,
+    ) -> datafusion::common::Result<Self> {
```
gabotechs: For `datafusion::common::Result`, prefer just using a normal import:

```suggestion
    ) -> Result<Self> {
```

iceberg/src/data_source.rs
```diff
@@ -125,6 +131,8 @@ pub struct IcebergDataSource {
     partitioning: Partitioning,
     fetch: Option<usize>,
     metrics: ExecutionPlanMetricsSet,
+    column_stats: Option<Arc<HashMap<i32, ColumnStatistics>>>,
+    projected_field_ids: Arc<Vec<i32>>,
     current_snapshot: Option<SnapshotRef>,
```
gabotechs: It should be fine to just store these fields without `Arc`. Cloning these should be negligible.

iceberg/src/data_source.rs
```diff
@@ -333,9 +373,138 @@ fn stats_from_snapshot(
...
+            Ok::<_, iceberg::Error>(col_stats)
+        });
+    }
+
+    let mut merged_col_stats = vec![ColumnStatistics::new_unknown(); fields_ids.len()];
+    while let Some(result) = join_set.join_next().await {
+        let manifest_col_stats = result
+            .map_err(|e| DataFusionError::External(Box::new(e)))?
+            .map_err(|e| DataFusionError::External(Box::new(e)))?;
+        for (i, cs) in manifest_col_stats.into_iter().enumerate() {
```
gabotechs: For Iceberg errors, we should map them with the `df_err` function, so that they are recoverable to the original Iceberg error if the method needs to be sent across the wire.

iceberg/src/data_source.rs
```diff
@@ -333,9 +373,138 @@ fn stats_from_snapshot(
...
+                    if let Some(&n) = df.null_value_counts().get(&id) {
+                        cs.null_count = cs.null_count.add(&Precision::Exact(n as usize));
+                    }
+                    if let Some(scalar) = df.lower_bounds().get(&id).and_then(datum_to_scalar) {
+                        cs.min_value = cs.min_value.min(&Precision::Inexact(scalar));
+                    }
+                    if let Some(scalar) = df.upper_bounds().get(&id).and_then(datum_to_scalar) {
+                        cs.max_value = cs.max_value.max(&Precision::Inexact(scalar));
+                    }
+                }
```
gabotechs: We are missing the two most important pieces of stats here:
- `byte_size`
- `distinct_values`

Is there any chance we can get them as well?

sandugood: Not that I know of particularly, following the Iceberg spec: https://iceberg.apache.org/spec/#field-level-metrics-and-statistics

gabotechs: I see there do is a `df.column_sizes()` method available, but found nothing for the distinct values... that's a shame, that's actually pretty important for estimating aggregations, filter selectivity, etc...

iceberg/src/data_source.rs
sandugood: Yeah, thank you for the suggestion.

Mainly refactored that part + added a couple of helpers to reduce boilerplate.

iceberg/tests/statistics.rs
sandugood: Removed this func completely.
Just added +1 line in every test that needs column stats.

iceberg/src/data_source.rs
sandugood: Yeah, without reading the underlying .parquet footers I think we can't get that info, sadly

alexanderbianchi: https://iceberg.apache.org/puffin-spec/#apache-datasketches-theta-v1-blob-type
We actually can get distinct count metadata from the puffin spec. I know internally at datadog we write these. So sometimes it's available depending on the system.
 ```json
   {
     "statistics": [
       {
         "snapshot-id": 123,
         "statistics-path": ".../stats.puffin",
         "blob-metadata": [
           {
             "type": "apache-datasketches-theta-v1",
             "snapshot-id": 123,
             "fields": [4],
             "properties": {
               "ndv": "975"
             }
           }
         ]
       }
     ]
   }
 ```

iceberg/src/data_source.rs
```diff
@@ -333,9 +363,170 @@ fn stats_from_snapshot(
...
+
+/// Reading table statistics from data-files concurrently
+pub async fn compute_column_stats(
+    table: Table,
+    fields_ids: Vec<i32>,
+) -> Result<Vec<ColumnStatistics>> {
+    let metadata = table.metadata();
+
+    // A table with no current snapshot has never had a commit. It was created, but zero data files were added
+    let Some(snapshot) = metadata.current_snapshot() else {
```
alexanderbianchi:
```rust
  #[tokio::test]
    async fn reports_statistics_for_the_selected_snapshot() -> Result<()> {
        let harness = IcebergTestHarness::new().await?;
        harness
            .ctx
            .sql(&format!(
                "CREATE EXTERNAL TABLE historical_taxi STORED AS ICEBERG \
                 LOCATION '{FIXTURE_URI}/metadata/time-travel-statistics.metadata.json' \
                 OPTIONS ('iceberg.snapshot_id' '42')"
            ))
            .await?
            .collect()
            .await?;

        let stats = source_statistics(&harness, "SELECT * FROM historical_taxi").await?;

        assert_eq!(stats.num_rows, Precision::Exact(42));
        assert_eq!(stats.total_byte_size, Precision::Exact(4_242));
        Ok(())
    }
```
I pushed a test [here](https://github.com/alexanderbianchi/datafusion-distributed/pull/1) . I think `current_snapshot` is going to provide incorrect statistics on timetravel queries that specify a snapshot.
Something like below which I'm using in https://github.com/datafusion-contrib/datafusion-distributed/pull/684/changes#diff-db45f0402754aed52af0ebf1cf1d876b6e79575355efd47fd15ac2efe7fae482
```rust
 let selected_snapshot = match opts.snapshot_id {
       Some(id) => table.metadata().snapshot_by_id(id),
       None => table.metadata().current_snapshot(),
   }
```

iceberg/src/data_source.rs
gabotechs: Looks pretty good now 👍 thanks!

gabotechs (approved):
Looks good! thanks @sandugood, really nice work.

+1 on my side, let's address @alexanderbianchi's and we are good to pull this in

iceberg/src/data_source.rs
sandugood: Yeah, that makes total sense, thanks

sandugood:
Added the part of getting NDV from the `BlobMetadata`'s properties

alexanderbianchi (approved):
NDV changes look good, we probably need a consistent story for testing across snapshots + creating 1 off metadata to query, doesn't have to be this PR IMO.

alexanderbianchi:
Rebased with conflicts from my own PR and merged

Issue #688, opened 2026-09-02
feat: introduce spell check CI job


Rich-T-kid:
Add typos spell checking to CI to catch typos in source code, comments, and documentation before they merge.

Why:
Spell checking is currently being added to arrow-rs and proved straightforward to implement. It keeps typos out of docs and comments, and removes one more thing reviewers need to manually catch.

PR #689, opened 2026-09-02
Feat: introduce type checker ci action

Rich-T-kid:
see https://github.com/datafusion-contrib/datafusion-distributed/issues/688

mostly identical to how datafusion handles it.

### DIFF
--- a/typos.toml
+++ b/typos.toml
@@ -1,0 +1,27 @@
+[default.extend-words]
+# ASCII art diagram in work_unit_feed.rs truncates "RemotWorkUnitFeed" to fit a narrow box
+Remot = "Remot"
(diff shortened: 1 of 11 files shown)


Rich-T-kid:
not super pressing but I thought I'd add it here while its still fresh in my memory

typos.toml
```diff
@@ -0,0 +1,27 @@
+[default.extend-words]
+# ASCII art diagram in work_unit_feed.rs truncates "RemotWorkUnitFeed" to fit a narrow box
```
Rich-T-kid, line 2: these would be more trouble than its worth to change so I kept them

gabotechs (approved):
This is actually very nice, thanks @Rich-T-kid!

PR #690, opened 2026-09-03
fix: preserve unknown Iceberg snapshot statistics

alexanderbianchi:
## Summary

An existing snapshot may legally omit total-records or total-files-size because they are optional snapshot summary fields. Missing statistics mean unknown, not zero. Therefore they should map to Precision::Absent, while valid values remain Precision::Exact(value).

- A table with no snapshot is genuinely empty, so Exact(0) is correct. 
- report absent statistics when an existing snapshot omits `total-records` or `total-files-size`
- report malformed snapshot summary values as absent instead of exact zero

[Iceberg defines these snapshot summary metrics as optional.](https://iceberg.apache.org/spec/#optional-snapshot-summary-fields) Treating a missing metric as exact zero can produce incorrect metadata.

## Validation

- `cargo clippy -p datafusion-distributed-iceberg --all-targets -- -D warnings` *(blocked by the inherited `collapsible_if` warning in `iceberg/tests/statistics.rs`)*
- `git diff --check origin/iceberg-0.10...HEAD`


Rich-T-kid (approved):
makes sense to me.

nit: I think we should combine  `table_without_snapshot_has_exact_zero_statistics` & `unavailable_snapshot_summary_statistics_are_absent` into a single test.

iceberg/src/data_source.rs
```diff
@@ -326,16 +326,76 @@ fn stats_from_snapshot(
...
+    use super::*;
+
+    #[test]
+    fn table_without_snapshot_has_exact_zero_statistics() {
+        let statistics =
+            stats_from_snapshot(None, &empty_schema()).expect("statistics should be available");
+
+        assert_eq!(statistics.num_rows, Precision::Exact(0));
+        assert_eq!(statistics.total_byte_size, Precision::Exact(0));
+    }
```
gabotechs: The `stats_from_snapshot` function is private, therefore an implementation detail of the stats collection system subject to change.

Changes to the implementation detail will require changes to the tests, which can make maintenance harder. Any chance we can test this through the existing public test harness?

gabotechs (approved):
👍 just non-blocking comment

iceberg/src/data_source.rs
alexanderbianchi:
```rust
    #[test]
    fn unavailable_snapshot_summary_statistics_are_absent() {
        for value in [None, Some("not-a-number")] {
            let snapshot = snapshot_with_summary_value(value);
            let statistics = stats_from_snapshot(Some(&snapshot), &empty_schema())
                .expect("statistics should be available");

            assert_eq!(statistics.num_rows, Precision::Absent);
            assert_eq!(statistics.total_byte_size, Precision::Absent);
        }
    }

    fn snapshot_with_summary_value(value: Option<&str>) -> SnapshotRef {
        let additional_properties = value
            .map(|value| {
                HashMap::from([
                    (TOTAL_RECORDS.to_string(), value.to_string()),
                    (TOTAL_FILE_SIZE.to_string(), value.to_string()),
                ])
            })
            .unwrap_or_default();
        Arc::new(
            Snapshot::builder()
                .with_snapshot_id(1)
                .with_sequence_number(1)
                .with_timestamp_ms(0)
                .with_manifest_list("metadata/snapshot.avro")
                .with_summary(Summary {
                    operation: Operation::Append,
                    additional_properties,
                })
                .build(),
        )
    }
}
```
It was nice to be able to build a `SnapshotRef`, using the factory as it is we have to check in a new metadata file. I'll follow up with a factory API update in https://github.com/datafusion-contrib/datafusion-distributed/pull/700

PR #691, opened 2026-09-04
fix: make broadcast insertion idempotent

### LINKED ISSUE #675: Make insert_broadcast_execs idempotent for existing broadcast inputs
## Problem

`insert_broadcast_execs` always inserts a new `BroadcastExec` beneath an eligible join, even when the build input already has the canonical broadcast shape:

```text
CoalescePartitionsExec
  BroadcastExec
    input
```

For a fetch-less `CoalescePartitionsExec`, the pass unwraps the coalesce and uses its input as the payload for a new broadcast. If that input is already a `BroadcastExec`, the resulting plan becomes:

```text
CoalescePartitionsExec
  BroadcastExec
    BroadcastExec
      input
```

Applying the pass repeatedly adds another broadcast layer each time.

## Impact

The nested broadcasts preserve query results, but they create unnecessary cache state and duplicate memory reservations for the same batches. This increases memory pressure and can make spilling or resource-exhaustion failures more likely.

It also means the optimizer pass is not idempotent, which is surprising when plans have already been processed or when an earlier custom optimizer has inserted the canonical broadcast shape.

## Proposed behavior

Before inserting a broadcast, inspect the build input:

- If it is already `CoalescePartitionsExec(BroadcastExec(...))` with no fetch, leave it unchanged.
- If the build is directly a `BroadcastExec`, add only the required coalesce wrapper rather than another broadcast layer, if that shape is supported.
- Preserve the existing handling for fetch-bearing `CoalescePartitionsExec`, where the fetch must remain below the broadcast.

The exact normalization can vary, but a second application of `insert_broadcast_execs` should produce the same plan as the first.

## Testing

Add focused tests for:

- `HashJoinExec` with an existing canonical broadcast build.
- `NestedLoopJoinExec` with an existing canonical broadcast build.
- `CrossJoinExec` with an existing canonical broadcast build.
- A direct `BroadcastExec` build input, if supported.
- A fetch-bearing coalesce to ensure its limit remains below the broadcast.
- Applying `insert_broadcast_execs` twice and asserting that the second result is structurally unchanged.

Plan assertions should verify that every eligible build contains exactly one `BroadcastExec` layer.

## Acceptance criteria

- Existing canonical broadcast build inputs are not wrapped again.
- Repeated optimizer application is idempotent.
- HashJoin, NestedLoopJoin, and CrossJoin are covered.
- Fetch-bearing coalesce semantics are preserved.
- Existing broadcast planning and correctness tests continue to pass.

## References

- Broadcast insertion: `src/distributed_planner/insert_broadcast.rs`
- NLJ and CrossJoin broadcast support: #459 / #484

WilliamK112:
## Summary

- preserve an existing canonical CoalescePartitionsExec wrapping BroadcastExec
- add only the missing coalesce when the build side is already a direct BroadcastExec
- keep fetch-bearing coalesces below newly inserted broadcasts so global limit semantics remain unchanged
- cover repeated insertion for hash, nested-loop, and cross joins, plus direct-broadcast normalization

## Validation

- cargo fmt --all -- --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test --lib (299 passed, 1 ignored)

Fixes #675

## AI assistance disclosure

I used OpenAI Codex to help inspect the planner behavior, implement the change, and run validation. I reviewed the final diff and test results.

### DIFF
--- a/src/distributed_planner/insert_broadcast.rs
+++ b/src/distributed_planner/insert_broadcast.rs
@@ -134,18 +134,20 @@ pub(super) fn insert_broadcast_execs(
             return Ok(Transformed::no(node));
         };
 
-        let broadcast_input = build_child
+        let new_build_child: Arc<dyn ExecutionPlan> = if let Some(coalesce) = build_child
             .downcast_ref::<CoalescePartitionsExec>()
             .filter(|coalesce| coalesce.fetch().is_none())
-            .map_or_else(
-                || Arc::clone(build_child),
-                |coalesce| Arc::clone(coalesce.input()),
-            );
-
-        // consumer_task_count=1 is a placeholder and will be corrected during optimizer rule.
-        let broadcast: Arc<dyn ExecutionPlan> = Arc::new(BroadcastExec::new(broadcast_input, 1));
-        let new_build_child: Arc<dyn ExecutionPlan> =
-            Arc::new(CoalescePartitionsExec::new(broadcast));
+        {
+            if coalesce.input().downcast_ref::<BroadcastExec>().is_some() {
+                return Ok(Transformed::no(node));
+            }
+            coalesced_broadcast(Arc::clone(coalesce.input()))
+        } else if build_child.downcast_ref::<BroadcastExec>().is_some() {
+            Arc::new(CoalescePartitionsExec::new(Arc::clone(build_child)))
+        } else {
+            // A fetch-bearing coalesce remains below the broadcast so the limit is global.
+            coalesced_broadcast(Arc::clone(build_child))
+        };
 
         let mut new_children: Vec<Arc<dyn ExecutionPlan>> = children.into_iter().cloned().collect();
         new_children[0] = new_build_child;
@@ -155,6 +157,12 @@ pub(super) fn insert_broadcast_execs(
         )?))
     })
     .map(|transformed| transformed.data)
+}
+
+fn coalesced_broadcast(input: Arc<dyn ExecutionPlan>) -> Arc<dyn ExecutionPlan> {
+    // consumer_task_count=1 is a placeholder and will be corrected during optimizer rule.
+    let broadcast: Arc<dyn ExecutionPlan> = Arc::new(BroadcastExec::new(input, 1));
+    Arc::new(CoalescePartitionsExec::new(broadcast))
 }
 
 fn can_broadcast_left_input(plan: &dyn ExecutionPlan) -> bool {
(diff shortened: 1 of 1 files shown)


src/distributed_planner/insert_broadcast.rs
```diff
@@ -348,15 +410,64 @@ mod tests {
         broadcast_enabled: bool,
         target_partitions: usize,
     ) -> String {
+        sql_to_plan_with_broadcast_applications(query, broadcast_enabled, target_partitions, 1)
+            .await
+    }
+
+    async fn sql_to_plan_with_broadcast_applications(
```
gabotechs: We should avoid adding many function variants with overlapping intentions for the tests, as we are at risk of maintaining one function signature per permutation of config values.

Rather than having several helper functions here, we should just rely on `TestPlanBuilder`'s builder-pattern to provide different tests with different configurations.

src/distributed_planner/insert_broadcast.rs
```diff
@@ -348,15 +410,64 @@ mod tests {
...
+            query,
+            broadcast_enabled,
+            target_partitions,
+            applications,
+        )
+        .await;
+        format!("{}", displayable(plan.as_ref()).indent(true))
+    }
+
+    async fn plan_with_broadcast_applications(
```
gabotechs: Same as above, let's aim for not having multiple functions here.

gabotechs (commented):
Looks good! mainly left a suggestion about tests.

src/distributed_planner/insert_broadcast.rs
```diff
@@ -286,14 +294,21 @@ mod tests {
...
+            .left()
+            .downcast_ref::<CoalescePartitionsExec>()
+            .expect("expected coalesce above broadcast");
+        let broadcast = coalesce
+            .input()
+            .downcast_ref::<BroadcastExec>()
+            .expect("expected broadcast below coalesce");
+
+        assert_eq!(coalesce.fetch(), None);
+        assert_eq!(broadcast.input().fetch(), Some(50));
```
gabotechs: Now that https://github.com/datafusion-contrib/datafusion-distributed/pull/698 is merged, this change should not be necessary.

WilliamK112:
Addressed the remaining review threads and refreshed onto current main in `b6932e9`.

- Restored the fetch-bearing snapshot now that #698 provides stable `<shard>` normalization, so the interim structural assertion is gone.
- Reworked the test setup to declare configuration through `TestPlanBuilder` at each call site.
- Removed the three overlapping helper variants; only one helper remains for the behavior under test (applying the rule repeatedly to an already-built test plan).
- Preserved the idempotence coverage for hash, nested-loop, and cross joins plus direct-broadcast normalization.

Local validation passed: all 10 `insert_broadcast` module tests, `cargo fmt --all -- --check`, and `cargo clippy --all-targets --all-features -- -D warnings`. The branch now merges cleanly with current main; hosted CI is running.

src/distributed_planner/insert_broadcast.rs
```diff
@@ -328,13 +378,91 @@ mod tests {
...
+            let once = displayable(once.as_ref()).indent(true).to_string();
+            let twice = displayable(twice.as_ref()).indent(true).to_string();
+
+            assert_eq!(once, twice);
+            assert_eq!(once.matches("BroadcastExec").count(), 1);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_insert_broadcast_wraps_direct_broadcast_without_nesting() {
```
gabotechs, line 417: This test is a bit complex and I don't see it providing extra coverage, let's delete this one and keep the rest.

src/distributed_planner/insert_broadcast.rs
```diff
@@ -134,18 +134,20 @@ pub(super) fn insert_broadcast_execs(
...
-        // consumer_task_count=1 is a placeholder and will be corrected during optimizer rule.
-        let broadcast: Arc<dyn ExecutionPlan> = Arc::new(BroadcastExec::new(broadcast_input, 1));
-        let new_build_child: Arc<dyn ExecutionPlan> =
-            Arc::new(CoalescePartitionsExec::new(broadcast));
+        {
+            if coalesce.input().downcast_ref::<BroadcastExec>().is_some() {
+                return Ok(Transformed::no(node));
+            }
+            coalesced_broadcast(Arc::clone(coalesce.input()))
+        } else if build_child.downcast_ref::<BroadcastExec>().is_some() {
```
gabotechs, line 145:
```suggestion
            if coalesce.input().is::<BroadcastExec>() {
                return Ok(Transformed::no(node));
            }
            coalesced_broadcast(Arc::clone(coalesce.input()))
        } else if build_child.is::<BroadcastExec>() {
```

gabotechs (approved):
Nice work, thanks @WilliamK112!

src/distributed_planner/insert_broadcast.rs
```diff
@@ -328,13 +378,91 @@ mod tests {
...
+        assert_eq!(plan.matches("BroadcastExec").count(), 1);
+        assert_snapshot!(plan, @"
+        HashJoinExec: mode=CollectLeft, join_type=Inner, on=[(RainToday@1, RainToday@1)], projection=[MinTemp@0, MaxTemp@2]
+          CoalescePartitionsExec
+            BroadcastExec: input_partitions=1, consumer_tasks=1, output_partitions=1
+              DataSourceExec: file_groups={1 group: [[/testdata/weather/result-000000.parquet, /testdata/weather/result-000001.parquet, /testdata/weather/result-000002.parquet]]}, projection=[MinTemp, RainToday], file_type=parquet
+          DataSourceExec: file_groups={1 group: [[/testdata/weather/result-000000.parquet, /testdata/weather/result-000001.parquet, /testdata/weather/result-000002.parquet]]}, projection=[MaxTemp, RainToday], file_type=parquet, predicate=DynamicFilter [ empty ], dynamic_rg_pruning=eligible
+        ");
+    }
+
```
gabotechs:
```suggestion
```

gabotechs (approved):

gabotechs:
Thanks @WilliamK112!

PR #692, opened 2026-09-04
Add async recursion helpers

gabotechs:
Adds a new recursion helper that resembles `.transform_up()` but with an `async` API:

```rust
    async fn transform_up_async<F, Fut>(self, f: F) -> Result<Transformed<Self>>
    where
        Self: Sized,
        F: Fn(Self) -> Fut + Send + Sync,
        Fut: Future<Output = Result<Transformed<Self>>> + Send;
```

To be used in future PRs in this same stack.

---

### DIFF
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -4,10 +4,39 @@ use datafusion::common::Result;
 use datafusion::common::tree_node::{Transformed, TreeNode, TreeNodeIterator, TreeNodeRecursion};
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::joins::{CrossJoinExec, HashJoinExec, NestedLoopJoinExec};
+use futures::future::{BoxFuture, try_join_all};
 use std::cell::RefCell;
+use std::future::Future;
 use std::sync::Arc;
 
 pub(crate) trait TreeNodeExt {
+    /// Recursively rewrite the node using `f` in a bottom-up (post-order)
+    /// fashion, awaiting `f` for each node.
+    ///
+    /// `f` is applied to the node's children first, and then to the node itself.
+    /// Sibling subtrees are transformed concurrently.
+    ///
+    /// The returned [`TreeNodeRecursion`] steers the traversal as in
+    /// [`TreeNode::transform_up`]: both [`TreeNodeRecursion::Jump`] and
+    /// [`TreeNodeRecursion::Stop`] prevent `f` from being applied to *any* ancestor of the
+    /// node that returned them.
+    ///
+    /// Because sibling subtrees run concurrently, [`TreeNodeRecursion::Stop`] cannot prune
+    /// them: unlike [`TreeNode::transform_up`], `f` has already been applied to the sibling
+    /// subtrees by the time the stop is observed, and their rewrites are then discarded. The
+    /// resulting plan is identical to [`TreeNode::transform_up`]'s, but `f` must not rely on
+    /// `Stop` to suppress its own side effects.
+    ///
+    /// Note that every child subtree is polled concurrently with no fan-out limit, so an `f`
+    /// that performs I/O will issue up to one request per node of the widest tree level at a
+    /// time.
+    #[allow(dead_code)]
+    async fn transform_up_async<F, Fut>(self, f: F) -> Result<Transformed<Self>>
+    where
+        Self: Sized,
+        F: Fn(Self) -> Fut + Send + Sync,
+        Fut: Future<Output = Result<Transformed<Self>>> + Send;
+
     /// Applies `f` to the node then each of its children, recursively (a
     /// top-down, pre-order traversal), propagating the [DistributedTaskContext] correctly
     /// across nodes that mutate this context, and ignoring nodes that do not belong to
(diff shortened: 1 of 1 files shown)


src/common/recursion.rs
```diff
@@ -4,10 +4,39 @@ use datafusion::common::Result;
...
+    /// The returned [`TreeNodeRecursion`] steers the traversal as in
+    /// [`TreeNode::transform_up`]: both [`TreeNodeRecursion::Jump`] and
+    /// [`TreeNodeRecursion::Stop`] prevent `f` from being applied to *any* ancestor of the
+    /// node that returned them.
+    ///
+    /// Because sibling subtrees run concurrently, [`TreeNodeRecursion::Stop`] cannot prune
+    /// them: unlike [`TreeNode::transform_up`], `f` has already been applied to the sibling
+    /// subtrees by the time the stop is observed, and their rewrites are then discarded. The
+    /// resulting plan is identical to [`TreeNode::transform_up`]'s, but `f` must not rely on
+    /// `Stop` to suppress its own side effects.
```
jayshrivastava, line 29: Subtle... but correct.

src/common/recursion.rs
```diff
@@ -269,14 +340,295 @@ mod tests {
...
+                    if Arc::ptr_eq(p, &jumper) {
+                        TreeNodeRecursion::Jump
+                    } else {
+                        TreeNodeRecursion::Continue
+                    }
+                }),
+            )
+            .await,
+            @r"
+        transformed=true tnr=Continue
```
jayshrivastava: Is it more useful to label which nodes were transformed than to put it at the top?

src/common/recursion.rs
```diff
@@ -269,14 +340,295 @@ mod tests {
...
+        CoalescePartitionsExec
+          UnionExec
+            CoalescePartitionsExec
+              CoalescePartitionsExec
+                EmptyExec
+            CoalescePartitionsExec
+              CoalescePartitionsExec
+                CoalescePartitionsExec
```
jayshrivastava: Why are these Coalesces here?

src/common/recursion.rs
```diff
@@ -269,14 +340,295 @@ mod tests {
...
+        .await
+        .lines()
+        .map(str::to_string)
+        .collect::<Vec<_>>();
+        // Concurrent siblings complete in a non-deterministic order.
+        visited.sort();
+        assert_snapshot!(visited.join("\n"), @r"
+        Leaf
+        Leaf [->stop]
+        Single
```
jayshrivastava, line 381: I feel like labelling each node with `transformed=true` or `transformed=false` makes this behavior more clear

src/common/recursion.rs
```diff
@@ -269,14 +340,295 @@ mod tests {
...
+        .await, @"Leaf [->jump]");
+    }
+
+    #[tokio::test]
+    async fn transform_up_async_sees_rewritten_children() {
+        // `f` must observe the children that its own recursive invocations produced.
+        let observed = Mutex::new(vec![]);
+        single(leaf())
+            .transform_up_async(async |plan| {
+                observed.lock().unwrap().push(format!(
```
jayshrivastava: This is fine but maybe saving the original child pointer and then asserting here that the pointer is different makes more sense.

src/common/recursion.rs
gabotechs: I think this is unlikely to bite us in the future. If it does, at least it's documented here.

src/common/recursion.rs
gabotechs: When we call `single()` here, it's CoalescePartitionsExec the node that gets used for representing a node with a single child,  just because it's easy to construct. 

`assert_matches_sync` will insert these for asserting against normal `.transform_up()`.

This is a bit messy though... I recognize that I might have let LLMs run too wildly when writing this tests, I'm going to rewrite them from scratch, I think they can be made way better.

src/common/recursion.rs
gabotechs: I'm going to rework this tests from scratch, hopefully this will become better then.

src/common/recursion.rs
gabotechs: Yeah, agree, I've reworked all this, let me know if this looks better

src/common/recursion.rs
gabotechs: This got a bit messy, I've reworked all this. The tests are now much more consistent with the ones we had before.

src/common/recursion.rs
```diff
@@ -4,10 +4,39 @@ use datafusion::common::Result;
...
+    /// Recursively rewrite the node using `f` in a bottom-up (post-order)
+    /// fashion, awaiting `f` for each node.
+    ///
+    /// `f` is applied to the node's children first, and then to the node itself.
+    /// Sibling subtrees are transformed concurrently.
+    ///
+    /// The returned [`TreeNodeRecursion`] steers the traversal as in
+    /// [`TreeNode::transform_up`]: both [`TreeNodeRecursion::Jump`] and
+    /// [`TreeNodeRecursion::Stop`] prevent `f` from being applied to *any* ancestor of the
+    /// node that returned them.
```
jayshrivastava: I would fix the wording here. "any ancestor" is technically wrong for `Jump`. The behavior is technically different from `Stop` because of `map_children` implementation:
```rust
fn map_children<F: FnMut(Self) -> Result<Transformed<Self>>>(
        self,
        f: F,
    ) -> Result<Transformed<Self>> {
        let children = self.arc_children();
        if !children.is_empty() {
            let new_children = children
                .into_iter()
                .cloned()
                .map_until_stop_and_collect(f)?;
```

We see this difference appear in your tests.

Ex. In this example, the union is transformed:
```
union
  single
     leaf <- jump
  single
      leaf
```
However in this example, the union will not be transformed:
```
union
  single
     leaf <- stop
  single
      leaf
```

jayshrivastava (approved):
Nice. This is much more clear. Just one comment to fix the comment to document the `Jump` behavior.

src/common/recursion.rs
gabotechs: reworded that section.

PR #693, opened 2026-09-04
Add ok_or_some_err macro

gabotechs:
Adds a new `ok_or_some_err!` macro to be used in event handlers that return
`Option<Result<T>>`.

Passing from something like this:

```rust
let value = match result {
    Ok(value) => value,
    Err(err) => return Some(Err(err))
};
```

To this

```rust
let value = ok_or_some_err!(result);
```

---


jayshrivastava (approved):

jayshrivastava:
I feel like you can merge this into main without the stack.

PR #694, opened 2026-09-04
Add dialer-based RouteTaskHandler

gabotechs:
Rewrite of the task routing API that unlocks the door for implementing retries and smarter routing.

# Key changes to the task-routing API:

## Fire on a per-task basis instead of per-stage

Previously, this API was fired once for each stage, and the expected returns where a list of URLs (one per task in the stage):

```
Original ExecutionPlan + 3 tasks -> [URL1, URL2, URL3]
```

Now, this is called once per task, and the passed plan is the task specialized one, not the original one.

```
Task-specialized ExecutionPlan + task index -> URL1 + Stablished connection
```

## Returns a successfully established connection

There was a problem with the previous task routing API:

It prompted users and/or default implementations to return a URL, without knowing if something is listening at the other side of the URL.

There is a chance that the routed machine is unavailable, in which case, the new API gives the option to implementations to retry on a different machine.

## The event handler is now `async`

Connections must be established in this event handler, which means that it must make network calls, therefore it's now made `async`.

## Introduction of `CoordinatorToWorkerDialer`

The `CoordinatorToWorkerDialer` is a trait that users are **not** supposed to implement, instead, they are supposed to call its `.dial()` method for attempting to establish a connection from coordinator to remote worker, so implementations of the `RouteTaskHandler` would look like this:

```rust
#[async_trait]
impl RouteTaskHandler for RandomRouteTaskHandler {
    async fn handle(&self, ev: RouteTaskEvent<'_>) -> Option<Result<RouteTaskEventResponse>> {
        // get a URL somehow
        
        Some(ev.dialer.dial(url.clone()).await)
    }
}
```

If there's an error while dialing the worker, the user can choose to retry on another URL.

The `CoordinatorToWorkerDialer` under the hood will handle all the API churn for making API calls to workers, leaving a clean and narrow API to the routing task event handler implementor, just requiring the URL to which it needs to connect to.

# What's not included yet

Any kind of retry mechanism on a different node in case of a connection failure.

---

### DIFF
--- a/docs/source/advanced/06-worker-routing.md
+++ b/docs/source/advanced/06-worker-routing.md
@@ -1,35 +1,68 @@
 # Routing tasks to workers
 
-By default, the planner spreads a stage's tasks across the available workers
-round-robin. When a task's data has a *home* — a worker that already holds it in a
-cache or on local disk — you can send the task **there** instead, so it reads
-locally instead of pulling data over the network.
+By default, each distributed task is routed to an available worker. When a
+task's data has a *home* (e.g, a worker that already holds it in a cache or on
+local disk) a custom routing handler can send the task there instead. As routing
+handlers are responsible for establishing connection to remote workers, they are
+`async`.
 
-Routing is handled by a registered `RouteTasksHandler`. It receives a
-`RouteTasksEvent` (the head plan of the stage, the task count, and the active
-`TaskContext`) and returns one worker URL per task, in task order:
+Implement `RouteTaskHandler` and register it on the coordinating session. The
+handler is called once per task with a `RouteTaskEvent`, providing contextual
+information about what task-specialized plan is getting routed, the task
+identifier to which its routed, etc..
+
+Return `None` when the handler does not apply, allowing the next custom or
+built-in handler to run. Otherwise, call `dialer.dial(url).await` and return the
+response for the selected connection. The dialer may be called more than once,
+sequentially or concurrently, to implement retries.
+
+`dialer.dial(url).await` connects to a remote worker under the hood, and returns
+the already established connection. If this call succeeds, it means that the
+worker is in a good state for being part of the query.
 
 ```rust
-fn route_tasks(event: RouteTasksEvent) -> Option<Result<RouteTasksEventResponse>>;
-```
+use async_trait::async_trait;
+use datafusion::common::{Result, exec_err};
+use datafusion::execution::SessionStateBuilder;
+use datafusion_distributed::{
+    DistributedExt, RouteTaskEvent, RouteTaskEventResponse, RouteTaskHandler,
+    ok_or_some_err
+};
 
-- `Some(Ok(RouteTasksEventResponse::new(urls)))` — task `i` is sent to `urls[i]`.
-- `None` — defer to the next handler; the built-in fallback keeps the round-robin behaviour.
+struct RetryRouteTaskHandler;
 
-Register it on the coordinating session builder:
+#[async_trait]
+impl RouteTaskHandler for RetryRouteTaskHandler {
+    async fn handle(&self, event: RouteTaskEvent<'_>) -> Option<Result<RouteTaskEventResponse>> {
+        let urls = match ok_or_some_err!(event.worker_resolver.get_urls());
+        if urls.is_empty() {
+            return Some(exec_err!("no workers available"));
+        }
 
-```rust
+        let start = event.task_key.task_number % urls.len();
+        let mut last_error = None;
+        for offset in 0..urls.len() {
+            let url = urls[(start + offset) % urls.len()].clone();
+            match event.dialer.dial(url).await {
+                Ok(response) => return Some(Ok(response)),
+                Err(error) => last_error = Some(error),
+            }
+        }
+        Some(Err(last_error.expect("at least one worker was attempted")))
+    }
+}
+
 SessionStateBuilder::new()
-    .with_distributed_route_tasks_handler(route_tasks);
+    .with_distributed_route_task_handler(RetryRouteTaskHandler);
 ```
 
 Routing pairs naturally with `ScaleUpLeafNodeHandler`: that decides *what* data
-task `i` reads, and `RouteTasksHandler` decides *where* it runs. If the leaf-scale
-handler returns a `DistributedLeafExec`, its `variants()` are in task order too,
-so you can line up each task's data with the worker that should serve it.
+task `i` reads, and `RouteTaskHandler` decides *where* that specialized task
+runs. The task index can be used to keep a stable slot-to-worker mapping for
+cache affinity.
 
-For a complete, runnable walkthrough — parquet files consistently routed to
-workers by rendezvous hashing of the file path, so each worker can serve them from
-an in-memory cache on repeat queries — see the
+For a complete, runnable walkthrough where parquet files consistently routed to
+workers by hashing the file path, so each worker can serve them from an
+in-memory cache on repeat queries, see the
 [custom_worker_url_routing.rs](https://github.com/datafusion-contrib/datafusion-distributed/blob/main/examples/custom_worker_url_routing.rs)
 example.
--- a/docs/upgrade/5.0.0.md
+++ b/docs/upgrade/5.0.0.md
@@ -1,0 +1,80 @@
+# Upgrading from 4.0.0 to 5.0.0
+
+This guide covers the source and behavioural changes merged after the
+`v4.0.0` tag.
+
+## 1. Task routing is asynchronous and runs once per task
+
+The stage-level `RouteTasksHandler` API has been replaced by the asynchronous,
+per-task `RouteTaskHandler` API:
+
+| 4.0 API                                      | 5.0 API                                     |
+|----------------------------------------------|---------------------------------------------|
+| `RouteTasksHandler`                          | `RouteTaskHandler`                          |
+| `RouteTasksEvent`                            | `RouteTaskEvent`                            |
+| `RouteTasksEventResponse`                    | `RouteTaskEventResponse`                    |
+| `{with,set}_distributed_route_tasks_handler` | `{with,set}_distributed_route_task_handler` |
+
+Previously, a handler ran once per stage and returned one worker URL for each
+task:
+
+```rust
+fn route_tasks(
+    event: RouteTasksEvent<'_>,
+) -> Option<Result<RouteTasksEventResponse>> {
+    let mut urls = event
+        .task_ctx
+        .session_config()
+        .get_distributed_worker_resolver()
+        .ok()?
+        .get_urls()
+        .ok()?;
+    urls.truncate(event.task_count);
+    Some(Ok(RouteTasksEventResponse::new(urls)))
+}
+
+let state = SessionStateBuilder::new()
+    .with_distributed_route_tasks_handler(route_tasks)
+    .build();
+```
+
+In 5.0, implement the async handler and select and dial one worker for the task
+described by the event:
+
+```rust
+use async_trait::async_trait;
+use datafusion::common::Result;
+use datafusion_distributed::{
+    DistributedExt, RouteTaskEvent, RouteTaskEventResponse, RouteTaskHandler,
+};
+
+struct MyRouteTaskHandler;
+
+#[async_trait]
+impl RouteTaskHandler for MyRouteTaskHandler {
+    async fn handle(
+        &self,
+        event: RouteTaskEvent<'_>,
+    ) -> Option<Result<RouteTaskEventResponse>> {
+        let urls = match event.worker_resolver.get_urls() {
+            Ok(urls) => urls,
+            Err(error) => return Some(Err(error)),
+        };
+        if urls.is_empty() {
+            return None;
+        }
+
+        let url = urls[event.task_key.task_number % urls.len()].clone();
+        Some(event.dialer.dial(url).await)
+    }
+}
+
+let state = SessionStateBuilder::new()
+    .with_distributed_route_task_handler(MyRouteTaskHandler)
+    .build();
+```
+
+`event.dialer.dial(url).await` is expected to be used inside the
+`RouteTaskHandler` implementation, therefore a worker is only chosen if a  
+coordinator-to-worker connection is currently established. Implementations can
+choose to retry on other worker if the current one is not available.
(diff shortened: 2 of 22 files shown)


docs/source/advanced/06-worker-routing.md
gabotechs, line 1: @JSOD11 would love some feedback about the new shape of this API.

docs/source/advanced/06-worker-routing.md
```diff
@@ -1,35 +1,68 @@
...
 
-By default, the planner spreads a stage's tasks across the available workers
-round-robin. When a task's data has a *home* — a worker that already holds it in a
-cache or on local disk — you can send the task **there** instead, so it reads
-locally instead of pulling data over the network.
+By default, each distributed task is routed to an available worker. When a
+task's data has a *home* (e.g, a worker that already holds it in a cache or on
+local disk) a custom routing handler can send the task there instead. As routing
+handlers are responsible for establishing connection to remote workers, they are
+`async`.
```
jayshrivastava, line 7: Just dumping this thought here at the start of my review.

One thing we're solving in our use case is how to get per-partition consistent routing. 

Say there's 9 partitions and 3 workers (a-c). Each partition may desire workers cacbacaab. In this case, it's optimal to rearrange the partitions so that they are grouped aaa abb ccc (so partitions desiring worker a are put in the same task and routed to worker a).

This can be done today as a caller/user of datafusion-distributed, but if possible, it would be nice to easily accommodate this natively.

docs/source/advanced/06-worker-routing.md
```diff
@@ -1,35 +1,68 @@
...
+    ok_or_some_err
+};
 
-- `Some(Ok(RouteTasksEventResponse::new(urls)))` — task `i` is sent to `urls[i]`.
-- `None` — defer to the next handler; the built-in fallback keeps the round-robin behaviour.
+struct RetryRouteTaskHandler;
 
-Register it on the coordinating session builder:
+#[async_trait]
+impl RouteTaskHandler for RetryRouteTaskHandler {
```
jayshrivastava, line 35: Nice.

src/distributed_ext.rs
```diff
@@ -659,40 +654,37 @@ pub trait DistributedExt: Sized {
...
+    /// struct AssignToWorker;
+    ///
+    /// #[async_trait]
+    /// impl RouteTaskHandler for AssignToWorker {
+    ///     async fn handle(
+    ///         &self,
+    ///         event: RouteTaskEvent<'_>,
+    ///     ) -> Option<Result<RouteTaskEventResponse>> {
+    ///         let url = Url::parse("http://worker1:50051").unwrap();
+    ///         Some(event.dialer.dial(url).await)
```
jayshrivastava: Worth using the worker resolver in this example IMO.

src/test_utils/routing.rs
```diff
@@ -292,19 +291,18 @@ pub fn url_emitter_scale_up_leaf_node(
...
+        let mut urls = ok_or_some_err!(ev.worker_resolver.get_urls());
 
         // Trivial routing policy: Assign tasks to URLs in reverse order.
-        routed_urls.reverse();
-        routed_urls.truncate(ev.task_count);
-        Ok(RouteTasksEventResponse::new(routed_urls))
-    })())
+        urls.reverse();
+        let url = urls.into_iter().nth(ev.task_key.task_number)?;
+        Some(ev.dialer.dial(url).await)
```
jayshrivastava, line 304: Clean

src/protocol/worker_channel.rs
```diff
@@ -64,6 +64,7 @@ pub struct TaskKey {
     pub task_number: usize,
 }
 
+#[derive(Clone)]
```
jayshrivastava: This can be copy. UUIDs are `[u8; 16]` which are copy.

src/events/route_tasks.rs
```diff
@@ -1,79 +1,107 @@
...
     /// The task context active for the query being coordinated.
-    pub task_ctx: Arc<TaskContext>,
-    /// The head execution plan of the stage whose tasks are being routed.
-    /// WARNING: this is never going to be a custom leaf node, this is the head node of the fragment
-    ///  of the plan that contains the custom leaf node.
-    pub plan: &'a Arc<dyn ExecutionPlan>,
-    /// The number of task slots that need a worker assignment.
+    pub task_ctx: &'a TaskContext,
+    /// [WorkerResolver] in scope.
+    pub worker_resolver: &'a dyn WorkerResolver,
```
jayshrivastava, line 17: You can get the resolver like so, so this feels unecessary.
```rust
let local_worker_context = ev
    .task_ctx
    .session_config()
    .get_distributed_worker_resolver();
```

src/events/route_tasks.rs
```diff
@@ -1,79 +1,107 @@
...
-    /// [`RouteTasksEvent::handle`].
-    pub fn new(urls: Vec<Url>) -> Self {
-        Self { urls }
-    }
+/// A worker connection selected by a [`RouteTaskHandler`].
+pub struct RouteTaskEventResponse {
+    /// The URL of the selected worker.
+    pub url: Url,
+    /// Messages streamed from the selected worker to the coordinator.
+    pub worker_to_coordinator_stream: BoxStream<'static, Result<WorkerToCoordinatorMsg>>,
```
jayshrivastava, line 42: I feel like this doesn't need comments on fields.

src/events/defaults/routing.rs
```diff
@@ -1,71 +1,91 @@
...
     }
+}
 
-    let start_idx = rand::rng().random_range(0..available_urls.len());
-    let urls = (0..ev.task_count)
-        .map(|i| available_urls[(start_idx + i) % available_urls.len()].clone())
-        .collect::<Vec<_>>();
+/// Chooses an item using a random starting offset shared by every task in this stage.
+/// Successive task numbers rotate through the list from that offset.
+fn stage_contiguous_rand_choice<T>(key: &TaskKey, list: impl IntoIterator<Item = T>) -> Option<T> {
```
jayshrivastava: nit: `stage_contiguous_rand_offset` is better imo

src/events/defaults/routing.rs
```diff
@@ -1,71 +1,91 @@
...
+    }
 
-    Some(Ok(RouteTasksEventResponse::new(urls)))
+    let mut seed = [0; 32];
+    seed[..16].copy_from_slice(key.query_id.as_bytes());
+    seed[16..24].copy_from_slice(&(key.stage_id as u64).to_le_bytes());
+    let mut rng = StdRng::from_seed(seed);
+    let start = rng.random_range(0..list.len());
+    let index = (start + key.task_number % list.len()) % list.len();
+    list.into_iter().nth(index)
```
jayshrivastava, line 38: 😆 

Smart

docs/source/advanced/06-worker-routing.md
jayshrivastava: Oh wow that's basically what the `CachedFileScanRouteTaskHandler` does.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -80,27 +80,30 @@ pub(super) async fn prepare_dynamic_plan(
...
-
+            let query_coordinator = Arc::clone(query_coordinator);
             let plans_for_viz = Arc::clone(&plans_for_viz);
+
             Ok(async move {
+                let mut stage_coordinator = query_coordinator.stage_coordinator(&input_stage);
+
+                let mut futures = Vec::with_capacity(input_stage.tasks);
+                for task_i in 0..input_stage.tasks {
+                    futures.push(stage_coordinator.send_plan_task(task_i));
```
jayshrivastava: I feel like `send_plan_task` should just be called `init_bidirectional_stream` and the comments say the first message sends the plan from coord to worker.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -80,27 +80,30 @@ pub(super) async fn prepare_dynamic_plan(
...
+                }
+                let results = futures::future::try_join_all(futures).await?;
+
+                let mut workers = Vec::with_capacity(input_stage.tasks);
+                for (task_i, (url, worker_tx, worker_rx)) in results.into_iter().enumerate() {
+                    workers.push(url);
+                    load_info_rxs.push({
+                        let rx = stage_coordinator.worker_to_coordinator_task(task_i, worker_rx);
```
jayshrivastava, line 101: Similarly `worker_to_coordinator_task` should just be the `load_info_collection_task` or something.

src/coordinator/query_coordinator.rs
```diff
@@ -151,66 +154,104 @@ impl<'a> StageCoordinator<'a> {
...
+        let metrics = self.metrics.clone();
+        let metrics_set = self.metrics_set.clone();
+        // Stores the last coordinator_to_worker_tx that was attempted for establishing a
+        // connection with the remote worker. If dialing a remote worker fails, and it's retried,
+        // this will hold the channel belonging to the last retry
+        let coordinator_to_worker_tx_slot = Mutex::new(None);
+
+        let dialer = new_coordinator_to_worker_dialer(|url| {
+            let (coordinator_to_worker_tx, coordinator_to_worker_rx) =
+                tokio::sync::mpsc::unbounded_channel();
```
jayshrivastava, line 178: nit: import `tokio::sync::mpsc::unbounded_channel` at the top

jayshrivastava (approved):
Minor comments. Otherwise, things look good.

docs/source/advanced/06-worker-routing.md
JSOD11: Super cool stuff

docs/upgrade/5.0.0.md
```diff
@@ -0,0 +1,80 @@
...
+use async_trait::async_trait;
+use datafusion::common::Result;
+use datafusion_distributed::{
+    DistributedExt, RouteTaskEvent, RouteTaskEventResponse, RouteTaskHandler,
+};
+
+struct MyRouteTaskHandler;
+
+#[async_trait]
+impl RouteTaskHandler for MyRouteTaskHandler {
```
JSOD11, line 54: I like this new API and that it's per task. Looks very nice

src/coordinator/prepare_dynamic_plan.rs
JSOD11: Is it a bidirectional stream? What's being streamed to worker, I thought we just send the index here?

src/events/defaults/routing.rs
JSOD11: I think choice is also okay, the caller doesn't necessarily assume or need to know that this method is round-robin over the list. But I like both

src/events/defaults/routing.rs
JSOD11: 😮‍💨

src/events/defaults/routing.rs
```diff
@@ -1,71 +1,91 @@
...
-    let available_urls = ok_or_some_err!(worker_resolver.get_urls());
+#[async_trait]
+impl RouteTaskHandler for RandomRouteTaskHandler {
+    async fn handle(&self, ev: RouteTaskEvent<'_>) -> Option<Result<RouteTaskEventResponse>> {
+        let urls = ok_or_some_err!(ev.worker_resolver.get_urls());
+        let url = stage_contiguous_rand_choice(&ev.task_key, &urls)?;
 
-    if available_urls.is_empty() {
-        return Some(exec_err!("0 URLs available during routing"));
+        Some(ev.dialer.dial(url.clone()).await)
```
JSOD11, line 20: I see that we usually dial and if the URL isn't available we return an error. Is it worth falling back to a different URL rather than failing? Or if a user wants this then they should just implement it

JSOD11 (approved):
Looks great to me!

src/coordinator/prepare_dynamic_plan.rs
gabotechs: It's bidirectional yes. We stream `WorkUnit`s to the worker in case of using `WorkUnitFeed`s.

About the naming: yes, agree, this has evolved quite a few times already, originally, this returned a spawned task, however it's been a few iterations where this has not been the case.

As this PR is significantly changing the signature of that function, I think we are good to do the renaming right now.

gabotechs: Done!

src/coordinator/prepare_dynamic_plan.rs
gabotechs: :thinking: this one is a bit more subtle. It does not only handle load info collection, it also spawns the task that gathers the metrics and the dynamic filters, although it does it under the hood.

src/coordinator/query_coordinator.rs
gabotechs: This one might look like the classical LLM import, but I actually wrote all this manually.

We typically use fully qualified imports for tokio primitives all across the project, mainly for differentiating them from other std primitives that have the same name.

for `unbounded_channel`, I think there's actually not an equivalent primitive in std (although there is in other queue libraries like `flume`), but I'm more inclined to keep it this was for consistency with the rest of the project.

src/events/route_tasks.rs
gabotechs: As it's part of the public API, we are forced by Clippy to have something here (I always forget if that's set up correctly, if not we should).

I think I can improve the comments with something better.

src/events/route_tasks.rs
gabotechs: Yeah, it was like this before, however, this tends to clutter the examples and the docs. I pretty much added this just for the sake of 3 line shorter examples, but otherwise yes, this is pretty much unnecessary.

src/distributed_ext.rs
gabotechs: Makes sense. Done!

src/events/defaults/routing.rs
gabotechs: Yes, this is implemented in the next PR in the stack:
- https://github.com/datafusion-contrib/datafusion-distributed/pull/712

src/events/defaults/routing.rs
gabotechs: I actually was a bit worried about the new API being per-task rather than per-stage, because it would make choosing a contiguous set of URLs like before difficult.

It took a little bit of creativity to get this.

PR #696, opened 2026-09-04
Use PeakMemoryUsed

gabotechs:
In DF 55 PeakMemoryUsed is available, so we can start using it


Rich-T-kid (approved):
makes sense to me

gabotechs:
Thanks @Rich-T-kid!

PR #698, opened 2026-09-04
Deflake non-deterministic early-exit due to `LIMIT`

stuhood:
Fix a flake introduced by https://github.com/datafusion-contrib/datafusion-distributed/pull/670


gabotechs (approved):
Good catch! thanks @stuhood

PR #700, opened 2026-09-05
refactor: build Iceberg test metadata at runtime

alexanderbianchi:
Tests can supply complete Iceberg table metadata, explicit table options, and in-memory fixture files through `IcebergTestHarness::builder()`.

- `with_table_metadata` accepts native Iceberg metadata without reconstructing its IDs, schemas, or snapshot history.
- `with_table_option` replaces duplicated registration SQL in the selected-snapshot and invalid-snapshot-ID tests.
- `with_file(uri, bytes)` overrides a fixture file, including manifests and manifest lists. Explicit files take precedence over generated metadata; unmodified paths fall back to the checked-in taxi fixture. An empty-table registration test verifies raw metadata override behavior.

`build()` now constructs the session and registers the table directly; there is no intermediate `create()` helper. The harness remains usable without distributed integration features.

Shared fixture helpers expose only the original taxi metadata and a native metadata builder starting before the first snapshot. Snapshot recipes stay test-local. The duplicated missing-summary JSON fixture is removed.

Rebased onto `iceberg-0.10` at `6507bf8`, including merged #687. The table-options cleanup and file-override support are separate commits. Test-case consolidation is intentionally kept in an independent PR.

Validation on #700 at `bb3872e`:
- `cargo test -p datafusion-distributed-iceberg --locked` — 90 tests passed, including the doctest
- `cargo clippy -p datafusion-distributed-iceberg --all-targets --locked -- -D warnings`
- `git diff --check`


iceberg/src/test_utils/harness.rs
```diff
@@ -82,13 +107,15 @@ impl StorageFactory for FixtureStorageFactory {
...
             root: self.root.clone(),
+            files: self.files.clone(),
         }))
     }
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct FixtureStorage {
     root: PathBuf,
+    files: HashMap<String, Vec<u8>>,
```
alexanderbianchi, line 177: I don't love having to add this state + checking it in every storage function going to see if there's a better way

iceberg/src/test_utils/harness.rs
```diff
@@ -158,3 +194,61 @@ impl Storage for FixtureStorage {
         Ok(OutputFile::new(Arc::new(self.clone()), path.to_string()))
     }
 }
+
+enum FixtureInput<'a> {
```
alexanderbianchi, line 257: First iteration required a duplicated check in each function above, i'd rather keep them 1 line and isolate the input to one place. If you think a struct for this is overkill happy to split it up into the `impl Storage`

iceberg/tests/statistics.rs
```diff
@@ -146,4 +144,57 @@ mod tests {
...
+                .remove("total-records")
+                .is_some()
+        );
+        assert!(
+            summary
+                .additional_properties
+                .remove("total-files-size")
+                .is_some()
+        );
+        let snapshot = Snapshot::builder()
```
alexanderbianchi, line 263: If we could do like `.into_builder()` on the current snapshot this wouldn't be a problem but the LoC for a new snapshot are needed right now.

iceberg/src/test_utils/harness.rs
```diff
@@ -78,6 +83,75 @@ impl IcebergTestHarness {
...
+    fn default() -> Self {
+        let metadata = serde_json::from_str(include_str!(
+            "../../../testdata/iceberg/taxi/metadata/v1.metadata.json"
+        ))
+        .expect("taxi metadata is valid JSON");
+        Self { metadata }
+    }
+}
+
+impl IcebergTestHarnessBuilder {
```
alexanderbianchi, line 86: Builder might be premature
```rust
let harness =
    IcebergTestHarness::with_current_snapshot_summary(|summary| {
        for key in ["total-records", "total-files-size"] {
            assert!(summary.additional_properties.remove(key).is_some());
        }
    })
    .await?;
```
somethiing like this would do the same thing. Thoughts?

alexanderbianchi: https://github.com/datafusion-contrib/datafusion-distributed/pull/702/changes
https://github.com/datafusion-contrib/datafusion-distributed/pull/703/changes
actually it think we are pretty quickly going to benefit from the builder. These are PR's containings tests that are today hard to do in the harness but would be useful for existing PR's

iceberg/tests/external_table.rs
```diff
@@ -31,6 +33,31 @@ mod tests {
...
+        let harness = IcebergTestHarness::builder()
+            .with_file(
+                format!("{FIXTURE_URI}/metadata/v1.metadata.json"),
+                serde_json::to_vec(&metadata).expect("metadata serializes"),
+            )
+            .build()
+            .await?;
+        let (_, batches) = harness.query("SELECT COUNT(*) AS trips FROM taxi").await?;
+
+        insta::assert_snapshot!(batches, @"
```
alexanderbianchi: Should we consider something like
```rust
   async fn run(
       ctx: &SessionContext,
       query: &str,
       Ok((plan, batches))
   }
 ```
from `tests/multi_task_collect_join_repros.rs`.

File: iceberg/src/test_utils/fixture.rs [hunk 6]
```diff
@@ -0,0 +1,18 @@
...
+        "../../../testdata/iceberg/taxi/metadata/v1.metadata.json"
+    ))
+    .expect("taxi metadata is valid JSON")
+}
+
+/// Starts a fixture history with the original schema and partition IDs and no snapshots.
+pub fn taxi_metadata_builder() -> TableMetadataBuilder {
+    let metadata = serde_json::from_str(include_str!(
+        "../../../testdata/iceberg/taxi/metadata/00000-00a113a6-47e0-4c4b-9522-4a7c44d74036.metadata.json"
+    ))
```
gabotechs: How can it be that this function is not just:

```rust
pub fn taxi_metadata_builder() -> TableMetadataBuilder {
    TableMetadataBuilder::new_from_metadata(taxi_metadata(), None)
}
```

iceberg/tests/external_table.rs
```diff
@@ -31,6 +33,31 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn explicit_metadata_file_overrides_the_default_fixture() -> Result<()> {
+        let metadata = taxi_metadata_builder()
+            .build()
+            .expect("valid empty table")
+            .metadata;
+        let harness = IcebergTestHarness::builder()
+            .with_file(
```
gabotechs: This tests seems to just be testing the harness, but not really any acctual feature from the project.

I'd not worry too much about tests that just cover the testing tooling, as it's going to end up being covered by itself as we add more real tests to this codebase.

iceberg/tests/statistics.rs
```diff
@@ -256,4 +256,27 @@ mod tests {
...
+            .with_manifest_list(current.manifest_list())
+            .schema_id_opt(current.schema_id())
+            .with_summary(Summary {
+                operation: Operation::Append,
+                additional_properties: Default::default(),
+            })
+            .build();
+        taxi_metadata_builder()
+            .set_branch_snapshot(snapshot, "main")
+            .expect("taxi snapshot can be added")
```
gabotechs, line 277: A bit annoying all this type builder nonsense.

gabotechs (approved):
:+1: no blockers, mainly just questions

iceberg/src/test_utils/fixture.rs
alexanderbianchi: The hash is the Iceberg table’s initial metadata file, created before its first data commit.

 It contains:
 - The taxi schema and Iceberg field IDs.
 - The partition specification and partition field IDs.
 - The table’s location and identity.
 - No snapshots.
 
 It gives tests the fixture’s existing schema and partition IDs, but a clean snapshot history.
 
 `empty_taxi_metadata_builder()` might be a better name.

iceberg/tests/statistics.rs
alexanderbianchi: Agreed, it's mostly a problem that upstream doesn't have like an `into_builder()` function but they do have `taxi_metadata().into_builder(None)` so it's a pattern they are familiar with. I can make a PR in `iceberg-rust` for it.

PR #704, opened 2026-09-06
feat: estimate Iceberg distributed task counts

### LINKED ISSUE #605: [Iceberg] Estimate desired task counts from table metadata

alexanderbianchi:
## Summary

Implements the Iceberg desired-task-count handler using the selected snapshot's `total-files-size` summary, read from the coordinator-local work-unit feed.

- Respect an explicitly selected snapshot, otherwise use the current snapshot.
- Estimate tasks using the configured bytes per partition and target partitions, with ceiling division and validation of zero divisors.
- Treat a table with no current snapshot as zero scan work; decline to estimate when size metadata is missing, invalid, or unavailable on a remote feed.
- Add focused tests using the runtime metadata fixtures from #700, including historical snapshots, empty tables, malformed summaries, and decoded remote feeds.

This is a whole-snapshot scan-work estimate, not a post-pruning byte estimate.

Closes #605.

## Dependency / scope

Stacked directly on #700's `codex/iceberg-runtime-metadata`.

Named `test-case` cases cover current and selected snapshots, empty tables, and missing or invalid file sizes. A separate distributed-execution test follows `tests/task_estimator_test.rs`: it executes a taxi aggregation through the in-memory workers. The core harness and its existing tests remain available without the `integration` feature. Distributed planning and worker setup are explicit builder choices through `.with_workers(...)`, requiring `integration`; CI enables that feature for the distributed test. The construction cleanup belongs to #700, so this PR no longer introduces and then removes `create()`. Existing plan and result snapshots remain unchanged.

## Validation

- `cargo test -p datafusion-distributed-iceberg --locked` — 101 tests passed, including harness-backed tests and the doctest.
- `cargo test -p datafusion-distributed-iceberg --features integration --locked` — 102 tests passed, including distributed execution and the doctest.
- `cargo clippy -p datafusion-distributed-iceberg --all-targets -- -D warnings` — passed, also with `--all-features`.
- `cargo check -p datafusion-distributed-iceberg --lib` — passed without the integration feature.
- `cargo fmt --all -- --check` — passed.
- `git diff --check origin/iceberg-0.10 HEAD` — passed.

### DIFF
--- a/docs/upgrade/5.0.0.md
+++ b/docs/upgrade/5.0.0.md
@@ -1,0 +1,8 @@
+# Upgrading from 4.0.0 to 5.0.0
+
+## Iceberg test harness
+
+`IcebergTestHarness::ctx` is private. Configure sessions before `build()` using
+`with_target_partitions`, `with_scan_bytes_per_partition`, and
+`with_column_stats_enabled`. Use the harness's `query`, `physical_plan`, `scan`,
+`estimate_task_count`, and `roundtrip_plan` methods instead of accessing its context.
--- a/docs/upgrade/5.0.0.md
+++ b/docs/upgrade/5.0.0.md
@@ -1,8 +1,0 @@
-# Upgrading from 4.0.0 to 5.0.0
-
-## Iceberg test harness
-
-`IcebergTestHarness::ctx` is private. Configure sessions before `build()` using
-`with_target_partitions`, `with_scan_bytes_per_partition`, and
-`with_column_stats_enabled`. Use the harness's `query`, `physical_plan`, `scan`,
-`estimate_task_count`, and `roundtrip_plan` methods instead of accessing its context.
--- a/iceberg/src/distributed_desired_task_count_handler.rs
+++ b/iceberg/src/distributed_desired_task_count_handler.rs
@@ -1,6 +1,7 @@
-use datafusion::common::stats::Precision;
+use std::num::NonZeroUsize;
+
 use datafusion::common::{Result, config_datafusion_err};
-use datafusion::datasource::source::{DataSource, DataSourceExec};
+use datafusion::datasource::source::DataSourceExec;
 use datafusion_distributed::{
     DesiredTaskCountEvent, DesiredTaskCountEventResponse, DistributedConfig,
 };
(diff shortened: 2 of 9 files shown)


iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    #[tokio::test]
+    async fn remote_feed_does_not_estimate_again() -> Result<()> {
+        let harness = IcebergTestHarness::new().await?;
+        let plan = harness.roundtrip_plan(scan(&harness).await?)?;
+        assert_eq!(estimate(&plan)?, None);
+        Ok(())
+    }
+
+    async fn run_distributed_query() -> Result<(String, String)> {
```
alexanderbianchi: Should this be in the harness?

iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    #[tokio::test]
+    async fn remote_feed_does_not_estimate_again() -> Result<()> {
+        let harness = IcebergTestHarness::new().await?;
+        let plan = harness.roundtrip_plan(scan(&harness).await?)?;
+        assert_eq!(estimate(&plan)?, None);
+        Ok(())
+    }
+
+    async fn scan(harness: &IcebergTestHarness) -> Result<Arc<dyn ExecutionPlan>> {
```
alexanderbianchi: here we are passing in the harness so we can grab the table provider and pass in the ctx - not sure if this is how the harness is meant to be used but it is easier to test the task count just calling .Scan then doing all of planning and having snapshots. open to recommendations.

iceberg/src/distributed_desired_task_count_handler.rs
```diff
@@ -1,21 +1,96 @@
...
+    };
+    let total_bytes = match snapshot {
+        Some(snapshot) => snapshot
+            .summary()
+            .additional_properties
+            .get("total-files-size")?
+            .parse()
+            .ok()?,
+        None => 0,
+    };
```
gabotechs: Here, I'd not access directly the guts of the `IcebergDataSource`, we should have the total byte size available through `node.partition_statistics(None)`. I'd just use that instead (this is what AQE will use anyways for computing CPU cost)

iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,159 @@
...
+    #[cfg(feature = "integration")]
+    #[tokio::test]
+    async fn executes_with_estimated_scan_tasks() -> Result<()> {
+        let mut harness = IcebergTestHarness::builder()
+            .with_workers(4)
+            .build()
+            .await?;
+        harness
+            .query("SET datafusion.execution.target_partitions = 2")
+            .await?;
```
gabotechs: It'd be nice to set this in the harness builder, that way the config under which the test runs is all gathered in the same place.

iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,159 @@
...
+        // 4,480,382 bytes / 1 MB / 2 partitions rounds up to 3 tasks, not all 4 workers.
+        harness
+            .ctx
+            .set_distributed_file_scan_config_bytes_per_partition(1_000_000)?;
```
gabotechs: Same as above

iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    #[test_case(metadata_with_file_size(Some("24000000")), None, Some(12); "current snapshot")]
+    #[test_case(metadata_with_file_size(Some("24000000")), taxi_metadata().current_snapshot_id(), Some(3); "selected snapshot")]
+    #[test_case(empty_metadata(), None, Some(0); "empty table")]
+    #[test_case(metadata_with_file_size(None), None, None; "missing size")]
+    #[test_case(metadata_with_file_size(Some("invalid")), None, None; "malformed size")]
+    #[test_case(metadata_with_file_size(Some("-1")), None, None; "negative size")]
+    #[test_case(metadata_with_file_size(Some("18446744073709551616")), None, None; "overflowing size")]
+    #[tokio::test]
```
gabotechs: I have to say I do not love this way of writing tests:
- this heavy use of macros allows to bypass the project's formatting rules, as `rustfmt` does not go into them
- IDEs cannot run the tests just by hovering over its definition, as test names are generated at compile time by the macros.
- It's hard to map from test_case parameter to function argument.

If you ask me, I'd just create 7 different tests and just reuse a common function in all of them. Each test case body should just be 1 LOC.

Not a blocker though if you like this approach.

iceberg/tests/desired_task_count.rs
```diff
@@ -0,0 +1,159 @@
...
+        harness
+            .ctx
+            .table_provider("taxi")
+            .await?
+            .scan(&harness.ctx.state(), None, &[], None)
+            .await
+    }
+
+    fn estimate(plan: &Arc<dyn ExecutionPlan>) -> Result<Option<usize>> {
+        let mut config = SessionConfig::new().with_target_partitions(2);
```
gabotechs: The main point of the harness is not needing to fill test-case code with helper functions. I think we should be able to bring most (if not all) these helper functions to the test harness (probably with a small tweaks).

gabotechs (commented):
Flushing a first round of comments, will continue tomorrow, but so far looking pretty good, just small things.

iceberg/tests/desired_task_count.rs
alexanderbianchi: I can rip it apart, I don't use an IDE so wasn't aware.

iceberg/src/test_utils/harness.rs
```diff
@@ -24,14 +30,14 @@ use iceberg::{Error, ErrorKind, Result as IcebergResult};
...
-use crate::{IcebergExt, IcebergIntegrationOptions};
+use crate::{IcebergExt, IcebergIntegrationOptions, iceberg_desired_task_count};
 
 pub const FIXTURE_URI: &str = "s3://iceberg-test/warehouse/taxi";
 const WAREHOUSE_URI: &str = "s3://iceberg-test/warehouse/";
 const FIXTURE_METADATA_URI: &str = "s3://iceberg-test/warehouse/taxi/metadata/v1.metadata.json";
 
 pub struct IcebergTestHarness {
-    pub ctx: SessionContext,
+    ctx: SessionContext,
```
alexanderbianchi, line 40: Make it private to avoid abuse in the future.

iceberg/tests/statistics.rs
```diff
@@ -45,8 +45,10 @@ mod tests {
 
     #[tokio::test]
     async fn reports_exact_row_count_and_byte_size_for_full_scan_w_col_stats() -> Result<()> {
```
alexanderbianchi, line 47: Changes needed if we are moving ctx private - can be new PR but I think it's a pretty minor change easy to include here.

docs/upgrade/5.0.0.md
```diff
@@ -0,0 +1,8 @@
+# Upgrading from 4.0.0 to 5.0.0
+
+## Iceberg test harness
+
```
alexanderbianchi: sorry, agent did this, will delete. not neede d

iceberg/src/test_utils/harness.rs
```diff
@@ -68,22 +90,50 @@ impl IcebergTestHarness {
 }
 
 pub struct IcebergTestHarnessBuilder {
+    config: SessionConfig,
+    distributed_config: DistributedConfig,
+    column_stats_enabled: bool,
```
gabotechs: Out of experience when dealing in the past with DataFusion config structs, typically the one that is most helpful for ergonomically building using a builder-pattern in `SessionStateBuilder`.

I can imagine how holding here a reference to `SessionStateBuilder` rather than `SessionConfig` or `DistributedConfig` can be a bit more future proof.

gabotechs (approved):
👍 Nice

PR #705, opened 2026-09-06
refactor: consolidate and strengthen Iceberg statistics tests

alexanderbianchi:
## Less test code, stronger statistics coverage

Based directly on `iceberg-0.10` at `f9340ef`, including merged #715 and #716. “Before” refers to that base. Only `iceberg/tests/statistics.rs` changes, with a net reduction of 24 lines.

| Test / edge case | Before | After |
|---|---|---|
| Row count and full-scan byte size | Separate enabled/disabled test bodies | Ordinary named tests share assertions and retain both modes |
| Full-schema column statistics | Vector length only, in another pair of tests | Complete expected vector: populated metrics and unknown columns |
| Reordered projection / nonconsecutive selected field IDs | Projected vector length and row count | Exact association of metrics with reordered columns using IDs 4, 1, and 5 |
| Null-count aggregation across files | **None** | Exact sum of 5 |
| Column-size aggregation across files | **None** | Inexact sums of 400 and 600 |
| Min/max bounds and scalar types | **None** | Int32 and Int64 bounds checked with their precision; opposing extrema order exercises both updating and retaining bounds |
| One file missing a column's null count | **None** | Aggregate stays `Absent`, rather than treating missing as zero |
| Entirely missing column metrics | Only an all-empty-metrics fixture | Unknown column alongside columns with known metrics |
| Column statistics disabled | Fixture had no metrics even when enabled | Populated fixture must return unknown column statistics when disabled |
| Explicit current-snapshot selection | Separate duplicate row-total test | Fixture selects the snapshot explicitly in every matrix case |
| Missing snapshot summary totals | Present | Retained |
| Filter and projection/sort propagation | Present | Retained |
| `COUNT(*)` skips the scan | Duplicated enabled/disabled bodies and snapshots | Both named cases retained, sharing assertions |
| Explain formatting | Two identical snapshots plus weak label checks | One diagnostic snapshot; computed values verified by the matrix |

## Structure

- Four ordinary named Tokio tests cover full scan / reordered projection × column statistics enabled / disabled. Two more named tests cover `COUNT(*)` scan elimination. No `test_case` macros or dependency.
- Session setup uses the merged harness's `configure_session(...)`; tests never access its private context.
- Native Iceberg writers create a manifest with two synthetic data-file entries and a matching manifest list. #700's `with_file` supplies the bytes to the harness.
- Assertions observe query-output statistics, removing the recursive search/downcast helper and also checking projection propagation.
- `insta::allow_duplicates!` is needed only for the shared, identical `COUNT(*)` inline snapshots.
- The fixture is planning-only: synthetic Parquet paths are not opened. Multi-manifest merging, delete files, NDV, and verification against actual Parquet contents are not claimed here.
- The genuinely historical-snapshot regression remains in #702; this PR consolidates the current-snapshot-only check from #687.

## Validation

- `cargo test -p datafusion-distributed-iceberg --test statistics --locked` — all 10 cases passed
- `cargo test -p datafusion-distributed-iceberg --locked` — 95 tests passed, including the doctest
- `cargo test -p datafusion-distributed-iceberg --features integration --locked` — 96 tests passed, including the doctest
- `cargo clippy -p datafusion-distributed-iceberg --all-targets --locked -- -D warnings` — passed, also with `--all-features`
- `git diff --check`
- Mutation check: replacing computed column statistics with unknown values fails both enabled cases; the other eight tests pass. Mutation reverted.
- Bounds mutation check: keeping the first minimum and taking the last maximum fails both enabled cases; the other eight tests pass. Mutation reverted.

### DIFF
--- a/iceberg/tests/statistics.rs
+++ b/iceberg/tests/statistics.rs
@@ -204,8 +204,8 @@ mod tests {
             file.file_path(format!("{FIXTURE_URI}/data/metrics-first.parquet"))
                 .null_value_counts([(1, 2), (4, 4)].into())
                 .column_sizes([(1, 100), (4, 200)].into())
-                .lower_bounds([(1, Datum::int(1)), (4, Datum::long(10))].into())
-                .upper_bounds([(1, Datum::int(5)), (4, Datum::long(20))].into())
+                .lower_bounds([(1, Datum::int(2)), (4, Datum::long(10))].into())
+                .upper_bounds([(1, Datum::int(9)), (4, Datum::long(20))].into())
                 .build()?,
             snapshot.sequence_number(),
         )?;
@@ -213,8 +213,8 @@ mod tests {
             file.file_path(format!("{FIXTURE_URI}/data/metrics-second.parquet"))
                 .null_value_counts([(1, 3)].into())
                 .column_sizes([(1, 300), (4, 400)].into())
-                .lower_bounds([(1, Datum::int(2)), (4, Datum::long(30))].into())
-                .upper_bounds([(1, Datum::int(9)), (4, Datum::long(40))].into())
+                .lower_bounds([(1, Datum::int(1)), (4, Datum::long(30))].into())
+                .upper_bounds([(1, Datum::int(5)), (4, Datum::long(40))].into())
                 .build()?,
             snapshot.sequence_number(),
         )?;
--- a/iceberg/tests/statistics.rs
+++ b/iceberg/tests/statistics.rs
@@ -17,7 +17,6 @@ mod tests {
         DataContentType, DataFileBuilder, DataFileFormat, Datum, Literal, ManifestListWriter,
         ManifestWriterBuilder, Operation, Snapshot, Struct, Summary, TableMetadata,
     };
-    use test_case::test_case;
 
     // Values from the checked-in taxi snapshot summary.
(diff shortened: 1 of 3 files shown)


iceberg/tests/statistics.rs
```diff
@@ -133,128 +118,40 @@ mod tests {
...
-                "CREATE EXTERNAL TABLE taxi_snapshot STORED AS ICEBERG \
-                 LOCATION '{FIXTURE_URI}/metadata/v1.metadata.json' \
-                 OPTIONS ('iceberg.snapshot_id' '{TAXI_SNAPSHOT_ID}')"
-            ))
-            .await?;
-        let stats = source_statistics(&harness, "SELECT * FROM taxi_snapshot").await?;
-
-        assert_eq!(stats.num_rows, Precision::Exact(TAXI_ROWS));
-        assert_eq!(stats.total_byte_size, Precision::Exact(TAXI_BYTES));
+        insta::allow_duplicates! {
```
alexanderbianchi, line 159: Should we switch the harness to something more like
```rust
   async fn run(
       ctx: &SessionContext,
       query: &str,
       Ok((plan, batches))
   }
 ```
from ` tests/multi_task_collect_join_repros.rs`

alexanderbianchi:
There's still some work to do on coverage for multi-manifest, delete-file, and NDV

Issue #601, opened 2026-08-09
[Iceberg] Add local benchmarks


gabotechs:
Part of #599.

Add repeatable single-node benchmarks for the Iceberg integration.

Scope:
- Reuse the repository benchmark harness.
- Run representative SQL workloads against a committed Iceberg fixture.
- Document how to run the benchmark locally.

Keep this scoped to local execution. Remote and distributed benchmark infrastructure is tracked separately.

I'm still not 100% sure what's the best way of handling these benchmarks. For example:
- Should they just have a separate `--iceberg` flag that switches to Iceberg instead of parquet files for existing benchmark datasets?
- Should they be treated as separate datasets, something like `--dataset iceberg_tpch_sf10`?

The second option sounds more future proof, but still not 100% sure

alexanderbianchi:
take

PR #711, opened 2026-09-06
feat: add local TPC-H Iceberg benchmarks

alexanderbianchi:
### Crate dependencies
datafusion-distributed and datafusion-distributed-benchmarks do not rely on datafusion-distributed-iceberg or iceberg-rust.
 ```text
   datafusion-distributed-benchmarks
   └── datafusion-distributed

   datafusion-distributed-iceberg
   ├── datafusion-distributed
   └── iceberg-rust

   datafusion-distributed-iceberg-benchmarks
   ├── datafusion-distributed-benchmarks
   └── datafusion-distributed-iceberg
       └── iceberg-rust
 ```

## Summary

Add local TPC-H Iceberg preparation and execution while keeping the generic benchmark crate independent of Iceberg.

```bash
# Prepare matching Parquet and Iceberg datasets
cargo run -p datafusion-distributed-benchmarks --release -- prepare-tpch \
  --output testdata/tpch/sf1 --scale-factor 1 --partitions 16
cargo run -p datafusion-distributed-iceberg-benchmarks --release -- prepare \
  --input testdata/tpch/sf1

# Run each representation
WORKERS=2 ./benchmarks/run.sh --dataset tpch/sf1 --threads 2 --partitions 2
WORKERS=2 ./iceberg/benchmarks/run.sh --dataset tpch/sf1_iceberg --threads 2 --partitions 2

# Compare saved results
./target/release/dfbench compare tpch/sf1 tpch/sf1_iceberg
```

The Iceberg benchmark package composes the shared `RunOpt`/`BenchmarkBackend` execution path with Iceberg registration and session configuration. The generic benchmark and root integration-test dependency trees contain no Iceberg dependency.

Preparation writes a sibling dataset such as `tpch/sf1_iceberg`. Each representation keeps the existing `.results/<branch>/` and `previous.json` layout. Iceberg benchmark runs load manifest column statistics.

While validating the two representations, the requested source partition count was found to multiply every `nation` and `region` row. This fixes generation so those fixed-size tables are emitted once.

## Validation

- Generic benchmark and Iceberg test suites
- TPC-H plan snapshots
- Clippy, formatting, and dependency-tree checks
- Local Parquet and Iceberg Q6 smoke runs through the split binaries

## Scope

Local storage and TPC-H only. Remote object storage and benchmark harness integration are follow-up work.

Refs #601.


benchmarks/tests/iceberg_workflow.rs
```diff
@@ -0,0 +1,215 @@
+#[cfg(test)]
+mod tests {
```
gabotechs: This is just benchmarking code, I don't think we need tests for these. The benchmarking itself is kind of a test, so it's self covered by us just using it.

benchmarks/tests/tpch_generation.rs
```diff
@@ -0,0 +1,52 @@
+#[cfg(test)]
+mod tests {
```
gabotechs: Same as above

benchmarks/Cargo.toml
```diff
@@ -12,6 +12,7 @@ datafusion-distributed = { path = "..", features = [
   "integration",
   "system-metrics",
 ] }
+datafusion-distributed-iceberg = { path = "../iceberg", features = ["benchmarks"] }
```
gabotechs: :thinking: there's a problem the comes to mind:

If `arrow` gets bumped, `iceberg-rust` is likely going to pick up very slowly, and that'll block compilation if the `datafusion-distributed-benchmarks` crate, which is necessary for benchmarks and integration tests.

We need to find a way of not depending on `datafusion-distributed-iceberg` here in this crate.

gabotechs: This might actually be a bigger problem than expected.

If the `datafusion-distributed-benchmarks` is the crate capable of running Iceberg benchmarks, it's forced to depend on `datafusion-distributed-iceberg`, and therefore, the dependency hell issues will be unavoidable.

One thing that comes to mind is:
- Move all the Iceberg dataset generation code to `datafusion-distributed-iceberg`
- Move all the Iceberg benchmark execution code to `datafusion-distributed-iceberg`
- Expose all the components from `datafusion-distributed-benchmarks` to `datafusion-distributed-iceberg`'s benchmarks so that we can still reuse as much code as possible.

gabotechs: This setup might make it a bit harder to wire up with the remote benchmarks, but I think it might be worth it.

gabotechs: Also, I wonder if having another `benchmarks-iceberg` crate can be worth it, rather than putting everything onto `datafusion-distributed-iceberg`

alexanderbianchi: alright back in the lab i'll play around with things. might need more substantial changes.

benchmarks/src/datasets/tpch.rs
```diff
@@ -138,13 +141,32 @@ fn generate_tpch_tables(
...
-    generate_tpch_table!(CustomerGenerator, CustomerArrow, "customer");
-    generate_tpch_table!(SupplierGenerator, SupplierArrow, "supplier");
-    generate_tpch_table!(PartGenerator, PartArrow, "part");
-    generate_tpch_table!(PartSuppGenerator, PartSuppArrow, "partsupp");
-    generate_tpch_table!(OrderGenerator, OrderArrow, "orders");
-    generate_tpch_table!(LineItemGenerator, LineItemArrow, "lineitem");
+    // These fixed-size generators ignore partition arguments and emit the entire table.
+    generate_tpch_table!(RegionGenerator, RegionArrow, "region", 1);
+    generate_tpch_table!(NationGenerator, NationArrow, "nation", 1);
+    generate_tpch_table!(CustomerGenerator, CustomerArrow, "customer", parts);
```
gabotechs: This make sense, as these are supposed to be very small tables. However, why did you find it necessary specifically now?

benchmarks/src/main.rs
```diff
@@ -1,33 +1,64 @@
...
+        /// Shared dataset for the existing two-branch comparison shorthand.
+        #[structopt(long)]
+        dataset: Option<String>,
+    },
     PrepareTpch(prepare_tpch::PrepareTpchOpt),
     PrepareTpcds(prepare_tpcds::PrepareTpcdsOpt),
     PrepareClickbench(prepare_clickbench::PrepareClickBenchOpt),
 }
 
+fn comparison_states(
```
gabotechs: I'm a bit confused about the location and name of this function, I think it might be worth it to:
- Move it to compare.rs, as this is just a comparison concern.
- name it something like parse_comparison_args or something that maches the other rename.

benchmarks/src/backend.rs
```diff
@@ -0,0 +1,34 @@
...
+use datafusion::error::Result;
+use datafusion::execution::SessionStateBuilder;
+use datafusion::prelude::SessionContext;
+use futures::future::BoxFuture;
+
+use crate::datasets::register_tables;
+
+pub type RegisterTables = for<'a> fn(&'a SessionContext, &'a Path) -> BoxFuture<'a, Result<()>>;
+pub type ConfigureSession = fn(SessionStateBuilder) -> SessionStateBuilder;
+
```
gabotechs: It's a weird way to invert control like this in Rust. A more Rusty way would be to have a trait here, something like this:

```rust
#[async_trait]
pub trait BenchmarkBackend {
    async fn register_tables(&mut self, ctx: &SessionContext, path: &Path) -> Result<()>;
    fn configure_session(&self, builder: SessionStateBuilder) -> SessionStateBuilder {
        builder
    }
}

struct ParquetBenchmarkBackend;

impl BenchmarkBackend for ParquetBenchmarkBackend {
    async fn register_tables(&mut self, ctx: &SessionContext, path: &Path) -> Result<()> {
        register_tables(ctx, path).await
    }
}
```

The idea of having an extension point for other benchmarks to implement is good :+1:

gabotechs (commented):
This is looking very good! at this point my suggestions are mostly cosmetic, but it does look like a good direction.

I just reviewed the code, I did not play around with it yet, so flushing comments. I'll play a bit with this now and report any findings.

iceberg/benchmarks/src/dataset.rs
```diff
@@ -0,0 +1,289 @@
...
+#[expect(
+    clippy::disallowed_types,
+    reason = "the iceberg catalog API requires std::collections::HashMap"
+)]
+type CatalogProperties = std::collections::HashMap<String, String>;
+
+/// The sibling dataset directory used for the Iceberg representation.
+pub(super) fn output_path(input: &Path) -> PathBuf {
+    let mut name = input.file_name().unwrap_or_default().to_os_string();
+    name.push("-iceberg");
```
gabotechs: Nit: We typically use `_` instead of `-` for benchmark folder names.

gabotechs (approved):
I still see some of the changes in-process, but this looks good to me 👍, nice work!

PR #712, opened 2026-09-07
feat: retry coordinator-to-worker channel setup + fix 2 bugs

gabotechs:
Builds on top of 

- https://github.com/datafusion-contrib/datafusion-distributed/pull/694

And adds 4 individual commits:
1. Some improvements to the chaos test, to try to make results a bit more catastrophic
2. Adds some retry mechanism that reacts to gRPC status codes and retries the coordinator->worker channels
3. Fix one bug surfaced by the chaos test: when retrying, the same TaskKey might be attempted to be written multiple times, making the `SingleWriteMultipleRead` struct fail on the second write. With this change, the second write just overrides the previous one.
4. Fix another bug surfaced by the chaos test: when retrying, there's a chance that two tasks fall on the same worker. If that happens there's a caching collision in the `WorkerConnectionPool`, where retrieving two different non-overlapping ranges of partitions will go through the same cached stream, trying to consume it multiple times.


src/protocol/grpc/worker_client.rs
```diff
@@ -72,7 +73,40 @@ impl WorkerChannel for pb::worker_service_client::WorkerServiceClient<BoxCloneSy
...
+                    // https://grpc.io/docs/guides/status-codes/#resource-exhausted
+                    // Admission pressure is local to this worker, so retry a different URL.
+                    Code::ResourceExhausted => RetryOutcome::OtherUrl.tag(err),
+                    // https://grpc.io/docs/guides/status-codes/#aborted
+                    // Routing retries this task setup at the higher level on the same URL.
+                    Code::Aborted => RetryOutcome::SameUrl.tag(err),
+                    // https://grpc.io/docs/guides/status-codes/#unavailable
+                    // This is a transient failure; retry the task setup on the same URL.
+                    Code::Unavailable => RetryOutcome::SameUrl.tag(err),
+                    Code::Ok => err,
```
gabotechs, line 96: tbh, I don't know what I'm doing here... there's probably a better retry policy for this.

Invoking @hermeGarcia, in the hopes that you have better suggestions about how to transparently handle retries here.

src/worker/single_write_multi_read.rs
```diff
@@ -42,15 +44,20 @@ impl std::error::Error for SingleWriteMultiReadError {}
...
-    /// subsequent calls overwrite silently.
+    /// Write the value. A written but unconsumed value may be replaced. Once a reader has begun
+    /// consuming the value, further writes fail.
     pub(crate) fn write(&self, item: T) -> Result<(), SingleWriteMultiReadError> {
+        let read_started = self.read_started.lock().unwrap();
         let mut already_written = false;
         self.tx.send_modify(|v| {
-            if v.is_none() {
+            if v.is_none() || !*read_started {
                 *v = Some(item);
```
gabotechs, line 61: Fix for bug 1: if nobody has read yet the value, we are good to replace it.

src/worker/worker_connection_pool.rs
```diff
@@ -13,17 +14,17 @@ use futures::stream::BoxStream;
...
 /// - Handles a range of partitions at a time in other to give the chance to [crate::WorkerChannel]
 ///   implementations to batch RecordBatch streams up, avoiding the overhead of multiple small
 ///   streams over an IO interface.
 /// - Short circuits to a local in-memory connection if the remote worker that should be reached
 ///   happens to be the same one issuing the request.
-/// - Lazy inits connections to a remote worker on first call to [WorkerConnectionPool::execute].
+/// - Lazily initializes connections per remote worker and requested partition range.
 pub(crate) struct WorkerConnectionPool {
-    lazy_stream_groups: Vec<OnceLock<SharedBoxFuture<StreamGroup>>>,
+    lazy_stream_groups: Vec<DashMap<Range<usize>, SharedBoxFuture<StreamGroup>>>,
```
gabotechs, line 27: Fix for bug 2: if two requests come for the same target task, we still need to discriminate between requested partition ranges.

src/events/defaults/routing.rs
```diff
@@ -15,10 +22,97 @@ pub(crate) struct RandomRouteTaskHandler;
...
+    d_cfg: &DistributedConfig,
+) -> Result<RouteTaskEventResponse> {
+    let max_retries = d_cfg.max_coordinator_channel_retries;
+    let initial_backoff = Duration::from_millis(d_cfg.coordinator_channel_retry_initial_backoff_ms);
+    let max_backoff = Duration::from_millis(d_cfg.coordinator_channel_retry_max_backoff_ms);
+
+    let mut url = url;
+    let mut attempted_urls = HashSet::from([url.clone()]);
+    retry_urls.retain(|candidate| attempted_urls.insert(candidate.clone()));
+    retry_urls.shuffle(&mut rand::rng());
```
jayshrivastava, line 45: Probably makes sense to re-use the seed used in `stage_contiguous_rand_choice`

src/events/defaults/routing.rs
```diff
@@ -15,10 +22,97 @@ pub(crate) struct RandomRouteTaskHandler;
...
+                    same_url_retries += 1;
+                }
+            }
+            Some(RetryOutcome::OtherUrl) => {
+                coordinator_to_worker_other_url_retries
+                    .get_or_insert_with(|| {
+                        MetricBuilder::new(metrics)
+                            .global_counter("coordinator_to_worker_other_url_retries")
+                    })
+                    .add(1);
```
jayshrivastava, line 83: This seems to work pretty well. 

You don't get to see how many times a specific worker was retried, but you can see how many times we tried a new worker, the total number of retries, and the total number retries to the same worker.

src/protocol/grpc/worker_client.rs
jayshrivastava: This all seems reasonable to me.

I was suspicious about aborted. This description makes it sound like an optimistic locking error and the canonical thing to do in that situation is retry. 
```
The operation was aborted, typically due to a concurrency issue such as a sequencer check failure or transaction abort. See the guidelines above for deciding between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE.
```

src/worker/single_write_multi_read.rs
jayshrivastava: Looks right. I would have called it a bug, not a feature though :)

src/worker/worker_connection_pool.rs
jayshrivastava: I also think this was a feature, prior to adding retry capabilities :)

tests/chaos.rs
```diff
@@ -38,10 +38,8 @@ mod tests {
     const TOTAL_QUERIES: usize = 100;
     const CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE: Range<usize> = 1..4;
     const CONCURRENT_CLIENTS: usize = 10;
-    const QUERIES: &[&str] = &[
```
jayshrivastava, line 41: Revert before merging?

tests/chaos.rs
```diff
@@ -70,7 +67,7 @@ mod tests {
         for _ in 0..TOTAL_QUERIES {
             let mut batch = vec![];
             for _ in 0..rng.random_range(CONCURRENT_QUERIES_PER_CLIENT_RANDOM_RANGE) {
-                let query = tpch::get_query(QUERIES.choose(&mut rng).unwrap())?;
+                let query = tpch::get_query("q20")?;
```
jayshrivastava, line 70: Ditto. Should revert?

jayshrivastava (approved):

src/protocol/grpc/worker_service.rs
```diff
@@ -107,14 +108,6 @@ impl pb::worker_service_server::WorkerService for Worker {
...
 
         let set_plan_request = decode_set_plan_request(set_plan_request)?;
-        let task_key = set_plan_request.task_key;
-        // Dynamic-filter reports may carry decoded physical expressions. Retain the worker's
-        // task data so the gRPC boundary can encode them with its configured codecs, even after
-        // the completed task has been removed from the worker cache.
-        let task_data_entry = self
-            .task_data_entries
-            .get_with(task_key, async { Default::default() })
-            .await;
```
gabotechs, line 117: @jayshrivastava, I needed to do a small tweak to the this code that recently changed because of the dynamic filtering work:

Here, we call `task_data_entry.read_now()` for retrieving the `TaskContext` used for encoding dynamic filters back to the coordinator, however, that has a side effect in the `SingleWriteMultipleRead` struct as we need to read it, marking it as no longer safe to be re-written.

Failure reproduced by the CI: https://github.com/datafusion-contrib/datafusion-distributed/actions/runs/34449713271/job/102782445764?pr=712.

I changed this a bit so that `self.coordinator_channle()` returns the built `TaskContext` directly, rather than retrieving it here. I think the result is also very nice, as we don't need to access private details of the `Worker` struct here for retrieving the `TaskContext` now, and it's less code.

Let me know what do you think

tests/chaos.rs
gabotechs: This actually is something I changed in the original PR, but I forgot that there was a reason for using just 1 specific query rather than a random distribution:

At the end of the test, we store in memory all the results from all the queries that have ever happened during the test, and we assert the results from all of them are exactly the same. This is to ensure that even if retries work, we do not produce duplicate or omitted data.

For this reason, we use just 1 relatively convoluted query, and then expect the results for all queries that have happened during the test to be the same. Forgot about this in the original PR, and got surfaced now.

tests/chaos.rs
gabotechs: Same as above

src/events/defaults/routing.rs
gabotechs: Well there's the `coordinator_to_worker_same_url_retries` metric as well, right next to this one, that should reflect this as well.

src/events/defaults/routing.rs
gabotechs: What I wanted to avoid is the same worker to get overloaded. Let's say for example that these are all the available URLs:

```
[http://1, http://2, http://3, http://4, http://5, http://6, http://7, http://8, http://9, http://10]
```

And the random stage-offseted choice is:

```
[http://5, http://6, http://7, http://8]
```

But it turns out that all workers 5, 6 and 7 are down. Then we'd be overloading the worker 8 with all tasks, so I thought we might as well just go truly random in case of retries.

jayshrivastava: I agree that we should shuffle. I meant that we should continue using the seed here: https://github.com/datafusion-contrib/datafusion-distributed/blob/4669347f16821dd559ead7ef94cd1cfeb2e3eea0/src/events/defaults/routing.rs?plain=1#L126

We have an instantiated rng:
```rust
let mut seed = [0; 32];
seed[..16].copy_from_slice(key.query_id.as_bytes());
seed[16..24].copy_from_slice(&(key.stage_id as u64).to_le_bytes());
let mut rng = StdRng::from_seed(seed);
```
and this can be re-used here
```rust
retry_urls.retain(|candidate| attempted_urls.insert(candidate.clone()));
retry_urls.shuffle(&mut rng); // don't use rand::rng(), which uses a new seed
```

src/protocol/grpc/worker_service.rs
jayshrivastava: Looks great 👍🏽

src/events/defaults/routing.rs
gabotechs: Ah ok, I thought you were referring to something different. TBH, just for the sake of avoiding threading one extra argument across functions, I'd just regenerated another random seed and call it a day, but if you feel strongly about it I can just do it.

PR #714, opened 2026-09-07
feat: support fractional desired task counts

### LINKED ISSUE #562: Allow DesiredTaskCountHandler to return f64 as desired task counts

WilliamK112:
## What this changes

- Change `TaskCountAnnotation::Desired` to carry `f64`, while keeping `Maximum` as an integer hard cap.
- Let `DesiredTaskCountEventResponse::desired` accept both integer and floating-point values, and add an explicit `unbounded()` helper.
- Preserve fractional desired hints through dynamic compute estimation, cardinality scaling, merge, and union aggregation, rounding only when a concrete task count is required.
- Compute built-in file-scan hints with true floating-point division.
- Use a neutral `Desired(0.0)` fallback when a normal node has no desired hint, while resolving `Desired(0)` to at least one concrete task so union task maps remain non-empty.
- Render non-integer desired annotations to two decimal places while preserving whole-number output.
- Update built-in handlers, examples, public documentation, and the 5.0 upgrade guide.

The focused union regressions cover both fractional child hints being added before resolution and two zero-valued leaf hints still producing a valid one-task execution plan.

This revision reapplies the maintainer-reviewed suggestions from #617. The four corresponding commits retain Nishchaya Sharma's original authorship, and the latest review follow-up is isolated in `df147d6`.

Closes #562.

## Validation

- `cargo test --lib` — 309 passed, 1 ignored
- `cargo test -p datafusion-distributed-iceberg --features integration` — 57 unit, 38 integration, and 1 doc test passed
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo clippy -p datafusion-distributed-iceberg --all-targets --features integration -- -D warnings`
- `git diff --check`

## AI assistance disclosure

I used OpenAI Codex to help inspect the planner paths, implement the change, reconcile the prior review suggestions with current `main`, and run validation. I reviewed the final diff and test results.

### DIFF
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -170,9 +170,7 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(
-        usize::MAX as f64,
-    )))
+    Some(Ok(DesiredTaskCountEventResponse::desired(f64::MAX)))
 }
 
 fn cached_file_scan_scale_up_leaf_node_handler(
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -170,7 +170,7 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(f64::MAX)))
+    Some(Ok(DesiredTaskCountEventResponse::unbounded()))
 }
 
 fn cached_file_scan_scale_up_leaf_node_handler(
--- a/examples/custom_worker_url_routing.rs
+++ b/examples/custom_worker_url_routing.rs
@@ -170,7 +170,9 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(usize::MAX)))
+    Some(Ok(DesiredTaskCountEventResponse::desired(
+        usize::MAX as f64,
+    )))
 }
 
 fn cached_file_scan_scale_up_leaf_node_handler(
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -188,10 +188,15 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             f(plan, ctx)?.visit_children(|| {
                 if let Some(ciu) = plan.downcast_ref::<ChildrenIsolatorUnionExec>() {
                     // Just recurse to children that will actually get executed by this
-                    // ChildrenIsolatorUnionExec.
-                    ciu.task_idx_map[ctx.task_index].iter().apply_until_stop(
-                        |(child_i, child_ctx)| recurse(&ciu.children[*child_i], *child_ctx, f),
-                    )
+                    // ChildrenIsolatorUnionExec. A zero-task (empty) stage has no slots.
+                    ciu.task_idx_map
+                        .get(ctx.task_index)
+                        .map(Vec::as_slice)
+                        .unwrap_or(&[])
+                        .iter()
+                        .apply_until_stop(|(child_i, child_ctx)| {
+                            recurse(&ciu.children[*child_i], *child_ctx, f)
+                        })
                 } else if plan.is_network_boundary() {
                     Ok(TreeNodeRecursion::Continue)
                 } else {
@@ -265,8 +270,10 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             let node = &transformed.data;
             if let Some(ciu) = node.downcast_ref::<ChildrenIsolatorUnionExec>() {
                 let mut child_ctxs = vec![None; ciu.children.len()];
-                for (child_idx, child_ctx) in &ciu.task_idx_map[dt_ctx.task_index] {
-                    child_ctxs[*child_idx] = Some(*child_ctx);
+                if let Some(children_in_task) = ciu.task_idx_map.get(dt_ctx.task_index) {
+                    for (child_idx, child_ctx) in children_in_task {
+                        child_ctxs[*child_idx] = Some(*child_ctx);
+                    }
                 }
                 stack.extend(child_ctxs.into_iter().rev());
             } else {
--- a/src/common/recursion.rs
+++ b/src/common/recursion.rs
@@ -188,15 +188,10 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             f(plan, ctx)?.visit_children(|| {
                 if let Some(ciu) = plan.downcast_ref::<ChildrenIsolatorUnionExec>() {
                     // Just recurse to children that will actually get executed by this
-                    // ChildrenIsolatorUnionExec. A zero-task (empty) stage has no slots.
-                    ciu.task_idx_map
-                        .get(ctx.task_index)
-                        .map(Vec::as_slice)
-                        .unwrap_or(&[])
-                        .iter()
-                        .apply_until_stop(|(child_i, child_ctx)| {
-                            recurse(&ciu.children[*child_i], *child_ctx, f)
-                        })
+                    // ChildrenIsolatorUnionExec.
+                    ciu.task_idx_map[ctx.task_index].iter().apply_until_stop(
+                        |(child_i, child_ctx)| recurse(&ciu.children[*child_i], *child_ctx, f),
+                    )
                 } else if plan.is_network_boundary() {
                     Ok(TreeNodeRecursion::Continue)
                 } else {
@@ -270,10 +265,8 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
             let node = &transformed.data;
             if let Some(ciu) = node.downcast_ref::<ChildrenIsolatorUnionExec>() {
                 let mut child_ctxs = vec![None; ciu.children.len()];
-                if let Some(children_in_task) = ciu.task_idx_map.get(dt_ctx.task_index) {
-                    for (child_idx, child_ctx) in children_in_task {
-                        child_ctxs[*child_idx] = Some(*child_ctx);
-                    }
+                for (child_idx, child_ctx) in &ciu.task_idx_map[dt_ctx.task_index] {
+                    child_ctxs[*child_idx] = Some(*child_ctx);
                 }
                 stack.extend(child_ctxs.into_iter().rev());
             } else {
--- a/src/coordinator/prepare_dynamic_plan.rs
+++ b/src/coordinator/prepare_dynamic_plan.rs
@@ -59,16 +59,17 @@ pub(super) async fn prepare_dynamic_plan(
                 "network_cost",
                 *cost.network.get_value().unwrap_or(&0),
             ));
-            let compute_based_task_count = cost
-                .cpu
-                .get_value()
-                .unwrap_or(&0)
-                .div_ceil(nb_ctx.d_cfg.dynamic_bytes_per_partition.max(1))
-                .div_ceil(input_stage.plan.output_partitioning().partition_count())
-                .clamp(1, nb_ctx.max_tasks()?);
+            let compute_based_task_count = *cost.cpu.get_value().unwrap_or(&0) as f64
+                / nb_ctx.d_cfg.dynamic_bytes_per_partition.max(1) as f64
+                / input_stage
+                    .plan
+                    .output_partitioning()
+                    .partition_count()
+                    .max(1) as f64;
+            let compute_based_task_count = compute_based_task_count.min(nb_ctx.max_tasks()? as f64);
             let task_count = nb_ctx
                 .task_count(&input_stage.plan)?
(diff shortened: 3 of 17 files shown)


gabotechs:
I see this is committing the same mistakes than https://github.com/datafusion-contrib/datafusion-distributed/pull/617. I recommend to go there, give it a look, and re-apply the same suggestions here.

WilliamK112:
Thanks for pointing me to #617. I went through every maintainer thread there and reapplied the resulting design fixes on top of current `main`:

- removed the public `DesiredTaskCountHandler for f64` convenience impl and the direct `ceil()` unit test
- switched the cache sentinel to `f64::MAX`
- made file-scan hints use true floating-point division
- limited fractional Debug output to two decimals
- made the dynamic consumer hint neutral with `Desired(0.0)`
- made `Desired(0)` a valid empty stage, including union allocation/traversal/execution safety
- added end-to-end union coverage for both fractional aggregation and two zero-valued leaves
- moved the breaking-change note to the 5.0 upgrade guide

I also rebased onto current `main`. The reviewed #617 follow-up commits retain Nishchaya Sharma's authorship, and the small final test reconciliation accounts for the newer cardinality-factor behavior from #713.

Fresh local validation passed: 305 library tests (1 ignored), all 25 network-boundary planner tests, fmt, all-target/all-feature clippy with warnings denied, and whitespace checks. Hosted CI is now running on `d79f887`.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -68,7 +68,7 @@ pub(super) async fn prepare_dynamic_plan(
                 .clamp(1, nb_ctx.max_tasks()?);
             let task_count = nb_ctx
                 .task_count(&input_stage.plan)?
-                .merge(Desired(compute_based_task_count));
+                .merge(Desired(compute_based_task_count as f64));
```
gabotechs: This division here should be made with `f64` values, the `compute_based_task_count` variable should never be a `usize` on the first place. This is so that it can take very low values like `0.004` for example.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -302,7 +302,7 @@ async fn _inject_network_boundaries(
         // broadcasts are unavailable.
         Maximum(1)
     } else {
-        let mut task_count = desired_task_count.unwrap_or(Desired(1));
+        let mut task_count = desired_task_count.unwrap_or(Desired(1.0));
```
gabotechs: This should be 0.0, not 1.0. We don't want to attribute any task count load artificially.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -443,6 +443,10 @@ impl InjectNetworkBoundaryContext<'_> {
     ) -> Result<Arc<dyn ExecutionPlan>> {
         // Handle leaf nodes.
         if plan.children().is_empty() {
+            // A zero-task stage (empty datasets) has nothing to distribute.
+            if task_count.as_usize() == 0 {
+                return Ok(self.plan_with_task_count(Arc::clone(plan), task_count));
+            }
```
gabotechs: This is not up to this code to decide. `ScaleUpLeafNodeHandler` implementations should be the ones deciding this.

src/distributed_planner/inject_network_boundaries.rs
```diff
@@ -650,7 +654,7 @@ impl NetworkBoundaryBuilder for CardinalityBasedNetworkBoundaryBuilder {
         let f = calculate_scale_factor(&input_stage.plan, nb_ctx.d_cfg);
 
         Ok(NetworkBoundaryBuilderResult {
-            consumer_task_count: Desired((f * input_stage.tasks as f64).ceil() as usize),
+            consumer_task_count: Desired((f * input_stage.tasks as f64).ceil()),
```
gabotechs: This `.ceil()` just makes sense if we convert to `usize`, otherwise it doesn't.

src/events/desired_task_count.rs
```diff
@@ -68,7 +71,10 @@ impl DesiredTaskCountEventResponse {
     /// - Other nodes providing a `DesiredTaskCountEventResponse::desired(M)` where `M` > `N`.
     /// - Any other node providing a `DesiredTaskCountEventResponse::maximum(M)` where `M` can be
     ///   anything.
-    pub fn desired(value: usize) -> Self {
+    ///
+    /// Fractional values let several isolated union children contribute less than one task each;
+    /// the planner combines those values before rounding the final task count up.
+    pub fn desired(value: f64) -> Self {
```
gabotechs: Let's do something like this:

```rust
    pub fn desired<T: AsPrimitive<f64> + 'static>(value: T) -> Self {
        DesiredTaskCountEventResponse {
            task_count: Desired(value.as_()),
        }
    }
```

So that we can afford to call `DesiredTaskCountEventResponse::desired` with any number castable as `f64`

examples/custom_worker_url_routing.rs
```diff
@@ -169,7 +169,7 @@ fn cached_file_scan_desired_task_count_handler(
     ev: DesiredTaskCountEvent,
 ) -> Option<Result<DesiredTaskCountEventResponse>> {
     ev.plan.downcast_ref::<CacheExec>()?;
-    Some(Ok(DesiredTaskCountEventResponse::desired(usize::MAX)))
+    Some(Ok(DesiredTaskCountEventResponse::desired(f64::MAX)))
```
gabotechs: This is a bit weird (it was already weird before), but I think now it's a good chance to fix it.

What comes to mind is that we can add a method like this:

```rust
impl DesiredTaskCountEventResponse {
    /// Some nice docs
    pub fn unbounded() -> Self {
        DesiredTaskCountEventResponse::desired(f64::MAX)
    }
}
```

To make it explicit.

src/common/recursion.rs
```diff
@@ -117,10 +117,15 @@ impl TreeNodeExt for Arc<dyn ExecutionPlan> {
...
+                    // ChildrenIsolatorUnionExec. A zero-task (empty) stage has no slots.
+                    ciu.task_idx_map
+                        .get(ctx.task_index)
+                        .map(Vec::as_slice)
+                        .unwrap_or(&[])
+                        .iter()
+                        .apply_until_stop(|(child_i, child_ctx)| {
+                            recurse(&ciu.children[*child_i], *child_ctx, f)
+                        })
                 } else if plan.is_network_boundary() {
```
gabotechs: This seems incorrect. A stage that computes a desired task count of 0.0 should not be empty, should just be executed in 1 task as normal.

src/execution_plans/children_isolator_union.rs
```diff
@@ -171,11 +171,15 @@ impl ChildrenIsolatorUnionExec {
...
             }
         }
-        let Some(partition_count) = partition_counts.iter().max() else {
-            return internal_err!(
-                "ChildrenIsolatorUnionExec built an empty task_idx_map. This is a bug in the distributed planning logic, please report it"
-            );
-        };
+        // A `Desired(0)` union (every leaf empty) produces an empty task map. Advertise the
+        // children's partition counts so the node stays a valid empty stage rather than erroring.
+        let partition_count = partition_counts.iter().copied().max().unwrap_or_else(|| {
```
gabotechs: This is not right, a `Desired(0.0)` union should not compute an empty task map. The task map should never be empty.

src/distributed_planner/inject_network_boundaries.rs
WilliamK112: Addressed in `df147d6`. A normal node with no handler response now starts from the neutral `Desired(0.0)` value instead of `Desired(1.0)`.

src/distributed_planner/inject_network_boundaries.rs
WilliamK112: Addressed in `df147d6`. The cardinality boundary now retains `f * input_stage.tasks as f64` directly and defers rounding until a concrete stage count is resolved.

src/distributed_planner/inject_network_boundaries.rs
WilliamK112: Addressed in `df147d6`. I removed the planner-side zero-task leaf special case, so every leaf goes through `ScaleUpLeafNodeHandler` as usual.

src/coordinator/prepare_dynamic_plan.rs
WilliamK112: Addressed in `df147d6`. Dynamic compute estimation now divides as `f64` throughout, so sub-one hints such as `0.004` are preserved; only the configured worker maximum remains an upper bound.

src/execution_plans/children_isolator_union.rs
WilliamK112: Addressed in `df147d6`. The zero-valued union regression now verifies both leaves receive `Maximum(1)`, and the temporary empty-map traversal/execution behavior has been removed. The union task map therefore stays non-empty.

src/common/recursion.rs
WilliamK112: Addressed in `df147d6`. `Desired(0.0)` is still neutral while annotations are combined, but concrete resolution clamps desired counts to at least one task, so the plan executes normally.

examples/custom_worker_url_routing.rs
WilliamK112: Addressed in `df147d6`. I added `DesiredTaskCountEventResponse::unbounded()`, implemented as `desired(f64::MAX)`, and updated the cache example to use it.

src/events/desired_task_count.rs
WilliamK112: Addressed in `df147d6`. `DesiredTaskCountEventResponse::desired` is now generic over `T: AsPrimitive<f64> + 'static`, and integer call sites no longer need casts.

WilliamK112:
Thanks for the detailed review. I addressed all eight threads in `df147d6` and rebased the branch onto current `main` (`afc2d5b`). In particular, fractional values now remain `f64` through dynamic and cardinality estimation, the default hint is neutral, `Desired(0)` resolves to one concrete task, and the public helper accepts integer or floating-point inputs with a separate `unbounded()` constructor.

Fresh validation on the rebased branch passes: `cargo test --lib` (309 passed, 1 ignored), `cargo clippy --all-targets --all-features -- -D warnings`, fmt, and `git diff --check`.

Would you please take another look when convenient?

WilliamK112:
The only failing hosted check was the newly added Iceberg empty-table expectation. It still expected a concrete task count of 0, while this revision deliberately resolves `Desired(0.0)` to the minimum concrete count of 1 so execution and union task maps stay valid.

I updated that regression in `ec68945` to assert the reviewed minimum-task behavior. The exact CI command now passes locally: 57 unit tests, 38 integration tests, and 1 doc test. Iceberg-targeted clippy with warnings denied, fmt, and whitespace checks also pass.

src/coordinator/prepare_dynamic_plan.rs
```diff
@@ -110,10 +111,13 @@ pub(super) async fn prepare_dynamic_plan(
...
                     metrics.extend(new_metrics);
-                    // returning Desired(1) here is our way to tell the planner that we don't care
+                    // returning Desired(0) here is our way to tell the planner that we don't care
                     // about the task count assigned to the network boundary in the consumer stage,
-                    // and we don't want it to affect other task count decisions.
-                    (Some(Arc::new(stats)), Desired(1))
+                    // and we don't want it to affect other task count decisions. A 0.0 hint does
+                    // not mask smaller fractional values via max-merge, and does not inflate
+                    // UNION sums when this boundary sits under a child of a
+                    // ChildrenIsolatorUnionExec.
```
gabotechs, line 119:
```suggestion
                    // and we don't want it to affect other task count decisions.
```

For comments, better to stick to the rationale about what the code does, not what it doesn't do.

gabotechs (approved):
This looks good. Thanks @WilliamK112! if we resolve the conflicts we should be good

WilliamK112:
Thanks for the approval. I merged current main through d2358fc and resolved the only conflict in the 5.0 upgrade guide: the upstream async WorkerPlanRewriteHandler migration remains section 2, and this PR’s fractional task-count migration is section 3. The PR is mergeable again.

Validation on the merged tree: cargo fmt --check, cargo metadata --locked --no-deps, and cargo test --locked --lib (311 passed, 1 ignored). The updated head is cd9b772; hosted checks are now queued.

WilliamK112:
Quick follow-up: the rebased branch is still conflict-free, and all 34 hosted checks have now passed. It is ready to merge from my side. Thanks again for the review!

src/events/defaults/file_scan_config.rs
```diff
@@ -26,9 +26,9 @@ pub(crate) fn file_scan_config_desired_task_count(
         }
     }
 
-    let task_count = total_bytes
-        .div_ceil(d_cfg.file_scan_config_bytes_per_partition)
-        .div_ceil(cfg.target_partitions());
+    let bytes_per_partition = d_cfg.file_scan_config_bytes_per_partition.max(1) as f64;
+    let target_partitions = cfg.target_partitions().max(1) as f64;
+    let task_count = total_bytes as f64 / bytes_per_partition / target_partitions;
```
gabotechs, line 31: [P2] Preserve fractional hints in the Iceberg estimator too

Please apply the same fractional calculation to `iceberg/src/distributed_desired_task_count_handler.rs`. Its `calculate_task_count` still returns `usize` using two `div_ceil` calls. Consequently, two Iceberg union children that each warrant `0.4` tasks contribute `1 + 1`, requesting two tasks instead of `ceil(0.8) = 1`. The new union behavior therefore still overestimates task counts for small Iceberg scans.

gabotechs: Just addressed this myself.

gabotechs:
Thanks @WilliamK112!

PR #721, opened 2026-09-10
feat: make WorkerPlanRewriteHandler async

toddtomkinson:
Rewrite of the worker plan rewrite handler API so implementations can perform async setup work before a worker-local plan is registered for execution.

Some setups need to make network or I/O calls while reacting to a decoded worker-local plan (e.g. registering tables the plan references) before the plan is handed off for execution. `WorkerSessionBuilder::build_session_state` runs too early for this: it builds the session before the plan is decoded, so it has no access to the plan being executed. `WorkerPlanRewriteHandler` is the only extension point that sees the decoded plan at worker-session setup time, so it is now `async`.

Because `WorkerPlanRewriteEvent` borrows the worker session's `SessionConfig`, a `Fn(WorkerPlanRewriteEvent<'_>) -> impl Future<...>` blanket implementation cannot express the required lifetimes without boxing. Implementations now define a named type and `#[async_trait] impl WorkerPlanRewriteHandler for ...`, matching the pattern already used by `RouteTaskHandler`.

No new capabilities are added for handlers themselves; this is purely an API change unlocking async work inside existing handlers.

### DIFF
--- a/src/events/worker_plan_rewrite.rs
+++ b/src/events/worker_plan_rewrite.rs
@@ -33,23 +33,37 @@ impl WorkerPlanRewriteEventResponse {
 /// previous handler. Returning an error aborts plan registration.
 ///
 /// The handler is `async` so implementations can perform setup work that needs to make network
-/// or I/O calls (e.g. registering tables in the worker session) before the plan executes.
+/// or I/O calls before the plan executes.
 #[async_trait]
 pub trait WorkerPlanRewriteHandler: Send + Sync + 'static {
     /// Returns the plan to pass to the next handler.
-    async fn rewrite_worker_plan(
+    async fn handle(
         &self,
         ev: WorkerPlanRewriteEvent<'_>,
     ) -> Result<WorkerPlanRewriteEventResponse>;
 }
 
 #[async_trait]
-impl WorkerPlanRewriteHandler for Arc<dyn WorkerPlanRewriteHandler> {
-    async fn rewrite_worker_plan(
+impl<F> WorkerPlanRewriteHandler for F
+where
+    F: Send + Sync + 'static,
+    F: for<'a> Fn(WorkerPlanRewriteEvent<'a>) -> Result<WorkerPlanRewriteEventResponse>,
+{
+    async fn handle(
         &self,
         ev: WorkerPlanRewriteEvent<'_>,
     ) -> Result<WorkerPlanRewriteEventResponse> {
-        self.as_ref().rewrite_worker_plan(ev).await
+        self(ev)
+    }
+}
+
+#[async_trait]
+impl WorkerPlanRewriteHandler for Arc<dyn WorkerPlanRewriteHandler> {
+    async fn handle(
+        &self,
+        ev: WorkerPlanRewriteEvent<'_>,
+    ) -> Result<WorkerPlanRewriteEventResponse> {
+        self.as_ref().handle(ev).await
     }
 }
 
@@ -66,7 +80,7 @@ impl WorkerPlanRewriteHandlers {
         if let Some(handlers) = session_config.get_extension::<WorkerPlanRewriteHandlers>() {
             for handler in handlers.iter() {
                 plan = handler
-                    .rewrite_worker_plan(WorkerPlanRewriteEvent {
+                    .handle(WorkerPlanRewriteEvent {
                         plan,
                         session_config,
                     })
--- a/src/events/worker_plan_rewrite.rs
+++ b/src/events/worker_plan_rewrite.rs
@@ -1,4 +1,5 @@
 use super::common::EventHandlerChain;
+use async_trait::async_trait;
 use datafusion::error::Result;
 use datafusion::execution::config::SessionConfig;
 use datafusion::physical_plan::ExecutionPlan;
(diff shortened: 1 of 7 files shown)


toddtomkinson:
@gabrielkerr This is the change that we needed.

@gabotechs Hoping this is a simple change that we can get in before the v5 release. Happy to rework if it isn't what you guys are thinking, but tried to follow the same patterns with the switch of RouteTaskHandler from sync to async in #694.

gabrielkerr:
@toddtomkinson this makes sense to me since I have context for our use case. The only thing I can think to add at this point is a test that exercises why we want this. Adding a network call might be overkill, but could we register a small in-memory table? The tests you updated illustrate the crux of the change imo, but this could be nice to see in the advanced user guide example.

gabotechs:
Hi @toddtomkinson and @gabrielkerr, glad to have you here!

This sounds completely reasonable, +1 to the idea. I'll take a look tomorrow at the implementation, but at first sight it does not look like this will raise controversy.

src/events/worker_plan_rewrite.rs
```diff
@@ -30,55 +31,49 @@ impl WorkerPlanRewriteEventResponse {
 ///
 /// Every registered handler runs in registration order and receives the plan returned by the
 /// previous handler. Returning an error aborts plan registration.
+///
+/// The handler is `async` so implementations can perform setup work that needs to make network
+/// or I/O calls (e.g. registering tables in the worker session) before the plan executes.
```
gabotechs, line 36: :thinking: this is a curious example in the comment (registering tables). Typically, this kind of registration is done at planning time, which is the main reason for `QueryPlanner` to be `async`, it feels a bit off to do it here.

You guys probably have very valid reasons for doing this, but on a general note I think we can avoid putting this example as it can mislead other people into doing table registration at execution time.

src/events/worker_plan_rewrite.rs
```diff
@@ -30,55 +31,49 @@ impl WorkerPlanRewriteEventResponse {
...
+#[async_trait]
 pub trait WorkerPlanRewriteHandler: Send + Sync + 'static {
     /// Returns the plan to pass to the next handler.
-    fn rewrite_worker_plan(
+    async fn rewrite_worker_plan(
         &self,
-        ev: WorkerPlanRewriteEvent,
+        ev: WorkerPlanRewriteEvent<'_>,
     ) -> Result<WorkerPlanRewriteEventResponse>;
 }
```
gabotechs, line 45: I'd still keep the callback-style auto `impl` here. A lot of people will still want to just provide a synchronous function for this, so it should be good to keep it with some small adaptations:

```rust
#[async_trait]
impl<F> WorkerPlanRewriteHandler for F
where
    F: Send + Sync + 'static,
    F: for<'a> Fn(WorkerPlanRewriteEvent<'a>) -> Result<WorkerPlanRewriteEventResponse>,
{
    async fn rewrite_worker_plan(
        &self,
        ev: WorkerPlanRewriteEvent<'_>,
    ) -> Result<WorkerPlanRewriteEventResponse> {
        self(ev)
    }
}
```

src/events/worker_plan_rewrite.rs
```diff
@@ -30,55 +31,49 @@ impl WorkerPlanRewriteEventResponse {
...
 /// Every registered handler runs in registration order and receives the plan returned by the
 /// previous handler. Returning an error aborts plan registration.
+///
+/// The handler is `async` so implementations can perform setup work that needs to make network
+/// or I/O calls (e.g. registering tables in the worker session) before the plan executes.
+#[async_trait]
 pub trait WorkerPlanRewriteHandler: Send + Sync + 'static {
     /// Returns the plan to pass to the next handler.
-    fn rewrite_worker_plan(
+    async fn rewrite_worker_plan(
```
gabotechs, line 40: :thinking: something I just noticed, and I think I'm the one to blame about this:

All event handlers in this project call their handling method just `handle()`, this one seems to be the only one that is calling it `rewrite_worker_plan()`. As we are changing the signature of this method. It might actually be a good moment to rename it to just `handle()`.

tests/worker_plan_hook.rs
```diff
@@ -115,39 +108,77 @@ mod tests {
...
-                if options.label != HOOK_LABEL {
-                    return internal_err!("unexpected plan hook label {}", options.label);
-                }
+    /// Fails the plan if [`PlanHookOptions::fail_in_hook`] is set.
+    struct FailingPlanHook;
+
+    #[async_trait]
+    impl WorkerPlanRewriteHandler for FailingPlanHook {
+        async fn rewrite_worker_plan(
+            &self,
```
gabotechs, line 117: With https://github.com/datafusion-contrib/datafusion-distributed/pull/721#discussion_r3989057653, I think we should be good to leave these implementations untouched, and they will still be valid out of the box.

gabotechs (commented):
:+1: Looks good! just left a couple of small suggestions, but this is a nice addition, thanks @toddtomkinson and @gabrielkerr!

src/events/worker_plan_rewrite.rs
toddtomkinson: Changed! Some of our execs require some "table" state (credentials, metadata) that we load into the context only when needed. Saying "registering tables" was a lazy way of saying that...

tests/worker_plan_hook.rs
toddtomkinson: I left one implementing the async trait directly to exercise the path. Let me know if you want me to switch that one back, too.

gabotechs (approved):
💯 All good here, thanks @toddtomkinson and @gabrielkerr!

tests/worker_plan_hook.rs
gabotechs: 👍 it looks good. Thanks!

Issue #611, opened 2026-08-09
[Iceberg] Add catalog and schema providers


gabotechs:
Part of #599.

Expose an Iceberg `Catalog` as DataFusion catalog and schema providers, so SQL can resolve Iceberg namespaces and tables without registering each table individually.

Upstream has this shape here:

https://github.com/apache/iceberg-rust/blob/f28ae7dca60904a5f135f3b6eecb03b09032b581/crates/integrations/datafusion/src/catalog.rs#L40-L90

The provider should resolve namespaces and tables through the configured Iceberg catalog.

sandugood:
take

PR #722, opened 2026-09-11
feat[iceberg]: add catalog and schema providers

sandugood:
Tackles #611.

# What's new?
Introduced two new structures: `IcebergCatalog` and `IcebergSchemaProvider`, which are responsible for resolving tables on-the-fly, without needing to manually register each table individually

Supports reading those tables as well as information_schema (which is explicitly set in the test harness)

### DIFF
--- a/iceberg/src/schema_provider.rs
+++ b/iceberg/src/schema_provider.rs
@@ -1,0 +1,75 @@
+use std::sync::Arc;
+
+use dashmap::DashMap;
+use datafusion::{
+    catalog::{SchemaProvider, TableProvider},
+    error::Result,
+};
+use futures::future::try_join_all;
+use iceberg::{Catalog, NamespaceIdent, Runtime};
+
+use crate::{IcebergCatalogTableProvider, common::df_err};
+
+#[derive(Debug)]
+pub struct IcebergSchemaProvider {
+    // Using Arc + DashMap for cheap clones of the tables in this schema
+    tables: Arc<DashMap<String, Arc<IcebergCatalogTableProvider>>>,
+}
+
+impl IcebergSchemaProvider {
+    pub async fn try_new(
+        client: Arc<dyn Catalog>,
+        namespace: NamespaceIdent,
+        iceberg_runtime: Runtime,
+    ) -> Result<Self> {
+        let table_names: Vec<_> = client
+            .list_tables(&namespace)
+            .await
+            .map_err(df_err)?
+            .into_iter()
+            .map(|ident| ident.name().to_string())
+            .collect();
+
+        let table_providers = try_join_all(
+            table_names
+                .iter()
+                .map(|name| {
+                    IcebergCatalogTableProvider::try_new(
+                        client.clone(),
+                        namespace.clone(),
+                        name,
+                        iceberg_runtime.clone(),
+                    )
+                })
+                .collect::<Vec<_>>(),
+        )
+        .await?;
+
+        let tables = Arc::new(DashMap::new());
+
+        // Getting a map of: <table_name, table_provider>
+        for (name, provider) in table_names.into_iter().zip(table_providers) {
+            tables.insert(name, Arc::new(provider));
+        }
+
+        Ok(IcebergSchemaProvider { tables })
+    }
+}
+
+#[async_trait::async_trait]
+impl SchemaProvider for IcebergSchemaProvider {
+    async fn table(&self, name: &str) -> Result<Option<Arc<dyn TableProvider>>> {
+        Ok(self
+            .tables
+            .get(name)
+            .map(|provider| provider.clone() as Arc<dyn TableProvider>))
+    }
+
+    fn table_names(&self) -> Vec<String> {
+        self.tables.iter().map(|k| k.key().clone()).collect()
+    }
+
+    fn table_exist(&self, name: &str) -> bool {
+        self.tables.contains_key(name)
+    }
+}
--- a/iceberg/src/schema_provider.rs
+++ b/iceberg/src/schema_provider.rs
@@ -18,11 +18,11 @@ pub struct IcebergSchemaProvider {
 
 impl IcebergSchemaProvider {
     pub async fn try_new(
-        client: Arc<dyn Catalog>,
+        catalog: Arc<dyn Catalog>,
         namespace: NamespaceIdent,
         iceberg_runtime: Runtime,
     ) -> Result<Self> {
-        let table_names: Vec<_> = client
+        let table_names: Vec<_> = catalog
             .list_tables(&namespace)
             .await
             .map_err(df_err)?
@@ -35,7 +35,7 @@ impl IcebergSchemaProvider {
                 .iter()
                 .map(|name| {
                     IcebergCatalogTableProvider::try_new(
-                        client.clone(),
+                        catalog.clone(),
                         namespace.clone(),
                         name,
                         iceberg_runtime.clone(),
--- a/iceberg/src/test_utils/harness.rs
+++ b/iceberg/src/test_utils/harness.rs
@@ -25,19 +25,26 @@ use iceberg::io::{
     FileMetadata, FileRead, FileWrite, InputFile, LocalFsStorage, OutputFile, Storage,
     StorageConfig, StorageFactory,
 };
+use iceberg::memory::{MEMORY_CATALOG_WAREHOUSE, MemoryCatalogBuilder};
 use iceberg::spec::TableMetadata;
-use iceberg::{Error, ErrorKind, Result as IcebergResult};
+use iceberg::{
+    Catalog, CatalogBuilder, Error, ErrorKind, NamespaceIdent, Result as IcebergResult, TableIdent,
+};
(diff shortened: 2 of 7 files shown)


sandugood:
By the way, I've got a question @alexanderbianchi (I think this is just nitpicking, but still)

I see that we mostly refer to `Arc<dyn Catalog>` as `catalog`, however I see it myself as a `client`, i.e:
- a user already has a running REST catalog or Hive Metastore (or any other catalog that resolves namespaces and tables and routes them to their respective metadata.json locations).

So this is essentially a `client`, I guess. 
But if that is not OK I can rename it to `catalog`.

iceberg/src/test_utils/harness.rs
```diff
@@ -25,19 +25,26 @@ use iceberg::io::{
...
 
+pub const FIXTURE_CATALOG: &str = "iceberg";
+pub const FIXTURE_NAMESPACE: &str = "nyc";
 pub const FIXTURE_URI: &str = "s3://iceberg-test/warehouse/taxi";
 const WAREHOUSE_URI: &str = "s3://iceberg-test/warehouse/";
 const FIXTURE_METADATA_URI: &str = "s3://iceberg-test/warehouse/taxi/metadata/v1.metadata.json";
 
 pub struct IcebergTestHarness {
     ctx: SessionContext,
+    catalog: Option<Arc<dyn Catalog>>,
```
alexanderbianchi: Maybe we should just be very explicit with iceberg_catalog in the actual variable name? and we should probably turn `Catalog` into `Iceberg::catalog` wherever we reference it.

alexanderbianchi:
@sandugood I share your view that this is is more of a client then the catalog itself. Hypothetically iceberg catalogs could be local so it's not always a catalog over RPC. IMO we should stick to the name they've chosen rather then add indirection, but we can be very explicit with an `iceberg_catalog: Option<Arc<dyn iceberg::Catalog>>`.

iceberg/src/test_utils/harness.rs
sandugood: Yeah, I've made it explicit in this struct

iceberg/src/schema_provider.rs
```diff
@@ -18,11 +18,11 @@ pub struct IcebergSchemaProvider {
 
 impl IcebergSchemaProvider {
     pub async fn try_new(
-        client: Arc<dyn Catalog>,
```
alexanderbianchi, line 21: We could use the iceberg_catalog iceberg::catalog everywhere, here as well, to be consistent.

sandugood: I think this really makes sense
wdyt @gabotechs?

alexanderbianchi:
We should be explicit in the docs with what the CatalogProvider can and can't be used for.
1. The catalog is instantiated entirely at startup so no updates if the tables change
2. There's no way to pass options or `iceberg::SessionContext` into the `iceberg::Catalog` from the CatalogProvider
3. No Auth or tenancy (still WIP upstream anyways https://github.com/apache/iceberg-rust/pull/3081)

So long as we make it clear this isn't a production ready catalog experience I think this PR is reasonable. It closely mirrors the upstream [provider](https://github.com/apache/iceberg-rust/blob/main/crates/integrations/datafusion/src/table/mod.rs) with the key difference that we call `load_table` in `IcebergCatalogTableProvider::try_new`. Although a `try_new_from_table` might be useful too [(link to iceberg-rust)](https://github.com/apache/iceberg-rust/blob/7b33321f34d9322ce5ede780ba3bc1f4e3741f47/crates/integrations/datafusion/src/table/mod.rs#L270).

Food for thought on how to implement a session aware catalog provider. 
- PR upstream: https://github.com/apache/iceberg-rust/pull/3000
- My own attempt: https://github.com/alexanderbianchi/iceberg-rust/pull/1

I have a couple comments there
1. Using `AsyncCatalogProvider` instead of the `CatalogProvider`[(example)](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-69910ed08a4c8430cd007a7297b8cee9b1924acd72b0492514956a54299fca68R53)
2. There could be a single enum for "resolved" or "session aware"[(example)](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-709b43501d3d88b5017cbb2aeb266d16f7d3b197620d05d537b2338dbd9d8a82R26), but different structs could make sense too. 
3. We need some sort of `iceberg_options` which I should could simply be a `iceberg::SessionContext` in the `extension_options` [example](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-69910ed08a4c8430cd007a7297b8cee9b1924acd72b0492514956a54299fca68R74) 

Given the most recent comment on https://github.com/apache/iceberg-rust/pull/3000 is to wait for that PR until we have `apache/datafusion-iceberg` I think porting it here for now would be reasonable if anyone needed it, since there's going to be other work to resolve with `apache/datafusion-iceberg`.

iceberg/src/schema_provider.rs
gabotechs: Not sure what the "iceberg_catalog iceberg::catalog" is

gabotechs: If you mean just the variable name, I'd stick to how iceberg-rust names things, that'll help with consistency

sandugood:
> We should be explicit in the docs with what the CatalogProvider can and can't be used for.
> 
> 1. The catalog is instantiated entirely at startup so no updates if the tables change
> 2. There's no way to pass options or `iceberg::SessionContext into the `iceberg::Catalog` from the CatalogProvider
> 3. No Auth or tenancy (still WIP upstream anyways [feat(rest_catalog::auth) Contextual Sessions apache/iceberg-rust#3081](https://github.com/apache/iceberg-rust/pull/3081))
> 
> So long as we make it clear this isn't a production ready catalog experience I think this PR is reasonable. It closely mirrors the upstream [provider](https://github.com/apache/iceberg-rust/blob/main/crates/integrations/datafusion/src/table/mod.rs) with the key difference that we call `load_table` in `IcebergCatalogTableProvider::try_new`. Although a `try_new_from_table` might be useful too [(link to iceberg-rust)](https://github.com/apache/iceberg-rust/blob/7b33321f34d9322ce5ede780ba3bc1f4e3741f47/crates/integrations/datafusion/src/table/mod.rs#L270).
> 
> Food for thought on how to implement a session aware catalog provider.
> 
> * PR upstream: [Datafusion session integration apache/iceberg-rust#3000](https://github.com/apache/iceberg-rust/pull/3000)
> * My own attempt: [Draft: resolve Iceberg session catalogs per DataFusion query alexanderbianchi/iceberg-rust#1](https://github.com/alexanderbianchi/iceberg-rust/pull/1)
> 
> I have a couple comments there
> 
> 1. Using `AsyncCatalogProvider` instead of the `CatalogProvider`[(example)](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-69910ed08a4c8430cd007a7297b8cee9b1924acd72b0492514956a54299fca68R53)
> 2. There could be a single enum for "resolved" or "session aware"[(example)](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-709b43501d3d88b5017cbb2aeb266d16f7d3b197620d05d537b2338dbd9d8a82R26), but different structs could make sense too.
> 3. We need some sort of `iceberg_options` which I should could simply be a `iceberg::SessionContext` in the `extension_options` [example](https://github.com/alexanderbianchi/iceberg-rust/pull/1/changes#diff-69910ed08a4c8430cd007a7297b8cee9b1924acd72b0492514956a54299fca68R74)
> 
> Given the most recent comment on [apache/iceberg-rust#3000](https://github.com/apache/iceberg-rust/pull/3000) is to wait for that PR until we have `apache/datafusion-iceberg` I think porting it here for now would be reasonable if anyone needed it, since there's going to be other work to resolve with `apache/datafusion-iceberg`.

Converting it to draft for now. 
Yeah, I think we might need to wait until a refreshable async catalog lands.

iceberg/src/schema_provider.rs
alexanderbianchi: They just use `catalog` which IMO get's confusing with `CatalogProvider` / the datafusion catalog. Happy to just have stick with `catalog` but `Iceberg::Catalog` IMO is still a win.

gabotechs: Ah yeah, this happens quite frequently that they use the same names. In case of name collision between DataFusion symbols, I'd lean towards fully qualifying them with `iceberg::`, and prefixing variable names with `iceberg_`

alexanderbianchi: > Ah yeah, this happens quite frequently that they use the same names. In case of name collision between DataFusion symbols, I'd lean towards fully qualifying them with iceberg::, and prefixing variable names with iceberg_

😆 alright so `iceberg_catalog : Arc<dyn iceberg::catalog>`

sandugood: think we ve come to an agreement😁

PR #724, opened 2026-09-12
refactor: use upstream Iceberg file scan task serialization

alexanderbianchi:
## Summary

Use iceberg-rust's `Serialize` and `Deserialize` implementation for `FileScanTask` work units instead of maintaining a parallel conversion layer.

- keep `FileScanTask` native for in-process feeds and serialize it only when the distributed transport encodes the work unit
- carry the upstream serialization as a small JSON payload inside the work-unit protobuf envelope
- remove the custom task/context protobuf schema, generated types, conversion code, validation tests, and context-reuse protocol
- retain the existing Iceberg data-source plan protobuf, which is unrelated to file scan task serialization
- use `Table::manifest_reader()` when collecting column statistics instead of opening and parsing manifests directly

The pinned iceberg-rust revision already contains both upstream APIs, so this requires no dependency update. This removes roughly 2,500 lines while leaving the public API unchanged. No upgrade guide entry is needed.

## Validation

- `cargo test -p datafusion-distributed-iceberg --locked`
- `cargo test -p datafusion-distributed-iceberg --features integration --locked`
- `cargo clippy -p datafusion-distributed-iceberg --all-targets --all-features --locked -- -D warnings`
- `git diff --check`


iceberg/src/work_unit_feed.rs
```diff
@@ -1,19 +1,108 @@
...
+struct SerializedFileScanTask {
+    #[prost(bytes = "vec", tag = "1")]
+    task: Vec<u8>,
+}
+
+#[derive(Clone, Debug)]
+enum FileScanTaskPayload {
+    Native(Box<FileScanTask>),
+    Serialized(SerializedFileScanTask),
+}
```
gabotechs, line 32: I saw the `FileScanConfig` serde code changed, and some files that were previously ignored they are now going over the wire, it seems to be done in https://github.com/apache/iceberg-rust/pull/3091.

iceberg/src/work_unit_feed.rs
```diff
@@ -1,19 +1,108 @@
...
+        Self {
+            payload: FileScanTaskPayload::Native(Box::new(task)),
+        }
+    }
+
+    fn serialized(&self) -> Cow<'_, SerializedFileScanTask> {
+        match &self.payload {
+            FileScanTaskPayload::Native(task) => Cow::Owned(SerializedFileScanTask {
+                task: serde_json::to_vec(task)
+                    .expect("Iceberg table scans produce serializable file scan tasks"),
```
gabotechs: Json serialization here is likely going to produce pretty big WorkUnits over the wire. Maybe msgpack is better here? this matters little now as we still don't have benchmarks.

gabotechs (approved):
👍 epic cleanup. Looks good!

iceberg/src/work_unit_feed.rs
alexanderbianchi: done

PR #725, opened 2026-09-13
feat: propagate typed session extensions

alexanderbianchi:
## Summary
A custom ExecutionPlan may already call task_ctx.session_config().get_extension::<T>(). After that plan is serialized for distributed execution, the worker must be able to recreate T without changing the execution-plan implementation.

- add an opt-in typed `SessionExtensionCodec` API for arbitrary `SessionConfig` extensions
- send named opaque extension payloads with task setup and decode them into task-local worker sessions
- keep codecs worker-local and reject request payloads that would replace existing worker extensions

## Testing

- `cargo check --all-targets`
- Integration tests were added but not run locally because the build volume ran out of space.

## AI Disclosure

Used an AI coding assistant for implementation and tests; the resulting changes were reviewed manually.

### DIFF
--- a/docs/upgrade/5.0.0.md
+++ b/docs/upgrade/5.0.0.md
@@ -79,7 +79,15 @@ let state = SessionStateBuilder::new()
 coordinator-to-worker connection is currently established. Implementations can
 choose to retry on other worker if the current one is not available.
 
-## 2. `WorkerPlanRewriteHandler::rewrite_worker_plan` renamed to `handle` and is now async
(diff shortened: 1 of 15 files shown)


alexanderbianchi:
Closing this, since beyond just "supporting what single node DF does" there isn't a strong reason to need it. One might come up with `datafusion-iceberg` around auth that shouldn't be stored in plain strings but even that I think is only relevant during planning and doesn't need a serialized extension. Once we get to writes, that becomes potentially interesting. Auth secrets being sent to the workers who are going to execute a commit.

Issue #602, opened 2026-08-09
[Iceberg] Add remote benchmarks


gabotechs:
Part of #599.

Add repeatable remote benchmarks for the Iceberg integration.

Scope:
- Extend the remote benchmark environment to register Iceberg tables from object storage.
- Document required object-store configuration and benchmark invocation.
- Run representative Iceberg workloads against a remotely accessible table.
- Establish a baseline for comparing local and remote execution.

Ideally, the same datasets that are supported for raw parquet files should also be supported for Iceberg.

They probably will have a different shape, and as the remote benchmarks are designed to run on AWS, they will probably need to be wired up to AWS Glue or something.

As this requires an expensive infra setup, this should be handled by a Datadog employee first, even if the remote setup should work for everyone at the end of the day.

gabotechs:
Take

PR #729, opened 2026-09-15
feat: add remote Iceberg benchmark worker

alexanderbianchi:
## Summary

Add an Iceberg-capable remote benchmark worker without introducing an Iceberg dependency into the generic benchmark or root integration-test paths.

```text
datafusion-distributed-benchmarks
└── datafusion-distributed

datafusion-distributed-iceberg-benchmarks
├── datafusion-distributed-benchmarks
└── datafusion-distributed-iceberg
    └── iceberg-rust
```

## Changes

- Extract the existing remote worker into `RemoteBenchmarkWorker::builder`.
- Make `RemoteBenchmarkWorkerBuilder<B>` generic over `BenchmarkBackend` so local and remote runners share the same format configuration.
- Add `IcebergBenchmarkBackend`, which installs the Iceberg integration and enables manifest column statistics.
- Apply the selected backend to both coordinator and worker sessions.
- Keep the existing `worker` binary on `ParquetBenchmarkBackend` and add an `iceberg-worker` binary backed by `IcebergBenchmarkBackend`.
- Keep dataset generation local and leave remote synchronization to the external benchmark tooling.

## Review order

1. `refactor: move remote worker into shared module` is a 100% rename; the temporary Cargo target keeps the moved worker buildable without changing it.
2. `refactor: expose configurable remote worker builder` adds the backend hook and thin worker binary after the move.
3. `feat: add remote Iceberg benchmark worker` adds the Iceberg worker.
4. `refactor: keep remote benchmark support scoped` removes remote generation/docs and moves backend code out of `lib.rs` following review.

## Validation

- Generic benchmark and Iceberg benchmark package tests
- Clippy with warnings denied
- Formatting and strict rustdoc
- Generic remote worker startup and discovery
- Iceberg remote worker static-table registration and distributed Q6 execution
- Verified the generic benchmark and root normal/dev dependency trees remain Iceberg-free

Refs #602.


iceberg/benchmarks/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+use datafusion::execution::SessionStateBuilder;
+use datafusion::prelude::SessionContext;
+use datafusion_distributed_benchmarks::backend::BenchmarkBackend;
+use datafusion_distributed_iceberg::{IcebergExt, IcebergIntegrationOptions};
+
+/// Iceberg registration and session extensions for local and remote benchmarks.
+pub struct IcebergBenchmarkBackend;
+
+#[async_trait]
+impl BenchmarkBackend for IcebergBenchmarkBackend {
```
gabotechs: Typically in this project we never place actual code in `lib.rs` or `mod.rs` files, we reserve them just for module plumbing. It should be fine and easy to just move this to a backend.rs file.

iceberg/benchmarks/src/dataset.rs
```diff
@@ -32,18 +34,38 @@ use parquet::file::properties::WriterProperties;
...
 type CatalogProperties = std::collections::HashMap<String, String>;
 
-/// The sibling dataset directory used for the Iceberg representation.
-pub(super) fn output_path(input: &Path) -> PathBuf {
-    let mut name = input.file_name().unwrap_or_default().to_os_string();
-    name.push("_iceberg");
-    input.with_file_name(name)
+/// Resolve the destination URI, defaulting to a local sibling dataset.
+pub(super) fn output_location(
+    input: &Path,
```
gabotechs: We don't need to worry about populating the remote buckets. The project in https://github.com/gabotechs/datafusion-distributed-dev-tools will just take whatever this project dumps locally and will sync it with the remote s3 bucket.

I think we should be fine without any changes to the dataset generation.

iceberg/benchmarks/README.md
```diff
@@ -15,3 +17,36 @@ WORKERS=2 ./iceberg/benchmarks/run.sh --dataset tpch/sf1_iceberg --threads 2 --p
 Preparation writes an immutable sibling dataset named `<input>_iceberg`. The source must be a
 non-empty local Parquet dataset, and the destination must be empty. See the
 [benchmark guide](../../benchmarks/README.md#iceberg-benchmarks) for generation and comparison.
+
+## Remote benchmarks
```
gabotechs: I'd not document remote benchmarks here. The current remote benchmarks are handled by https://github.com/gabotechs/datafusion-distributed-dev-tools, and deployed under Datadog's infrastructure, so in this project the official stance is to not provide support for remote benchmarks at all (besides the worker binary for other projects to do their benches)

gabotechs (commented):
👍 Approving for unblocking, but please take a look at the comments first. The approach looks good thanks!

gabotechs (approved):

iceberg/benchmarks/README.md
alexanderbianchi: Removed the remote benchmark documentation and restored this guide to local-only scope.

iceberg/benchmarks/src/dataset.rs
alexanderbianchi: Restored local-only dataset generation and removed the object-store output option and its extra dependencies.

iceberg/benchmarks/src/lib.rs
alexanderbianchi: Moved the backend implementation to `backend.rs`; `lib.rs` now contains module plumbing only.

Issue #606, opened 2026-08-09
[Iceberg] Support partitioning in work-unit feeds


gabotechs:
Part of #599.

Extend `IcebergWorkUnitFeed` beyond `Partitioning::UnknownPartitioning` to support hash and range partitioning, including routing scan tasks according to the requested output partitioning.

Essentially, implementing these TODO comments in the code:

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/work_unit_feed.rs#L93-L94

https://github.com/datafusion-contrib/datafusion-distributed/blob/72203a0faf105ca9c6d5775a184ca4e1e1cd79d1/iceberg/src/work_unit_feed.rs#L187-L188

This needs to define which requested partitioning schemes can actually be satisfied from Iceberg scan tasks, then route the tasks without violating those semantics.

alexanderbianchi:
@sandugood if you are looking for work, this one is relevant to DFD regardless the broader iceberg-rust + datafusion stuff.

sandugood:
take

Issue #119, opened 2025-09-02
Improvement Epic: Enhancing Join Efficiency and Reducing Network Overhead Through Data Layout Awareness


NGA-TRAN:
This is a broad topic, so we'll explore it through two examples

Optimizing distributed joins involves several factors, including join order—which itself is influenced by data layout and execution strategy. To simplify the discussion, we'll assume a fixed join order and focus on optimizing repartitioning and data shuffling.

# Single Join Case

We begin with a query performing an inner join between two tables, with a fixed assignment of build and probe sides. Since our focus is on join behavior, unrelated operators are omitted for clarity.

<img width="348" height="195" alt="Image" src="https://github.com/user-attachments/assets/04ab3719-1bc9-4dda-9ade-994eb011c874" />

Assuming both tables are large, we typically partition their data by the join key to enable efficient joining, as illustrated below. 

### Simplified Textual Plan

```
┌───── Stage 4   Task: partitions: 0
│. . . . . .
│     ArrowFlightReadExec: Stage 3 
└─────────────────────────── ──────────
  ┌───── Stage 3   Task: partitions: 0, Task: partitions: 1
  │RepartitionExec: Send as-is *
  │    HashJoinExec: mode=Partitioned, join_type=Inner, on=[(c_custkey, o_custkey)]
  │        . . . . . .
  │            ArrowFlightReadExec: Stage 1
  │        . . . . . .
  │            ArrowFlightReadExec: Stage 2
  └─────────────────────────────────────
    ┌───── Stage 1   Task: partitions: 0, Task: partitions: 1
    │RepartitionExec: partitioning=Hash (c_custkey)
    │    . . . . .
    │        DataSourceExec: file_groups={2 groups of customer}
    └─────────────────────────────────────
    ┌───── Stage 2   Task: partitions: 0, Task: partitions: 1
    │RepartitionExec: partitioning=Hash (o_custkey)
    │    . . . . .
    │        DataSourceExec: file_groups={2 groups of orders}
    └─────────────────────────────────────
```

### Equivalent Graphical Plan

<img width="855" height="759" alt="Image" src="https://github.com/user-attachments/assets/f811eaaf-a3f8-4812-b972-a3acba570e62" />

Let’s walk through key details of this plan that will inform further optimization in the next section.

[...]

PR #733, opened 2026-09-17
feat[iceberg]: hash-partitioning during file scan tasks on identity transformed partitions

sandugood:
Tackles #606 for the identity partition columns.

### What's new?

- now during every file scan task instead of simply throwing `UnknownPartitioning` we can enable `Hash` over the identity transformed columns.

### How does it work?
- identity transform leaves column values essentially without any transformation. So if the values are low-cardinality or of `DATE` type (just an example) it will be partitioned on the disk as: `value='my_low_card_value'` or `date='2026-16-09'`.
This information is being used during both grouping and joins, as we can throw off the `RepartitionExec` node, replacing it with direct scans of needed partitions (in distributed context this would replace overhead of shuffling during joins/group by).

#### Tests
- this feature was tested against the existing taxi dataset, as it is partitioned by the `pickup_date`:
```

  "partition-specs": [
    {
      "spec-id": 0,
      "fields": [
        {
          "source-id": 13,
          "field-id": 1000,
          "name": "pickup_date",
          "transform": "identity"
        }
      ]
    }
  ],

```


iceberg/src/common/scalar.rs
```diff
@@ -0,0 +1,42 @@
...
+use iceberg::spec::{Datum, PrimitiveLiteral, PrimitiveType};
+
+/// Conversion function of iceberg's Datum
+pub(crate) fn datum_to_scalar(d: &Datum) -> Option<ScalarValue> {
+    primitive_to_scalar(d.data_type(), d.literal())
+}
+
+/// Converts an Iceberg primitive literal of the given type into a [ScalarValue]. Returns `None`
+/// for the types that have no supported conversion.
+pub(crate) fn primitive_to_scalar(
```
alexanderbianchi, line 11: https://iceberg.apache.org/spec/?h=primiti#primitive-types for reference. Looks like maybe there's more timestamp types we could support. Others could be added over time, but `binary` could probably be supported as well.

iceberg/src/table_provider/catalog.rs
```diff
@@ -71,19 +70,20 @@ impl TableProvider for IcebergCatalogTableProvider {
...
             table.clone(),
             self.schema.clone(),
-            Partitioning::UnknownPartitioning(state.config().target_partitions()),
+            state.config().target_partitions(),
             IcebergDataSourceOptions {
                 snapshot_id: None,
                 projection,
                 filters,
                 fetch: limit,
+                hash_partitioning_enabled: iceberg_config.hash_partitioning_enabled,
```
alexanderbianchi, line 83: @gabotechs should we maybe just use the benchmarks now to determine if we should just have this enabled by default?

iceberg/src/partitioning.rs
```diff
@@ -0,0 +1,247 @@
...
+use iceberg::spec::{
+    Literal, PartitionSpec, PrimitiveType, Schema, TableMetadata, Transform, Type,
+};
+use iceberg::table::Table;
+
+use crate::common::{df_err, primitive_to_scalar};
+
+/// Returns the output partitioning that a scan producing `output_schema` can declare.
+///
+/// Only one scheme can be satisfied from Iceberg metadata alone: [Partitioning::Hash] over
```
alexanderbianchi:
```suggestion
/// Only one schema can be satisfied from Iceberg metadata alone: [Partitioning::Hash] over
```

iceberg/src/partitioning.rs
```diff
@@ -0,0 +1,247 @@
...
+/// identity-transformed partition columns. All the rows of a data file share the same value for
+/// those columns, and that value is known from the file's partition metadata, so
+/// [FileScanTaskRouter] can place each file in the partition where a hash repartition would have
+/// placed its rows.
+///
+/// Any other Iceberg transform does not qualify. `bucket[N]` in particular hashes with murmur3,
```
alexanderbianchi, line 27: This issue has come up other places before. `RepartitionExec` is tied to the DF specific hash function. Not for this PR but:
- We could follow up with something like a `iceberg_bucket(N, expression)` that feeds the DF hash so repartioning becomes `DF_HASH(iceberg_bucket(N, key)) % P`
- upstream we could allow HashPartitioning to take in a hash function that matches the iceberg hash properties. (which would either need to be applied to the whole query plan, or the planner would need to know when to default to a DF repartition if for example joining with a non-iceberg table) 

https://iceberg.apache.org/spec/#bucket-transform-details

iceberg/src/partitioning.rs
```diff
@@ -0,0 +1,247 @@
...
+                    exec_datafusion_err!(
+                        "Iceberg file scan task for {} has no identity partition value for field id {}",
+                        task.data_file_path(),
+                        key.source_id
+                    )
+                })?;
+            let scalar = match value {
+                None => ScalarValue::try_from(&key.arrow_type)?,
+                Some(Literal::Primitive(literal)) => {
+                    primitive_to_scalar(&key.iceberg_type, literal)
```
alexanderbianchi, line 167: https://iceberg.apache.org/spec/#schema-evolution
Iceberg allows promoting an identity-partitioned `int` column to `long`.
The literal here comes from iceberg partitioning, which could be `PrimitiveLiteral::Int`. `key.iceberg_type` comes from `scan_schema` which if the schema has evolved could be ` PrimitiveType::Long` 
That can produce a mismatched pair
 ```rust
   (PrimitiveType::Long, PrimitiveLiteral::Int(v))
 ```
Which would return `None`. 
1. Historical manifests retain an `PrimitiveLiteral::Int` partition value
2. `key.iceberg_type` comes from the snapshot schema and is now `PrimitiveType::Long`.
3. `primitive_to_scalar(&key.iceberg_type, literal)` therefore returns `None`
4. Enabling this feature makes an otherwise valid evolved table fail with “unsupported Iceberg partition value”. 

Something like `ScalarValue::Int64(Some(i64::from(*v)))` for the supported evolved schema's should do the trick.

iceberg/src/partitioning.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        // Same hashing as RepartitionExec, so that this scan is co-partitioned with any other
+        // input that was hash repartitioned by the same keys.
+        let mut hashes = vec![0; 1];
+        create_hashes(
+            &arrays,
+            REPARTITION_RANDOM_STATE.random_state(),
+            &mut hashes,
+        )?;
+        Ok((hashes[0] % channels as u64) as usize)
```
alexanderbianchi, line 193: This unconditionally routes every file with the same identity value to one channel. For a table with one partition value and 64 target partitions, enabling the option changes the scan from 64 round-robin readers to one non-empty reader and 63 empty readers, collapsing DataFusion/distributed partition parallelism and creating a worker hotspot.

The Parquet/listing implementation guards this trade-off with `optimizer.preserve_file_partitions`: it retains partition-aware grouping only when the number of resulting groups reaches the configured threshold; otherwise, it falls back to normal file splitting without declaring hash partitioning.

Could this use an equivalent cardinality/parallelism guard rather than a boolean? Because `route` runs after the scan has advertised its output partitioning, the decision likely needs to gate the `Partitioning::Hash` returned by `scan_partitioning`.

- Partition grouping: https://github.com/apache/datafusion/blob/55.0.0/datafusion/datasource/src/file_groups.rs#L486-L542
- Cardinality threshold and fallback: https://github.com/apache/datafusion/blob/55.0.0/datafusion/catalog-listing/src/table.rs#L878-L900
- Declared scan partitioning: https://github.com/apache/datafusion/blob/55.0.0/datafusion/datasource/src/file_scan_config/mod.rs#L590-L614

This also seems related to the data-layout and worker-colocation design in https://github.com/datafusion-contrib/datafusion-distributed/issues/119.

PR #734, opened 2026-09-17
chore: Add tests and benchmarks for Range partitioning

stuhood:
## Summary

Adds integration tests and benchmarks for range partitioning, multi-way joins, and range shuffle pipelines, unblocking #628.

This PR adds fixtures, benchmark scenarios, and baselines for range partitioning across stages. Support for `Partitioning::Range` across `NetworkShuffleExec` and `NetworkCoalesceExec` is stacked atop this PR in #730.

## Changes

- Integration tests and fixtures:
  - `tests/range_partitioning.rs`: 7 integration tests covering 2-way and 3-way distributed range join configurations (pre-partitioned co-partitioned joins, stream adaptation, under/over-parallelism, and hybrid broadcast pipelines).
  - `testdata/join/`: Added `services` dimension dataset (CSV, SQL generation script, Parquet) for 3-way join scenarios.
  - `src/test_utils/mock_exec.rs`: Added `with_partitioning` builder to `MockExec`.

- Range shuffle and join benchmarks:
  - `benchmarks/benches/range_join_scenarios.rs`: Benchmarks for 5 real-world range join scenarios (derived from ParadeDB optimizations in paradedb/paradedb#6342).
  - `benchmarks/benches/shuffle.rs`: Added range shuffle configurations across 1-to-1, 4-to-1, 4-to-4, 8-to-8, and 8-to-4 producer/consumer task ratios.
  - `src/execution_plans/benchmarks/shuffle_bench.rs`: Added `ShufflePartitioningMode` enum (`Hash` and `Range`) and synthetic range partitioning helpers.

PR #735, opened 2026-09-17
fix: preserve `NetworkShuffleExec` output ordering across tasks via streaming merge

stuhood:
## Summary

When `NetworkShuffleExec` consumes an upstream stage with `input_task_count > 1`, worker streams were multiplexed non-deterministically via `futures::stream::select_all`. If an upstream stage maintained sort order (e.g. from an ordered scan into a hash repartition), downstream operators such as `SortExec` evaluated `ordering_satisfy == true` and bypassed sorting at runtime, yielding out-of-order batches.

`NetworkShuffleExec` now honors its declared output ordering by sort-merging incoming worker streams across tasks:

- In `src/execution_plans/network_shuffle.rs`:
  - In `execute`, merge worker streams using `StreamingMergeBuilder` when `output_ordering` is present, short-circuiting on single/empty streams and falling back to `select_all` when unordered.
  - In `compute_properties`, retain `output_ordering` while clearing per-partition constants across multiple tasks.
  - In `DisplayAs`, display `sort_exprs=[...]` when performing a sorted merge across multiple input tasks (`input_tasks > 1`).
  - Add unit tests verifying that `output_ordering` is preserved across single and multi-task configurations.
- In `src/codec/distributed_codec.rs`:
  - Add `ordering: Vec<PhysicalSortExprNode>` (tag 5) to `NetworkShuffleExecProto` so sort expressions are transmitted over the wire to remote workers.
- In `tests/shuffle.rs`:
  - Add integration test `test_sorted_network_shuffle` comparing a multi-worker distributed shuffle against single-node DataFusion to verify identical content and sort order.

Originally extracted from a flaky test in https://github.com/datafusion-contrib/datafusion-distributed/pull/734.


src/execution_plans/network_shuffle.rs
```diff
@@ -108,9 +108,38 @@ pub struct NetworkShuffleExec {
 }
 
 impl NetworkShuffleExec {
+    /// Computes the properties advertised by this [NetworkShuffleExec].
+    ///
+    /// When `input_task_count > 1`, worker streams are merged concurrently via
+    /// `futures::stream::select_all`, yielding record batches non-deterministically across
+    /// upstream tasks. Therefore, any ordering guarantees or partition-local equivalence
+    /// constants from upstream tasks cannot be preserved across the network boundary and are cleared.
```
gabotechs: 🤔 This does look like a problem. Even if we correctly clear the orderings and expose no ordering to upstream nodes, the planner rule that enforces ordering runs before placing these NetworkShuffleExec nodes, so we'd still be missing some `SortExec`s that ensure this ordering, as the plan was created originally under the assumption that ordering do was preserved.

gabotechs: Do you think we can replicate this with an integration test?

src/execution_plans/network_shuffle.rs
```diff
@@ -266,3 +297,76 @@ impl ExecutionPlan for NetworkShuffleExec {
...
+
+    #[test]
+    fn clears_output_ordering_when_multiple_input_tasks() {
+        let repart = sample_hash_repart(true);
+        assert!(repart.properties().output_ordering().is_some());
+
+        // Multiple producer tasks: ordering and per-partition constants should be cleared
+        let shuffle = NetworkShuffleExec::try_new(repart, 3).unwrap();
+        assert_eq!(shuffle.properties().output_ordering(), None);
+    }
```
gabotechs: For these things, we typically prefer integration tests that span a real query against one of the parquet datasets committed to the repo. 

Maybe adding a new one here `tests/shuffle.rs`?

src/execution_plans/network_shuffle.rs
```diff
@@ -108,9 +108,38 @@ pub struct NetworkShuffleExec {
...
+    ///
+    /// When `input_task_count <= 1`, all batches are received from a single upstream task stream,
+    /// so the upstream equivalence properties and sort order are preserved.
+    pub(crate) fn compute_properties(
+        input_properties: &Arc<PlanProperties>,
+        input_task_count: usize,
+    ) -> Arc<PlanProperties> {
+        if input_task_count > 1 {
+            let mut eq_properties = input_properties.eq_properties.clone();
+            eq_properties.clear_orderings();
```
gabotechs: I'm wondering that, rather than clearing the orderings here, we might actually need to honor them, and stop doing the `select_all()` in favor of something that does sort merge the streams.

stuhood: Should we be adding `SortPreservingMergeExec`?

I'll take a look at it.

gabotechs: Something similar yeah. Maybe not `SortPreservingMergeExec` itself, as that's a full blown node, but probably it's internal machinery for merging multiple streams into one in a sorted manner.

src/execution_plans/network_shuffle.rs
gabotechs: This would actually be beautifully replicated with an integration test that executes single-node VS distributed and asserts the same results.

src/execution_plans/network_shuffle.rs
stuhood: Looks like `StreamingMergeBuilder`. On it.

stuhood:
@gabotechs : Ready for another look!

src/execution_plans/network_shuffle.rs
```diff
@@ -266,3 +336,83 @@ impl ExecutionPlan for NetworkShuffleExec {
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
gabotechs: I think we should be good with the integration tests. We typically have avoided placing tests here, as they've been a mantainance burden more than anything, without really covering anyhting that is not already covered by the integration tests.

As long as the integration tests do cover failure scenarios, we should be good.

src/execution_plans/network_shuffle.rs
```diff
@@ -232,22 +273,51 @@ impl ExecutionPlan for NetworkShuffleExec {
...
+            return Ok(streams.pop().unwrap());
+        }
+
+        if let Some(ordering) = self.properties.output_ordering()
+            && !ordering.is_empty()
+        {
+            let reservation = MemoryConsumer::new(format!("NetworkShuffleExec[{partition}]"))
+                .register(&context.runtime_env().memory_pool);
+            let batch_size = context.session_config().batch_size();
+            let metrics = BaselineMetrics::new(&self.worker_connections.metrics, partition);
```
gabotechs: I think we are going to be double-counting metrics here. The `worker_connections` already bundle all the relevant metrics, so we should be fine to not pass them.

src/execution_plans/network_shuffle.rs
```diff
@@ -232,22 +273,51 @@ impl ExecutionPlan for NetworkShuffleExec {
...
+            StreamingMergeBuilder::new()
+                .with_streams(streams)
+                .with_schema(self.schema())
```
gabotechs, line 314: Nice! pretty convenient helper.

tests/shuffle.rs
```diff
@@ -0,0 +1,153 @@
...
+        ");
+
+        let single_node_formatted = pretty_format_batches(&single_node_results)?.to_string();
+        let distributed_formatted = pretty_format_batches(&distributed_results)?.to_string();
+
+        assert_eq!(
+            distributed_formatted, single_node_formatted,
+            "Distributed results must match single-node results exactly in content and sort order"
+        );
+
```
gabotechs, line 138: I see this assertion passes even on `main`, although the plan does look like it's exercising the new path :thinking:.

Any ideas on how to tweak this test so that it demonstrates the failure on `main`?

stuhood: Hm... tricky. This was originally reproduced with Range partitioning... let me see if I can still get a relevant plan with Hash.

stuhood: Ok, was able to get it to reproduce, albeit flakily: please take another look.

stuhood:
Was able to get the test to repro (flakily) on `main`.

This does now need some proto changes though. Please take another look!

src/codec/distributed_codec.rs
```diff
@@ -516,12 +538,14 @@ pub struct NetworkShuffleExecProto {
     input_stage: Option<StageProto>,
     #[prost(message, repeated, tag = "4")]
     equivalence_classes: Vec<EquivalenceClassProto>,
+    /// Sort expressions preserved across tasks and used by workers to sort-merge streams.
+    #[prost(message, repeated, tag = "5")]
+    ordering: Vec<protobuf::PhysicalSortExprNode>,
```
gabotechs, line 543: 👍 This is fair and expected, so all good here.

gabotechs (approved):
All good here @stuhood. Good catch, and thanks for the PR!

PR #742, opened 2026-09-22
refactor(iceberg): adopt Apache datafusion-iceberg

alexanderbianchi:
## Summary

- add the Apache `datafusion-iceberg` crate as a pinned dependency of the DFD Iceberg integration
- use its public `convert_filters_to_predicate` implementation and remove DFD's copied implementation and tests
- align the direct `iceberg` and `iceberg-storage-opendal` revisions with the revision used by `datafusion-iceberg`

DFD still depends directly on iceberg-rust for core Iceberg model, runtime, scan-task, and storage types because `datafusion-iceberg` does not re-export them. DFD's table providers and data source also remain custom: they build an `IcebergDataSource` backed by distributed work-unit feeds, whereas the upstream providers build a local `IcebergTableScan`.

Ideally we see datafusion-iceberg and DFD converge such that DFD has it's own data source but it's swapped out during planning. This is because DFD relies on work-unit feeds which are not a general datafusion concept.

## Validation

- `cargo check -p datafusion-distributed-iceberg`
- `cargo test -p datafusion-distributed-iceberg --lib`
- `git diff --check`


gabotechs (approved):
Nice, thanks @alexanderbianchi!

